PR #5, opened 2018-04-16
fixed minor typo in begin_circuit()

cgogolin:
Fix for a inconsequential typo I found while browsing the code.


josh146:
Hey Christian, the raw string noted in this pull request is intentional - while there are currently no backslash characters in that particular docstring that would need to be escaped, any docstring that contains mathjax we have written as a raw string to avoid Sphinx documentation rendering errors down the line

cgogolin:
Sorry. Got it. The r""" """ seemed unnecessary, but you are right.

PR #6, opened 2018-04-16
Minor improvements in `fock_prob()`

cgogolin:
Some minor improvements to `fock_prob()` that should make the code a bit more readable and which, in test runs with cProfile, reduce the average run-time of the unite tests of the `gaussian` backend by a few percent (for larger systems the difference should be more substantial). Hope you find these useful!


josh146:
Looks good!

Perhaps @nquesada could have a look over the change, especially the rounding of `Amat` - if I recall, the rounding was intentional to prevent an edge case bug where numerical precision affected the symmetry of Amat?

cgogolin:
If that's the case, then maybe it would be a good idea to add a test that fails when the rounding is removed. Anyway, the (minor) performance improvement comes from not inverting a matrix twice. Sorry for failing to add spaces around ==.

Issue #16, opened 2018-06-20
New state preparations


co9olguy:
We'd like to add the possibility for users to prepare arbitrary states in the Fock basis. Currently we allow a single mode pure state. We will now extend this to allow the following cases:
- single mode density matrix
- multimode pure state
- multimode mixed state

I've added three new methods to `backends/fockbackend/backend.py` that specify the API:
`prepare_dm_state`
`prepare_multimode_ket_state`
`prepare_multimode_dm_state`

@cgogolin can you modify the corresponding Fock circuit simulator in `backends/fockbackend/circuit.py` so that it implements these functions?

Also you should add new tests for these (see `tests/test_state_preparations.py:test_prepare_ket_state for an example to follow)

smite:
I would suggest combining the _multimode_ functions with the corresponding single mode functions, since the 'mode' parameter can uniquely determine the intent. E.g. mode=4 (or mode=[4]) for a single-mode preparation, and mode=[0,4,3] for a multimode preparation.

co9olguy:
Yep, that is the plan for the final API. The multimode state preps will probably require new methods in the circuit simulators, which is what @cgogolin will implement. But the final user-facing API will likely combine all functionalities into one method

josh146:
Resolved by pull request #19

PR #19, opened 2018-07-08
New state preps multimode

cgogolin:
This fully implements the single and multi mode state preparations in the fock and tf backends as specified in #16 and adds various corresponding unit tests tests. The implementation is a little bit more general than would have been necessary, as it allows the target mode list to be out of order and then correctly reorders the modes.
**TODO**
- The multimode functions are not yet included in the base.py
- The API needs some thought. It might make sense to remove the single mode versions of the state preparation methods (wich already now simply call through to the multi mode variants) and rename the multimode variants
**Benefits:**
- Allows the preparation of arbitrary single and multi mode states or arbitrary modes 
- The order of the specified modes is taken into account when interpreting the input
- The state() methods that returns the state can now also return correctly reordered states of a list of modes that needs not be sorted.
**Possible Drawbacks:**
- The newly implemented code does not always use the `tensor()`, `insert()`, and similar methods in the corresponding ops.py files, but rather calls numpy/tensor flow directly, as these were written in a way that was hard to generalize.  
**Related GitHub Issues:**
#16


cgogolin:
The last few commits implement the API changes we discussed earlier. This resolves the two open TODO points mentioned in the initial pull request.

cgogolin:
If all tests pass, I will, as a final step, run local test with `tensorflow==1.3` and then give green light for merging if all goes well.

cgogolin:
Turns out that the way I am now using `tf.tensordot()` in `/tfbackend/ops.py` in line 744, in replace_modes, namely:
```revised_modes = tf.tensordot(reduced_state, replacement, axes=0)```
raises a
```ValueError: 'axes' must be at least 1.```
with `tensorflow<=1.4`. From `1.5` onwards all works.

Would a dependency on `tensorflow>=1.5` be acceptable, or shall I look for a workaround? Alternatively one can probably add a fake dimension and then `tensordot` that away conditional on `tensorflow<=1.4`.

co9olguy:
Hmm... It would definitely be preferable to find a workaround that doesn't require checking which version of Tensorflow is installed

cgogolin:
As python packages have no numerical version numbers, comparing version is indeed a not a good idea (tough there are solutions: https://stackoverflow.com/questions/20180543/how-to-check-version-of-python-modules). 

It turns out that one can indeed replace
```            revised_modes = tf.tensordot(reduced_state, replacement, axes=0)```
by
```            revised_modes = tf.tensordot(tf.expand_dims(reduced_state, 0), tf.expand_dims(replacement, 0), axes=[[0],[0]])```
and the latter works with `tensorflow>=1.3`.

I am sure this hack of adding extra dimensions it not exactly efficient, so I put a todo so that we can remember to remove it once `tensorflow>=1.6` becomes a dependency.

PR #34, opened 2018-09-26
Added arbitrary second order phase space expectation/variances

josh146:

* Added the method `BaseState.poly_quad_expectations(A, d, k, phi=0)`. This returns the expectation `E[P(r)]` and variance `var[P(r)]` of arbitrary second order polynomials of the form `P(r) = r^T A r/2 + r^T d+kI`, where the vector of quadrature operators  `r = (x1, x2, ..., x_n, p1, ..., p_n)` has been optionally rotated by angle `phi` in the phase space. In particular,
  - `A` is a  2Nx2N real symmetric matrix containing the quadratic coefficients,
  - `d` is a length-2N real vector containing the linear coefficients, and
  - `k` is the constant term.

  In the Gaussian backend, the formulas for the [quadratic form](https://en.wikipedia.org/wiki/Quadratic_form_(statistics)) are used. In the Fock backends, `E[P(r)] = Tr(P(r)*rho)` is calculated directly using `np.einsum`.

* Added unit tests for `poly_quad_expectations`. These test that, for `E[x]` and `E[p]`, that the result agrees with `quad_expectations`. In addition, it also tests against the known result for `E[x^2]`, `E[p^2]`, `E[xp]`, as well as various arbitrary quadratic forms - both single mode, and multi mode.

* `BaseState.mean_photon` now returns a tuple of both the mean and the variance, bringing it in line with `BaseState.quad_expectations`.

* Modified the `GaussianState.mean_photon` method to use the equations

  `<n> = [Tr(cov)+muT.mu]/(2\hbar) - 1/2`
  `\sigma_n^2 = [Tr(cov^2)+2muT.cov.mu]/(2\hbar^2) - 1/4`

* Added additional unit tests for `mean_photon`, expanding to displaced squeezed state, and also testing the variance.

**Benefits:**

* Added the ability to return arbitrary second order expectations

* The output of `mean_photon` now matches those of the other expectation methods

* Calculation of the mean photon number is now much more efficient and accurate for Gaussian states

**Possible Drawbacks:**

* At the moment, `poly_quad_expectations` does not work on the TF backend in batch mode - this will need to be subclassed and fixed within the `TFFockState` class. In addition, symbolic evaluation will also fail until the NumPy function calls are replaced with their TF equivalent.

* In the Fock backends:
  - the `poly_quad_expectations` calculation scales quite badly as the number of operators acting on _different_ modes increases. This is due to the increasing number of dimensions required in the einsum calculation. This can be somewhat alleviated by reducing the Fock cutoff, at the expense of accuracy. Alternatively, the use of sparse matrices might solve this problem.
  - There is a current limit of ~15 modes in the multimode expectation, due to the use of `einsum`.

**Related GitHub Issues:**

n/a

### DIFF
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -61,6 +61,7 @@ inheriting classes.
     fidelity_coherent
     wigner
     quad_expectation
+    poly_quad_expectation
 
 
 Base Gaussian state
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -405,7 +405,7 @@ class BaseState(abc.ABC):
                 perform the calculation. The calculation will take place within a Fock basis
                 dimension of cutoff_dim+worksize. Increasing the worksize will result in
                 increased numerical accuracy, at the expense of a longer computational time.
-                By default, worksize is set to 5.
+                By default, worksize is set to 1.
 
         Returns:
             tuple (float, float): expectation value and variance
@@ -767,7 +767,7 @@ class BaseFockState(BaseState):
 
         # expand the cutoff dimension in approximating the x and p
         # operators in the Fock basis, to reduce numerical inaccuracy.
-        worksize = kwargs.get('worksize', 5)
+        worksize = kwargs.get('worksize', 1)
         dim = self._cutoff + worksize
 
         # construct the x and p operators
@@ -851,8 +851,7 @@ class BaseFockState(BaseState):
 
         # add constant term
         if k != 0:
-            for n in range(num_modes):
-                poly_op += k*expand_dims(np.eye(dim), n, num_modes)
+            poly_op += k*expand_dims(np.eye(dim), 0, num_modes)
 
         # calculate Op^2
         ind = "{},{}->{}".format(ind1[1:], ind2[1:], ind3)
@@ -871,8 +870,8 @@ class BaseFockState(BaseState):
         # For 3 modes, this gives the einsum '(ab)(cd)(ef),(ba)(dc)(fe)->'
         mean = np.einsum(ind, poly_op, rho).real
 
-        # calculate variance Tr(Op^2 @ rho) - Tr(Op @ rho)^2 + |A| + 2k/hbar^2
-        var = np.einsum(ind, poly_op_sq, rho).real - mean**2 + np.linalg.det(2*quad_coeffs) + 2*k/(self._hbar**2)
+        # calculate variance Tr(Op^2 @ rho) - Tr(Op @ rho)^2
+        var = np.einsum(ind, poly_op_sq, rho).real - mean**2
 
         return mean, var
 
@@ -1108,7 +1107,6 @@ class BaseGaussianState(BaseState):
         covphi = rot.T @ cov @ rot
         return (muphi[0], covphi[0, 0])
 
-
     def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
         if A is None:
             A = np.zeros([2*self._modes, 2*self._modes])
@@ -1128,6 +1126,12 @@ class BaseGaussianState(BaseState):
         # determine modes with quadratic expectation values
         nonzero = np.concatenate([np.mod(A.nonzero()[0], self._modes), np.mod(d.nonzero()[0], self._modes)])
         ex_modes = list(set(nonzero))
+
+        # reduce the size of A so that we only consider modes
+        # which we need to calculate the expectation value for
+        rows = ex_modes + [i+self._modes for i in ex_modes]
+        num_modes = len(ex_modes)
+        quad_coeffs = A[:, rows][rows]
 
         if not ex_modes:
             # only a constant term was provided
@@ -1157,7 +1161,14 @@ class BaseGaussianState(BaseState):
         mean = np.trace(A @ cov) + k2
         # variance Var[P(r)]_{mu=0} = 2tr(A.cov.A.cov) + 4*muT.A.cov.A.mu + dT.cov.d|_{mu=0}
         #                           = 2tr(A.cov.A.cov) + dT.cov.d
-        var = 2*np.trace(A @ cov @ A @ cov) + d2.T @ cov @ d2 + k/2
+        var = 2*np.trace(A @ cov @ A @ cov) + d2.T @ cov @ d2
+
+        # correction term to account for incorrect symmetric ordering in the variance
+        # this occurs because Var[S(P(r))] = Var[P(r)] - Σ_{m1, m2} |hbar*A_{(m1, m1+N),(m2, m2+N)}|
+        # where m1, m2 are all possible mode numbers, and N is the total number of modes.
+        # Therefore, the correction term is the sum of the determinants of 2x2 submatrices of A.
+        modes = np.arange(2*num_modes).reshape(2, -1).T
+        var -= np.sum([np.linalg.det(self._hbar*quad_coeffs[:, m][n]) for m in modes for n in modes])
 
         return mean, var
 
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -833,7 +833,7 @@ class BaseFockState(BaseState):
             poly_op += k*expand_dims(np.eye(dim), 0, num_modes)
 
         # truncate down
-        sl = [slice(0, self._cutoff)]*(2*num_modes)
+        sl = tuple([slice(0, self._cutoff)]*(2*num_modes))
         poly_op = poly_op[sl]
 
         # calculate Op^2
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -388,13 +388,13 @@ class BaseState(abc.ABC):
 
         as well as the variance
 
-        .. math:: \Delta P(\mathbf{r})^2 = \langle P(\mathbf{r})^2\rangle - \braket{P(\mathbf{r})}^2
+        .. math:: \Delta P(\mathbf{r})^2 = \braket{P(\mathbf{r})^2} - \braket{P(\mathbf{r})}^2
 
         Args:
             A (array): a real symmetric 2Nx2N NumPy array, representing the quadratic
                 coefficients of the second order quadrature polynomial.
-            d (array): a real symmetric length-2N NumPy array, representing the linear
-                coefficients of the second order quadrature polynomial. Default the zero vector.
+            d (array): a real length-2N NumPy array, representing the linear
+                coefficients of the second order quadrature polynomial. Defaults to the zero vector.
             k (float): the constant term. Default 0.
             phi (float): quadrature angle, clockwise from the positive :math:`x` axis. If provided,
                 the vectori of quadrature operators :math:`\mathbf{r}` is first rotated
@@ -751,7 +751,7 @@ class BaseFockState(BaseState):
         if A.shape != (2*self._modes, 2*self._modes):
             raise ValueError("Matrix of quadratic coefficients A must be of size 2Nx2N.")
 
-        if np.any(A.T != A):
+        if not np.allclose(A.T, A):
             raise ValueError("Matrix of quadratic coefficients A must be symmetric.")
 
         if d is None:
@@ -825,7 +825,7 @@ class BaseFockState(BaseState):
         ind3 = ''.join([str(i)+str(j) for i, j in zip(ind1[1::2], ind2[2::2])])
         ind = "{},{}->{}".format(ind1, ind2, ind3)
 
-        if np.all(quad_coeffs == 0.):
+        if np.allclose(quad_coeffs, 0.):
             poly_op = np.zeros([dim]*(2*num_modes), dtype=np.complex128)
         else:
             # Einsum above applied to to r,Ar
@@ -1102,7 +1102,7 @@ class BaseGaussianState(BaseState):
         if A.shape != (2*self._modes, 2*self._modes):
             raise ValueError("Matrix of quadratic coefficients A must be of size 2Nx2N.")
 
-        if np.any(A.T != A):
+        if not np.allclose(A.T, A):
             raise ValueError("Matrix of quadratic coefficients A must be symmetric.")
 
         if d is not None:
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -119,10 +119,12 @@ from itertools import chain
 from copy import copy
 
 import numpy as np
+from scipy.linalg import block_diag
 from scipy.stats import multivariate_normal
 from scipy.special import factorial
 
 from .shared_ops import rotation_matrix as _R
+from .shared_ops import changebasis
 
 indices = string.ascii_lowercase
 
@@ -353,6 +355,56 @@ class BaseState(abc.ABC):
 
                 * :math:`\phi=0` corresponds to the :math:`x` expectation and variance (default)
                 * :math:`\phi=\pi/2` corresponds to the :math:`p` expectation and variance
+
+        Returns:
+            tuple (float, float): expectation value and variance
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
+        r"""The multi-mode expectation values and variance of arbitrary 2nd order polynomials
+        of quadrature operators.
+
+        An arbitrary 2nd order polynomial of quadrature operators over $N$ modes can always
+        be written in the following form:
+
+        .. math:: P(\mathbf{r}) = \frac{1}{2}\mathbf{r}^T A\mathbf{r} + \mathbf{r}^T \mathbf{d} + k I
+
+        where:
+
+        * :math:`A\in\mathbb{R}^{2N\times 2N}` is a symmetric matrix
+          representing the quadratic coefficients,
+        * :math:`\mathbf{d}\in\mathbb{R}^{2N}` is a real vector representing
+          the linear coefficients,
+        * :math:`k\in\mathbb{R}` represents the constant term, and
+        * :math:`\mathbf{r} = (\x_1,\dots,\x_N,\p_1,\dots,\p_N)` is the vector
+          of quadrature operators in :math:`xp`-ordering.
+
+        This method returns the expectation value of this second-order polynomial,
+
+        .. math:: \langle P(\mathbf{r})\rangle,
+
+        as well as the variance
+
+        .. math:: \Delta P(\mathbf{r})^2 = \langle P(\mathbf{r})^2\rangle - \braket{P(\mathbf{r})}^2
+
+        Args:
+            A (array): a real symmetric 2Nx2N NumPy array, representing the quadratic
+                coefficients of the second order quadrature polynomial.
+            d (array): a real symmetric length-2N NumPy array, representing the linear
+                coefficients of the second order quadrature polynomial. Default the zero vector.
+            k (float): the constant term. Default 0.
+            phi (float): quadrature angle, clockwise from the positive :math:`x` axis. If provided,
+                the vectori of quadrature operators :math:`\mathbf{r}` is first rotated
+                by angle :math:`\phi` in the phase space.
+
+        Keyword args:
+            worksize (int): For Fock backends, the additional Fock basis truncation used to
+                perform the calculation. The calculation will take place within a Fock basis
+                dimension of cutoff_dim+worksize. Increasing the worksize will result in
+                increased numerical accuracy, at the expense of a longer computational time.
+                By default, worksize is set to 5.
 
         Returns:
             tuple (float, float): expectation value and variance
@@ -689,6 +741,110 @@ class BaseFockState(BaseState):
 
         return mean, var
 
+    def poly_quad_expectation(self, A, d=None, k=0, phi=0, worksize=5, **kwargs):
+        if A.shape != (2*self._modes, 2*self._modes):
+            raise ValueError("Matrix of quadratic coefficients A must be of size 2Nx2N.")
+
+        if d is None:
+            d = np.zeros([2*self._modes])
+
+        # d[self._modes:] = -d[self._modes:]
+
+        if d.shape != (2*self._modes,):
+            raise ValueError("Vector of linear coefficients d must be of length 2N.")
+
+        # expand the cutoff dimension in approximating the x and p
+        # operators in the Fock basis, to reduce numerical inaccuracy.
+        worksize = kwargs.get('worksize', 5)
+        dim = self._cutoff + worksize
+
+        # construct the x and p operators
+        a = np.diag(np.sqrt(np.arange(1, dim)), 1)
+        x = np.sqrt(self._hbar/2) * (a + a.T)
+        p = -1j * np.sqrt(self._hbar/2) * (a - a.T)
+
+        if phi != 0:
+            # rotate the quadrature operators
+            xold, pold = x, p
+            x = np.cos(phi)*xold + np.sin(phi)*pold
+            p = -np.sin(phi)*xold + np.cos(phi)*pold
+
+        def expand_dims(op, n, modes):
+            """Expand quadrature operator to act on nth mode"""
+            I = np.identity(dim)
+            allowed_indices = zip(indices[:2*modes:2], indices[1:2*modes:2])
+            ind = ','.join(a+b for a, b in allowed_indices)
+            ops = [I]*n + [op] + [I]*(modes-n-1)
+            # the einsum 'ij,kl,mn->ijklmn' (for 3 modes)
+            return np.einsum(ind, *ops)
+
+        # determine modes with quadratic expectation values
+        nonzero = np.concatenate([np.mod(A.nonzero()[0], self._modes), np.mod(d.nonzero()[0], self._modes)])
+        ex_modes = list(set(nonzero))
+        num_modes = len(ex_modes)
+
+        if not ex_modes:
+            # only a constant term was provided
+            return k, 0.
+
+        # There are non-zero elements of A and/or d
+        # therefore there are quadratic and/or linear terms.
+        # find the reduced density matrix
+        rho = self.reduced_dm(modes=ex_modes)
+
+        # generate vector of quadrature operators
+        # this array will have shape [2*num_modes] + [dim]*(2*num_modes)
+        r = np.empty([2*num_modes] + [dim]*(2*num_modes), dtype=np.complex128)
+        for n in range(num_modes):
+            r[n] = expand_dims(x, n, num_modes)
+            r[num_modes+n] = expand_dims(-p, n, num_modes)
+
+        # reduce the size of A so that we only consider modes
+        # which we need to calculate the expectation value for
+        rows = ex_modes + [i+self._modes for i in ex_modes]
+        A = A[:, rows][rows]/2
+
+        # compute the polynomial
+        # For 3 modes, this gives the einsum (with brackets denoting modes):
+        # 'a(bc)(de)(fg),a(ch)(ei)(gj)->(bh)(di)(fj)' applied to r, A@r
+        ind1 = indices[:2*num_modes+1]
+        ind2 = ind1[0] + ''.join([str(i)+str(j) for i, j in zip(ind1[2::2], indices[2*num_modes+1:3*num_modes+1])])
+        ind3 = ''.join([str(i)+str(j) for i, j in zip(ind1[1::2], ind2[2::2])])
+        ind = "{},{}->{}".format(ind1, ind2, ind3)
+
+        # Einsum above applied to to r,Ar
+        # This einsum sums over all quadrature operators, and also applies matrix
+        # multiplication between the same mode of each operator
+        poly_op = np.einsum(ind, r, np.tensordot(A, r, axes=1))
+
+        # add linear term
+        poly_op += r.T @ d[rows]
+
+        # add constant term
+        if k != 0:
+            poly_op += k*expand_dims(np.eye(dim), 0, num_modes)
+
+        # truncate down
+        sl = [slice(0, self._cutoff)]*(2*num_modes)
+        poly_op = poly_op[sl]
+
+        # calculate Op^2
+        ind = "{},{}->{}".format(ind1[1:], ind2[1:], ind3)
+        poly_op_sq = np.einsum(ind, poly_op, poly_op)[sl]
+
+        ind1 = ind1[:-1]
+        ind2 = ''.join([str(j)+str(i) for i, j in zip(ind1[::2], ind1[1::2])])
+        ind = "{},{}".format(ind1, ind2)
+
+        # calculate expectation value, Tr(Op @ rho)
+        # For 3 modes, this gives the einsum '(ab)(cd)(ef),(ba)(dc)(fe)->'
+        mean = np.einsum(ind, poly_op, rho).real
+
+        # calculate variance Tr(Op^2 @ rho) - Tr(Op @ rho)^2
+        var = np.einsum(ind, poly_op_sq, rho).real - mean**2
+
+        return mean, var
+
 
 class BaseGaussianState(BaseState):
     r"""Class for the representation of quantum states using the Gaussian formalism.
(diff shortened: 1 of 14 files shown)


strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+        An arbitrary 2nd order polynomial of quadrature operators over $N$ modes can always
+        be written in the following form:
+
+        .. math:: P(\mathbf{r}) = \frac{1}{2}\mathbf{r}^T A\mathbf{r} + \mathbf{r}^T \mathbf{d} + k I
+
+        where:
+
+        * :math:`A\in\mathbb{R}^{2N\times 2N}` is a symmetric matrix
+          representing the quadratic coefficients,
+        * :math:`\mathbf{d}\in\mathbb{R}^{2N}` is a real vector representing
```
co9olguy: In the PR overview, it says that d is a 2N-dimensional complex vector, not real. Just to confirm, it should be real?

strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+          the linear coefficients,
+        * :math:`k\in\mathbb{R}` represents the constant term, and
+        * :math:`\mathbf{r} = (\x_1,\dots,\x_N,\p_1,\dots,\p_N)` is the vector
+          of quadrature operators in :math:`xp`-ordering.
```
co9olguy: Just curious, what would it be called if we used the ordering (x1,p1,x2,p2,...)?

strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+
+        This method returns the expectation value of this second-order polynomial,
+
+        .. math:: \langle P(\mathbf{r})\rangle,
+
+        as well as the variance
+
+        .. math:: \Delta P(\mathbf{r})^2 = \langle P(\mathbf{r})^2\rangle - \braket{P(\mathbf{r})}^2
```
co9olguy: not sure how this renders, but how come we are mixing `\langle\rangle` and `\braket`?

strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+
+        Args:
+            A (array): a real symmetric 2Nx2N NumPy array, representing the quadratic
+                coefficients of the second order quadrature polynomial.
+            d (array): a real symmetric length-2N NumPy array, representing the linear
```
co9olguy: symmetric?

strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+                coefficients of the second order quadrature polynomial. Default the zero vector.
```
co9olguy: typo: 'Default is the zero vector' or 'Defaults to the zero vector'

strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+            k (float): the constant term. Default 0.
+            phi (float): quadrature angle, clockwise from the positive :math:`x` axis. If provided,
+                the vectori of quadrature operators :math:`\mathbf{r}` is first rotated
```
co9olguy: typo: 'vectori'

strawberryfields/backends/states.py
```diff
@@ -359,6 +362,56 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+                by angle :math:`\phi` in the phase space.
+
+        Keyword args:
+            worksize (int): For Fock backends, the additional Fock basis truncation used to
```
co9olguy: I think we can fix the worksize, no? For instance, truncation of the x^2 operator causes only the last matrix element to be inaccurate. I think @nquesada can probably tell you hard bounds on the maximal worksize needed

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
 
+    def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
+        if A is None:
+            A = np.zeros([2*self._modes, 2*self._modes])
+
+        if A.shape != (2*self._modes, 2*self._modes):
+            raise ValueError("Matrix of quadratic coefficients A must be of size 2Nx2N.")
+
+        if np.any(A.T != A):
+            raise ValueError("Matrix of quadratic coefficients A must be symmetric.")
```
co9olguy: We should probably use `isclose` here instead, because these matrices may come downstream from some other numerical computation and may not be *exactly* symmetric

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+
+        if d is None:
+            linear_coeff = np.zeros([2*self._modes])
+        else:
+            linear_coeff = d.copy()
+            linear_coeff[self._modes:] = -d[self._modes:]
```
co9olguy: Couldn't understand what this line is for

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+        if linear_coeff.shape != (2*self._modes,):
+            raise ValueError("Vector of linear coefficients d must be of length 2N.")
+
+        # expand the cutoff dimension in approximating the x and p
+        # operators in the Fock basis, to reduce numerical inaccuracy.
+        worksize = kwargs.get('worksize', 5)
+        dim = self._cutoff + worksize
+
+        # construct the x and p operators
+        a = np.diag(np.sqrt(np.arange(1, dim)), 1)
```
co9olguy: can we implement these as sparse matrices? Might give significant savings

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+        # this array will have shape [2*num_modes] + [dim]*(2*num_modes)
+        r = np.empty([2*num_modes] + [dim]*(2*num_modes), dtype=np.complex128)
+        for n in range(num_modes):
+            r[n] = expand_dims(x, n, num_modes)
+            r[num_modes+n] = expand_dims(p, n, num_modes)
+
+        # reduce the size of A so that we only consider modes
+        # which we need to calculate the expectation value for
+        rows = ex_modes + [i+self._modes for i in ex_modes]
+        quad_coeffs = A[:, rows][rows]/2
```
co9olguy: what is the advantage of putting the 1/2 in the convention for the quadratic form? This forces users to compensate for it in A, which they might forget to do...

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+        quad_coeffs[num_modes:, :num_modes] = -quad_coeffs[num_modes:, :num_modes]
+        quad_coeffs[:num_modes, num_modes:] = -quad_coeffs[:num_modes, num_modes:]
```
co9olguy: See question above. What is the role of these lines? Why do we make these particular elements of the matrix negative?

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+
+        # compute the polynomial
+        # For 3 modes, this gives the einsum (with brackets denoting modes):
+        # 'a(bc)(de)(fg),a(ch)(ei)(gj)->(bh)(di)(fj)' applied to r, A@r
+        ind1 = indices[:2*num_modes+1]
+        ind2 = ind1[0] + ''.join([str(i)+str(j) for i, j in zip(ind1[2::2], indices[2*num_modes+1:3*num_modes+1])])
+        ind3 = ''.join([str(i)+str(j) for i, j in zip(ind1[1::2], ind2[2::2])])
+        ind = "{},{}->{}".format(ind1, ind2, ind3)
+
+        if np.all(quad_coeffs == 0.):
```
co9olguy: Should we use `isclose` here as well?

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+        # reduce the size of A so that we only consider modes
+        # which we need to calculate the expectation value for
+        rows = ex_modes + [i+self._modes for i in ex_modes]
+        quad_coeffs = A[:, rows][rows]/2
+        quad_coeffs[num_modes:, :num_modes] = -quad_coeffs[num_modes:, :num_modes]
+        quad_coeffs[:num_modes, num_modes:] = -quad_coeffs[:num_modes, num_modes:]
+
+        # compute the polynomial
+        # For 3 modes, this gives the einsum (with brackets denoting modes):
+        # 'a(bc)(de)(fg),a(ch)(ei)(gj)->(bh)(di)(fj)' applied to r, A@r
```
co9olguy: Just for my own understanding, what does index `a` correspond to?

strawberryfields/backends/states.py
```diff
@@ -917,10 +1090,55 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
 
+
+    def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
+        if A is None:
+            A = np.zeros([2*self._modes, 2*self._modes])
+
+        if A.shape != (2*self._modes, 2*self._modes):
+            raise ValueError("Matrix of quadratic coefficients A must be of size 2Nx2N.")
+
+        if np.any(A.T != A):
```
co9olguy: `isclose`

strawberryfields/backends/states.py
```diff
@@ -917,10 +1090,55 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+        # E[P(r)]_(mu,cov) = E(Q(r+mu)]_(0,cov)
+        #                  = E[rT.A.r + rT.(2A.mu+d) + (muT.A.mu+muT.d+cI)]_(0,cov)
+        #                  = E[rT.A.r + rT.d' + k']_(0,cov)
+        d2 = 2*H @ mu + d
+        k2 = mu.T @ H @ mu + mu.T @ d + k
+
+        # expectation value E[P(r)] = tr(A.cov) + muT.A.mu + muT.d + k
+        mean = np.trace(H @ cov) + k2
+        # variance Var[P(r)] = 2tr(A.cov.A.cov) + 4*muT.A.cov.A.mu + dT.cov.d
+        var = 2*np.trace(H @ cov @ H @ cov) + d2.T @ cov @ d2
```
co9olguy: equations in comments don't obviously appear to match implementation

strawberryfields/backends/tfbackend/states.py
```diff
@@ -501,14 +501,20 @@ def mean_photon(self, mode, **kwargs):
             flat_rho = tf.reshape(rho, [-1, self.cutoff_dim ** 2])
 
             nbar = tf.real(tf.reduce_sum(flat_rho * flat_n, axis=1)) # implements a batched tr(rho * n)
+            nbarSq = tf.real(tf.reduce_sum(flat_rho * flat_n**2, axis=1)) # implements a batched tr(rho * n)
```
co9olguy: comment should have `n**2`

strawberryfields/backends/tfbackend/states.py
```diff
@@ -608,6 +614,64 @@ def wigner(self, mode, xvec, pvec):
...
+
+        as well as the variance
+
+        .. math:: \Delta P(\mathbf{r})^2 = \langle P(\mathbf{r})^2\rangle - \braket{P(\mathbf{r})}^2
+
+        Args:
+            A (array): a real symmetric 2Nx2N NumPy array, representing the quadratic
+                coefficients of the second order quadrature polynomial.
+            d (array): a real symmetric length-2N NumPy array, representing the linear
+                coefficients of the second order quadrature polynomial. Default the zero vector.
```
co9olguy: adjust 'Default the zero vector' to whatever we change this sentence to elsewhere

strawberryfields/backends/tfbackend/states.py
```diff
@@ -608,6 +614,64 @@ def wigner(self, mode, xvec, pvec):
...
+                perform the calculation. The calculation will take place within a Fock basis
+                dimension of cutoff_dim+worksize. Increasing the worksize will result in
+                increased numerical accuracy, at the expense of a longer computational time.
+                By default, worksize is set to 5.
+
+        Returns:
+            tuple (float, float): expectation value and variance
+        """
+        if self._eval and not self.batched:
+            return super().poly_quad_expectation(A, d, k, phi, **kwargs)
```
co9olguy: This might cause confusion in the cas that `self._eval==True` and `self.batched==True`, but where the user passes `eval=False` or a session within `kwargs` when they call this method. In that case, the class should override the default `self._eval`.

For now, we can catch this by calling checking `ops.check_for_eval(kwargs)`

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
         self.assertAllAlmostEqual(res.flatten(), res_exact.flatten(), delta=self.tol)
 
 
+class PolyQuadExpectationSingleMode(BaseTest):
```
co9olguy: Just noticed this now, but the test classes in `test_states.py` all have names without `Test` at the end, while in all other test files (that I'm aware of), we use the convention `SomeKindOfTest(ParentTest)`

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+    num_subsystems = 3
+    cutoff = 12
```
co9olguy: We should prefer not to have one of the user-adjustable test parameters explicitly bypassed in the tests (at least, not without warning).

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+    cov = R(qphi).T @ squeezed_cov(r, phi, hbar=hbar) @ R(qphi)
+
+    def sample_one_mode_normal_expectations(self, mu, cov, func):
+        """Returns the expectation value E(f) and the variance var(f)
+        for some normal distribution X~N(mu, cov).
+
+        Args:
+            mu (array): means vector
+            cov (array): covariance matrix
+            func (function): function acting on the random variables X, P, XP,
```
co9olguy: only X, P, and XP?

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        self.logTestName()
+        state = self.circuit.state()
+
+        A = np.zeros([6, 6])
+        d = np.zeros([6])
+        k = 0
+
+        mean, var = state.poly_quad_expectation(A, d, k)
+        self.assertEqual(mean, 0)
+        self.assertEqual(var, 0)
```
co9olguy: We probably want to set these to `assertAlmostEqual` if we change the backend implementations to use `np.isclose`

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+
+        # prepare a squeezed displaced state
+        self.circuit.reset(pure=self.kwargs['pure'])
+        self.circuit.prepare_displaced_squeezed_state(a, r, phi, 0)
+
+        state = self.circuit.state()
+        mean, var = state.poly_quad_expectation(A, d, k, phi=self.qphi)
+
+        self.assertAlmostEqual(mean, self.mu[0], delta=self.tol)
+        self.assertAlmostEqual(var, self.cov[0, 0], delta=self.tol)
```
co9olguy: Are these expected results independent of the `hbar` value?

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+
+        # prepare a squeezed displaced state
+        self.circuit.reset(pure=self.kwargs['pure'])
+        self.circuit.prepare_displaced_squeezed_state(a, r, phi, 0)
+
+        state = self.circuit.state()
+        mean, var = state.poly_quad_expectation(A, d, k, phi=self.qphi)
+
+        self.assertAlmostEqual(mean, self.mu[1], delta=self.tol)
+        self.assertAlmostEqual(var, self.cov[1, 1], delta=self.tol)
```
co9olguy: same question about `hbar`

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        k = 0
+
+        # prepare a squeezed displaced state
+        self.circuit.reset(pure=self.kwargs['pure'])
+        self.circuit.prepare_displaced_squeezed_state(a, r, phi, 0)
+
+        state = self.circuit.state()
+        mean, var = state.poly_quad_expectation(A, d, k, phi=self.qphi)
+
+        # E(2x+p) = 2E(x) + E(p)
```
co9olguy: update this comment to use `a` and `b`

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        mean_expected = d[0]*self.mu[0] + d[3]*self.mu[1]
+        self.assertAlmostEqual(mean, mean_expected, delta=self.tol)
+
+        # var(2x+p) = 4var(x)+var(p)+4cov(x,p)
```
co9olguy: same

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        self.assertAlmostEqual(mean, mean_ex, delta=self.tol)
+        self.assertAlmostEqual(var, var_ex, delta=self.tol)
+
+    def test_xp(self):
+        """Test that the correct result is returned for E(xp)"""
+        self.logTestName()
+
+        # set quadratic coefficient
+        A = np.zeros([6, 6])
+        A[3, 0] = A[0, 3] = 1
```
co9olguy: This is more `xp+px` than `xp`, no?

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+
+        if isinstance(self.backend, backends.TFBackend):
+            raise unittest.SkipTest('The poly_quad_expectation method is not yet supported on the TF backend.')
+
+        mu = np.zeros([6])
+        cov = np.zeros([6, 6])
+
+        # construct the vector of means and covariance matrix for each mode
+        # note that we use (x0,p0,x1,p1,...) ordering here.
+        for i, (x, y, z) in enumerate(zip(self.a_list, self.r_list, self.phi_list)):
```
co9olguy: Might help for code clarify to use `i, (a, r, phi)`

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        t = 1/np.sqrt(2)
+        BS = np.array([[ t,  0,  -t,  0],
+                       [ 0,  t,  0,  -t],
+                       [ t,  0,  t,  0],
+                       [ 0,  t,  0,  t]])
+
+        S1 = block_diag(BS, np.identity(2))
+
+        if self.linear_modes == 3:
+            S2 = block_diag(np.identity(2), BS)
```
co9olguy: unused? (same question applies throughout)

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+            d = np.array([x0, x1, x2, p0, 0, 0])
+        elif self.linear_modes == 2:
+            d = np.array([x0, x1, 0, p0, 0, 0])
+
+        k = 0
+
+        self.circuit.reset(pure=self.kwargs['pure'])
+
+        # squeeze and displace each mode
+        for i, (x, y, z) in enumerate(zip(self.a_list, self.r_list, self.phi_list)):
```
co9olguy: see comment above

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        if self.linear_modes == 3:
+            S2 = block_diag(np.identity(2), BS)
+        elif self.linear_modes == 2:
+            S2 = np.identity(6)
+
+        self.mu = S2 @ S1 @ mu
+        self.cov = S2 @ S1 @ cov @ S1.T @ S2.T
+
+    def test_multi_mode_linear_combination(self):
+        """Test that the correct result is returned for E(2x0 + x1 - p0)"""
```
co9olguy: docstring should be changed to reflect actual parameters used

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        # apply a beamsplitter to the modes
+        self.circuit.beamsplitter(1/np.sqrt(2), 1/np.sqrt(2), 0, 1)
+
+        if self.linear_modes == 3:
+            self.circuit.beamsplitter(1/np.sqrt(2), 1/np.sqrt(2), 1, 2)
+
+        state = self.circuit.state()
+        mean, var = state.poly_quad_expectation(A, d, k, phi=self.qphi)
+
+        # E(2x0 + x1 - p0) = 2E(x0) + E(x1) - E(p0)
```
co9olguy: update comment to reflect parameter values

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        mean_expected = np.array([x0, p0, x1, 0, x2, 0]) @ self.mu
+        self.assertAlmostEqual(mean, mean_expected, delta=self.tol)
+
+        # var(x0 + x1 + p0) = 4var(x0)+var(x1)+var(p0)-4cov(x0,p0) + 4cov(x0, x1) - 2*cov(x1, p0)
```
co9olguy: update

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        state = self.circuit.state()
+        mean, var = state.poly_quad_expectation(A, d, k, phi=self.qphi)
+
+        mean_ex, var_ex = self.sample_one_mode_normal_expectations(
+            self.mu, self.cov, lambda X, P, XP: A[3, 3]*P**2/2)
+        self.assertAlmostEqual(mean, mean_ex, delta=self.tol)
+        self.assertAlmostEqual(var, var_ex, delta=self.tol)
+
+    def test_xp(self):
+        """Test that the correct result is returned for E(xp)"""
```
co9olguy: Since we force the polynomials to be symmetric, we shouldn't be testing non-symmetric operators

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+
+        # var(x0 + x1 + p0) = 4var(x0)+var(x1)+var(p0)-4cov(x0,p0) + 4cov(x0, x1) - 2*cov(x1, p0)
+        var_expected = self.cov[0, 0]*x0**2 + self.cov[1, 1]*p0**2 + self.cov[2, 2]*x1**2 \
+            + self.cov[4, 4]*x2**2 + 2*self.cov[0, 1]*x0*p0 + 2*self.cov[0, 2]*x0*x1 \
+            + 2*self.cov[0, 4]*x0*x2 + 2*self.cov[1, 2]*p0*x1 + 2*self.cov[1, 4]*p0*x2 \
+            + 2*self.cov[2, 4]*x1*x2
+        self.assertAlmostEqual(var, var_expected, delta=self.tol)
+
+    def test_multi_mode_quadratic_combination(self):
+        """Test that the correct result is returned for E(x0x1+x0p1)"""
```
co9olguy: See earlier comment about testing only symmetric operators

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+
+        A = (A + A.T)/2
+
+        d = None
+        k = 0
+
+        self.circuit.reset(pure=self.kwargs['pure'])
+
+        # squeeze and displace each mode
+        for i, (x, y, z) in enumerate(zip(self.a_list, self.r_list, self.phi_list)):
```
co9olguy: see above comment about variable names

strawberryfields/backends/states.py
josh146: Good question, @nquesada? I just copied and pasted this from somewhere else in the SF codebase. I've seen one called the symmetric ordering, and one called the `xp`-ordering.

strawberryfields/backends/states.py
josh146: Ah, I'll fix that. `\braket` is just a macro for `\langle#1\rangle`

strawberryfields/backends/states.py
josh146: This actually really confuses me - for some reason, the einsum Fock calculation doesn't provide the same answers as the Guassian backend/theoretical result _unless_ I first modify `d` and `A` such that

1. The linear coefficients on the momentum operators are negated: `d' = (d_1,...,d_n,-d_{n+1},...,-d_{2n})`, and

2. The block-off diagonal elements of `A` (i.e. so the coefficients of the `x_np_m` terms) are also negated.

The only reason I know that this works is because I had the _exact_ same issue in OpenBoson, and this is the trick that allowed the Fock backends to provide the correct result.

strawberryfields/backends/states.py
```diff
@@ -689,6 +744,124 @@ def quad_expectation(self, mode, phi=0, **kwargs):
...
+
+        # construct the x and p operators
+        a = np.diag(np.sqrt(np.arange(1, dim)), 1)
+        x = np.sqrt(self._hbar/2) * (a + a.T)
+        p = -1j * np.sqrt(self._hbar/2) * (a - a.T)
+
+        if phi != 0:
+            # rotate the quadrature operators
+            xold = x.copy()
+            pold = p.copy()
```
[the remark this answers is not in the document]
josh146: Only because we have simultaneous linear equations in the next two lines, using the same variable name. Since `x` and `p` are NumPy arrays, using temporary variables does not work, as it provides a view to the same array.

strawberryfields/backends/states.py
josh146: For 3 modes (for example), array `r` has dimensions `aijklmn`. `a` corresponds to the index in the vector of quadrature operators `r = (x_1,...,x_n,p_1,...,p_n)`, and the remaining indices `ijklmn` are the elements of the operator acting on a 3 mode density matrix.

josh146: So, in effect, matrix of quadratic coefficients `A` acts only on index `a`, this index is then summed, and then each mode of `r`, `A@r` undergoes matrix multiplication

strawberryfields/backends/tfbackend/states.py
josh146: :+1: Do you want to make this change? It will probably also need to be changed in the `TFFockState.wigner` method.

tests/test_states.py
josh146: I originally tried that; unfortunately, a cutoff of 5 (the default) causes the expectation of the quadratic terms to have too much error, causing the assertions to fail

tests/test_states.py
josh146: Because you can get X^2 from X, etc. :laughing: Maybe I was a bit lazy there

tests/test_states.py
josh146: Yep, self.mu and self.cov are calculated taking hbar into account

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+        c3 = 1/3.
+        c4 = -1
+
+        # define the arbitrary quadratic
+        A = np.zeros([6, 6])
+        A[0, 0] = 2*c0
+        A[3, 3] = 2*c1
+
+        A[3, 0] = c2
+        A[0, 3] = c2
```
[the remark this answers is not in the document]
josh146: The factor of 1/2 only need to be accounted for on the diagonal of A, as the off-diagonals are counted twice by default

tests/test_states.py
```diff
@@ -220,6 +247,394 @@ def test_squeezed_coherent(self):
...
+    hbar = 2
+    qphi = 0.78
+
+    # testing parameters
+    linear_modes = 2
+
+    # define circuit parameters
+    a_list = [0.344+0.123j, 0.432+0.123j, 0.1231-0.543j][:linear_modes]
+    r_list = [0.1065, 0.32, 0.4123][:linear_modes]
+    phi_list = [0.897, 0.31, 0.9643][:linear_modes]
```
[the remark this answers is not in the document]
josh146: That's correct; the `linear_modes=3` case I added just as a sanity check, to ensure that this works locally. The only reason I didn't include it as a separate test is that it takes quite long to calculate on the Fock backend, approximately 30s.

tests/test_states.py
josh146: I originally had that, unfortunately the linter complains that we are modifying global variables

tests/test_states.py
josh146: See https://github.com/XanaduAI/strawberryfields/pull/34#discussion_r221158250

tests/test_states.py
josh146: see https://github.com/XanaduAI/strawberryfields/pull/34#discussion_r221158339

tests/test_states.py
josh146: Bit of lazy writing on my half. The expectation value of the polynomial xp in the phase space is identical/equivalent to the expectation value of the operator (xp+px) after Weyl quantization

tests/test_states.py
josh146: Same as above; in this case, E(x0x1+x0p1) is the polynomial in the phase space, not the quantized operators

tests/test_states.py
josh146: See https://github.com/XanaduAI/strawberryfields/pull/34#discussion_r221158339

strawberryfields/backends/states.py
josh146: Yes I guess it should be, otherwise the observable would be non-Hermitian

strawberryfields/backends/states.py
josh146: What are your thoughts on changing it?

strawberryfields/backends/states.py
co9olguy: Perhaps @nquesada knows why there is a discrepency between backends? Is this something that the SF frontend compensates for?

strawberryfields/backends/states.py
co9olguy: Let's remember to do it before bumping the release number

strawberryfields/backends/states.py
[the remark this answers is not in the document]
co9olguy: What if we just use different names for the transformed operators? Then we wouldn't need to copy, right?

```
x_ = np.sqrt(...)
p_ = -1j * ...

if phi != 0:
    x = np.cos(phi) * x - np.sin(phi) * p
    p = np.sin(phi) * x + np.sin(phi) * p
else:
    x = x_
    p = p_

File: strawberryfields/backends/states.py (hunk 11 above)
Comment (co9olguy) [reply strawberryfields/backends/states.py]: I'd prefer to not have the 1/2 in the convention. This would likely cause less surprises

File: strawberryfields/backends/states.py (hunk 14 above)
Comment (co9olguy) [reply strawberryfields/backends/states.py]: Makes sense. Maybe add this short explanation to the code comment directly?

File: tests/test_states.py (hunk 40 above)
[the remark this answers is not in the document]
Comment (co9olguy) [reply tests/test_states.py]: k then let's remove from tests

File: tests/test_states.py (hunk 29 above)
Comment (co9olguy) [reply tests/test_states.py]: `(a_, r_, phi_)`?

File: strawberryfields/backends/tfbackend/states.py (hunk 19 above)
Comment (co9olguy) [reply strawberryfields/backends/tfbackend/states.py]: Will do. 
The `wigner` method is ok, since it explicitly states both in the documentation and an exception that it only supports that case

File: strawberryfields/backends/tfbackend/states.py [hunk 41]
```diff
@@ -550,7 +556,7 @@ def dm(self, **kwargs):
                 * If eval is not present in kwargs, then state falls back to the an internal evaluation behaviour,
                   which is specified at initialization.
                 * A Tensorflow Session or feed_dict may also be passed via the keys ``session`` or ``feed_dict``, respectively.
-                  These are ignored when ``eval=False``; when ``eval=True``, they are used when numerically evaluating the underlying quantity.
+                  If a Session is supplied, then `eval` is overriden and the numerical evaluation takes place in the provided Session.
```
josh146: Should be ``eval``

josh146:
Update: added additional `mean_photon` tests (as there were bugs in the previous implementation that weren't picked up from the tests), and added an additional `poly_quad_expectation` test to ensure that `E(n)` and `var(n)` match that of `mean_photon`.

strawberryfields/backends/states.py
josh146: @nquesada:
> worksize of cutoff+1 should be fine

strawberryfields/backends/states.py
josh146: @nquesada didn't know either

tests/test_states.py
josh146: Let's leave this for a future PR, and perhaps make an issue for it?

tests/test_states.py
co9olguy: We should either change the parameters of the test to fit within that cutoff, or add a warning msg inside the test that the user-provided value is overridden (if it is lower than 12)

tests/test_states.py
co9olguy: This was quick, just took care of it

tests/test_states.py
co9olguy: I added a logger warning to the relevant tests

co9olguy:
Not sure why the codecov stuff failed, but tests are passing, so I'm now merging

PR #36, opened 2018-10-11
Allow turning circuits into operators

jleni:
I still need to add a couple of unit tests, etc. and but I thought it made sense to submit an early PR to check if the approach meets your expectations.

You can check `examples/custom_operator.py` to see how the new feature is supposed to be used.

### DIFF
--- a/examples/custom_operator.py
+++ b/examples/custom_operator.py
@@ -15,10 +15,13 @@ from strawberryfields.ops import *
 from strawberryfields.utils import operator
 
 
-@operator(ns=4)
-def prepare_op(q):
-    """
-    :param q: This operator is applied to 4 registers
+@operator(4)
+def prepare_squeezing(q):
+    """This operator prepares modes 0 to 4
+    as squeezed states with r = -1.
+
+    Args:
+        q (register): the qumode register.
     """
     S = Sgate(-1)
     S | q[0]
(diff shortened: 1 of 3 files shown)


examples/custom_operator.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    S = Sgate(-1)
+    S | q[0]
+    S | q[1]
+    S | q[2]
+    S | q[3]
+
+
+@operator(ns=3)
+def circuit_op(v1, v2, q):
```
josh146: Nice that it works with parameters! Does the operator decorator always require the qumode register `q` as the last argument?

jleni: It works with parameters and also for class methods, like the example in the unit test: https://github.com/XanaduAI/strawberryfields/pull/36/files#diff-c487e7ec5c066aeaecbe95c769fdccaaR31

Yes, the last argument are always the register(s).

josh146:
Hi Juan, thanks for your contribution! This is a really high quality PR - it's well linted, and the unittests are coming along well. Let me know when you have finished, and I'll do a short code review before merging.

jleni:
Thanks! :) I would say the main functionality is there and it could be reviewed.
I am not sure if you would like to document the feature in this very same PR. But maybe it is a good idea that you review the code/approach before that.

josh146:
Hmm, I'm not sure why Travis/Codecov didn't pick up this pull request. Let me see if I can trigger a build manually, and then I'll merge

jleni:
Yes, I noticed that too. In my experience, circleci behaves a lot better than travis

josh146:
Very odd. I have downloaded it locally, tested it, and checked the coverage, so I am happy to perform the merge now.

PR #37, opened 2018-12-13
Slight improvement of Interferometer and related code in decompositions

cgogolin:

* do clements decompose lazily
* fixed seed to make decompositions unit test deterministic
* fixed docstring of Interferometer

**Benefits:**

* saves some memory and computational time
* decompositions unit test no longer can fail randomly

**Possible Drawbacks:**

none

**Related GitHub Issues:**

none

### DIFF
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -131,7 +131,7 @@ def clements(V, tol=11):
 
     Args:
         V (array): Unitary matrix of size n_size
-        tol (int): the number of decimal places to use when determining if the matrix is symmetric
+        tol (int): the number of decimal places to use when determining whether the matrix is unitary
 
     Returns:
         tuple[array]: returns a tuple of the form ``(tilist,tlist,np.diag(localV))``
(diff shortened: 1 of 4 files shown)


josh146:
I just did a quick read through, and I'm happy to merge this PR. Just a quick question regarding the lazy evaluation beforehand, though.

The original reason behind doing the eager evaluation of the Williamson/Bloch-Messiah/Clements decompositions is that:

1. These decompositions are very sensitive to the matrix argument used, due to the large number of validation checks (i.e. checking if the matrix is unitary, symplectic, etc.). Performing an eager evaluation let's the user know that their is an issue with their input matrix at circuit construction time, rather than waiting for potentially an intensive simulation to fail halfway due to input validation.

2. The amount of computation time needed for calculating the decomposition is generally insignificant compared to the amount of computation time to implement it.

Since matrix input validation is an issue that trips up a lot of people (it can be as simple as passing a covariance matrix with a different value of hbar to the simulation, or a matrix that is unitary to 6 significant figures instead of 11), I would prefer to leave this in circuit construction.

The options are then as follows:

1. Move the input validations into `ops.py`. However, the reason that they are in `decompositions.py` is that they are used in numerous places - moving them to `ops.py` would require significant code duplication.

2. Keep the eager evaluation (a question of whether the computation/memory is a significant issue or not).

cgogolin:
These are very good reasons for the eager decomposition. I reverted the that change. Thanks!

PR #42, opened 2019-01-17
Feature/graphembedding

nquesada:
Adds GraphEmbed in the frond end and graph_embed in the decompositions module to embed an arbitrary (in general complex valued) adjacency matrix of a graph.

**Benefits:**
Simplifies a number of simulation when the user want embed a graph into a GBS setup

**Possible Drawbacks:**
None
**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -21,7 +21,7 @@ from scipy.linalg import block_diag, sqrtm, polar, schur
 from .backends.shared_ops import sympmat, changebasis
 
 
-def takagi(N, tol=13):
+def takagi(N, tol=1e-13):
     r"""Computes the Autonne-Takagi decomposition of a complex symmetric (not Hermitian!) matrix.
 
     Note that singular values of N are considered equal if they are equal after np.round(values, tol).
@@ -39,7 +39,7 @@ def takagi(N, tol=13):
     (n, m) = N.shape
     if n != m:
         raise ValueError("The input matrix must be square")
-    if np.round(np.linalg.norm(N-np.transpose(N)), tol) != 0:
+    if np.linalg.norm(N-np.transpose(N)) >= tol:
         raise ValueError("The input matrix is not symmetric")
 
     v, l, ws = np.linalg.svd(N)
@@ -78,7 +78,8 @@ def takagi(N, tol=13):
 
 
 def graph_embed(mat, max_mean_photon=1.0, make_traceless=True, tol=1e-6):
-    r""" Given an symmetric adjacency matrix (that can be complex in general),
+    r""" Given an symmetric adjacency matrix (in general, with arbitrary complex off-diagonal and
+    real diagonal entries),
     it returns the squeezing parameters and interferometer necessary for
     creating the Gaussian state whose off-diagonal parts are proportional to that matrix.
 
@@ -331,7 +332,7 @@ def bloch_messiah(S, tol=1e-10):
     return ut1, st1, v1
 
 
-def covmat_to_hamil(V, tol=10): # pragma: no cover
+def covmat_to_hamil(V, tol=1e-10): # pragma: no cover
     r"""Converts a covariance matrix to a Hamiltonian.
 
     Given a covariance matrix V of a Gaussian state :math:`\rho` in the xp ordering,
@@ -354,8 +355,8 @@ def covmat_to_hamil(V, tol=10): # pragma: no cover
     (n, m) = V.shape
     if n != m:
         raise ValueError("Input matrix must be square")
-    if np.round(np.linalg.norm(V-V.T), tol) != 0:
-        raise ValueError("Input matrix must be symmetric")
+    if np.linalg.norm(V-np.transpose(V)) >= tol:
+        raise ValueError("The input matrix is not symmetric")
 
     n = n//2
     omega = sympmat(n)
@@ -372,7 +373,7 @@ def covmat_to_hamil(V, tol=10): # pragma: no cover
     return H
 
 
-def hamil_to_covmat(H, tol=10): # pragma: no cover
+def hamil_to_covmat(H, tol=1e-10): # pragma: no cover
     r"""Converts a Hamiltonian matrix to a covariance matrix.
 
     Given a Hamiltonian matrix of a Gaussian state H, finds the equivalent covariance matrix
@@ -390,8 +391,8 @@ def hamil_to_covmat(H, tol=10): # pragma: no cover
     (n, m) = H.shape
     if n != m:
         raise ValueError("Input matrix must be square")
-    if np.round(np.linalg.norm(H-H.T), tol) != 0:
-        raise ValueError("Input matrix must be symmetric")
+    if np.linalg.norm(H-np.transpose(H)) >= tol:
+        raise ValueError("The input matrix is not symmetric")
 
     vals = np.linalg.eigvalsh(H)
     for val in vals:
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -80,7 +80,7 @@ def takagi(N, tol=13):
 def graph_embed(mat, max_mean_photon=1.0, make_traceless=True, tol=1e-6):
     r""" Given an symmetric adjacency matrix (that can be complex in general),
     it returns the squeezing parameters and interferometer necessary for
-    implementing it in GBS.
+    creating the Gaussian state whose off-diagonal parts are proportional to that matrix.
 
     Args:
         mat (array): square symmetric complex (or real or integer) array representing a (weighted) adjacency matrix of a graph
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -81,26 +81,28 @@ def graph_embed(mat, max_mean_photon=1.0, make_traceless=True, tol=6):
     r""" Given an symmetric adjacency matrix (that can be complex in general),
     it returns the squeezing parameter and interferometer necessary for
     implementing it in GBS.
+
     Args:
         mat (array): square symmetric complex (or real or integer) array
-        max_mean_photon (float): therhold value. I guarantees that the mode with
-            the largest squeezing has max_men_photon as mean photon number.
-            i.e. sinh(r_max)^2 == max_mean_photon
-        make_tracelss (boolean): removes the trace of the input matrix.
+        max_mean_photon (float): threshold value. It guarantees that the mode with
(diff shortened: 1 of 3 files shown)


tests/test_decompositions.py
```diff
@@ -49,6 +49,18 @@ def test_takagi_random_symm(self):
...
+    def test_graph_embed_random_symm(self):
+        self.logTestName()
+        error=np.empty(nsamples)
+        max_mean_photon = 2
+        for i in range(nsamples):
+            X=random_degenerate_symmetric()
+            sc, U = dec.graph_embed(X, max_mean_photon=max_mean_photon)
+            error[i] = np.sinh(np.max(np.abs(sc)))**2 - max_mean_photon
+
+        self.assertAlmostEqual(error.mean() , 0)
```
co9olguy: These tests seem a bit puzzling, and possibly strongly tied to the particular random seed.
What is the `sampled` part here? Is `graph_embed` stochastic?

tests/test_decompositions.py
```diff
@@ -210,6 +227,35 @@ def test_gaussian_transform(self):
         state = self.eng.run()
         self.assertAllAlmostEqual(state.cov(), self.S@self.S.T*self.hbar/2, delta=self.tol)
 
+    def test_graph_embed(self):
+        """Test that embedding a traceless adjacency matrix A
+        reults in the property Amat/A = c J, where c is a real constant,
```
co9olguy:
```suggestion
        results in the property Amat/A = c J, where c is a real constant,
```

tests/test_decompositions.py
```diff
@@ -210,6 +227,35 @@ def test_gaussian_transform(self):
...
+        self.eng.reset()
+        q = self.eng.register
+
+        with self.eng:
+            GraphEmbed(self.A) | q
+
+        state = self.eng.run()
+        Amat = self.eng.backend.circuit.Amat()
+
+        ratio = np.real_if_close(Amat[self.num_subsystems:, self.num_subsystems:]/self.A)
```
co9olguy: Question: why do we index only the bottom-right block?

strawberryfields/ops.py
```diff
@@ -1576,6 +1577,41 @@ def decompose(self, reg):
...
+
+    This operation uses the Takagi decomposition to decompose
+    a adjacency matrix into a sequence of squeezers and beamsplitters and
+    rotation gates.
+
+    Args:
+        A (array): an :math:`N\times N` complex or real symmetric matrix.
+    """
+    ns = None
+    def __init__(self, A, max_mean_photon=1.0, make_traceless=True, tol=6):
```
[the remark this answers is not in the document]
josh146: Nope, 6 is correct. This is because the tolerance is tested using the `np.round()` function, which accepts the number of decimal places to round to (6 in this case). Basically equivalent to 1e-6 precision.

strawberryfields/ops.py
```diff
@@ -1576,6 +1577,41 @@ def decompose(self, reg):
...
+            self.sq, self.U = graph_embed(A, max_mean_photon=max_mean_photon, make_traceless=make_traceless, tol=tol)
+            self.ns = self.U.shape[0]
+
+    def decompose(self, reg):
+        cmds = []
+
+        if not self.identity:
+            for n, s in enumerate(self.sq):
+                if np.round(s, 13) != 0:
+                    cmds.append(Command(Sgate(s), reg[n], decomp=True))
```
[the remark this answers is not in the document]
josh146: Exactly. This allows the frontend to distinguish between gates that came about via decompositions, and those that didn't, allowing for the user to print out both the gates they queued, vs the gates that were applied.

tests/test_decompositions.py
josh146: The matrix Amat is a 2Nx2N block diagonal matrix, by construction composed of [[B^, 0], [0, B]]. So it is sufficient to only consider the bottom right NxN block.

strawberryfields/ops.py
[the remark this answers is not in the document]
co9olguy: ok, then this works differently than basically every other tolerance in SF. To avoid confusion, can we either restructure to not use `np.round` or rename to something like `tol_decimal_places`?

tests/test_decompositions.py
co9olguy: then it shouldn't hurt to add another some other asserts for the other 3 blocks to make sure that the `by construction` steps worked as intended

strawberryfields/decompositions.py
```diff
@@ -77,29 +77,29 @@ def takagi(N, tol=13):
     return rl, U
 
 
-def graph_embed(mat, max_mean_photon=1.0, make_traceless=True, tol=6):
+def graph_embed(mat, max_mean_photon=1.0, make_traceless=True, tol=1e-6):
     r""" Given an symmetric adjacency matrix (that can be complex in general),
-    it returns the squeezing parameter and interferometer necessary for
+    it returns the squeezing parameters and interferometer necessary for
```
co9olguy: Just below: we should change "for implementing it in GBS" to something more generic like "for creating the Gaussian state whose off-diagonal parts are proportional to that matrix"

strawberryfields/decompositions.py
```diff
@@ -77,6 +77,40 @@ def takagi(N, tol=13):
...
+    r""" Given an symmetric adjacency matrix (that can be complex in general),
+    it returns the squeezing parameter and interferometer necessary for
+    implementing it in GBS.
+
+    Args:
+        mat (array): square symmetric complex (or real or integer) array
+        max_mean_photon (float): threshold value. It guarantees that the mode with
+            the largest squeezing has ``max_mean_photon`` as the mean photon number
+            i.e. :math:`sinh(r_{max})^2 == max_mean_photon`
+        make_traceless (boolean): removes the trace of the input matrix.
```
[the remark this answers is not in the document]
co9olguy: Note: when this is clarified, the same thing appears in strawberryfields/ops.py

tests/test_decompositions.py
```diff
@@ -240,9 +240,17 @@ def test_graph_embed(self):
 
         state = self.eng.run()
         Amat = self.eng.backend.circuit.Amat()
+        N = self.num_subsystems
 
-        ratio = np.real_if_close(Amat[self.num_subsystems:, self.num_subsystems:]/self.A)
-        self.assertAllAlmostEqual(ratio, np.ones([self.num_subsystems, self.num_subsystems])*ratio[0, 0], delta=self.tol)
+        # check that the matrix Amat is constructed to be of the form
+        # Amat = [[B^, 0], [0, B]]
```
co9olguy: github won't let me suggest fix, but `B^` seems to be missing a dagger

co9olguy:
Hey guys, trying to clear the backlog of PRs. What's the status on this one? I can see there's still a few small points to fix up.

josh146:
From my point of view this one is all ready to go as soon as @nquesada has made the suggested changes in decompositions.py

strawberryfields/decompositions.py
```diff
@@ -77,6 +77,40 @@ def takagi(N, tol=13):
     return rl, U
 
 
+def graph_embed(mat, max_mean_photon=1.0, make_traceless=True, tol=6):
+    r""" Given an symmetric adjacency matrix (that can be complex in general),
+    it returns the squeezing parameter and interferometer necessary for
+    implementing it in GBS.
+
+    Args:
+        mat (array): square symmetric complex (or real or integer) array
```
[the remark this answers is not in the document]
josh146: @nquesada: Yes it works for weighted graphs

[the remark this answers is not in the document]
josh146:
```suggestion
        mat (array): square symmetric complex (or real or integer) array representing a (weighted) adjacency matrix of a graph
```

strawberryfields/decompositions.py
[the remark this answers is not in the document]
josh146:
```suggestion
        make_traceless (boolean): removes the trace of the input matrix, by performing the transformation
            :math:`\tilde{A} = A-\mathrm{tr}(A) \I/n`. This may reduce the amount of squeezing needed to encode
            the graph.
```

tests/test_decompositions.py
```diff
@@ -49,6 +49,18 @@ def test_takagi_random_symm(self):
 
         self.assertAlmostEqual(error.mean() , 0)
 
+    def test_graph_embed_random_symm(self):
```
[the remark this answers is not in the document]
josh146:
```suggestion
    def test_takagi_fixed_random_symm(self):
```

tests/test_decompositions.py
```diff
@@ -49,6 +49,18 @@ def test_takagi_random_symm(self):
...
+        self.logTestName()
+        error=np.empty(nsamples)
+        max_mean_photon = 2
+        for i in range(nsamples):
+            X=random_degenerate_symmetric()
```
josh146:
```suggestion
            X = random_degenerate_symmetric()
```

tests/test_decompositions.py
```diff
@@ -49,6 +49,18 @@ def test_takagi_random_symm(self):
 
         self.assertAlmostEqual(error.mean() , 0)
 
+    def test_graph_embed_random_symm(self):
+        self.logTestName()
```
josh146:
```suggestion
        """This test verifies that the maximum amount of squeezing used to encode the graph is indeed capped by the parameter max_mean_photon"""
        self.logTestName()
```

tests/test_decompositions.py
josh146: This test verifies that the maximum amount of squeezing used to encode the graph is indeed capped by the parameter max_mean_photon (the docstring has been updated to reflect this). There is only randomness in that we are checking this for random adjacency matrices - since we are looping over an ensemble and only asserting that the mean is equal to 0, the test is ~deterministic (using the same approach used across the SF test suite)

josh146:
All remaining comments addressed in the latest commit :slightly_smiling_face:

strawberryfields/decompositions.py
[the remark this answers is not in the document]
co9olguy: So the matrix `mat` is the one that is made traceless? 
If so, it might confuse users that we earlier call it an `adjacency matrix` (usually this means the diagonal entries are already zero)

tests/test_decompositions.py
co9olguy: could this test fail if the sample mean is, by chance, not within tolerance of zero?

josh146: Updated this to add the argument `delta=self.tol`, so that it will use the default SF test suite tolerance.

PR #44, opened 2019-01-24
Better decompositions: Clements et al. (and possibly Reck?)

ishdhand:
Creating this pull request to get the conversation going about better support for linear optics decompositions. 

The first commits add support for the final form of the Clements et al. decomposition, i.e. having the phases at the end (as in Equation 5) rather than in the middle (as in Equation 4, which is the current implementation). This Equation 5 decomposition is more symmetric than the intermediate one, which could make it more robust. This form also makes it easier to implement a tensor-network simulation of linear optics. 

This new code is for now only in an additional function, but could replace the current function. As this is quite central to some of the sf functionality, it's better to be careful. Would love to learn more about what you folks think is the best way to go about this.

I also have code for the Reck et al. decomposition, which I could add if useful, e.g., for a temporal-mode implementation. What do you think?

### DIFF
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -15,7 +15,6 @@
 
 from itertools import groupby
 
-import math as mt
 import numpy as np
 from scipy.linalg import block_diag, sqrtm, polar, schur
 
@@ -216,7 +215,7 @@ def clements(V, tol=1e-11):
 
     return tilist, tlist, np.diag(localV)
 
-def clements_eq5(V, tol=11):
+def clements_phase_end(V, tol=1e-11):
     r"""Clements decomposition of unitary matrix
 
     See Clements et al. Optica 3, 1460 (2016) [10.1364/OPTICA.3.001460]
@@ -247,7 +246,7 @@ def clements_eq5(V, tol=11):
 
         # The new parameters required for D',T' st. T^(-1)D = D'T'
         new_theta = theta
-        new_phi = mt.fmod((alpha - beta + np.pi), 2*np.pi)
+        new_phi = np.fmod((alpha - beta + np.pi), 2*np.pi)
         new_alpha = beta - phi + np.pi
         new_beta = beta
 
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -168,7 +168,8 @@ def nullT(n, m, U):
 
 
 def clements(V, tol=1e-11):
-    r"""Performs the Clements decomposition of a Unitary complex matrix.
+    r"""Performs the Clements decomposition of a unitary complex matrix, with local
+    phase shifts applied between two interferometers.
 
     See Clements et al. Optica 3, 1460 (2016) [10.1364/OPTICA.3.001460]
     for more details.
--- a/tests/test_decompositions.py
+++ b/tests/test_decompositions.py
@@ -130,14 +130,14 @@ class DecompositionsModule(BaseTest):
         self.assertAlmostEqual(error.mean() , 0)
 
 
-    def test_clements_eq5_random_unitary(self):
+    def test_clements_phase_end_random_unitary(self):
         self.logTestName()
         error=np.empty(nsamples)
         for k in range(nsamples):
             n=20
             U=haar_measure(n)
 
-            new_tlist, new_diags = dec.clements_eq5(U)
+            new_tlist, new_diags = dec.clements_phase_end(U)
 
             U_rec=np.identity(n)
             for i in new_tlist:


co9olguy:
Great, thanks @ishdhand. @nquesada is the expert on the decompositions in SF, and @josh146 can provide some insight on any hiccups that might occur while making this a first-level feature in SF

strawberryfields/decompositions.py
```diff
@@ -15,6 +15,7 @@
 
 from itertools import groupby
 
+import math as mt
```
josh146: Since only one function is being used from the math library, perhaps better to be explicit here and have `from math import fmod`

strawberryfields/decompositions.py
```diff
@@ -163,10 +164,19 @@ def nullT(n, m, U):
 
     return [n-1, n, thetar, phir, nmax]
 
-def clements(V, tol=1e-11):
-    r"""Performs the Clements decomposition of a Unitary complex matrix.
+def clements(V, tol=11):
+    r"""Clements decomposition with phases in the middle of interferomter
```
```suggestion
    r"""Performs the Clements decomposition of a Unitary complex matrix, with local
    phase shifts applied in the middle of the interferometer.
```

strawberryfields/decompositions.py
```diff
@@ -202,6 +212,47 @@ def clements(V, tol=1e-11):
 
     return tilist, tlist, np.diag(localV)
 
+def clements_eq5(V, tol=11):
+    r"""Clements decomposition of unitary matrix
```
```suggestion
    r"""Performs the Clements decomposition of a Unitary complex matrix, with local
    phase shifts applied after the interferometer.
```

strawberryfields/decompositions.py
```diff
@@ -202,6 +212,47 @@ def clements(V, tol=1e-11):
 
     return tilist, tlist, np.diag(localV)
 
+def clements_eq5(V, tol=11):
```
josh146: Is there an alternative function name we could use that is a bit more descriptive here? Maybe, `clements_phase_end` or `clements_final_phase`. Those are pretty bad, so feel free to use your own suggestions!

tests/test_decompositions.py
```diff
@@ -129,6 +129,23 @@ def test_clements_random_unitary(self):
...
+            n=20
+            U=haar_measure(n)
+
+            new_tlist, new_diags = dec.clements_eq5(U)
+
+            U_rec=np.identity(n)
+            for i in new_tlist:
+                U_rec=dec.T(*i) @ U_rec
+            U_rec = np.diag(new_diags) @ U_rec
+            self.assertAlmostEqual(np.linalg.norm(U_rec-U), 0)
```
josh146:
```suggestion
            self.assertAlmostEqual(np.linalg.norm(U_rec-U), 0, delta=self.tol)
```
This ensures that the unittest method uses the SF test suite default tolerance.

josh146:
Hi @ishdhand, it looks great! I've left some very minor comments.

Another quick note: yesterday, we merged in a PR that replaces the integer rounding tolerance `tol=13` in the `clements` function with a floating point tolerance `tol=1e-13`, to better match our existing conventions - see the following line changes:
* https://github.com/XanaduAI/strawberryfields/blob/master/strawberryfields/decompositions.py#L166, 
* https://github.com/XanaduAI/strawberryfields/blob/master/strawberryfields/decompositions.py#L173, and
* https://github.com/XanaduAI/strawberryfields/blob/master/strawberryfields/decompositions.py#L188.

If you could bring the PR in line with those changes, that would be a massive help in avoiding merge conflicts.

With regards to the frontend integration, I propose we first merge this PR, and then start a second PR for frontend changes.

strawberryfields/decompositions.py
nquesada: What @josh146  says sounds like a better idea than importing all of math. Why do you prefer the fmod from python over the fmod from numpy?

strawberryfields/decompositions.py
nquesada: I like the fact that you labelled the transformations according to the  equation numbering of Clements paper. I am assuming there is no difference between the Optica and arxiv.

ishdhand:
Thanks, Josh and Nico for the suggestions! Have implemented these.

co9olguy:
Awesome, thanks @ishdhand!
@josh146 @nquesada where do we currently stand on this? Is it ready for merge (after a final round of code review)?

strawberryfields/decompositions.py
```diff
@@ -170,7 +170,16 @@ def nullT(n, m, U):
 def clements(V, tol=1e-11):
     r"""Performs the Clements decomposition of a Unitary complex matrix.
```
```suggestion
    r"""Performs the Clements decomposition of a unitary complex matrix.
```

strawberryfields/decompositions.py
josh146: Hi @ishdhand, wondering on the status of this suggested docstring change? Just curious why it was marked resolved.

josh146:
Only two (minor) docstring changes, and then all ready for merging :slightly_smiling_face:

PR #46, opened 2019-02-13
Petty pep8 changes

quantshah:

* Minor pep8 fixes.
I noticed some pep8 issues and thought I'd fix them to get started with the codebase.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
Looks great, thanks @quantshah! The Codacy CI integration is very slow, so I've been testing using code factor instead, so I won't worry about it not returning a result.

PR #48, opened 2019-02-25
Extract methods

ziofil:

Added two new utilities to extract a numerical representation of a circuit from an Engine object: `extract_unitary` and `extract_channel`.
The first returns a numerical array that represents a unitary transformation (and it's applicable to the case where in the circuit specification there are only ops of type `Gate`).
The second returns a numerical array that represents a channel (and it's applicable to the case where in the circuit specification there are only ops of type `Gate` or `Channel`).

Both functions support an option to vectorise the indices that correspond to different modes. This makes `extract_unitary` return a matrix, and `extract_channel` return a rank-4 tensor (if the option `representation` is 'choi' or 'liouville') or a rank-3 tensor (if the option `representation` is 'kraus')

`extract_unitary` also supports the option `backend` which can take the value 'fock' (default), which results in a numpy array output, or 'tf' which results in a TensorFlow tensor output.

**Benefits:**
Now a user can specify a circuit and ask for a numerical representation of the transformation that it represents.

**Possible Drawbacks:**
What drawbacks? xD

**Related GitHub Issues:**
None (that I know of)

### DIFF
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -850,7 +850,7 @@ def _interleaved_identities(num_subsystems: int, cutoff_dim: int):
 
 
 def _engine_with_CJ_cmd_queue(engine, cutoff_dim: int):
-    """Doubles the number of modes of an engine objects and prepends to its command queue
+    """Doubles the number of modes of an engine object and prepends to its command queue
     the operation that creates the interleaved identities ket.
 
     Here, CJ is from Choi-Jamiolkowski, as under the hood we are expliting the CJ isomorphism.
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -963,7 +963,7 @@ def extract_channel(engine, cutoff_dim: int, representation: str = 'choi', vecto
       - ``representation='choi'`` and ``representation='liouville'`` return an array
         with :math:`4N` indices
       - ``representation='kraus'`` returns an array of Kraus operators with :math:`2N` indices each,
-        where :math:`N` is the number of modes that the engine is created with.
+        where :math:`N` is the number of modes that the engine is created with
 
     Note that the Kraus representation automatically returns only the non-zero Kraus operators.
     One can reduce the number of operators by discarding Kraus operators with small norm (thus approximating the channel).
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -979,7 +979,7 @@ def extract_channel(engine, cutoff_dim: int, representation: str = 'choi', vecto
             \mathcal{C}(\rho) = \mathrm{Tr}[(\rho_A^T\otimes\mathbb{1}_B)\Lambda_{AB}]
 
     The indices of the non-vectorized Choi operator match exactly those of the state, so that the action
-    of the channel can be computed as (e.g. for one mode or for ``vectorize_modes=True``):
+    of the channel can be computed as (e.g., for one mode or for ``vectorize_modes=True``):
 
     >>> rho_out = np.einsum('ab,abcd', rho_in, choi)
 
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -771,7 +771,7 @@ def _vectorize_dm(tensor):
 
     Raises:
         ValueError: if the input tensor's dimensions are not all equal or if the number
-            of its indices is not a multiple of 4.
+            of its indices is not a multiple of 4
     """
     dims = tensor.ndim
 
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -1019,7 +1019,7 @@ def extract_channel(engine, cutoff_dim: int, representation: str = 'choi', vecto
     >>> rho_out = np.einsum('abcdefgh,fbhd->eagc', liouville, rho_in)
 
     The Liouville representation has the property that if the channel is unitary, the operator is separable.
-    Whereas even if the channel were the identity, the Choi operator would correspond to a maximally entangled state.
+    On the other hand, even if the channel were the identity, the Choi operator would correspond to a maximally entangled state.
 
     The choi and liouville operators in matrix form (i.e. with two indices) can be found as follows, where
     ``D`` is the dimension of each vectorized index (i.e. for :math:`N` modes, ``D=cutoff_dim**N``):
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -872,7 +872,7 @@ def _engine_with_CJ_cmd_queue(engine, cutoff_dim: int):
 
 
 def extract_unitary(engine, cutoff_dim: int, vectorize_modes: bool = False, backend: str = 'fock'):
-    r"""Returns the array representation of an queued unitary circuit
+    r"""Returns the array representation of a queued unitary circuit
     as an NumPy ndarray (``'fock'`` backend) or as a TensorFlow Tensor (``'tf'`` backend).
 
     Note that the circuit must only include operations of the :class:`strawberryfields.ops.Gate` class.
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -949,6 +949,8 @@ def extract_channel(engine, cutoff_dim: int, representation: str = 'choi', vecto
     The representation choices include the Choi state representation, the Liouville representation, and
     the Kraus representation.
 
+    .. note:: Channel extraction can currently only be performed using the ``'fock'`` backend.
+
     **Tensor shapes**
 
     * If ``vectorize_modes=True``:
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -725,7 +725,9 @@ def is_channel(engine):
 
 def _vectorize_dm(tensor):
     """Given a tensor with 4N indices of dimension :math:`D` each, it returns the vectorized
-    tensor with 4 indices of dimension :math:`D^N` each.
+    tensor with 4 indices of dimension :math:`D^N` each. This is the inverse of the procedure
+    given by :func:`_unvectorize_dm`.
+    Caution: this private method is intended to be used only for Choi and Liouville operators.
 
     For example, :math:`N=2`,
     ::
@@ -768,9 +770,9 @@ def _vectorize_dm(tensor):
 
 def _unvectorize_dm(tensor, num_subsystems):
     """Given a tensor with 4 indices, each of dimension :math:`D^N`, return the unvectorized
-    tensor with 4N indices of dimension D each.
-
-    This is the inverse of the procedure given by :func:`_vectorize_dm`.
+    tensor with 4N indices of dimension D each. This is the inverse of the procedure
+    given by :func:`_vectorize_dm`.
+    Caution: this private method is intended to be used only for Choi and Liouville operators.
 
     Args:
         tensor (array): a tensor with :math:`4` indices of dimension :math:`D^N`
@@ -817,6 +819,11 @@ def _interleaved_identities(num_subsystems: int, cutoff_dim: int):
     for _ in range(1, num_subsystems):
         I = np.tensordot(I, np.identity(cutoff_dim), axes=0)
 
+    # This is just some index gymnastics, the listcomp creates a sequence that advances by alternating steps like so:
+    # for num_subsystems=3 it's [0, 3, 1, 4, 2, 5]
+    # for num_subsystems=4 it's [0, 4, 1, 5, 2, 6, 3, 7]
+    # for num_subsystems=5 it's [0, 5, 1, 6, 2, 7, 3, 8, 4, 9]
+    # and so on...
     return np.einsum(I, [int(n) for n in np.arange(2*num_subsystems).reshape((2, num_subsystems)).T.reshape([-1])])
 
 
@@ -864,7 +871,19 @@ def extract_unitary(engine, cutoff_dim: int, vectorize_modes: bool = False, back
 
     Raises:
         TypeError: if the operations used to construct the circuit are not all unitary.
+
+    Example:
+        This shows the Hong-Ou-Mandel effect by extracting the unitary of a 50/50 beam splitter and then
+        computing the output given by one photon at each input (notice the order of the indices: out1, in1, out2, in2, ...).
+        The result tells us that the two photons always emerge together from a random output port and never one per port.
+
+    >>> engine, (A, B) = sf.Engine(num_subsystems=2)
+    >>> with engine:
+    >>>    BSgate(np.pi/4) | (A, B)
+    >>> U = extract_unitary(engine, cutoff_dim=3)
+    >>> print(abs(U[:,1,:,1])**2)
     """
+
     if not is_unitary(engine):
         raise TypeError("The circuit definition contains elements that are not of type Gate")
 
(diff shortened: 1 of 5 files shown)


strawberryfields/utils.py
```diff
@@ -673,3 +675,239 @@ def f_proxy(*args):
...
+def is_unitary(engine):
+    "Returns true if every command in the queue is of type Gate"
+    return all(isinstance(cmd.op, Gate) for cmd in engine.cmd_queue)
+
+def is_channel(engine):
+    "Returns true if every command in the queue is either of type Gate or type Channel"
+    return all(isinstance(cmd.op, (Channel, Gate)) for cmd in engine.cmd_queue)
+
+def vectorize_dm(tensor):
+    """Given a tensor with 4N indices each of dimension D each,
```
ziofil: there's an "each" too many

co9olguy:
Great, thanks for the PR @ziofil!

I notice some of the tests are currently failing (https://travis-ci.org/XanaduAI/strawberryfields/builds/498260170?utm_source=github_status&utm_medium=notification). Is this still a work-in-progress, or do you expect to resolve those shortly? If so, we can get started on code review.

ziofil:
Hi Nathan,
Ah, it's because I used f-strings, and you're also checking that the code works for python 3.5
I'll update the code to use the `format()` syntax.

josh146:
Ignore the codecov bot for now - the comment will be automatically updated with the correct coverage once the build passes

ziofil:
The remaining issue is that I need to use the Engine's method `_add_subsystems()`, and it begins with an underscore... I don't see a way out other than refactor the code and make it part of the public interface, but it's probably even worse.

josh146:
@ziofil, all good - we're okay with ignoring that linting rule :slightly_smiling_face:

Everything looks like it's passing now :champagne:

Will you be adding any more commits, or is it okay to begin the code review?

ziofil:
No, I'm done. 👍

strawberryfields/utils.py
```diff
@@ -673,3 +691,334 @@ def f_proxy(*args):
...
+
+    Here, CJ is from Choi-Jamiolkowski, as under the hood we are expliting the CJ isomorphism.
+
+    Args:
+        engine (Engine): a Strawberry Fields engine instance
+        cutoff_dim (int): the Fock basis truncation
+
+    Returns:
+        strawberryfields.engine.Engine: returns the same Engine instance, with the number of subsystems
+        increased and the interleaved identities ket prepended to the queue
```
josh146: Just double checking, but because there is no deepcopy here, this is an 'in-place' engine operation?

ziofil: Hi Josh, I'm not sure I understand your question. Could you rephrase?

strawberryfields/utils.py
```diff
@@ -673,3 +691,334 @@ def f_proxy(*args):
...
+    elif representation.lower() == 'liouville':
+        result = np.einsum('abcd -> dbca', choi)
+        if not vectorize_modes:
+            result = _unvectorize_dm(result, N)
+
+    elif representation.lower() == 'kraus':
+        eigvals, eigvecs = np.linalg.eig(np.einsum('abcd -> cadb', choi).reshape([cutoff_dim**(2*N), cutoff_dim**(2*N)]))
+        eigvecs = eigvecs[:, ~np.isclose(abs(eigvals), 0)]
+        eigvals = eigvals[~np.isclose(abs(eigvals), 0)]
+        result = np.einsum('abc->cab', np.einsum('b,ab->ab', np.sqrt(eigvals), eigvecs).reshape([cutoff_dim**N, cutoff_dim**N, -1]))
```
josh146: @ziofil, your docstring is fantastic, but could you also add some comments here (especially to the Kraus lines) just to clarify what is happening on each line?

tests/test_utils.py
```diff
@@ -290,13 +292,211 @@ def test_random_interferometer(self):
...
+        with tf.Session() as sess:
+            sess.run(tf.global_variables_initializer())
+            in_state = tf.constant(initial_state.reshape([-1]), dtype=tf.complex64)
+            final_state_tf_backend = sess.run(tf.einsum('ab,b', matrix_extract_tf_backend, in_state))
+        final_state_sf = eng_sf.run("fock", cutoff_dim=cutoff_dim).ket().reshape([-1])
+        final_state_extract = matrix_extract.dot(initial_state)
+        final_state_sf = eng_sf.run("fock", cutoff_dim=cutoff_dim).ket()
+
+        self.assertAllAlmostEqual(final_state_extract, final_state_sf, delta=defaults.TOLERANCE)
+        self.assertAllAlmostEqual(final_state_tf_backend, final_state_sf, delta=defaults.TOLERANCE)
```
josh146: Is there a reason here why you are testing the circuit twice, once with the tf backend and once with the Fock backend? The way we have the tests set up, the backend should be dynamically chosen based on an environment variable (`BACKEND=fock` or `BACKEND=tf`) in `defaults.py`, allowing us to run the same tests with different backends

ziofil: No, no reason, I just had not understood that your tests can do that switch automatically 😅

strawberryfields/utils.py
josh146: in this function, even though the engine is returned, it is also modified directly - so even if nothing was returned, the passed engine will still be modified.

I see that further down on line 875 where it is being used, you are using deepcopy, so it should be an issue there :slightly_smiling_face:

ziofil: Yeah, I didn't want to mess up the current Engine, so I work on a deepcopy.

strawberryfields/utils.py
ziofil: Done

josh146:
Hi @ziofil, I've done a code review, as well as pushing a small commit to integrate the new functions into the SF documentation, and make sure everything renders correctly (let me know if you think anything should be changed).

It looks really fantastic! Especially the documentation, really in-depth and well covered. I think only a couple of minor commits are needed before it gets merged.

1. The comments you added to the `extract_channel` function are great, and will really help with maintainability if someone else is working on this part of the codebase in the future. Do you mind quickly adding similar comments to `extract_unitary`, or where ever you think they might be useful?

2. Tests: the test should be modified as per https://github.com/XanaduAI/strawberryfields/pull/48#discussion_r260674838 to correctly allow the environment variables to set the backends. This is something I can quickly do.

3. Documentation: It would be nice to have a short example in the documentation of usage. Another thing that crossed my mind - do you think defining equations could be useful, for example when discussing the channel representation?

ziofil:
Hi @josh146 
1. sure, no problem, I'll do that.
2. ok, please go ahead
3. sure, how would you like the example to be formatted? Something like a doctest?

josh146:
>     3. sure, how would you like the example to be formatted? Something like a doctest?

I was thinking something very simple, like Sphinx code-block example at the top of the Engine page in the documentation:

https://strawberryfields.readthedocs.io/en/latest/code/engine.html

>  What kind of equations do you have in mind? The definitions of channel action like
>        C(ρ) = tr_A[(ρᵀᴬ ⊗ Iᴮ)Λᴬᴮ]?

Yeah, just something like you have written, that clearly defines the various channels etc.

ziofil:
Is the current version enough, or should I add more?

josh146:
I was a bit late with my reply, what you have looks perfect :slightly_smiling_face:

tests/test_utils.py
```diff
@@ -290,13 +299,381 @@ def test_random_interferometer(self):
...
+        self.assertAllAlmostEqual(final_rho_liouville, final_rho_sf, delta=self.tol)
+        self.assertAllAlmostEqual(final_rho_kraus, final_rho_sf, delta=self.tol)
+
+    def test_vectorize_unvectorize(self):
+        """Test vectorize utility function"""
+        self.logTestName()
+
+        cutoff = 4
+
+        dm = np.random.rand(*[cutoff]*8) + 1j*np.random.rand(*[cutoff]*8) # 4^8 -> (4^2)^4 -> 4^8
```
co9olguy: make (arbitrary) choices here instead of random

tests/test_utils.py
```diff
@@ -290,13 +299,381 @@ def test_random_interferometer(self):
...
+    def test_extract_unitary_2_modes_fock(self):
+        """Test that unitary extraction works for 2 modes using the fock backend"""
+        self.logTestName()
+
+        eng_extract, q_extract = sf.Engine(num_subsystems=2)
+
+        S = sf.ops.Sgate(2)
+        B = sf.ops.BSgate(2.234, -1.165)
+
+        initial_state = np.complex64(np.random.rand(self.D, self.D) + 1j*np.random.rand(self.D, self.D))
```
co9olguy: fix choice of initial state

tests/test_utils.py
```diff
@@ -290,13 +299,381 @@ def test_random_interferometer(self):
...
+    def test_extract_unitary_2_modes_tf(self):
+        """Test that unitary extraction works for 2 modes using the tf backend"""
+        self.logTestName()
+
+        eng_extract, q_extract = sf.Engine(num_subsystems=2)
+
+        S = sf.ops.Sgate(2)
+        B = sf.ops.BSgate(2.234, -1.165)
+
+        initial_state = np.complex64(np.random.rand(self.D, self.D) + 1j*np.random.rand(self.D, self.D))
```
co9olguy: fix choice of initial state

tests/test_utils.py
```diff
@@ -290,13 +299,381 @@ def test_random_interferometer(self):
...
+    def test_extract_channel_2_modes(self):
+        """Test that channel extraction works for 2 modes"""
+        self.logTestName()
+
+        eng_extract, q_extract = sf.Engine(num_subsystems=2)
+
+        S = sf.ops.Sgate(2)
+        B = sf.ops.BSgate(2.234, -1.165)
+
+        initial_state = np.random.rand(self.D, self.D, self.D, self.D) + 1j*np.random.rand(self.D, self.D, self.D, self.D)
```
co9olguy: fix initial state

tests/test_utils.py
```diff
@@ -290,13 +299,381 @@ def test_random_interferometer(self):
...
+    def test_extract_channel_2_modes_vectorize(self):
+        """Test that channel extraction works for 2 modes vectorized"""
+        self.logTestName()
+
+        eng_extract, q_extract = sf.Engine(num_subsystems=2)
+
+        S = sf.ops.Sgate(2)
+        B = sf.ops.BSgate(2.234, -1.165)
+
+        initial_state = np.random.rand(self.D, self.D, self.D, self.D) + 1j*np.random.rand(self.D, self.D, self.D, self.D)
```
co9olguy: fix initial state

co9olguy:
I would commit all the minor suggestions myself, but it says I don't have push permissions to the repo 🤔

strawberryfields/utils.py
```diff
@@ -673,3 +713,415 @@ def f_proxy(*args):
...
+        4 --|    |-- 5
+        6 --|____|-- 7
+
+    goes to
+    ::
+        (0,2) --|‾‾‾‾|-- (1,3)
+        (4,6) --|____|-- (5,7)
+
+    Args:
+        tensor (array): a tensor with :math:`4N` indices of dimension :math:`D` each
```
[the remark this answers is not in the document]
ziofil: That's not the best choice of name on my part. I just wanted to indicate that it's not acting on vectors, but it's not acting on density matrices either. Feel free to change the name (or just remove "_dm" since the name is already beginning with an underscore)

strawberryfields/utils.py
```diff
@@ -673,3 +713,415 @@ def f_proxy(*args):
...
+
+    # here we rearrange the indices to go back to the order [in1, out1, in2, out2, etc...]
+    if backend == 'fock':
+        return np.transpose(result, [int(n) for n in np.arange(2*N).reshape((2, N)).T.reshape([-1])])
+
+    return tf.transpose(result, [int(n) for n in np.arange(2*N).reshape((2, N)).T.reshape([-1])])
+
+
+def extract_channel(engine, cutoff_dim: int, representation: str = 'choi', vectorize_modes: bool = False):
+    r"""Returns a numerical array representation of a channel from a queued Engine circuit.
```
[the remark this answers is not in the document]
ziofil: Because `extract_channel` requires a lot of index manipulations that I would have to rewrite in TensorFlow and duplicate a lot of code. in `extract_unitary` it's easier because it was just a couple of lines (which I still didn't like duplicating). 
It could be possible to use functions that have the same name in TensorFlow and Numpy, and call the one we need depending on the required backend (but it's not the best for maintainability).

strawberryfields/utils.py
```diff
@@ -673,3 +713,415 @@ def f_proxy(*args):
...
+            otherwise it returns a rank-4N tensor, where N is the number of modes.
+
+    Returns:
+        The numerical array of the channel, according to the specified representation
+        and vectorization options.
+
+    Raises:
+        TypeError: if the gates used to construct the circuit are not all unitary or channels.
+    """
+    # if is_unitary(engine):
```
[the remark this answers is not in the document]
ziofil: I left it there as a reminder that it's a possible check that could be performed, but raising a warning seemed excessive.

tests/test_utils.py
ziofil: What's the right way of fixing the initial array if we want it to be quite large? Here it has D^4 complex numbers (1296 as D=6)

josh146: You could use `np.random.seed(42)` directly above the call to `np.random.rand`. NumPy has a guarantee that it will produce the same result anywhere:

> Compatibility Guarantee A fixed seed and a fixed series of calls to ‘RandomState’ methods using the same parameters will always produce the same results up to roundoff error except when the values were incorrect.

See
1. https://stackoverflow.com/questions/40676205/cross-platform-numpy-random-seed
2. https://stackoverflow.com/questions/5836335/consistently-create-same-random-numpy-array

ziofil: Ah yes, that's a good idea. Is it sufficient to set it once for the whole test suite? Will it break/upset other parts of the tests?

strawberryfields/utils.py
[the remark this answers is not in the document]
ziofil: Ok, I removed the "_dm" termination

tests/test_utils.py
ziofil: I set the seed after importing numpy

tests/test_utils.py
ziofil: I set the seed after importing numpy

tests/test_utils.py
ziofil: I set the seed after importing numpy

tests/test_utils.py
ziofil: I set the seed after importing numpy

tests/test_utils.py
ziofil: I set the seed after importing numpy

strawberryfields/utils.py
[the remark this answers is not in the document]
co9olguy: No problem. It's something we can target for future enhancements, but not a deal-breaker in terms of this PR. can we add a line near the top of the docstring "Note: currently this function is only supported for the 'fock' backend"?

strawberryfields/utils.py
```diff
@@ -673,3 +713,415 @@ def f_proxy(*args):
...
+    # here we rearrange the indices to go back to the order [in1, out1, in2, out2, etc...]
+    if backend == 'fock':
+        return np.transpose(result, [int(n) for n in np.arange(2*N).reshape((2, N)).T.reshape([-1])])
+
+    return tf.transpose(result, [int(n) for n in np.arange(2*N).reshape((2, N)).T.reshape([-1])])
+
+
+def extract_channel(engine, cutoff_dim: int, representation: str = 'choi', vectorize_modes: bool = False):
+    r"""Returns a numerical array representation of a channel from a queued Engine circuit.
+
```
co9olguy:
```suggestion
Note: currently this function is only supported for the 'fock' backend

```

strawberryfields/utils.py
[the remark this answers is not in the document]
co9olguy: Ok got it. Probably best to just remove it.

co9olguy:
```suggestion
```

strawberryfields/utils.py
```diff
@@ -673,3 +713,415 @@ def f_proxy(*args):
...
+    Returns:
+        The numerical array of the channel, according to the specified representation
+        and vectorization options.
+
+    Raises:
+        TypeError: if the gates used to construct the circuit are not all unitary or channels.
+    """
+    # if is_unitary(engine):
+    #     #raise Warning(f"This circuit is unitary and you could use extract_unitary for a more compact representation")
```
co9olguy:
```suggestion
```

strawberryfields/utils.py
```diff
@@ -673,3 +713,415 @@ def f_proxy(*args):
...
+    #     pass
```
co9olguy:
```suggestion
```

co9olguy:
Just one or two tiny suggestions to commit, but all good on my side :)

tests/test_utils.py
josh146: :+1:

Thinking about it, probably the best place to put it is in the `setUp` method of each test class.

PR #64, opened 2019-04-22
Version update

josh146:
Pull request to increment the version number to 0.10.0, as well as a few maintenance issues to get ready for a minor release:

* Update the changelog
* Remove the old test folder
* Update Makefile and travis configs for pytest
* Update documentation on how to run the test suite


TESTS.rst
```diff
@@ -1,19 +1,54 @@
...
+
+
 To ensure that Strawberry Fields is working correctly after installation, the test suite can be run by navigating to the source code folder and running
 ::
 
-  make test
+    $ make test
 
-Note that this runs *all* of the tests, using *all* available backends, so can be quite slow (it should take around 40 minutes to complete). Alternatively, you can run the full test suite for a particular backend by running
+Note that this runs *all* of the tests, using *all* available backends, so can be quite slow (it should take around 40 minutes to complete). Alternatively, you can run the full test suite for a particular component by running
```
co9olguy: Is the 40 minute estimate accurate for tests that a user would run locally (i.e., in sequence)?

TESTS.rst
```diff
@@ -1,19 +1,54 @@
 Software tests
```
co9olguy: Is this file linked anywhere else in docs?

TESTS.rst
josh146: Actually, you're right. It would be about an hour now. This is because the previous version of `make test` *said* it was running all the tests, but would only run the tests in pure mode. Now, they will run in both pure and mixed mode.

TESTS.rst
josh146: No, this is purely a top level readme-esque file. This used to be in the `README.rst` file, but we spun it out into it's own file so it wouldn't render on github and take too much space.

josh146: But, the same information is also duplicated inside the docs (in the installation page).

TESTS.rst
josh146:
```suggestion
Note that this runs *all* of the tests, using *all* available backends, in both pure *and* mixed modes, so can be quite slow (it should take around 1 hour to complete). Alternatively, you can run the full test suite for a particular component by running
```

PR #65, opened 2019-04-23
Engine refactor

smite:

This PR splits the former `Engine` class into two new classes: `Program`, which represents a quantum circuit or a fragment thereof, and `Engine`, which executes `Program` instances.

* Most of the frontend test have been updated and run successfully. A few are skipped for now.
* A few `utils.py` functions (`is_unitary()` etc) haven't been converted yet. 

**Benefits:**

Improves the SF syntax and program fragment handling.

Enables the further development of Engine into the Executor classes that take care of the communication between the user and the various backends. The current Engine class more or less corresponds to a local backend executor.

**Note:**

The files `test_utils.py` and `test_ops_decompositions.py` had Windows-style CRLF line endings which I changed to just LF, which is why it seems that every line in those files has changed. in reality both files have only a small number of actual changes.


strawberryfields/__init__.py
```diff
@@ -25,14 +25,15 @@
...
-and :file:`utils.py`) and backend components (all found within the :file:`strawberryfields.backends` submodule).
+These can be separated into frontend components
+and backend components (all found within the :file:`strawberryfields.backends` submodule).
 
 Software components
 -------------------
 
 **Frontend:**
 
+* Quantum programs: :mod:`strawberryfields.program`
```
josh146: 🏆

strawberryfields/__init__.py
```diff
@@ -50,20 +51,18 @@
...
 
 Code details
 ~~~~~~~~~~~~
 """
 
-from __future__ import unicode_literals
 from ._version import __version__
-from .engine import Engine as _Engine
-from .engine import _convert as convert
+from .program import (Program, _convert as convert)
```
josh146: I've forgotten what the reason is - why are we importing `_convert` as `convert`, and not just calling it `convert` to begin with?

strawberryfields/engine.py
```diff
@@ -127,870 +78,173 @@
...
-    they enter the command queue in order to be able to report register
-    reference errors as soon as they happen.
-
-    The backend, however, only becomes aware of subsystem changes when
-    the circuit is run. See :meth:`reset_queue` and :meth:`reset`.
-
     Args:
-        num_subsystems (int): Number of subsystems in the quantum register.
+        backend (str): name of the backend
+        host    (str): URL of the remote backend host, or None if a local backend is used
```
josh146: Since this existing `Engine` class will likely become `LocalEngine`, the `host` argument is not needed, right?

strawberryfields/engine.py
```diff
@@ -127,870 +78,173 @@
...
-        #: list[list[Command]]: list of lists of commands that have been run (one list per run)
-        self.cmd_applied = []
-        self.backend = None       #: BaseBackend: backend for executing the quantum circuit
-        #: float: Numerical value of hbar in the (implicit) units of position and momentum.
+    def __init__(self, backend, host=None, *, hbar=2, **kwargs):
+        #: list[Program]: list of Programs that have been run
+        self.run_progs = []
+        #: str: URL of the remote backend host
+        self.host = host
+        #: float: numerical value of hbar in the (implicit) units of position * momentum
```
josh146: Are these pycharm style type hints? I was chatting to Tom about this today, and the thinking is we will start allowing/highly suggesting Python 3.5 style type hints from now on in our style guide. So these should probably be moved as so:

```python
# list of Programs that have been run
self.run_progs = [] # type: List[Program]
```

Alternatively, it could be written as a attribute docstring which would get picked up by Sphinx automatically:

```python
self.run_progs = [] # type: List[Program]
"""list[Program]: list of Programs that have been run"""
```

strawberryfields/engine.py
```diff
@@ -127,870 +78,173 @@
...
-                    except NotImplementedError as err:
-                        # simplify the error message by suppressing the previous exception
-                        raise err from None
+                    temp = cmd.op.decompose(cmd.reg)
+                    # run the decomposition
+                    applied_cmds = self._run_command_list(temp)
+                    applied.extend(applied_cmds)
+                except NotImplementedError as err:
+                    # simplify the error message by suppressing the previous exception
+                    raise err from None
```
josh146: This logic for applying quantum operations seems by and large unchanged - how come it broke the `Decomposition` subclass of operations?

strawberryfields/engine.py
smite: Right.

strawberryfields/engine.py
```diff
@@ -127,870 +78,173 @@
...
+            for p in program:
+                if self.run_progs:
+                    prev = self.run_progs[-1]
+
+                if prev is not None and not p.can_follow(prev):
+                    raise RuntimeError("Register mismatch after Program {}, '{}'.".format(len(self.run_progs)-1, prev.name))
+
+                # if the program hasn't been compiled for this backend, do it now
+                if compile and p.backend != self.backend._short_name:
+                    p = p.compile(self.backend._short_name)
```
josh146: What happens if `compile=False`, but the program isn't compiled properly for the target backend? I guess this is something we will need to think about for the new backend API

strawberryfields/ops.py
```diff
@@ -73,11 +73,11 @@
...
   The information is directly available only after the simulation has been run up to the point of measurement::
 
-    eng, q = sf.Engine(2)
+    eng, (alice, bob) = sf.Engine(2)
     with eng:
-      Measure       | q[0]
+      Measure       | alice
       eng.run()
-      Dgate(q[0].val) | q[1]
+      Dgate(alice.val) | bob
```
josh146: Was this something you changed before the program/engine split 😆

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+        func (function): function to be converted to a :class:`RegRefTransform`.
+    """
+    @functools.wraps(func)
+    def wrapper(*args):
+        "Unused docstring."
+        return RegRefTransform(args, func)
+    return wrapper
+
+
+# TODO replace this mock backend capability database
```
josh146: Looks good for now 👍 but we'll definitely need something more complicated to allow us to also specify restricted topologies

strawberryfields/engine.py
smite: They are attribute comments which Sphinx does pick up:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

josh146: Oh nice, I didn't know that. I wonder if mypy also picks them up.

strawberryfields/engine.py
smite: It's the hbar handling which is the issue with `ops.Decomposition`

strawberryfields/engine.py
smite: Yes. This was to make some tests work, since compilation can change the relative order of gates that commute.

strawberryfields/ops.py
smite: Yeah. The rest of the module docstring still needs to be updated.

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+        # accept a single RegRef in addition to a Sequence
+        if not isinstance(reg, Sequence):
+            reg = [reg]
+
+        #: Operation: quantum operation to apply
+        self.op = op
+        #: Sequence[RegRef]: subsystems to which the operation is applied
+        self.reg = reg
+        #: bool: is this Command a part of a decomposition?
+        self.decomp = decomp
```
josh146: This (was?) currently used by the engine to mark where decompositions were being applied, as a convenience to the user. This avoided a bug where calling `S2gate` would later show `S2gate`, alongside `BSgate` and `Sgate` in the `print_applied` list.

By putting the `S2gate` in a command with `decomp=False`, the engine then knows to skip it from the `print_applied` list.

Not sure if this is still needed after the refactor or not.

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+    def __init__(self, ind):
+        self.ind = ind   #: int: subsystem index
+        self.val = None  #: Real: measured eigenvalue. None if the subsystem has not been measured yet
+        self.active = True  #: bool: True at construction, False after the subsystem is deleted
+
+    def __str__(self):
+        return 'q[{}]'.format(self.ind)
+
+    def __hash__(self):
+        return hash((self.ind, self.active))
```
josh146: haha the temptation to use `__eq__` was too much :laughing:

strawberryfields/engine.py
josh146: Ah, it looks like GVM does not support adding doctring type hinting support to mypy 🙁

josh146:
Re: continuous-integration failures

* TravisCI: Looks like there are still a lot of places in the tests where `eng, q = Engine(num_subsystems)` is being called.

* CodeFactor: There are 7 undefined variables now being used, probably old names that no longer exist after the refactor and were missed.

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+        self.circuit.append(Command(op, reg))
+        return reg
+
+    def compile(self, backend='fock', optimize=True):
+        """Compile the program for the given backend.
+
+        The compilation step validates the program, making sure all the Operations used are accepted by the backend.
+        Additionally it may decompose certain gates into sequences of simpler gates.
+
+        TODO for now the compilation happens in-place, i.e. it modifies self.
```
josh146: Was this because there was an issue with getting `prog_new = prog.compile('chip0')` to work?

strawberryfields/program.py
smite: Yeah, this way I can compare two dictionaries with ==. I'll add the missing docstring.

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+
+    def evaluate(self):
+        """Evaluates the numeric value of the function if all the measurement values are available.
+
+        Returns:
+            Number: function value
+        """
+        temp = [r.val for r in self.regrefs]
+        if any(v is None for v in temp):
+            # NOTE: "if None in temp" causes an error if temp contains arrays,
```
co9olguy: Is this just a note to explain the logical expression above, or is the comment about something more important (i.e., a warning about something that will lead to bugs in other parts of the code)?

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+            # since it uses the == comparison in addition to "is"
+            raise CircuitError("Trying to use a nonexistent measurement result (e.g. before it can be measured).")
```
co9olguy:
```suggestion
            raise CircuitError("Trying to use a nonexistent measurement result (e.g., before it can be measured).")
```

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+    """Represents a quantum circuit.
+
+    A quantum circuit is a set of quantum operations applied in a specific order
+    to a set of subsystems (represented by wires in the circuit diagram) of the quantum register.
+    The Program class represents a quantum circuit in general as a directed acyclic graph (DAG)
+    whose nodes are :class:`Command` instances, and each (directed) edge in the graph
+    corresponds to a specific wire along which the two associated Commands are connected.
+
+    Program instances also act as context managers (and the context itself) for inputting
+    quantum circuits using a `with` block and :class:`Operation` instances.
```
co9olguy:
```suggestion
    quantum circuits using a :code:`with` block and :class:`Operation` instances.
```

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+        Returns:
+            int: number of Commands in the program
+        """
+        return len(self.circuit)
+
+    def print(self, print_fn=print):
+        """Print the program contents using Blackbird syntax.
+
+        Args:
+            print_fn (function): optional custom function to use for string printing.
```
co9olguy:
```suggestion
            print_fn (function): optional custom function to use for string printing
```

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+        The locking happens when the program is run, or a successor Program is constructed,
+        in order to ensure that the RegRef state of the Program does not change anymore.
+        """
+        self.locked = True
+
+    def can_follow(self, prev):
+        """Checks if this program can follow the given program.
+
+        This requires that the final RegRef state of the first program matches
+        the initial RegRef state of the second program, i.e. they have the same number
```
co9olguy:
```suggestion
        the initial RegRef state of the second program, i.e., they have the same number
```

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+        """
+        temp = []
+        for rr in reg:
+            # must be either an integer or a RegRef
+            if isinstance(rr, RegRef):
+                # regref must be found in the dict values (the RegRefs are compared using __eq__, which, since we do not define it, defaults to "is")
+                if rr not in self.reg_refs.values():
+                    raise RegRefError('Unknown RegRef.')
+                if self.reg_refs[rr.ind] is not rr:
+                    raise RegRefError('Should never happen!')
```
co9olguy: Should we make the error message more clear? What exactly should never happen?

strawberryfields/program.py
co9olguy: This shows up in docs. We should switch to a comment-based TODO

co9olguy:
```suggestion
        TODO for now the compilation happens in-place, i.e., it modifies self.
```

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+
+
+    def draw_circuit(self, tex_dir='./circuit_tex', write_to_file=True):
+        r"""Draw the circuit using the Qcircuit :math:`\LaTeX` package.
+
+        This will generate the tex code required to display the queued or applied
+        quantum operations as a quantum circuit.
+
+        Args:
+            tex_dir (str): relative directory for latex document output.
```
co9olguy:
```suggestion
            tex_dir (str): relative directory for latex document output
```

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+            write_to_file (bool): if False, no output file is created.
```
co9olguy:
```suggestion
            write_to_file (bool): if False, no output file is created
```

tests/frontend/test_engine.py
```diff
@@ -164,228 +144,150 @@ def test_print_commands(self, backend):
...
+    def test_corrupt_index(self, prog):
         """This should never happen!"""
-        eng, q = sf.Engine(2)
-        q[0].ind = 1
-
-        with eng:
-            with pytest.raises(engine.RegRefError, match="Should never happen!"):
+        with prog.context as q:
+            q[0].ind = 1
+            with pytest.raises(program.RegRefError, match="Should never happen!"):
```
co9olguy: Will need to update once we update the error message to be more explicit about the behaviour that should never happen

tests/frontend/test_ops_gate.py
```diff
@@ -53,7 +53,7 @@ def Q(self):
     def G(self, gate):
         """Initialize each gate"""
         if gate in ops.zero_args_gates:
-            return gate
+            return gate()
```
co9olguy: What caused the change here?

tests/frontend/test_utils.py
```diff
@@ -1,686 +1,660 @@
...
+
+# ===================================================================================
+# Engine utility tests
+# ===================================================================================
+
+
+class TestEngineUtilityFunctions:
+    """Tests for some engine auxiliary functions"""
+
+    @pytest.mark.skip('FIXME is_unitary')
```
co9olguy: Can we make a separate ticket/issue for specific related fixes that need to be made going forward (e.g., I've seen `is_unitary`, `is_channel`, and one about hbar issues)

strawberryfields/engine.py
```diff
@@ -127,870 +78,173 @@
...
+        for cmd in prog.circuit:
+            try:
+                # try to apply it to the backend
+                cmd.op.apply(cmd.reg, self.backend, hbar=self.hbar, **kwargs)
+                applied.append(cmd)
+            except NotApplicableError:
+                # command is not applicable to the current backend type
+                raise NotApplicableError('The operation {} cannot be used with {}.'.format(cmd.op, self.backend)) from None
+            except NotImplementedError:
+                # command not directly supported by backend API, try a decomposition instead
```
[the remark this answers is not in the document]
josh146: I think so. Leave it for a separate pre-validation PR, out of scope for this one.

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+.. currentmodule:: strawberryfields.program
+
+This module implements the :class:`Program` class which acts as a representation for quantum circuits.
+The Program object also acts as a context for defining the quantum circuit using the Python-embedded Blackbird syntax.
+
+A typical use looks like
+::
+
+  prog = sf.Program(num_subsystems)
+  with prog.context as q:
```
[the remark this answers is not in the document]
josh146: `prog.declare` :smiling_imp:

strawberryfields/program.py
```diff
@@ -0,0 +1,897 @@
...
+# TODO replace this mock backend capability database
+backend_database = {
+    'fock': {
+        'New_modes': True,
+        'Delete': True,
+        'Rgate': True,
+        'Sgate': True,
+        'Dgate': True,
+        'BSgate': True,
+        'S2gate': False,  # use a decomposition
```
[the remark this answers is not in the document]
josh146: Yep, I agree. But lets make this change in a separate pre-validation PR.

tests/frontend/test_ops_gate.py
josh146: `Fouriergate` (with shortcut variable `Fourier = Fouriergate()`) is the only zero argument gate.

In this loop, it used to receive `Fourier`, which was pre-initialized. Now it receives `Fouriergate` instead, which need to be initialized.

strawberryfields/program.py
smite: Now that the compilation/optimization step is explicit I think `Command.decomp` could be removed.

josh146: :+1: I agree.
Let's make an issue for this, and do this in a new PR as it is out of scope.

strawberryfields/program.py
smite: This turned out to be a bit simpler, and one can always make an explicit copy of the original program if you want to keep it. I'm not opposed to changing it though.

josh146: I'm on the fence, but lean towards whatever seems clearer from the user perspective. i.e., if you see this,

```python
prog.compile('chip')
```

what would the user expect the behaviour to be - in place, or return a new program?

Put another way, which use case is _less_ surprising to the user? @co9olguy, thoughts?

strawberryfields/ops.py
```diff
@@ -73,11 +73,11 @@
 * :class:`Measurement` operations manipulate the register state and produce classical information.
   The information is directly available only after the simulation has been run up to the point of measurement::
```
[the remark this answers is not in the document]
smite: The double colon here is an intentional reST trick, it acts both as a single colon and a literal block separator.

tests/frontend/test_utils.py
smite: Sure

strawberryfields/program.py
smite: It's an explanation of the expression above.

strawberryfields/program.py
smite: Yeah, we should. I'll fix it.

strawberryfields/program.py
co9olguy: I think I prefer
```prog_compiled = prog.compile('chip')```
That way the original program is always available (e.g., for comparison)

File: tests/integration/test_engine_integration.py [hunk 27]
```diff
@@ -96,84 +91,82 @@ class TestProperExecution:
...
             subroutine(bob, alice)
 
-        state = eng.run()
+        state = eng.run(prog)
 
         # state norm must be invariant
         if isinstance(eng.backend, BaseFock):
             assert np.allclose(state.trace(), 1, atol=tol, rtol=0)
 
-    def test_checkpoints(self, setup_eng, tol):
```
co9olguy: @smite Is this test dropped because it is no longer relevant? i.e., because we no longer have history checkpoints?

smite: We do, but they work very differently now. Each Program object stores the state of the RegRefs both at its beginning and at its end. The test no longer tested anything relevant.

smite: The Engine only keeps track of the execution history, i.e. the list of Programs executed so far.

co9olguy:
@josh146 @smite 

There is only one outstanding message from previous review, which refers to a TODO anyway, so there doesn't seem to be anything significant to deal with on that front.

However, there are still a large number of tests failing. In order to move other parts of the dev process forward, I propose we should merge this -- knowing it is still broken -- and move on to new items. New tickets can be made to deal with failing tests.

strawberryfields/program.py
smite: Now compile returns a copy.

smite:
@josh146 @co9olguy 

> However, there are still a large number of tests failing. In order to move other parts of the dev process forward, I propose we should merge this -- knowing it is still broken -- and move on to new items. New tickets can be made to deal with failing tests.

I've fixed some more tests since, but many still remain broken, and fixing some of them does not make much sense before we've decided how the backend API will work in the future.

PR #68, opened 2019-04-25
Adds the ability to save/load SF Programs to/from Blackbird files

josh146:
* Adds the module `io.py`, containing the functions:

  - `to_blackbird()`: converts a SF Program object to a Blackbird Program object (for internal use only)

  - `to_program()`: converts a BB Program object to a SF Program object (for internal use only)

  - `save()`: saves a quantum program to a Blackbird .xbb file.

  - `load()`: loads a quantum program from a Blackbird .xbb file.

* Import `save` and `load` into `__init__` to make them top level functions

* Adds the blackbird library as a dependency of SF.


strawberryfields/io.py
```diff
@@ -0,0 +1,208 @@
...
+    """Convert a Strawberry Fields Program to a Blackbird Program.
+
+    Args:
+        prog (Program): the Strawberry Fields program
+
+    Returns:
+        blackbird.BlackbirdProgram:
+    """
+    bb = blackbird.BlackbirdProgram(name=prog.name, version=1.0)
+
```
smite: Make version either an int or a string

strawberryfields/__init__.py
```diff
@@ -57,12 +57,12 @@
...
-
-from ._version import __version__
-from .program import (Program, _convert as convert)
 from .engine import Engine
+from .io import save, load
+from .program import Program, _convert as convert
+from ._version import __version__
 
-__all__ = ['Engine', 'Program', 'convert', 'version']
+__all__ = ["Engine", "Program", "convert", "version", "save"]
```
smite:
```suggestion
__all__ = ["Engine", "Program", "convert", "version", "save", "load"]
```

strawberryfields/io.py
```diff
@@ -0,0 +1,208 @@
...
+
+def to_program(bb):
+    """Convert a Blackbird Program to a Strawberry Fields Program.
+
+    Args:
+        bb (blackbird.BlackbirdProgram): the input Blackbird program object
+
+    Returns:
+        program:
```
smite:
```suggestion
        Program: corresponding SF program
```

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+
+        op["op"] = cmd.op.__class__.__name__
+        op["modes"] = [i.ind for i in cmd.reg]
+
+        if "Measure" in op["op"]:
+            # special case to take into account 'select' keyword argument
+            if cmd.op.select is not None:
+                op["kwargs"]["select"] = cmd.op.select
+
+            if cmd.op.p:
```
co9olguy: This line won't work if we add new parameterized measurements in the future, correct?

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+                # argument is quadrature phase
+                op["kwargs"]["phi"] = cmd.op.p[0].x
+
+        else:
+            if cmd.op.p is not None:
```
co9olguy: When looking at this piece of the code, it's hard to know what the attribute `p` is. Should we refactor to give a better name (`params`?)

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+                for a in cmd.op.p:
+                    # check if reg ref transform
+                    if isinstance(a.x, ops.RegRefTransform):
```
co9olguy: Same comment as above, but for `x` now. Anyone looking at this code will not know what `cmd.op.p.x` is supposed to mean

co9olguy: Could replace `x` with `val`. 

```cmd.op.params[0].val``` is a lot clearer

File: strawberryfields/io.py [hunk 7]
```diff
@@ -0,0 +1,214 @@
...
+                            # as the RegRefTransform string cannot be checked
+                            # to determine if it is a valid function for serialization!
+                            #
+                            # Possible solutions:
+                            #
+                            #   * Use SymPy to represent functions, as
+                            #     SymPy provides methods for converting to
+                            #     Python functions as well as serialization
+                            #
+                            #   * Don't allow classical processing of measurements
```
co9olguy: Seems like the lowest overhead way to do it. All post-processing can happen locally. Of course, this may mean more data transfer cost

co9olguy: Perhaps any processing which does not change the size of a value in memory should be deferred to local device?

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+                # argument is quadrature phase
+                op["kwargs"]["phi"] = cmd.op.p[0].x
+
+        else:
+            if cmd.op.p is not None:
+                for a in cmd.op.p:
+                    # check if reg ref transform
+                    if isinstance(a.x, ops.RegRefTransform):
+                        if a.x.func_str is not None:
+                            # Note: not currently guaranteed to work,
```
co9olguy: What isn't guaranteed to work? Can you give a failure case?

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+        raise ValueError("Blackbird program contains no quantum operations!")
+
+    prog = Program(max(bb.modes)+1, name=bb.name)
+
+    # append the quantum operations
+    with prog.context as q:
+        for op in bb.operations:
+            try:
+                if 'args' in op:
+                    getattr(ops, op["op"])(*op["args"], **op["kwargs"]) | [q[i] for i in op["modes"]] #pylint:disable=expression-not-assigned
```
co9olguy: Yikes! Hard to parse this one!

co9olguy: suggest refactoring. even if leads to some extra assignment lines, at least it will be readable

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+                else:
+                    getattr(ops, op["op"]) | [q[i] for i in op["modes"]] #pylint:disable=expression-not-assigned
+            except AttributeError:
+                raise NameError("Quantum operation {} not defined!".format(op["op"]))
+
+    # compile the program if a target exists
```
co9olguy: Do we want to do an implicit compilation of a program? Or do we want to just convert from bb program to sf program, and leave it up to user to compile?

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+    if bb.target["name"] is not None:
+        return prog.compile(bb.target["name"])
+
+    return prog
+
+
+def save(file, prog):
```
co9olguy: Github is highlighting `file`. I checked and `file` seems to be a builtin in Python. However, it looks like it is a python2 thing: https://stackoverflow.com/questions/24942358/is-file-a-keyword-in-python,
so I guess it's ok to use `file` instead of `file_`?

strawberryfields/io.py
```diff
@@ -0,0 +1,214 @@
...
+            # argument file is a file-object
+            fid = file
+        else:
+            # argument file is a Path or string
+            f = os.fspath(file)
+            fid = open(f, "r")
+            own_file = True
+
+    except TypeError:
+        raise ValueError("file must be a string, pathlib.Path, or file-like object")
```
co9olguy: Something feels wrong about this exception, specifically to do with the appearance of both `file` and `file-like`. I'm not really sure how to improve though.

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+r"""Unit tests for the IO module"""
+import pytest
+
+pytestmark = pytest.mark.frontend
```
co9olguy: the linter doesn't complain about this?

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+            "op": "Dgate",
+            "modes": [1],
+            "args": [0.54 + 0.324j, 0],
+            "kwargs": {},
+        }
+
+        assert bb.operations[0] == expected
+
+    def test_two_mode_gate(self):
+        """Test gate with keyword argument converts"""
```
co9olguy: wrong docstring?

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+        expected = {
+            "op": "MeasureHomodyne",
+            "modes": [0],
+            "args": [],
+            "kwargs": {"phi": 0.43, "select": 0.543},
+        }
+
+        assert bb.operations[0] == expected
+
+        # repeat with kwargs only
```
co9olguy: Nice!

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+        version 1.0
+        """
+
+        bb = blackbird.loads(bb_script)
+
+        with pytest.raises(ValueError, match="contains no quantum operations"):
+            io.to_program(bb)
+
+    def test_gate_not_defined(self):
+        """Test empty program raises error"""
```
co9olguy: wrong docstring

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+        bb = blackbird.loads(bb_script)
+        prog = io.to_program(bb)
+
+        assert len(prog) == 1
+        assert prog.circuit[0].op.__class__.__name__ == "Dgate"
+        assert prog.circuit[0].op.p[0].x == 0.54
+        assert prog.circuit[0].reg[0].ind == 0
+
+    def test_gate_multimode(self):
+        """Test gate with keyword argument converts"""
```
co9olguy: wrong docstring?

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+
+        assert len(prog) == 1
+        assert prog.circuit[0].op.__class__.__name__ == "BSgate"
+        assert prog.circuit[0].op.p[0].x == 0.54
+        assert prog.circuit[0].op.p[1].x == np.pi
+        assert prog.circuit[0].reg[0].ind == 0
+        assert prog.circuit[0].reg[1].ind == 2
+
+    def test_compilation(self):
+        """Test automatic compilation using the target keyword"""
```
co9olguy: Not sure we want circuits to be implicitly compiled

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+        with open(filename, "w") as f:
+            f.write(test_prog_not_compiled)
+
+        res = sf.load(filename)
+
+        # check loaded program is the same as expected
+        for cmd1, cmd2 in zip(res.circuit, prog.circuit):
+            assert cmd1.op.__class__.__name__ == cmd2.op.__class__.__name__
+            assert cmd1.reg[0].ind == cmd2.reg[0].ind
+            if cmd1.op.p:
```
co9olguy: Why is the `if` necessary? Aren't these known circuits?

co9olguy: Same question for two tests below

tests/frontend/test_io.py
```diff
@@ -0,0 +1,543 @@
...
+
+        with open(filename, "r") as f:
+            res = sf.load(f)
+
+        # check loaded program is the same as expected
+        for cmd1, cmd2 in zip(res.circuit, prog.circuit):
+            assert cmd1.op.__class__.__name__ == cmd2.op.__class__.__name__
+            assert cmd1.reg[0].ind == cmd2.reg[0].ind
+            if cmd1.op.p:
+                assert np.all(cmd1.op.p[0].x == cmd2.op.p[0].x)
```
co9olguy: probably worth writing a separate function for this to avoid code duplication

josh146:
> Looks good. A couple minor questions, but the main thing that caught my attention is that programs are automatically and implicitly compiled when a bb file is loaded and it has a target. Do we want to do this?

@smite and I had a big discussion about this as I coded it, and went through all possible user expectations, and this one made kind of made the most sense.

One alternative would be to 'trick' the program into thinking it was already compiled, i.e.,
```python
prog.backend = bb.target['name']
```
but we thought this would lead to too much unexpected behaviour.

The final alternative would be to just ignore the `target` keyword, but this also seems unexpected from a user point of view.

strawberryfields/io.py
josh146: What do you mean by parametrized measurements?

strawberryfields/io.py
josh146: I definitely think so. Even `p.x` is not very explanatory, `p.val` would be better.

However, it's still worth thinking about the future of the parameter class - do we want to replace it with something like SymPy, which is now easy to pip install, and potentially could do what the existing parameter class does and more.

strawberryfields/io.py
josh146: lol you had the same thought as me two comments up

strawberryfields/io.py
josh146: Ah, but this is not post-processing, but mid-quantum simulation classical processing. So the remote backend will need the serialized function, it's not something that can be done locally.

I'm leaning towards a mixture of (a) SymPy approach for serializing any small function allowed by blackbird (i.e., basic arithmetic, sqrt, trig functions, exp, log), and (b) raising an error ("Function not supported on remote backend, please use a local backend") if the RegRefTransform includes anything more complicated.

strawberryfields/io.py
josh146: Because the `func_str` is set by the user, and doesn't need to correlate to the RegRefTransform function at all :laughing:.

For example, `Dgate(RR(q[0], lambda x: x**2, func_str='hi'))`.

This is really something that can only be solved by allowing symbolic functions, i.e.,
```python
q0 = sympy.Symbol('q0')
Dgate(q0**2) | q[1]
```

We can then access a string representation of the function,

```python
>>> func = q0**2`
>>> str(func)
q0**2
```

or use the Lambdify ability of SymPy to turn it into a evaluate-able Python function:

```python
>>> f = sympy.lambdify(q[0], func)
>>> f(5)
25
```

strawberryfields/io.py
josh146: Oops. My syntax highlighter doesn't highlight it. Okay to leave it as is?

strawberryfields/io.py
josh146: I copied this from the NumPy documentation, where they use the term `file-like` to describe any object that supports the `.write()` method.

tests/frontend/test_io.py
josh146: It does ('all imports should come at the top of the file'), but this has always looked cleaner to me. Want me to move it down to below the imports?

strawberryfields/io.py
josh146: I would argue for implicit conversion, personally.

strawberryfields/io.py
josh146: I'll change the argument to `f`?

josh146: Done.

tests/frontend/test_io.py
josh146: Resolving this comment since we have moved the discussion below.

tests/frontend/test_io.py
josh146: Solely because one of the gates `Vac` has no arguments, so `cmd1.op.p == []`. This was just a quick way to exclude gates that fit this criteria (no args).

tests/frontend/test_io.py
josh146: done

strawberryfields/io.py
josh146: done

strawberryfields/io.py
josh146: image

strawberryfields/io.py
co9olguy: I'm not convinced :thinking:

tests/frontend/test_io.py
```diff
@@ -320,18 +319,18 @@ def test_empty_program(self):
...
     def test_gate_not_defined(self):
-        """Test empty program raises error"""
+        """Test unknown gate raises error"""
 
         bb_script = """\
         name test_program
         version 1.0
 
-        Toffoli | [0, 1]
+        np | [0, 1]
```
co9olguy: How come you changed this?

tests/frontend/test_io.py
```diff
@@ -493,6 +492,14 @@ def test_invalid_file(self, prog, tmpdir):
         with pytest.raises(ValueError, match="must be a string, path"):
             sf.load(1)
 
+    def assert_programs_equal(self, prog1, prog2):
```
co9olguy: Nice!

strawberryfields/io.py
co9olguy: I mean a new measurement operator that might take a parameter via a keyword argument. Currently the only real kwarg is `select`, but if we add a new one, we'll have to adjust the code here as well, no?

tests/frontend/test_io.py
josh146: I realised there was a vulnerability with using
```python
try:
    getattr(ops, a_user_provided_string)(*args, **kwargs)
except Attribute error:
    raise Exception
```

in that 

1. Python turns _everything_ defined in the module as an accessible module attribute, i.e. if NumPy is imported, then `ops.np` exists.

2. The try/except pattern always causes the `getattr` line to be executed, and only fails if the attribute doesn't exist.

So, a malicious user would be able to do something like this:

```
np.io.something_malicious('rm -rf /') | [0, 1, 2]
```
and it could potentially be executed.

The new logic uses the `__all__` dictionary to verify the quantum operation is whitelisted, _before_ calling `getattr`.

strawberryfields/io.py
co9olguy: Let's make an issue to refactor the names `p` and `x` to make them more clear.

The other question about the future of the Parameter class and sympy seems unrelated to the original comment and PR, so I won't answer here.

strawberryfields/io.py
josh146: Oh. Yes, that's right. `select` is a weird edge case in that, it is the only kwarg used anywhere, and for some reason we store it as a named attribute.

strawberryfields/io.py
josh146: :+1:

strawberryfields/io.py
co9olguy: I think we should not allow inline processing for remote backends atm

strawberryfields/io.py
co9olguy: Based on further discussion, it makes more sense for the basic load methods to just do deserialization, and no validation. Any target that appears in the blackbird program can be added as an attribute to the corresponding `Program` instance

tests/frontend/test_io.py
co9olguy: yup

josh146:
Autocompilation has been removed. Now, if the target is provided, the attribute `program.backend` is set.

Todo: add ability to Program.compile() to use existing `prog.backend` attribute if no argument is provided

strawberryfields/io.py
josh146: Removed from the PR.

tests/frontend/test_io.py
josh146: done

PR #69, opened 2019-04-25
Utils module fixed.

smite:

The `utils` module and the associated tests should now work.
Also updated some docs.

### DIFF
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -138,7 +138,6 @@ from scipy.special import factorial as fac
 
 import strawberryfields as sf
 from .program import _convert, Command
-from .backends import load_backend
 from .ops import Gate, Channel, Ket
 # pylint: disable=abstract-method,ungrouped-imports,
 
--- a/strawberryfields/utils.py
+++ b/strawberryfields/utils.py
@@ -88,31 +88,30 @@ Decorators
 ----------
 
 The :class:`~.strawberryfields.utils.operation` decorator allows functions containing quantum operations
-acting on a qumode to be used as an operation itself within an engine context.
+acting on a qumode to be used as an operation itself within a :class:`.Program` context.
 
 .. autosummary::
    operation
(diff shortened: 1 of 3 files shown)


strawberryfields/utils.py
```diff
@@ -121,31 +120,26 @@
...
-from numpy.polynomial.hermite import hermval as H
-
+from numpy.polynomial.hermite import hermval
 import scipy as sp
 from scipy.special import factorial as fac
 
-from .program import _convert
+import strawberryfields as sf
+from .program import _convert, Command
 from .backends import load_backend
```
josh146: Since `load_backend` is no longer used, this line can be removed.

tests/integration/test_utils_integration.py
```diff
@@ -436,28 +426,22 @@ def test_extract_arbitrary_unitary_two_modes_vectorized(
...
-            S | q[0]
-            B | q
-            S | q[1]
-            B | q
-
-        eng, q = sf.Engine(2)
 
-        with eng:
+        prog = sf.Program(p0)
+        with prog.context as q:
```
co9olguy: How come it was split up this way?

tests/integration/test_utils_integration.py
```diff
@@ -590,52 +564,47 @@ def setup_two_mode_circuit(self, setup_eng, cutoff):
...
             S | q[0]
             B | q
             S | q[1]
             B | q
 
-        rho = eng_ref.run().dm()
-        return eng, rho, initial_state
+        rho = eng_ref.run([p0, prog]).dm()
```
co9olguy: same question as above about splitting up into two pieces

josh146:
This got merged at the same time your review appeared @co9olguy! I'm trying to think what the best approach is.

I could revert the merge, which rolls-back master pre-merge, and creates a new PR for Ville to address your changes in.

PR #72, opened 2019-04-25
Created a pytest mark for broken tests

josh146:
To mark tests that will currently fail due to the underlying code-base being broken, I've created the new pytest mark `pytest.mark.broken`, as opposed to manually skipping the test. This allows us to easily find and fix the broken tests in the future as the code base regains stability.

Use this at the top of a test module:
```python
pytestmark = pytest.mark.broken
```
or as a decorator on a test function/test class:
```python
@pytest.mark.broken
```
along with a comment as to _why_ the test is currently broken.


josh146:
Great, thanks @smite!

PR #74, opened 2019-04-26
Make sure tests run succesfully with a nonstandard hbar value

smite:

Try `hbar = 1.7` (say), see which tests fail, fix them.

### DIFF
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1828,8 +1828,11 @@ class Gaussian(Preparation, Decomposition):
             th, self.S = williamson(V, tol=tol)
             self.pure = np.abs(np.linalg.det(V) - (self.hbar/2)**(2*self.ns)) < tol
             self.nbar = np.diag(th)[:self.ns]/self.hbar - 0.5
+            super().__init__([V, r])
+        else:
+            # scale state
+            super().__init__([V*2/hbar, r/np.sqrt(2*hbar)])
 
-        super().__init__([V, r])
         # FIXME merge() probably does not work for Gaussians if r is not zero?
 
     def _apply(self, reg, backend, **kwargs):
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1824,14 +1824,16 @@ class Gaussian(Preparation, Decomposition):
         self.p_disp = r[self.ns:]
 
         self.decomp = decomp
+
         if decomp:
             th, self.S = williamson(V, tol=tol)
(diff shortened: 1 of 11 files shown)


tests/conftest.py
```diff
@@ -30,7 +30,7 @@
 TOL = 1e-3
 CUTOFF = 6
 ALPHA = 0.1
-HBAR = 2
+HBAR = 1.7
```
josh146:
```suggestion
HBAR = 2
```

strawberryfields/ops.py
```diff
@@ -1828,8 +1828,11 @@ def __init__(self, V, r=None, decomp=True, hbar=2, tol=1e-6):
             th, self.S = williamson(V, tol=tol)
             self.pure = np.abs(np.linalg.det(V) - (self.hbar/2)**(2*self.ns)) < tol
             self.nbar = np.diag(th)[:self.ns]/self.hbar - 0.5
+            super().__init__([V, r])
+        else:
+            # scale state
+            super().__init__([V*2/hbar, r/np.sqrt(2*hbar)])
```
smite: Can this be right? Previously when hbar=2 the tests ran fine, but this becomes `[V, r/2]`, not  `[V, r]` like above.

josh146: The issue is that the tests aren't testing displacement, see `test_decompositions_integration.py`:

image

tests/conftest.py
```diff
@@ -30,7 +30,7 @@
 TOL = 1e-3
 CUTOFF = 6
 ALPHA = 0.1
-HBAR = 1.7
+HBAR = 2
```
smite: Is there a reason for using HBAR=2 in the tests? I would think a randomly picked (non-half-integer) value would be the best choice since it has the highest chance of flushing out bugs in the future.

josh146: No reason, apart from `HBAR = 2` being the default for Strawberry Fields. I like this idea - do you think we could remove the `test_hbar` test? This test is one of the slower ones, due to using sampling.

strawberryfields/ops.py
josh146: This is now fixed @smite

tests/conftest.py
smite: Yeah, the sampling takes a long time, I'll see if it could be sped up or avoided somehow.

tests/conftest.py
co9olguy: Was this set back to 2 in merged version?

Issue #75, opened 2019-04-30
Why add batch on last dimension?


ziofil:
It may not matter in the end, but shouldn't [these lines](https://github.com/XanaduAI/strawberryfields/blob/80191c7a0d6b3430ba31228e0e53ffb5c496dfab/strawberryfields/backends/tfbackend/ops.py#L292) be 

```python
t = tf.expand_dims(t, 0)
r = tf.expand_dims(r, 0)
```

to match what is done in the other ops?

co9olguy:
Thanks @ziofil! 
I don't expect it to make any functional difference, but it is clearer this way when people are reading code.
Fixed in latest master

PR #77, opened 2019-04-30
Changed index in ops.py

co9olguy:
**Description of the Change:** Tiny update to the `ops.py` file in tf backend

**Benefits:** code is clearer

**Possible Drawbacks:** None that I can see

**Related GitHub Issues:** #75


co9olguy:
Closes #75

PR #78, opened 2019-05-06
Eliminate hbar from backend API.

smite:

This PR simplifies the way hbar is handled. Previously many different objects kept track of the hbar value associated with the object data. This enabled practices that are not useful and that could be outright harmful, like combining gates with different hbar values in the same circuit, and made the call interfaces cluttered.

Changes:

* The backend API is now entirely hbar-independent, i.e. every backend API method is defined in terms of `a` and `a^\dagger` only, not x and p.
* The backends always explicitly use hbar=2 internally.
* hbar is now a global, frontend-only variable that the user can set at the beginning of the session. It is used at the `Operation.apply()` level to scale the inputs and outputs of the backend API calls as needed, and inside the `State` objects.
* The only backend API calls that need to do hbar scaling for the input parameters are the X, Z, and V gates, the Gaussian state decomposition, and homodyne measurements (both the returned value and postselection argument are scaled).

**Benefits:**

Makes the code more readable, debuggable and maintainable. Simplifies the UX.

**Possible Drawbacks:**

None as far as I can tell.

### DIFF
--- a/strawberryfields/backends/base.py
+++ b/strawberryfields/backends/base.py
@@ -30,7 +30,7 @@ This module implements the backend API. It contains the classes
 
 as well as a few methods which apply only to the Gaussian backend.
 
-.. note:: The backend API is hbar-independent.
+.. note:: The backend API is :math:`\hbar` independent.
           Internally the Strawberry Fields backends use :math:`\hbar=2`.
 
 
@@ -463,7 +463,7 @@ class BaseBackend:
            \hat{q}_\phi = \sqrt{2/\hbar}(\cos(\phi)\hat{x} +\sin(\phi)\hat{p}) = e^{-i\phi} \hat{a} +e^{i\phi} \hat{a}^\dagger.
 
         .. note::
-           This method is hbar-independent.
+           This method is :math:`\hbar` independent.
            The returned values can be converted to conventional position/momentum
            eigenvalues by multiplying them with :math:`\sqrt{\hbar/2}`.
 
@@ -592,16 +592,16 @@ class BaseFock(BaseBackend):
         to the specified mode.
 
         .. note::
-           This method is hbar-independent.
-           The usual definition of the cubic phase gate is hbar-dependent:
+           This method is :math:`\hbar` independent.
+           The usual definition of the cubic phase gate is :math:`\hbar` dependent:
 
            .. math::
               V(\gamma) = \exp\left(i \frac{\gamma}{3\hbar} \hat{x}^3\right) = \exp\left(i \frac{\gamma \sqrt{\hbar/2}}{6} (\hat{a} +\hat{a}^\dagger)^3\right).
 
            Hence the cubic phase gate `V(gamma)` is executed on a backend by scaling the
            `gamma` parameter by :math:`\sqrt{\hbar/2}` and then passing it to this method,
-           much in the way the hbar-dependent `X` and `Z` gates are implemented through the
-           hbar-independent :meth:`~BaseBackend.displacement` method.
+           much in the way the :math:`\hbar` dependent `X` and `Z` gates are implemented through the
+           :math:`\hbar` independent :meth:`~BaseBackend.displacement` method.
 
         .. warning::
             The cubic phase gate can suffer heavily from numerical inaccuracies
@@ -694,7 +694,7 @@ class BaseGaussian(BaseBackend):
         The requested modes are traced out and replaced with the given Gaussian state.
 
         .. note::
-           This method is hbar-independent.
+           This method is :math:`\hbar` independent.
            The input arrays are the means and covariance of the
            :math:`a+a^\dagger` and :math:`-i(a-a^\dagger)` operators.
            They are obtained by dividing the xp means by :math:`\sqrt{\hbar/2}`
--- a/strawberryfields/backends/base.py
+++ b/strawberryfields/backends/base.py
@@ -30,8 +30,9 @@ This module implements the backend API. It contains the classes
 
 as well as a few methods which apply only to the Gaussian backend.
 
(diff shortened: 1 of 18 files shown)


strawberryfields/backends/base.py
```diff
@@ -30,8 +30,9 @@
 
 as well as a few methods which apply only to the Gaussian backend.
 
-.. note:: The Strawberry Fields backends by default assume :math:`\hbar=2`, however
-    different conventions may be chosen when calling :meth:`~.BaseBackend.begin_circuit`
+.. note:: The backend API is hbar-independent.
```
josh146:
```suggestion
.. note:: The backend API is :math:`hbar` independent.
```

strawberryfields/backends/base.py
```diff
@@ -452,8 +455,17 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
...
-        :math:`f(q) = \bra{q}_x R^\dagger(\phi) \rho R(\phi) \ket{q}_x`
+        :math:`f(q) = \bra{q_\phi} R^\dagger(\phi) \rho R(\phi) \ket{q_\phi}`
         and returns the sampled value.
+        Here :math:`\ket{q_\phi}` is the eigenstate of the operator
+
+        .. math::
+           \hat{q}_\phi = \sqrt{2/\hbar}(\cos(\phi)\hat{x} +\sin(\phi)\hat{p}) = e^{-i\phi} \hat{a} +e^{i\phi} \hat{a}^\dagger.
+
+        .. note::
+           This method is hbar-independent.
```
josh146:
```suggestion
           This method is :math:`\hbar` independent.
```

strawberryfields/backends/base.py
```diff
@@ -494,7 +506,7 @@ def state(self, modes=None, **kwargs):
                 requesting the modes=[3,1] results in a two mode state being returned with the first mode being
                 subsystem 3 and the second mode being subsystem 1 of simulator.
         Returns:
-            An instance of the Strawberry Fields State class, suited to the particular backend.
+            BaseState: state description, suited to the particular backend
```
josh146: Good catch

strawberryfields/backends/base.py
```diff
@@ -569,9 +581,28 @@ def prepare_dm_state(self, state, modes):
...
 
+        Applies the operation
+
+        .. math::
+           \exp\left(i \frac{\gamma'}{6} (\hat{a} +\hat{a}^\dagger)^3\right)
+
+        to the specified mode.
+
+        .. note::
+           This method is hbar-independent.
```
josh146:
```suggestion
           This method is :math:`\hbar` independent.
```

strawberryfields/ops.py
```diff
@@ -1844,7 +1838,8 @@ def _apply(self, reg, backend, **kwargs):
             raise NotImplementedError
 
         p = _unwrap(self.p)
-        backend.prepare_gaussian_state(p[1], p[0], reg)
+        s = sqrt(sf.hbar / 2)  # scaling factor, since the backend API call is hbar-independent
+        backend.prepare_gaussian_state(p[1]/s, p[0]/(s*s), reg)
```
josh146: Nice!

strawberryfields/ops.py
```diff
@@ -1799,17 +1797,17 @@ class Gaussian(Preparation, Decomposition):
...
     # pylint: disable=too-many-instance-attributes
     ns = None
 
-    def __init__(self, V, r=None, decomp=True, hbar=2, tol=1e-6):
+    def __init__(self, V, r=None, decomp=True, tol=1e-6):
         # TODO NOTE: there is no contextual hbar value anymore, is the hbar actually necessary here?
-        self.hbar = hbar
+
+        #: float: value of :math:`\hbar` used in the definition of the :math:`\x` and :math:`\p` quadrature operators
+        # TODO can we just divide V by hbar/2 here and remove hbar from further expressions?
```
josh146: Probably, I think

tests/integration/test_algorithms.py
```diff
@@ -35,8 +35,8 @@ def test_teleportation_fidelity(setup_eng, pure):
         BS | (q[1], q[2])
 
         BS | (q[0], q[1])
-        MeasureHomodyne(0, select=0) | q[0]
-        MeasureHomodyne(np.pi / 2, select=0) | q[1]
+        MeasureHomodyne(0, select=0.07) | q[0]
+        MeasureHomodyne(np.pi / 2, select=0.1) | q[1]
```
josh146: How come this was changed?

strawberryfields/ops.py
```diff
@@ -898,14 +894,6 @@ class Catstate(Preparation):
 
     where :math:`N = \sqrt{2 (1+\cos(\phi)e^{-2|\alpha|^2})}` is the normalization factor.
 
-    This is a Strawberry Fields quantum gate operator, and thus is used within an engine
```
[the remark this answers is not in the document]
smite: I guess. I removed the example because no other Operation has them, and it's yet another item to be updated if/when the syntax changes.

strawberryfields/ops.py
smite: Maybe I'll give it a go later, but for now it ought to work.

tests/integration/test_algorithms.py
smite: If the postselected value is zero, scaling errors do not show up.

josh146: :+1:

smite: Wait no, that's wrong. At the moment `test_teleportation_fidelity` does not use the measurement results to displace the output mode, so if the postselected value is nonzero the result is wrong. I'll fix it.

smite: Fixed, I added the required displacements.

PR #79, opened 2019-05-07
Adds test file for program unit tests

josh146:

* Reorders the test suite, creates a new `frontend/test_program.py` file containing program unit tests
* Removes program specific unit tests from `test_engine.py`
* Fixes the `test_circuitdrawer.py` tests to work with the new program object


tests/frontend/test_program.py
```diff
@@ -0,0 +1,308 @@
...
+class TestRegRefs:
+    """Testing register references."""
+
+    def test_corrupt_index(self, prog):
+        """User messes up the RegRef indices."""
+        with prog.context as q:
+            q[0].ind = 1
+            with pytest.raises(
+                program.RegRefError, match="RegRef state has become inconsistent"
+            ):
```
smite: Could we increase the allowed line length to 100--120 chars to prevent black splitting up lines like this?

smite:
Looks good, I'll add some more Engine-specific tests once I have the LocalEngine class done.

tests/frontend/test_program.py
smite: I mean this is not the worst formatting but I tend to use slightly longer lines myself...

josh146: We should pick a standard Black line length to ensure that we are all using the same formatting, @co9olguy @trbromley @jswinarton

tests/frontend/test_circuitdrawer.py
```diff
@@ -942,15 +941,15 @@ def test_compile(self, tmpdir):
...
 
-        with eng:
+        with prog.context as q:
             ops.Dgate(1) | (q[1])
             ops.Rgate(1) | (q[1])
             ops.S2gate(1) | (q[0], q[1])
 
         subdir = tmpdir.join("subdir")
-        document = eng.draw_circuit(print_queued_ops=True, tex_dir=subdir)[0]
+        document = prog.draw_circuit(tex_dir=subdir)[0]
```
co9olguy: didn't realize before how comprehensive this file was. 1000 lines :)

tests/frontend/test_program.py
```diff
@@ -0,0 +1,308 @@
...
+            D | r
+        assert cont.can_follow(prog)
+        assert prog.reg_refs == cont.init_reg_refs
+        assert prog.unused_indices == cont.init_unused_indices
+
+    def test_print_commands(self, eng, prog):
+        """Program.print and Engine.print_applied return correct strings."""
+        prog = sf.Program(2)
+        res = []
+        print_fn = lambda x: res.append(x.__str__())
```
co9olguy: our usage of `res` and `print_fn` is a bit confusing to follow in the tests; it's not easy to follow the flow. Perhaps some brief explanatory comment could be useful here

tests/frontend/test_program.py
co9olguy: go ahead and just pick one

josh146: Can I make an executive decision and say 100 chars per line

tests/frontend/test_program.py
josh146: done

PR #80, opened 2019-05-09
LocalEngine class

smite:

Introduces the BaseEngine ABC and the LocalEngine child class. Engine is kept as an alias for LocalEngine.

The Engine API has been changed slightly:

* `LocalEngine.run()` returns a `Result` object that contains both a state object and measurement samples.
* The complicated and error-prone way kwargs were used has been simplified by introducing the new kwargs-like arguments `backend_options` and `state_options`.  `LocalEngine.run()` passes the actual kwargs only to `Operation.apply()`

**Benefits:**

Now we can start working on the remote engines.

### DIFF
--- a/doc/op_conventions.rst
+++ b/doc/op_conventions.rst
@@ -34,7 +34,10 @@ In this section, we provide the definitions of various quantum operations used b
    conventions/others
 
 
-.. note:: In Strawberry Fields we use the convention :math:`\hbar=2` by default, but other conventions can also be chosen on engine :func:`initialization <strawberryfields.Engine>`. In this document we keep :math:`\hbar` explicit.
+.. note:: In Strawberry Fields we use the convention :math:`\hbar=2` by default, but other
+   conventions can also be chosen by setting the global variable :py:data:`strawberryfields.hbar`
+   at the beginning of a session.
+   In this document we keep :math:`\hbar` explicit.
 
 
 .. note::
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -94,8 +94,7 @@ class Result:
 
     def __str__(self):
         
-        return "Result: {} subsystems, state: {}\n samples: {}'.format(len(self.samples), self.state, self.samples)
-            + '\n  samples: ' + str(self.samples)
+        return 'Result: {} subsystems, state: {}\n samples: {}'.format(len(self.samples), self.state, self.samples)
 
 
 class BaseEngine(abc.ABC):
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -234,7 +234,7 @@ class BaseEngine(abc.ABC):
             # program execution was successful
             pass
 
-        return Result(self.samples)
+        return Result(self.samples.copy())
 
 
 class LocalEngine(BaseEngine):
@@ -293,9 +293,13 @@ class LocalEngine(BaseEngine):
         Extends :meth:`BaseEngine._run`.
 
         Args:
+            program (Program, Sequence[Program]): quantum programs to run
+            compile_options (Dict[str, Any]): keyword arguments for :meth:`.Program.compile`
             modes (None, Sequence[int]): Modes to be returned in the ``Result.state`` :class:`.BaseState` object.
                 An empty sequence [] means no state object is returned. None returns all the modes.
             state_options (Dict[str, Any]): keyword arguments for :meth:`.BaseBackend.state`
+
+        The ``kwargs`` keyword arguments are passed to :meth:`_run_program`.
 
         Returns:
             Result: results of the computation
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -93,7 +93,8 @@ class Result:
         self.samples = samples
 
     def __str__(self):
-        return self.__class__.__name__ + ': {} subsystems, state: {}'.format(len(self.samples), self.state)\
+        
+        return "Result: {} subsystems, state: {}\n samples: {}'.format(len(self.samples), self.state, self.samples)
             + '\n  samples: ' + str(self.samples)
 
 
@@ -183,7 +184,7 @@ class BaseEngine(abc.ABC):
         * The Program instance is compiled and optimized.
         * The compiled program is executed on the backend, updating the backend state.
         * The latest measurement results of each subsystem are stored
-          in the :class:`.RegRef` instances of the corresponding Program, as well as in self.samples.
+          in the :class:`.RegRef` instances of the corresponding Program, as well as in :attr:`~.samples`.
         * The compiled program is appended to self.run_progs.
 
         Finally, the result of the computation is returned.
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -104,7 +104,10 @@ class BaseEngine(abc.ABC):
         backend (str): backend short name
         backend_options (Dict[str, Any]): keyword arguments for the backend
     """
-    def __init__(self, backend, backend_options={}):
+    def __init__(self, backend, backend_options=None):
+        if backend_options is None:
+            backend_options = {}
+
         #: str: short name of the backend
         self.backend_name = backend
         #: Dict[str, Any]: keyword arguments for the backend
@@ -263,7 +266,10 @@ class LocalEngine(BaseEngine):
     def __str__(self):
         return self.__class__.__name__ + '({})'.format(self.backend_name)
 
-    def reset(self, backend_options={}):
+    def reset(self, backend_options=None):
+        if backend_options is None:
+            backend_options = {}
+
         super().reset(backend_options)
         self.backend_options.pop('batch_size', None)  # HACK to make tests work for now
         self.backend.reset(**self.backend_options)
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 """
-Quantum program executor engine
-===============================
+Execution engine
+================
 
 **Module name:** :mod:`strawberryfields.engine`
 
@@ -93,7 +93,7 @@ class Result:
         self.samples = samples
 
     def __str__(self):
-        
+        """String representation."""
         return 'Result: {} subsystems, state: {}\n samples: {}'.format(len(self.samples), self.state, self.samples)
 
 
@@ -175,14 +175,16 @@ class BaseEngine(abc.ABC):
             list[Command]: commands that were applied to the backend
         """
 
-    def _run(self, program, *, compile=True, **kwargs):
+    def _run(self, program, *, compile_options={}, **kwargs):
         """Execute the given programs by sending them to the backend.
 
+        If multiple Programs are given they will be executed sequentially as
+        parts of a single computation.
         For each :class:`Program` instance given as input, the following happens:
 
-        * The Program instance is compiled and optimized.
-        * The compiled program is executed on the backend, updating the backend state.
-        * The latest measurement results of each subsystem are stored
+        * The Program instance is compiled and optimized for the target backend.
+        * The compiled program is executed on the backend.
+        * The measurement results of each subsystem (if any) are stored
           in the :class:`.RegRef` instances of the corresponding Program, as well as in :attr:`~.samples`.
         * The compiled program is appended to self.run_progs.
 
@@ -190,7 +192,7 @@ class BaseEngine(abc.ABC):
 
         Args:
             program (Program, Sequence[Program]): quantum programs to run
-            compile (bool): If True, compile the Program instances before sending them to the backend.
+            compile_options (Dict[str, Any]): keyword arguments for :meth:`.Program.compile`
 
         The ``kwargs`` keyword arguments are passed to :meth:`_run_program`.
 
@@ -217,8 +219,8 @@ class BaseEngine(abc.ABC):
                         p.reg_refs[k].val = v
 
                 # if the program hasn't been compiled for this backend, do it now
-                if compile and p.backend != self.backend_name:
-                    p = p.compile(self.backend_name)
+                if p.backend != self.backend_name:
+                    p = p.compile(self.backend_name, **compile_options)
                 p.lock()
 
                 self._run_program(p, **kwargs)
@@ -281,18 +283,11 @@ class LocalEngine(BaseEngine):
                 # command is not applicable to the current backend type
                 raise NotApplicableError('The operation {} cannot be used with {}.'.format(cmd.op, self.backend)) from None
             except NotImplementedError:
-                # command not directly supported by backend API, try a decomposition instead
-                try:
-                    temp = cmd.op.decompose(cmd.reg)
-                    # run the decomposition
-                    applied_cmds = self._run_program(temp, **kwargs)
-                    applied.extend(applied_cmds)
-                except NotImplementedError as err:
-                    # simplify the error message by suppressing the previous exception
-                    raise err from None
+                # command not directly supported by backend API
+                raise NotImplementedError('The operation {} has not been implemented for {}.'.format(cmd.op, self.backend)) from None
         return applied
 
-    def run(self, program, *, compile=True, modes=None, state_options={}, **kwargs):
+    def run(self, program, *, compile_options={}, modes=None, state_options={}, **kwargs):
         """Execute the given programs by sending them to the backend.
 
         Extends :meth:`BaseEngine._run`.
@@ -306,7 +301,7 @@ class LocalEngine(BaseEngine):
             Result: results of the computation
         """
 
-        result = super()._run(program, compile=compile, **kwargs)
+        result = super()._run(program, compile_options=compile_options, **kwargs)
         if isinstance(modes, Sequence) and not modes:
             # empty sequence
             pass
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -20,10 +20,11 @@ Quantum program executor engine
 
 .. currentmodule:: strawberryfields.engine
 
-The :class:`BaseEngine` subclasses are responsible for communicating
-quantum programs represented by :class:`.Program` objects
-to a backend that could be e.g., a simulator, a hardware quantum processor, or a circuit drawer,
-and returning the result to the user.
+This module implements :class:`BaseEngine` and its subclasses that are responsible for
+communicating quantum programs represented by :class:`.Program` objects
+to a backend that could be e.g., a simulator, a hardware quantum processor,
+or a circuit drawer, and returning the result to the user.
+One can think of each BaseEngine instance as a separate quantum computation.
 
 A typical use looks like
 
@@ -39,10 +40,10 @@ Classes
    Result
 
 
-BaseEngine methods
-------------------
+LocalEngine methods
+-------------------
 
-.. currentmodule:: strawberryfields.engine.BaseEngine
+.. currentmodule:: strawberryfields.engine.LocalEngine
 
 .. autosummary::
    run
@@ -55,6 +56,7 @@ BaseEngine methods
     .. autosummary::
        _init_backend
        _run_program
+       _run
 
 
 .. currentmodule:: strawberryfields.engine
@@ -91,7 +93,8 @@ class Result:
         self.samples = samples
 
     def __str__(self):
-        return self.__class__.__name__ + ': {} subsystems, state: {}'.format(len(self.samples), self.state) + '\n' + str(self.samples)
+        return self.__class__.__name__ + ': {} subsystems, state: {}'.format(len(self.samples), self.state)\
+            + '\n  samples: ' + str(self.samples)
 
 
 class BaseEngine(abc.ABC):
@@ -116,7 +119,7 @@ class BaseEngine(abc.ABC):
         """String representation."""
 
     @abc.abstractmethod
-    def reset(self,  backend_options):
+    def reset(self, backend_options):
         r"""Re-initialize the quantum computation.
 
         Resets the state of the engine and the quantum circuit represented by the backend.
@@ -130,7 +133,7 @@ class BaseEngine(abc.ABC):
 
         Args:
            backend_options (Dict[str, Any]): keyword arguments for the backend,
-              overriding the old values
+              updating (overriding) old values
         """
         self.backend_options.update(backend_options)
         for p in self.run_progs:
@@ -162,7 +165,7 @@ class BaseEngine(abc.ABC):
 
     @abc.abstractmethod
     def _run_program(self, prog, **kwargs):
-        """Execute a program on the backend.
+        """Execute a single program on the backend.
 
         This method should not be called directly.
 
@@ -172,23 +175,24 @@ class BaseEngine(abc.ABC):
             list[Command]: commands that were applied to the backend
         """
 
-    def run(self, program, *, modes=None, compile=True, **kwargs):
-        """Execute the given program by sending it to the backend.
+    def _run(self, program, *, compile=True, **kwargs):
+        """Execute the given programs by sending them to the backend.
 
-        * Each executed Program is compiled and optimized
-        * The compiled versions of the executed Programs are appended to self.run_progs.
-        * The backend state is updated.
-        * The result of the computation is returned.
-        * Additionally, the latest measurement results of each subsystem are stored
-          in the :class:`.RegRef` instances of the corresponding Program.
+        For each :class:`Program` instance given as input, the following happens:
+
+        * The Program instance is compiled and optimized.
+        * The compiled program is executed on the backend, updating the backend state.
+        * The latest measurement results of each subsystem are stored
+          in the :class:`.RegRef` instances of the corresponding Program, as well as in self.samples.
+        * The compiled program is appended to self.run_progs.
+
+        Finally, the result of the computation is returned.
 
         Args:
             program (Program, Sequence[Program]): quantum programs to run
-            modes (None, Sequence[int]): Modes to be returned in the state object.
-                An empty sequence [] means no state object is returned. None returns all the modes.
             compile (bool): If True, compile the Program instances before sending them to the backend.
 
-        The keyword args are passed to :meth:`_run_program`.
+        The ``kwargs`` keyword arguments are passed to :meth:`_run_program`.
 
         Returns:
             Result: results of the computation
@@ -203,13 +207,14 @@ class BaseEngine(abc.ABC):
                     # initialize the backend
                     self._init_backend(p.init_num_subsystems)
                 else:
+                    # there was a previous program segment
                     if not p.can_follow(prev):
-                        raise RuntimeError("Register mismatch after Program {}, '{}'.".format(len(self.run_progs)-1, prev.name))
+                        raise RuntimeError("Register mismatch: program {}, '{}'.".format(len(self.run_progs), p.name))
 
-                    # todo copy vals from self.samples in case same prog is used in two engines....
-                    # copy the measured values in RegRefs over from prev
-                    for k, v in prev.reg_refs.items():
-                        p.reg_refs[k].val = v.val
+                    # Copy the latest measured values in the RegRefs of p.
+                    # We cannot copy from prev directly because it could be used in more than one engine.
+                    for k, v in enumerate(self.samples):
+                        p.reg_refs[k].val = v
 
                 # if the program hasn't been compiled for this backend, do it now
                 if compile and p.backend != self.backend_name:
@@ -218,16 +223,14 @@ class BaseEngine(abc.ABC):
 
                 self._run_program(p, **kwargs)
                 self.run_progs.append(p)
+                # store the latest measurement results
+                self.samples = [p.reg_refs[k].val for k in sorted(p.reg_refs)]
                 prev = p
         except Exception as e:
             raise e
         else:
             # program execution was successful
             pass
-
-        if prev:
-            # store the latest measurement results
-            self.samples = [prev.reg_refs[k].val for k in sorted(prev.reg_refs)]
 
         return Result(self.samples)
 
@@ -259,19 +262,6 @@ class LocalEngine(BaseEngine):
         return self.__class__.__name__ + '({})'.format(self.backend_name)
 
     def reset(self, backend_options={}):
-        r"""Re-initialize the backend state to vacuum.
-
-        Resets the state of the quantum circuit represented by the backend.
-
-        * The original number of modes is restored.
-        * All modes are reset to the vacuum state.
-        * All RegRefs of previously run Programs are cleared of measured values.
-        * List of previously run Progams is cleared.
-
-        Note that the reset does nothing to any Program objects in existence, beyond erasing the measured values.
-
-        The keyword args are passed on to :meth:`strawberryfields.backends.base.BaseBackend.reset`.
-        """
         super().reset(backend_options)
         self.backend_options.pop('batch_size', None)  # HACK to make tests work for now
         self.backend.reset(**self.backend_options)
@@ -285,7 +275,7 @@ class LocalEngine(BaseEngine):
         for cmd in prog.circuit:
             try:
                 # try to apply it to the backend
-                res = cmd.op.apply(cmd.reg, self.backend, **kwargs)  # TODO we could also handle storing the measurement vals here
+                cmd.op.apply(cmd.reg, self.backend, **kwargs)  # NOTE we could also handle storing measured vals here
                 applied.append(cmd)
             except NotApplicableError:
                 # command is not applicable to the current backend type
@@ -302,13 +292,26 @@ class LocalEngine(BaseEngine):
                     raise err from None
         return applied
 
-    def run(self, program, *, modes=None, compile=True, **kwargs):
-        result = super().run(program, modes=modes, compile=compile, **kwargs)
+    def run(self, program, *, compile=True, modes=None, state_options={}, **kwargs):
+        """Execute the given programs by sending them to the backend.
+
+        Extends :meth:`BaseEngine._run`.
+
+        Args:
+            modes (None, Sequence[int]): Modes to be returned in the ``Result.state`` :class:`.BaseState` object.
+                An empty sequence [] means no state object is returned. None returns all the modes.
+            state_options (Dict[str, Any]): keyword arguments for :meth:`.BaseBackend.state`
+
+        Returns:
+            Result: results of the computation
+        """
+
+        result = super()._run(program, compile=compile, **kwargs)
         if isinstance(modes, Sequence) and not modes:
             # empty sequence
             pass
         else:
-            result.state = self.backend.state(modes=modes)  # TODO tfbackend.state can use kwargs
+            result.state = self.backend.state(modes, **state_options)  # tfbackend.state can use kwargs
         return result
 
 
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -13,49 +13,47 @@
 # limitations under the License.
 
 """
-Quantum compiler engine
-========================================================
+Quantum program executor engine
+===============================
 
 **Module name:** :mod:`strawberryfields.engine`
 
 .. currentmodule:: strawberryfields.engine
 
-The :class:`Engine` class is responsible for communicating
+The :class:`BaseEngine` subclasses are responsible for communicating
 quantum programs represented by :class:`.Program` objects
-to a backend that could be e.g., a simulator, a hardware quantum processor, or a circuit drawer.
+to a backend that could be e.g., a simulator, a hardware quantum processor, or a circuit drawer,
+and returning the result to the user.
 
 A typical use looks like
-::
 
-  prog = sf.Program(num_subsystems)
-  with prog.context as q:
-      Coherent(0.5)  | q[0]
-      Vac            | q[1]
-      Sgate(2)       | q[1]
-      Dgate(0.5)     | q[0]
-      BSgate(1)      | q[0:2]
-      Dgate(0.5).H   | q[0]
-      Measure        | q
-  eng = sf.Engine(backend='fock')
-  eng.run(prog, cutoff_dim=5)
-  v1 = prog.register[1].val
+.. include:: example_use.rst
 
 
-Engine methods
---------------
+Classes
+-------
 
-.. currentmodule:: strawberryfields.engine.Engine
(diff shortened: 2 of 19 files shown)


josh146:
Just doing code-review now

smite:
I fixed the failing test (TF backend returns an array of samples with batching on).
I can't find a complete explanation on how the batching is supposed to work, just hints here and there. Is it like multi-shot simulation, except that gate parameters can also be given as vectors with parameter `a[i]` corresponding to shot `i`?

strawberryfields/engine.py
```diff
@@ -182,58 +285,34 @@ def _run_program_locally(self, prog, **kwargs):
...
-            prev = None  # previous program segment
-            for p in program:
-                if self.run_progs:
-                    prev = self.run_progs[-1]
-                else:
-                    # initialize the backend TODO where should this happen? c.f. the fixtures in conftest.py
-                    self.backend.begin_circuit(num_subsystems=p.init_num_subsystems, **self.kwargs)
+            modes (None, Sequence[int]): Modes to be returned in the ``Result.state`` :class:`.BaseState` object.
+                An empty sequence [] means no state object is returned. None returns all the modes.
+            state_options (Dict[str, Any]): keyword arguments for :meth:`.BaseBackend.state`
```
josh146: What `state_options` are there? I don't remember this being in the previous iteration (from what I recall, only the backends themselves ever pass args to the State constructor)

strawberryfields/engine.py
```diff
@@ -182,58 +285,34 @@ def _run_program_locally(self, prog, **kwargs):
...
-            # TODO reset the backend to the last checkpoint here?
-            raise e
-        else:
-            # program execution was successful
+        result = super()._run(program, compile=compile, **kwargs)
+        if isinstance(modes, Sequence) and not modes:
+            # empty sequence
             pass
+        else:
+            result.state = self.backend.state(modes, **state_options)  # tfbackend.state can use kwargs
```
josh146: Oh, `state_options` is for the TF backend only?

strawberryfields/ops.py
```diff
@@ -24,7 +24,7 @@
 .. note::
 
   In the :mod:`strawberryfields.ops` API we use the convention :math:`\hbar=2` by default, however
-  this can be changed on engine initialisation.
+  this can be changed using the global variable :py:data:`strawberryfields.hbar`.
```
josh146: nice :+1:

strawberryfields/ops.py
```diff
@@ -33,101 +33,97 @@
...
   by supplying registers as the argument to an :class:`Operation`, in which case the measurement may be deferred,
   i.e., we may symbolically use the measurement result before it exists::
 
-    eng, (alice, bob) = sf.Engine(2)
-    with eng:
-      Measure   | alice
-      Dgate(alice) | bob
-    eng.run()
+    with prog.context as q:
+        alice, bob = q
```
josh146: Due to tuple unpacking, would
```python
with prog.context as alice, bob:
```
also work?

strawberryfields/ops.py
```diff
@@ -1603,7 +1598,7 @@ class Interferometer(Decomposition):
     Args:
         U (array): an :math:`N\times N` complex unitary matrix.
         tol (float): the tolerance used when checking if the matrix is unitary:
-            :math:`|UU^\dagger-I| \leq tol`
+            :math:`|UU^\dagger-I| \leq` tol
```
josh146: Should this be
```suggestion
            :math:`|UU^\dagger-I| \leq \text{tol}`
```

strawberryfields/ops.py
```diff
@@ -1652,10 +1647,10 @@ class GraphEmbed(Decomposition):
         A (array): an :math:`N\times N` complex or real symmetric matrix
         max_mean_photon (float): threshold value. It guarantees that the mode with
             the largest squeezing has ``max_mean_photon`` as the mean photon number
-            i.e., :math:`sinh(r_{max})^2 == max_mean_photon`
+            i.e., :math:`sinh(r_{max})^2 ==` max_mean_photon
```
josh146: Likewise,
```suggestion
            i.e., :math:`sinh(r_{max})^2 = \text{max_mean_photon}`
```
It looks a bit odd having `max_mean_photon` outside the equation in normal font.

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
 
-* Once the entire program is inputted, Engine optimizes it. This involves merging and commuting Commands
-  inside the circuit graph, the building of which requires knowledge of their dependencies, both direct and Parameter-based.
+* Once the entire program is inputted, it is compiled and optimized. This involves
+  merging and commuting Commands inside the circuit graph. The circuit graph is built
+  using the knowledge of which subsystems the Commands act and depend on.
 
-* Merging two :class:`~ops.Gate` instances of the same subclass involves
+* Merging two :class:`.Gate` instances of the same subclass involves
   adding their first parameters after equality-comparing the others. This is easily done if
```
josh146: This is not strictly true - gates which accept matrix parameters require matrix multiplication, and some require addition of more than one parameter (`GaussianState`). We should mark this an issue to fix/implement.

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
-  RegRefTransform-based Parameters using :func:`Parameter.evaluate` (other types of Parameters are simply passed through).
-  The parameter values and the subsystem indices are passed to :func:`~ops.Operation._apply`.
+* :meth:`Operation.apply` evaluates the numeric value of any
+  RegRefTransform-based Parameters using :meth:`Parameter.evaluate` (other types of Parameters are simply passed through).
+  The parameter values and the subsystem indices are passed to :meth:`Operation._apply`.
 
-* :func:`~ops.Operation._apply` "unwraps" the Parameter instances. There are three different cases:
+* :meth:`~ops.Operation._apply` "unwraps" the Parameter instances. There are three different cases:
 
   1. We still need to do some arithmetic, unwrap after it is done using p.x.
```
josh146:
```suggestion
  1. We still need to do some arithmetic, unwrap after it is done using `p.x`.
```

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
   2. No arithmetic required, use :func:`~parameters._unwrap`.
   3. No parameters are used, do nothing.
```
josh146: 1. What is the difference between 'arithmetic required' and 'no arithmetic required'? In almost all cases, I have simply used `p.x`.

2. Let's mark to change this to `p.val`, as per @co9olguy's suggestion. It is more descriptive than `p.x`.

tests/frontend/test_engine.py
```diff
@@ -27,61 +27,54 @@
 @pytest.fixture
```
josh146: This is more a comment on the file, ignore this particular line.

Can we rename this test to `test_local_engine.py`? The tests for the remote engines are likely to be a lot more complex with significant network imports and fixtures, so for ease of development, better to have those in separate files.

tests/frontend/test_engine.py
```diff
@@ -137,36 +131,31 @@ def inspect():
...
-
-        assert inspect() == expected
+        assert inspect() == expected2
 
         # reapply history
         eng.reset()
         eng.run([p1, p2])
-
-        assert inspect() == expected
+        assert inspect() == expected2
```
josh146: Nice, looks like the tests required minimal changes to work!

tests/integration/test_tf_symbolic.py
```diff
@@ -81,6 +82,7 @@ def test_eng_run_eval_false_measurements_are_tensors(self, setup_eng):
         val = q[0].val
         assert isinstance(val, tf.Tensor)
 
+    @pytest.mark.broken('FIXME')
```
josh146: Why is this test broken? Might be better to add something to the mark string

josh146:
> I fixed the failing test (TF backend returns an array of samples with batching on).
> I can't find a complete explanation on how the batching is supposed to work, just hints here and there. Is it like multi-shot simulation, except that gate parameters can also be given as vectors with parameter `a[i]` corresponding to shot `i`?

I _think_ this is correct, but @co9olguy would know in more detail.

strawberryfields/engine.py
```diff
@@ -73,50 +73,56 @@
...
+    Returned by :meth:`.BaseEngine.run`.
+    """
+    def __init__(self, samples):
+        #: BaseState: quantum state object returned by a local backend, if any
+        self.state = None
+        #: List[List[Number]]: measurement samples, shape == (modes, shots)
+        self.samples = samples
+
+    def __str__(self):
+        return self.__class__.__name__ + ': {} subsystems, state: {}'.format(len(self.samples), self.state)\
```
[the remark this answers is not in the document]
smite: Probably not in the final version, especially if you have several shots, but it's good for debugging.

strawberryfields/engine.py
```diff
@@ -73,50 +73,56 @@
...
-            self.backend_name = backend._short_name
-            self.backend = backend
-        else:
-            raise TypeError('backend must be a string or a BaseBackend instance.')
+        #: List[List[Number]]: latest measurement results, shape == (modes, shots)
+        self.samples = None
 
+    @abc.abstractmethod
     def __str__(self):
         """String representation."""
```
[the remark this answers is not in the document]
smite: I'd say no, abc.abstractmethod was introduced to make that unnecessary (and also to make it impossible to instantiate the ABC).

strawberryfields/ops.py
smite: No, but `with prog.context as (alice, bob):` does! That's nicer, I'll change it.

strawberryfields/ops.py
smite: I tried that first, but this way `tol` is rendered in the same font as above.

tests/frontend/test_engine.py
smite: Sure.

smite: Although the tests in this file mostly apply to the ABC, the other engines should pass them too.

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+                    if not p.can_follow(prev):
+                        raise RuntimeError("Register mismatch: program {}, '{}'.".format(len(self.run_progs), p.name))
+
+                    # Copy the latest measured values in the RegRefs of p.
+                    # We cannot copy from prev directly because it could be used in more than one engine.
+                    for k, v in enumerate(self.samples):
+                        p.reg_refs[k].val = v
+
+                # if the program hasn't been compiled for this backend, do it now
+                if compile and p.backend != self.backend_name:
```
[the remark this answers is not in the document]
smite: Maybe the engine should compile it if it hasn't been compiled yet for the targeted backend? I.e. remove "compile and" here.

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+                self._run_program(p, **kwargs)
+                self.run_progs.append(p)
+                # store the latest measurement results
+                self.samples = [p.reg_refs[k].val for k in sorted(p.reg_refs)]
+                prev = p
+        except Exception as e:
+            raise e
+        else:
+            # program execution was successful
+            pass
```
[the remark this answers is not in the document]
smite: For nothing now, I just left the try/except structure there for the time being. Whether we need it depends on how the engine should handle individual programs raising exceptions.

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+    def _init_backend(self, init_num_subsystems):
+        self.backend.begin_circuit(init_num_subsystems, **self.backend_options)
+
+    def _run_program(self, prog, **kwargs):
         applied = []
         for cmd in prog.circuit:
             try:
                 # try to apply it to the backend
-                cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                cmd.op.apply(cmd.reg, self.backend, **kwargs)  # NOTE we could also handle storing measured vals here
```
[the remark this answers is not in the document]
smite: Currently the measured vals are stored in the RegRefs in `Measurement.apply()`, but it could be done here as well. I don't know if it is any simpler, so I just left a remark.

strawberryfields/engine.py
smite: At least the TF backend requires this, for substituting numerical values to symbolic variables using `feed_dict`

strawberryfields/engine.py
smite: Yeah, I think so.

strawberryfields/parameters.py
smite: See `Dgate._apply()` for an example of arithmetic required. I guess the `_unwrap` function is a bit unnecessary since in most cases you only have a single parameter to unwrap.

tests/integration/test_tf_symbolic.py
smite: It's fixed now.

co9olguy:
> > I fixed the failing test (TF backend returns an array of samples with batching on).
> > I can't find a complete explanation on how the batching is supposed to work, just hints here and there. Is it like multi-shot simulation, except that gate parameters can also be given as vectors with parameter `a[i]` corresponding to shot `i`?
> 
> I _think_ this is correct, but @co9olguy would know in more detail.

Since `batch_size` must be set, the underlying computation is always batched or unbatched, as selected. All gate parameters can be given either as a scalar, or as a vector. Scalars are broadcasted to `batch_size` length vectors. For measurements, it returns a batched vector, with each element corresponding to an appropriate measurement result/sample for the corresponding batch index input arguments. e.g., if I do Dgate([alpha0,alpha1]) as part of a batched computation, then MeasureHomodyne should return a vector with [val0, val1], where `valX` is an appropriate measurement value for `alphaX`

strawberryfields/engine.py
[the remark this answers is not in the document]
josh146: Yep, that would work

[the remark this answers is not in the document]
smite: I just remembered, the reason it's required is that some tests fail if the program is compiled (because compilation can change the order of gates that commute). Maybe we need more nuanced compilation options like "preserve gate order".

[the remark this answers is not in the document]
josh146: Ah, that's annoying. I guess it doesn't matter too much for local engines, but for remote engines we can write the tests to get around this problem.

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+        * The compiled program is executed on the backend, updating the backend state.
+        * The latest measurement results of each subsystem are stored
+          in the :class:`.RegRef` instances of the corresponding Program, as well as in self.samples.
+        * The compiled program is appended to self.run_progs.
+
+        Finally, the result of the computation is returned.
+
+        Args:
+            program (Program, Sequence[Program]): quantum programs to run
+            compile (bool): If True, compile the Program instances before sending them to the backend.
```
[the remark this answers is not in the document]
co9olguy: This way the user has to specifically overwrite the default option. Since `compile` actually means `compile_and_optimize`, I can see a use-case for this (e.g., comparing two programs which are valid, but one is unoptimized)

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
         Returns:
             list[Command]: commands that were applied to the backend
         """
+
+    def _run(self, program, *, compile=True, **kwargs):
+        """Execute the given programs by sending them to the backend.
```
co9olguy: We should clarify what the scope of `program` is when it is a sequence. Is it a list of programs, each of which should be run independently one after the other, or is it a list of program segments which should be linked together sequentially (without returning intermediate measurements) as part of a longer computation?

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+            raise TypeError('backend must be a string or a BaseBackend instance.')
+
+    def __str__(self):
+        return self.__class__.__name__ + '({})'.format(self.backend_name)
+
+    def reset(self, backend_options={}):
+        super().reset(backend_options)
+        self.backend_options.pop('batch_size', None)  # HACK to make tests work for now
+        self.backend.reset(**self.backend_options)
+        # TODO should backend.reset and backend.begin_circuit be combined?
```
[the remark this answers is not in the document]
co9olguy: The difference between `begin_circuit` and `reset` is that the second allows you to specify whether the computational graph should be cleared out. Programmatically, I don't think there is huge barrier to combine them, but from user/test perspective it is cleaner to have a `reset` method (since a lot of tests need to clear out old junk from tf graph)

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+            self.backend = backend
+        else:
+            raise TypeError('backend must be a string or a BaseBackend instance.')
+
+    def __str__(self):
+        return self.__class__.__name__ + '({})'.format(self.backend_name)
+
+    def reset(self, backend_options={}):
+        super().reset(backend_options)
+        self.backend_options.pop('batch_size', None)  # HACK to make tests work for now
```
co9olguy: It shouldn't be too tricky to allow this to work the same way as, e.g., `cutoff_dim` works. Perhaps there is some implicit logic that needs to be made explicit, but I don't see any huge reason why we can't get this working like any other backend option

strawberryfields/engine.py
```diff
@@ -182,58 +285,34 @@ def _run_program_locally(self, prog, **kwargs):
...
-        # TODO unsuccessful run due to exceptions should ideally have no effect on the backend state (state checkpoints?)
-        try:
-            prev = None  # previous program segment
-            for p in program:
-                if self.run_progs:
-                    prev = self.run_progs[-1]
-                else:
-                    # initialize the backend TODO where should this happen? c.f. the fixtures in conftest.py
-                    self.backend.begin_circuit(num_subsystems=p.init_num_subsystems, **self.kwargs)
+            modes (None, Sequence[int]): Modes to be returned in the ``Result.state`` :class:`.BaseState` object.
```
co9olguy: Seems to be missing docstrings for other args?

strawberryfields/engine.py
```diff
@@ -182,58 +285,34 @@ def _run_program_locally(self, prog, **kwargs):
...
+                    applied_cmds = self._run_program(temp, **kwargs)
                     applied.extend(applied_cmds)
                 except NotImplementedError as err:
                     # simplify the error message by suppressing the previous exception
                     raise err from None
         return applied
 
-    def run(self, program, return_state=True, modes=None, compile=True, **kwargs):
-        """Execute the given program by sending it to the backend.
+    def run(self, program, *, compile=True, modes=None, state_options={}, **kwargs):
```
co9olguy: If you have a `state_options` argument, then is there any more need for `kwargs`?

strawberryfields/engine.py
co9olguy: Yup. Maybe we can think of a better name though

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
+    def _init_backend(self, init_num_subsystems):
+        self.backend.begin_circuit(init_num_subsystems, **self.backend_options)
+
+    def _run_program(self, prog, **kwargs):
         applied = []
         for cmd in prog.circuit:
             try:
                 # try to apply it to the backend
-                cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                cmd.op.apply(cmd.reg, self.backend, **kwargs)  # NOTE we could also handle storing measured vals here
```
co9olguy: What remaining `kwargs` do we expect here? Just tensorflow and hbar stuff?

tests/conftest.py
```diff
@@ -189,8 +188,9 @@ def setup_eng(setup_backend_pars):  # pylint: disable=redefined-outer-name
     def _setup_eng(num_subsystems, **kwargs):
         """Factory function"""
         prog = Program(num_subsystems)
-        setup_backend_pars.update(kwargs)  # override defaults with kwargs
-        eng = Engine(backend=setup_backend_pars['backend_name'], **setup_backend_pars)
+        backend, backend_options = setup_backend_pars
+        backend_options.update(kwargs)  # override defaults with kwargs
+        eng = LocalEngine(backend=backend, backend_options=backend_options)
```
co9olguy: At some point we'll have to make more of a distinction that this fixture only applies to LocalEngines. Any tests that look at the `state` can remain as-is. But we'll have to think about how to test other tests on remote engines

tests/conftest.py
```diff
@@ -174,8 +174,7 @@ def setup_backend_pars(
     use the ``@pytest.mark.backends()`` fixture. For example, for a test that
     only works on the TF and Fock backends, ``@pytest.mark.backends('tf', 'fock').
     """
-    return {
-        'backend_name': request.param,
+    return request.param, {
```
co9olguy: Done for convenience I assume?

tests/frontend/test_engine.py
```diff
@@ -27,61 +27,54 @@
...
+        eng.run(prog)
         assert len(eng.run_progs) == 1
 
         eng.reset()
         assert not eng.run_progs
-
         p2 = sf.Program(3)
         with p2.context as q:
-            D | q[2]
+            ops.Rgate(1.0) | q[2]
```
co9olguy: any reason for this change?

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
-or a TensorFlow object.
+The :class:`Parameter` class encapsulates a parameter passed to the
+quantum operations represented by :class:`.Operation`.
+The parameter objects can represent a
 
-.. currentmodule:: strawberryfields
+* number
+* NumPy array
+* value measured from the quantum register (:class:`.RegRefTransform`)
+* TensorFlow object
```
co9olguy:
```suggestion
* TensorFlow object (a :code:`Variable` or :code:`placeholder`)
```

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
-  This creates a :class:`~engine.Command` instance that wraps
-  the Operation and the RegRefs it acts on, which is appended to the Engine command queue.
+* The Operation instance is applied using its :meth:`~ops.Operation.__or__`
+  method inside an :class:`.Program` context.
+  This creates a :class:`.Command` instance that wraps
+  the Operation and the RegRefs it acts on, which is appended to the Program command queue.
 
-* Once the entire program is inputted, Engine optimizes it. This involves merging and commuting Commands
-  inside the circuit graph, the building of which requires knowledge of their dependencies, both direct and Parameter-based.
+* Once the entire program is inputted, it is compiled and optimized. This involves
```
co9olguy: It is only compiled/optimized during `eng.run`, not directly after inputting (`with prog.context() as q:`)

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
+  merging and commuting Commands inside the circuit graph. The circuit graph is built
```
co9olguy: Will it be clear to reader what a circuit graph is?

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
-* Merging two :class:`~ops.Gate` instances of the same subclass involves
+* Merging two :class:`.Gate` instances of the same subclass involves
   adding their first parameters after equality-comparing the others. This is easily done if
-  all the parameters have an immediate value. RegRefTransforms and TensorFlow objects are more complicated,
-  but could in principle be handled.
-
-For now, we simply don't do the merge if RegRefTransforms or TensorFlow objects are involved.
+  all the parameters have an immediate numerical value.
+  RegRefTransforms and TensorFlow objects are more complicated, but could in principle be handled.
+  For now, we simply don't do the merge if RegRefTransforms or TensorFlow objects are involved.
```
co9olguy: so, if tensorflow objects are lurking around, the new argument `compile=True` (which normally does compiled & optimize) will just do compilation (i.e., probably nothing)?

strawberryfields/parameters.py
```diff
@@ -1,64 +1,78 @@
...
 
-* The optimized command queue is run by Engine, which calls the :func:`~ops.Operation.apply` method
-  of each Operation in turn (and tries :func:`~ops.Operation.decompose`
+* The compiled Program is run by :class:`.Engine`, which calls the :meth:`~ops.Operation.apply` method
+  of each Operation in turn (and tries :meth:`~ops.Operation.decompose`
   if a :py:exc:`NotImplementedError` exception is raised).
```
co9olguy: It no longer does this, does it?

strawberryfields/engine.py
```diff
@@ -182,58 +285,34 @@ def _run_program_locally(self, prog, **kwargs):
                 try:
```
co9olguy: With a program validation stage, this step should never happen. I think we should remove it so that our tests alert us to validation errors rather than fall back to some implicity decomposition

strawberryfields/ops.py
```diff
@@ -33,101 +33,94 @@
...
+.. code-block:: python
 
-Here :samp:`eng` is an instance of :class:`strawberryfields.engine.Engine` and defines the context in which
-the commands are executed.
+  prog = sf.Program(3)
+  with prog.context as q:
+      G(params) | q
+      F(params) | (q[1], q[6], q[2])
+
+Here :samp:`prog` is an instance of :class:`strawberryfields.program.Program` and defines the context in which
```
```suggestion
Here :samp:`prog` is an instance of :class:`strawberryfields.program.Program` which defines the context where
```

strawberryfields/ops.py
```diff
@@ -33,101 +33,94 @@
...
+        return q ** 2
 
-    eng, q = sf.Engine(2)
-    with eng:
-      Measure           | q[0]
-      Dgate(square(q[0])) | q[1]
-    eng.run()
+    with prog.context as q:
+        Measure           | q[0]
+        Dgate(square(q[0])) | q[1]
```
co9olguy: Nice!

strawberryfields/ops.py
```diff
@@ -463,15 +456,17 @@ def _apply(self, reg, backend, **kwargs):
             reg (Sequence[int]): subsystem indices the operation is
                 acting on (this is how the backend API wants them)
             backend (BaseBackend): backend to execute the operation
+
+        Returns:
+            array[Number] or None: Measurement results, if any. shape == (len(reg), shots).
```
```suggestion
            array[Number] or None: Measurement results, if any; shape == (len(reg), shots).
```

strawberryfields/ops.py
```diff
@@ -463,15 +456,17 @@ def _apply(self, reg, backend, **kwargs):
...
         """
-        raise NotImplementedError(
-            'Missing direct implementation: {}'.format(self))
+        raise NotImplementedError('Missing direct implementation: {}'.format(self))
```
co9olguy: lol obviously people's editors have different standards for acceptable line length. @josh146 can we just go ahead and pick one and use `black` to enforce it?

doc/op_conventions.rst
```diff
@@ -34,7 +34,10 @@ In this section, we provide the definitions of various quantum operations used b
    conventions/others
 
 
-.. note:: In Strawberry Fields we use the convention :math:`\hbar=2` by default, but other conventions can also be chosen on engine :func:`initialization <strawberryfields.Engine>`. In this document we keep :math:`\hbar` explicit.
+.. note:: In Strawberry Fields we use the convention :math:`\hbar=2` by default, but other
+   conventions can also be chosen by setting the global variable :py:data:`strawberryfields.hbar`
```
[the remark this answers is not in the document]
josh146: http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py:data

> Describes global data in a module, including both variables and values used as “defined constants.” Class and object attributes are not documented using this environment.

Basically module constants.

strawberryfields/engine.py
```diff
@@ -168,11 +174,108 @@ def _run_program_locally(self, prog, **kwargs):
...
             list[Command]: commands that were applied to the backend
         """
+
+    def _run(self, program, *, compile=True, **kwargs):
+        """Execute the given programs by sending them to the backend.
+
+        For each :class:`Program` instance given as input, the following happens:
+
+        * The Program instance is compiled and optimized.
+        * The compiled program is executed on the backend, updating the backend state.
```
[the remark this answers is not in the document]
josh146: I think we should remove the statement `updating the backend state.` altogether. It is entirely up to the backend how it treats incoming quantum jobs, independently of the engine.

With some, there will be an implicit reset between consecutive jobs (hardware backends, hardware simulators). Meanwhile, other backends will allow you to preserve and act on the existing backend state (`gaussian`, `fock`, `tf`).

strawberryfields/engine.py
josh146: Yep, curious about this too

strawberryfields/parameters.py
josh146: It currently does this still. But it is marked for deprecation, and we should tackle this in a new PR.

strawberryfields/engine.py
josh146: :+1: thumb

strawberryfields/engine.py
```diff
@@ -13,50 +13,50 @@
 # limitations under the License.
 
 """
-Quantum compiler engine
-========================================================
+Quantum program executor engine
```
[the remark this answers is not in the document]
josh146: Would prefer not to have `Program` in there (or the first word), it would lead to a confusing table of contents.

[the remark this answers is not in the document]
smite: True, the header appears in the TOC so it should be short.

strawberryfields/engine.py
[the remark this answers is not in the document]
smite: How about if I replace the `compile` parameter with a kwargs-like dictionary `compile_options` which is passed on to `Programs.compile()`?

strawberryfields/engine.py
```diff
@@ -73,50 +73,56 @@
...
 from collections.abc import Sequence
 
 from .backends import load_backend
 from .backends.base import (NotApplicableError, BaseBackend)
 
 
-class Engine:
-    r"""Quantum program executor engine.
+class Result:
+    """Result of a quantum computation.
```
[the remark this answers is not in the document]
smite: The ABC does not exist yet, I'll add it when we design RemoteEngine.

strawberryfields/engine.py
```diff
@@ -73,50 +73,56 @@
...
-            self.backend_name = backend
-            #: BaseBackend: backend for executing the quantum circuit
-            self.backend = load_backend(backend)
-        elif isinstance(backend, BaseBackend):
-            self.backend_name = backend._short_name
-            self.backend = backend
-        else:
-            raise TypeError('backend must be a string or a BaseBackend instance.')
+        #: List[List[Number]]: latest measurement results, shape == (modes, shots)
+        self.samples = None
```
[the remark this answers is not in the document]
smite: The engine can execute several Program instances in succession, and after each one it stores the measurement results in `self.samples`. The engine needs to do this because the same Program instance could be next executed by a different engine (possibly with a different Program history), changing the measurement results stored in its RegRefs.

strawberryfields/engine.py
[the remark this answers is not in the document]
smite: Removed it.

strawberryfields/engine.py
smite: It's the latter, I'll clarify it.

strawberryfields/engine.py
smite: It's on purpose, there's a link to the parent class method that `LocalEngine.run()` extends. But I can also copy the docs for the parent method args here if you think it's clearer.

strawberryfields/engine.py
smite: `kwargs` get passed on to `Operation.apply()`, which uses `eval_params` if it appears, and then on to the `backend.measure_*` methods that look for various other options in them. It seems these include at least feed dictionaries for TF, homodyne sampling options for the Fock backend and quadrature eigenstate approximation parameters for the Gaussian backend.

strawberryfields/engine.py
smite: hbar was eliminated from the backend API, but all kinds of other stuff remains. See https://github.com/XanaduAI/strawberryfields/pull/80#discussion_r283754124

tests/conftest.py
smite: Yes, made updating the tests easier.

tests/frontend/test_engine.py
smite: I guess I wanted to make the second program clearly different from the first one.

strawberryfields/parameters.py
smite: Only the Commands that directly involve RR or TF objects are exempt from merging, and we could probably even merge RR and TF objects if we added more symbolic computation capabilities to the Parameter class, using SymPy for example. We've been discussing this a bit with @josh146

strawberryfields/parameters.py
smite: Right, the backend capabilities database should be designed first.

strawberryfields/engine.py
smite: Agreed.

strawberryfields/ops.py
smite: Fixed. BTW this isn't implemented yet, since the backends cannot handle multiple shots, yet.

strawberryfields/engine.py
```diff
@@ -73,50 +73,56 @@
...
+    """
+    def __init__(self, samples):
+        #: BaseState: quantum state object returned by a local backend, if any
+        self.state = None
+        #: List[List[Number]]: measurement samples, shape == (modes, shots)
+        self.samples = samples
+
+    def __str__(self):
+        
+        return 'Result: {} subsystems, state: {}\n samples: {}'.format(len(self.samples), self.state, self.samples)
```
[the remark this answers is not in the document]
smite: Agreed, but I'd leave it like this for a while as it helps with debugging.

strawberryfields/parameters.py
smite: Clarified.

strawberryfields/engine.py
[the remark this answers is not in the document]
josh146: How about just `Execution engine`?

strawberryfields/engine.py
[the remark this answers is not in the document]
josh146: I like that solution :+1:

tests/conftest.py
josh146: Yep, I imagine almost all existing engine fixtures will be renames to be local engine fixtures, with new constructs required for remote engines :+1:

strawberryfields/ops.py
josh146: On GitHub, the displayed linewidth varies depending on the browser, OS, and monitor display settings, and can be anywhere from 100-130 chars per line.

I suggest 100 line width, if everyone agrees.

strawberryfields/engine.py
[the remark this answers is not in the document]
smite: Changed.

strawberryfields/engine.py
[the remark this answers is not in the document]
co9olguy: Sounds good @smite

tests/frontend/test_engine.py
co9olguy: sounds good

strawberryfields/ops.py
co9olguy: :+1:

josh146:
@smite, what is the latest status, have all comments been addressed and is the PR now ready for merging?

strawberryfields/parameters.py
smite: Ah, but those are not `Gate` subclasses, but rather `Decomposition` subclasses. `Gate` subclasses are defined to behave as one-parameter unitary groups wrt. p[0].

strawberryfields/parameters.py
smite: Fixed.

strawberryfields/ops.py
smite: 100 chars is good.

strawberryfields/parameters.py
josh146: ah, got it

strawberryfields/parameters.py
smite: Added.

strawberryfields/engine.py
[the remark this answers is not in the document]
co9olguy: ok that makes sense for RegRefs, but won't there also be a `Results` class (possibly under a different name) returned by an engine which also stores this information? Isn't it redundant to have the engine also store it? 

I feel like there's some special use-case involving `RegRefTransform`s here, but I couldn't figure it out

strawberryfields/engine.py
co9olguy: Are we leaving this for a future PR?

strawberryfields/engine.py
co9olguy: Might be clearer to copy over docs related to `Args`, since people who are directly reading the source code won't see descriptions of all the args

strawberryfields/engine.py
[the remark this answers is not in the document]
smite: Each engine instance represents a single quantum computation, which IMO makes it the primary place to store that information. If a Program contains RRTs (gates that depend on previously measured values) the engine needs to be able to access those samples before it can execute the Program.

OTOH `Result.samples` is copy of `BaseEngine.samples`. The `Result` class is a bit redundant in the sense that all that information could be accessed manually through the engine instance, but I think it's good to have it from a usability perspective. One can think of a Result instance as a snapshot of the state of the computation at one point in time.

strawberryfields/engine.py
smite: I think so, I made an issue for it: #84

strawberryfields/engine.py
smite: Ok, I'll copy them over.

josh146:
Before merging, pylint is picking up the following:
image

It would be better to replace the default value as `None`, to avoid unintended bugs

Issue #59, opened 2019-04-11
Degenerate symplectic eigenvalues still seems to break Bloch-Messiah decompositions


josh146:
See the following comment for the specific test that doesn't seem to be working:

_Originally posted by @co9olguy in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjI1MTM2NDE1/pull_request_reviews/more_threads_

ishdhand:
Interesting! @josh146, @co9olguy, I am not able to use that link. Could you share some more information about this?

josh146:
Oops, look like the link is broken. It is from here:
https://github.com/XanaduAI/strawberryfields/blob/master/tests/frontend/test_decompositions.py#L508

I ported over the test you wrote from unittest to pytest, and for some reason could not get those two commented out assertions to pass. I had a look at the two orthogonal matrices `O1` and `O2`, and it looks like they were no longer symplectic, so I assume it is the return of that bug?

The actual `decompositions.py` file is unchanged, so it could be three things:

1. The old test had a bug in it that made it look like it was working, and this was removed by porting to pytest

2. The ported test has a bug in it

3. The new test is using a different function argument that has uncovered an edge case bug in the decompositions.py file

ishdhand:
@josh146, thanks for sharing. If I understand correctly, this is a different bug that pops up when we input passive matrices. The S matrix is sensible (identity) but some submatrices of the O1,O2 matrices are somehow rotated, as was happening in the original bug. 

I'm not sure how Bloch-Messiah should respond to passive inputs. We could in principle deal with passive matrices separately. One option could be to return O1 = S_in and S = O2 = Identity. Another could be S identity and O1, O2 = sqrt(U).

josh146:
Ah, thanks for catching that. I saw the rotated O1 and O2 matrices and assumed it was the same issue.

> One option could be to return O1 = S_in and S = O2 = Identity. Another could be S identity and O1, O2 = sqrt(U).

In most cases, the Clements or Reck decomposition will be called directly after the Bloch-Messiah decomposition. In this case, it probably makes more sense to do O1 = S_in and S = O2 = Identity, to reduce the gate count?

ishdhand:
Sounds good! Implemented. @josh146, could you please review? I also think that this function is getting a little too so suggestions for refactoring it are welcome!

PR #81, opened 2019-05-11
Bugfix for Bloch-Messiah returning non-symplectic matrices when input is passive

ishdhand:
This bugfix resolves issue #59, in which  Bloch-Messiah was returning non-symplectic matrices when input is passive. Now, bloch_messiah checks if the input is passive, and treats it separately. I.e., if S is a passive transformation, then return the S as the first passive transformation, and set the the squeezing and second unitary matrices to identity. This choice is clearly not unique but perhaps suitable for simulation/hardware decisions. If S is active, then action is same as before.


josh146:
Thanks @ishdhand, that's fantastic! Merging in now.

PR #83, opened 2019-05-15
Frontend validation

josh146:
* Adds frontend validation database, the `strawberryfields.devicespecs` submodule

* Defines the abstract base class `DeviceSpecs`, a pre-Python 3.7 'dataclass' for storing validation information for a backend. Note: after playing around with both Python 3.7 `dataclass` module and the Python 3+ `attrs` module, it seems to make more sense in the short term to simply take the approach used for device validation in PennyLane.

* Each supported backend by Strawberry Fields should have a class defined in the `data` submodule that inherits from `DeviceSpecs`, with required class attributes provided.

* `Program` class modified to validate against the new backend databases

### DIFF
--- a/strawberryfields/data/fock.py
+++ b/strawberryfields/data/fock.py
@@ -26,7 +26,7 @@ class Fock(DeviceData):
         # meta operations
         "All",
         "New",
-        "Del",
+        "Delete",
         # state preparations
         "Vacuum",
         "Coherent",
--- a/strawberryfields/data/fock.py
+++ b/strawberryfields/data/fock.py
@@ -15,7 +15,7 @@
 from .device import DeviceData
 
 
-class Fock(DeviceData):
+class FockData(DeviceData):
     """Validation data for the Fock backend"""
 
     modes = 0
--- a/strawberryfields/data/fock.py
+++ b/strawberryfields/data/fock.py
@@ -1,0 +1,67 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fock backend validation data"""
+from .device import DeviceData
+
+
+class Fock(DeviceData):
+    """Validation data for the Fock backend"""
+
+    modes = 0
+    remote = True
+    interactive = True
+
+    primitives = {
+        # meta operations
+        "All",
+        "New",
+        "Del",
+        # state preparations
+        "Vacuum",
+        "Coherent",
+        "Squeezed",
+        "DisplacedSqueezed",
+        "Thermal",
+        "Fock",
+        "Catstate",
+        "Ket",
+        "DensityMatrix",
+        # measurements
+        "MeasureFock",
+        "MeasureHomodyne",
+        # channels
+        "LossChannel",
+        # single mode gates
+        "Dgate",
+        "Xgate",
+        "Zgate",
+        "Sgate",
+        "Rgate",
+        "Vgate",
+        "Kgate",
+        "Fouriergate",
+        "BSgate",
+        "CKgate",
+    }
+
+    decompositions = {
+        "Interferometer": {},
+        "GraphEmbed": {},
+        "GaussianTransform": {},
+        "Gaussian": {},
+        "Pgate": {},
+        "S2gate": {},
+        "CXgate": {},
+        "CZgate": {},
+    }
--- a/strawberryfields/data/fock.py
+++ b/strawberryfields/data/fock.py
@@ -25,7 +25,7 @@ class Fock(DeviceData):
     primitives = {
         # meta operations
         "All",
-        "New",
+        "New_modes",
         "Delete",
         # state preparations
         "Vacuum",
--- a/strawberryfields/devicespecs/__init__.py
+++ b/strawberryfields/devicespecs/__init__.py
@@ -44,13 +44,14 @@ DeviceSpecs methods
 
 .. autosummary::
    modes
+   local
    remote
    interactive
    primitives
    decompositions
    parameter_ranges
-   topology
-   blackbird_template
+   graph
+   circuit
 
 Code details
 ^^^^^^^^^^^^
--- a/strawberryfields/devicespecs/__init__.py
+++ b/strawberryfields/devicespecs/__init__.py
@@ -12,18 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Device validation submodule"""
-from .chip0 import Chip0Data
-from .fock import FockData
-from .gaussian import GaussianData
-from .tensorflow import TFData
+from .chip0 import Chip0Specs
+from .fock import FockSpecs
+from .gaussian import GaussianSpecs
+from .tensorflow import TFSpecs
 
 
 backend_databases = {
-    "base": FockData,
-    "chip0": Chip0Data,
-    "fock": FockData,
-    "gaussian": GaussianData,
-    "tf": TFData,
+    "base": FockSpecs,
+    "chip0": Chip0Specs,
+    "fock": FockSpecs,
+    "gaussian": GaussianSpecs,
+    "tf": TFSpecs,
 }
-"""dict[str, DeviceData]: dictionary mapping backend shortname
+"""dict[str, DeviceSpecs]: dictionary mapping backend shortname
 to the validation dataclass."""
--- a/strawberryfields/devicespecs/__init__.py
+++ b/strawberryfields/devicespecs/__init__.py
@@ -27,8 +27,8 @@ These details are used by the :class:`~.Program` class when validating and
 compiling quantum programs. By querying the data class corresponding to the
 requested device/backend, the :class:`~.Program` will be able to:
 
-1. **Validate** that the program has the correct number of modes, valid
-   quantum operations,
+1. **Validate** that the program has the correct number of modes, and consists
+   of valid quantum operations for that device.
 
 2. **Compile** the program to match the backend topology, making use
    of allowed decompositions along the way.
@@ -56,6 +56,7 @@ Code details
 ^^^^^^^^^^^^
 """
 from .device_specs import DeviceSpecs
+from .base import BaseSpecs
 from .chip0 import Chip0Specs
 from .fock import FockSpecs
 from .gaussian import GaussianSpecs
@@ -63,7 +64,7 @@ from .tensorflow import TFSpecs
 
 
 backend_specs = {
-    "base": FockSpecs,
+    "base": BaseSpecs,
     "chip0": Chip0Specs,
     "fock": FockSpecs,
     "gaussian": GaussianSpecs,
--- a/strawberryfields/devicespecs/__init__.py
+++ b/strawberryfields/devicespecs/__init__.py
@@ -11,14 +11,58 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Device validation submodule"""
+"""
+Backend specifications
+======================
+
+**Module name:** :mod:`strawberryfields.devicespecs`
+
+.. currentmodule:: strawberryfields.devicespecs
+
+This module implements the :class:`~.DeviceSpecs` class, an abstract base
+data class used to store details and specifications of the Strawberry Fields
+backend and devices.
+
+These details are used by the :class:`~.Program` class when validating and
+compiling quantum programs. By querying the data class corresponding to the
+requested device/backend, the :class:`~.Program` will be able to:
+
+1. **Validate** that the program has the correct number of modes, valid
+   quantum operations,
+
+2. **Compile** the program to match the backend topology, making use
+   of allowed decompositions along the way.
+
+To access the correct specifications dataclass, :attr:`~.backend_specs` provides
+a dictionary mapping the backend shortname to the correct dataclass.
+
+
+DeviceSpecs methods
+-------------------
+
+.. currentmodule:: strawberryfields.devicespecs.DeviceSpecs
+
+.. autosummary::
+   modes
+   remote
+   interactive
+   primitives
+   decompositions
+   parameter_ranges
+   topology
+   blackbird_template
+
+Code details
+^^^^^^^^^^^^
+"""
+from .device_specs import DeviceSpecs
 from .chip0 import Chip0Specs
 from .fock import FockSpecs
 from .gaussian import GaussianSpecs
 from .tensorflow import TFSpecs
 
 
-backend_databases = {
+backend_specs = {
     "base": FockSpecs,
     "chip0": Chip0Specs,
     "fock": FockSpecs,
@@ -27,3 +71,5 @@ backend_databases = {
 }
 """dict[str, DeviceSpecs]: dictionary mapping backend shortname
 to the validation dataclass."""
+
+__all__ = ["backend_specs", "DeviceSpecs"]
--- a/strawberryfields/devicespecs/chip0.py
+++ b/strawberryfields/devicespecs/chip0.py
@@ -20,7 +20,7 @@ from .device_specs import DeviceSpecs
 class Chip0Specs(DeviceSpecs):
     """Validation data for the Chip0 simulator"""
 
-    modes = 0
+    modes = 4
     remote = True
     interactive = True
 
@@ -33,13 +33,12 @@ class Chip0Specs(DeviceSpecs):
     # TODO: update the below to specify the rectangular_symmetric
     # mapping for the interferometer when #87 is merged.
     # The current topology defined below is just for demonstration
-    blackbird_template = textwrap.dedent(
+    circuit = textwrap.dedent(
         """\
         name chip0_template
-        version 1.0
+        version 0.0
         target chip0 (shots=1)
 
-        # for n spatial degrees, first n signal modes, then n idler modes, phase zero
         S2gate({sq0}, 0.0) | [0, 2]
         S2gate({sq1}, 0.0) | [1, 3]
 
--- a/strawberryfields/devicespecs/chip0.py
+++ b/strawberryfields/devicespecs/chip0.py
@@ -15,42 +15,47 @@
 import textwrap
 
 import networkx as nx
-import blackbird
-from blackbird.utils import to_DiGraph
 
-from .device import DeviceData
+from .device_specs import DeviceSpecs
 
 
-class Chip0Data(DeviceData):
+class Chip0Specs(DeviceSpecs):
     """Validation data for the Chip0 simulator"""
 
     modes = 0
     remote = True
     interactive = True
 
-    primitives = {"S2gate", "BSgate", "MeasureFock"}
+    primitives = {"S2gate", "Interferometer", "MeasureFock", "Rgate", "BSgate"}
 
-    decompositions = {"BSgate": {}, "Rgate": {}, "Interferometer": {}}
+    # TODO: update the below to specify the rectangular_symmetric
+    # mapping for the interferometer when #87 is merged
+    decompositions = {"Interferometer": {}}
 
     blackbird_template = textwrap.dedent(
         """\
         name chip0_template
         version 1.0
-        target chip0 (shots={shots})
+        target chip0 (shots=1)
 
         # for n spatial degrees, first n signal modes, then n idler modes, phase zero
-        S2gate({squeezing0}, 0.0) | [0, 2]
-        S2gate({squeezing1}, 0.0) | [1, 3]
+        S2gate({sq0}, 0.0) | [0, 2]
+        S2gate({sq1}, 0.0) | [1, 3]
 
-        # standard 2x2 interferometer for the signal modes (the lower ones in frequency)
-        BSgate(0.0, {external_phase}) | [0, 1]
-        BSgate({internal_phase}, 0.0) | [0, 1]
+        # this will need to be updated to use the new
+        # rectangular_symmetric mapping when #87 is merged
+        Rgate({phase}) | 0
+        BSgate({theta}, {phi}) | [0, 1]
+        Rgate({phase}) | 0
+        Rgate({phase}) | 1
 
-        #duplicate the interferometer for the idler modes (the higher ones in frequency)
-        BSgate(0.0, {external_phase}) | [2, 3]
-        BSgate({internal_phase}, 0.0) | [2, 3]
+        # this will need to be updated to use the new
+        # rectangular_symmetric mapping when #87 is merged
+        Rgate({phase}) | 2
+        BSgate({theta}, {phi}) | [2, 3]
+        Rgate({phase}) | 2
+        Rgate({phase}) | 3
 
-        # Measurement in Fock basis
         MeasureFock() | [0]
         MeasureFock() | [1]
         MeasureFock() | [2]
@@ -58,15 +63,3 @@ class Chip0Data(DeviceData):
         """
     )
 
-    # returned DAG has all parameters set to 0
-    topology = to_DiGraph(
-        blackbird.loads(blackbird_template)(
-            squeezing0=0,
-            squeezing1=0,
-            external_phase=0,
-            internal_phase=0,
-        )
-    )
-
-    for i in sorted(topology.nodes().data()):
-        print(i)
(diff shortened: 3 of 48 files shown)


strawberryfields/data/fock.py
```diff
@@ -0,0 +1,67 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fock backend validation data"""
+from .device import DeviceData
+
+
+class Fock(DeviceData):
```
smite: I'd name the DeviceData classes more descriptively, we have so many object names with 'Fock' in them. Maybe "FockData" or "FockCapabilities"?

josh146: `FockSpecs`?

strawberryfields/program.py
```diff
@@ -772,49 +702,90 @@ def compile(self, backend='fock', optimize=True):
         Returns:
             Program: compiled program
         """
-        db = backend_database[backend]
+        # TODO: Avoid circular import
+        from strawberryfields.data import backend_databases
```
smite: I think circular imports can be avoided if you use "import X" instead of "from X import Y".

josh146: I tried `import strawberryfields.data` but no luck. From what I can tell, the issue is that `data.chip0` imports `io`, which then imports `Program`.

strawberryfields/data/fock.py
smite: Not bad, maybe also rename the base class to `DeviceSpecs` for consistency?

josh146: Yep. We could also update the module to `strawberryfields.devicespecs`, for consistency, using `data` feels too general

strawberryfields/program.py
josh146: Fixed in the latest commit

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+            int: number of supported modes
+        """
+
+    @property
+    @abc.abstractmethod
+    def remote(self) -> bool:
+        """Whether the backend supports remote execution.
+
+        Returns:
+            bool: returns ``True`` if the backend supports remote execution.
```
co9olguy:
```suggestion
            bool: ``True`` if the backend supports remote execution
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+
+    @property
+    @abc.abstractmethod
+    def interactive(self) -> bool:
+        """Whether the backend can be used interactively, that is,
+        the backend state is not reset between engine executions.
+
+        Returns:
+            bool: returns ``True`` if the backend supports interactive use.
```
co9olguy:
```suggestion
            bool: ``True`` if the backend supports interactive use
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+
+            {'operation_name': {'option1': val, 'option2': val,...}}
+
+        For each operation specified in the dictionary, the
+        :meth:`~Operation.decompose` method will be called during
+        :class:`Program` compilation, with keyword arguments
+        given by the dictionary value.
+
+        Returns:
+            dict[str, dict]: the quamtum operations that are supported
```
co9olguy:
```suggestion
            dict[str, dict]: the quantum operations that are supported
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    def parameter_ranges(self) -> Dict[str, List[List[float]]]:
+        """Allowed parameter ranges for supported quantum operations.
+
+        This property is optional.
+
+        Returns:
+            dict[str, list]: a dictionary mapping an allowed quantum operation
+            to a nested list of the form ``[[p0_min, p0_max], [p1_min, p0_max], ...]``.
+            where ``pi`` corresponds to the ``i``th gate parameter.z
```
co9olguy:
```suggestion
            where ``pi`` corresponds to the ``i``th gate parameter.
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        return dict()
```
co9olguy: How come this one returns something? :thinking:

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    def topology(self):
+        """The allowed circuit topology of the backend device as a directed
+        acyclic graph.
+
+        This property is optional; if arbitrary topologies are allowed by the device,
+        this will simply return ``None``.
+
+        Returns:
+            networkx.DAGGraph: a directed acyclic
```
co9olguy:
```suggestion
            networkx.DAGGraph: a directed acyclic graph
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        if self.blackbird_template is None:
+            return None
```
co9olguy: To confirm logic here: non-existance of `blackbird_template` implies that arbitrary topologies are allowed? Is it better to make this explicit? i.e., have a property like `all_topologies` or something?

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # returned DAG has all parameters set to 0
+        bb = blackbird.loads(self.blackbird_template)
+
+        if bb.is_template():
```
co9olguy: isn't it confusing that something found under `self.blackbird_template` isn't guaranteed to be a template?

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @property
+    def blackbird_template(self):
+        """The Blackbird template that will be accepted by the backend device.
+
+        This property is optional; if arbitrary topologies are allowed by the device,
+        this will simply return ``None``.
+
+        If the device expects a specific template for the recieved Blackbird
+        script, this method will return the seralized Blackbird template in string
```
co9olguy:
```suggestion
        script, this method will return the serialized Blackbird template in string
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+        form.
+
+        Returns:
+            str: Blackbird template
```
co9olguy: want to also put that `None` is a valid return type here?

strawberryfields/ops.py
```diff
@@ -1814,30 +1816,24 @@ def __init__(self, V, r=None, decomp=True, tol=1e-6):
...
-            raise NotImplementedError
-
         p = _unwrap(self.p)
         s = sqrt(sf.hbar / 2)  # scaling factor, since the backend API call is hbar-independent
         backend.prepare_gaussian_state(p[1]/s, p[0]/(s*s), reg)
 
     def _decompose(self, reg):
         # pylint: disable=too-many-branches
-        if not self.decomp:
-            return None  # refuse to decompose
```
co9olguy: Why was this here before?

strawberryfields/program.py
```diff
@@ -765,49 +696,99 @@ def compile(self, backend='fock', **kwargs):
         Returns:
             Program: compiled program
         """
-        db = backend_database[backend]
+        if backend in specs.backend_databases:
```
co9olguy: do we want to go with `database`? might confuse people that we have a relational DB or something?

strawberryfields/program.py
```diff
@@ -765,49 +696,99 @@ def compile(self, backend='fock', **kwargs):
...
+                    # by passing the kwarg `decomp=False` to more
+                    # operations (currently only ops.Gaussian allows this).
+                    #
+                    # For example, the 'gaussian' backend supports setting the state
+                    # via passing directly the (mu, cov) OR by first having the
+                    # frontend decompose into other primitive gates.
+                    # That is, ops.Gaussian is both a primitive _and_ a decomposition
+                    # for the 'gaussian' backend, and it's behaviour can be chosen
+                    # by the user.
+                    if (op_name in db.primitives) and hasattr(cmd.op, 'decomp'):
```
co9olguy: Just want to follow logic here: this block just checks the case that an op might be both decomposable and also a primitive, and confirms user-override before performing decomp?

strawberryfields/program.py
```diff
@@ -765,49 +696,99 @@ def compile(self, backend='fock', **kwargs):
...
+            # check topology
+            grid = self._list_to_grid(seq)
+            DAG = self._grid_to_DAG(grid)
+
+            # relabel the DAG nodes to integers, with attributes
+            # specifying the operation name. This allows them to be
+            # compared, rather than using Command objects.
+            mapping = {i: n.op.__class__.__name__ for i, n in enumerate(DAG.nodes())}
+            circuit = nx.convert_node_labels_to_integers(DAG)
+            nx.set_node_attributes(circuit, mapping, name='name')
```
co9olguy: Getting a bit confused. Does 'name' refer to the generic name of the Command (e.g., 'BSgate') or to a unique name associated with each node?

strawberryfields/devicespecs/chip0.py
```diff
@@ -0,0 +1,61 @@
...
+    # mapping for the interferometer when #87 is merged
+    decompositions = {"Interferometer": {}}
+
+    # TODO: update the below to specify the rectangular_symmetric
+    # mapping for the interferometer when #87 is merged.
+    # The current topology defined below is just for demonstration
+    blackbird_template = textwrap.dedent(
+        """\
+        name chip0_template
+        version 1.0
```
co9olguy: Can we just set everything throughout the library to the placeholder `version 0.0` until everything is more stable?

strawberryfields/devicespecs/chip0.py
```diff
@@ -0,0 +1,61 @@
...
+        target chip0 (shots=1)
+
+        # for n spatial degrees, first n signal modes, then n idler modes, phase zero
```
co9olguy: This comment, and earlier `modes` variable indicates this `devicespec` supports an arbitrary number of modes. However, the actual circuit has 4 modes.

strawberryfields/devicespecs/fock.py
```diff
@@ -0,0 +1,67 @@
...
+        "Rgate",
+        "Vgate",
+        "Kgate",
+        "Fouriergate",
+        "BSgate",
+        "CKgate",
+    }
+
+    decompositions = {
+        "Interferometer": {},
```
co9olguy: Can you remind me what effect a blank dictionary has? Does that mean it supports any possible decomposition the frontend might try?

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+
+        If the device supports arbitrary number of modes, set this to 0.
+
+        Returns:
+            int: number of supported modes
+        """
+
+    @property
+    @abc.abstractmethod
+    def remote(self) -> bool:
```
co9olguy: I suppose we also want to have a `local` property as well then, since various combinations or remote/local execution are possible depending on the device

tests/frontend/test_devicespecs.py
```diff
@@ -0,0 +1,146 @@
...
+            the abstract base class"""
+            modes = 0
+            remote = False
+            interactive = True
+            primitives = set()
+            decompositions = set()
+
+        dummy = DummyDevice()
+
+        assert not dummy.parameter_ranges
```
co9olguy: Per docstring, do we also want to test that it is a dictionary that is returned?

tests/frontend/test_devicespecs.py
```diff
@@ -0,0 +1,146 @@
...
+        # check gates are correct
+        assert circuit[0]['name'] == 'Sgate'
+        assert circuit[1]['name'] == 'Dgate'
+        assert circuit[2]['name'] == 'BSgate'
+        assert circuit[3]['name'] == 'MeasureFock'
+        assert circuit[4]['name'] == 'MeasureFock'
+
+        # check arguments
+        assert circuit[0]['args'] == [0, 0]
+        assert circuit[1]['args'] == [-7.123]
```
co9olguy: Just curious why the Dgate parameter does not get converted to zero?

tests/frontend/test_program.py
```diff
@@ -305,3 +309,269 @@ def test_merge_incompatible(self):
...
+        prog = sf.Program(3)
+        r = 0.453
+        cov = np.array([[np.exp(-2*r), 0], [0, np.exp(2*r)]])*sf.hbar/2
+
+        with prog.context:
+            ops.Gaussian(cov, decomp=True) | 0
+
+        new_prog = prog.compile(backend='gaussian')
+
+        print([i.op.__class__.__name__ for i in new_prog.circuit])
```
co9olguy:
```suggestion
```

tests/frontend/test_program.py
```diff
@@ -305,3 +309,269 @@ def test_merge_incompatible(self):
...
+        assert len(new_prog) == 1
+
+        # decomposed gate should be an Sgate
+        circuit = new_prog.circuit
+        assert circuit[0].op.__class__.__name__ == "Sgate"
+        assert circuit[0].op.p[0] == r
+
+    def test_topology_validation(self, monkeypatch):
+        """Test compilation properly matches the device toplogy"""
```
co9olguy:
```suggestion
        """Test compilation properly matches the device topology"""
```

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+"""Abstract base class for storing device data for validation"""
+from typing import List, Set, Dict
+import abc
+
+import blackbird
+from blackbird.utils import to_DiGraph
+
+
+class DeviceSpecs(abc.ABC):
+    """Abstract base class for backend data"""
```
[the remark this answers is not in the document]
josh146: :+1: I'll extend this with proper documentation

strawberryfields/devicespecs/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+"""Device validation submodule"""
+from .chip0 import Chip0Specs
+from .fock import FockSpecs
+from .gaussian import GaussianSpecs
+from .tensorflow import TFSpecs
+
+
+backend_databases = {
+    "base": FockSpecs,
```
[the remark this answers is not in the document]
josh146: Yeah, I wasn't sure what to do for the `BaseBackend`? I could just make another dataclass for the base backend, that pretends to support all gates

strawberryfields/devicespecs/device_specs.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @property
+    @abc.abstractmethod
+    def modes(self) -> int:
+        """The supported number of modes of the device.
+
+        If the device supports arbitrary number of modes, set this to 0.
```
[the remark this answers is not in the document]
josh146: I deliberated over this and eventually found a reason why `None` wouldn't work (I had https://github.com/XanaduAI/pennylane/issues/178 in the back of my mind). But for the life of me, I can't remember why :thinking: I'll look back into this.

strawberryfields/devicespecs/device_specs.py
josh146: lol oops

strawberryfields/devicespecs/device_specs.py
josh146: Should it return `None`? I figured that

1. This shouldn't be an abstract base class, as only a very small subset of devices will need to set this, and

2. Better to be consistent, and just return an empty dictionary instead of `None`. It also leads to clearer code, i.e., `if db.parameter_range:` as opposed to `if db.parameter_range is not None:`

strawberryfields/devicespecs/device_specs.py
josh146: Yeah, I don't like the name `blackbird_template` for the property anyway. Some suggestions:

* `DeviceSpec.topology` for the blackbird script

* `DeviceSpec.graph` for the DAG

strawberryfields/ops.py
josh146: This was required to 'hack' around the previous try-except logic for decompositions. Now that that has been removed, it is no longer needed.

strawberryfields/program.py
josh146: How about `backend_specs`?

strawberryfields/program.py
josh146: Yep!

strawberryfields/program.py
josh146: You're correct, name refers to the operation name `BSgate`.

strawberryfields/devicespecs/fock.py
josh146: Essentially, yes. So for instance, once #87 is merged, the `chip0` device will look like this:
```python
   decompositions = {
        "Interferometer": {"geometry": "mach_zehnder"},
   }
```
specifying that compilation of the interferometer will use the `mach_zehnder` decomposition function, as opposed to `rectangular` or `triangular`.

strawberryfields/devicespecs/device_specs.py
josh146: Ah, I didn't think of this. How would this look? I assumed `remote` and `local` would be mutually exclusive

strawberryfields/devicespecs/device_specs.py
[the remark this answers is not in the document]
josh146: Changed to `None` for arbitrary number of modes.

strawberryfields/devicespecs/device_specs.py
josh146: Since this these two class attributes are mutually exclusive, it might be better to not allow both to be defined?

strawberryfields/devicespecs/device_specs.py
josh146: Made the change above :+1:

strawberryfields/program.py
josh146: Done

strawberryfields/devicespecs/device_specs.py
josh146: Done

strawberryfields/devicespecs/chip0.py
josh146: Done

strawberryfields/devicespecs/chip0.py
josh146: fixed

tests/frontend/test_devicespecs.py
josh146: Done

tests/frontend/test_devicespecs.py
josh146: Because it is hardcoded into the circuit attribute:
```python
circuit = """
    name test
    version 0.0

    Sgate({sq}, 0) | 0
    Dgate(-7.123) | 1
    BSgate({theta}) | 0, 1
    MeasureFock() | 0
    MeasureFock() | 2
    """
```
This takes into account devices that potentially have fixed gate parameters, i.e., fixed squeezing.

strawberryfields/devicespecs/device_specs.py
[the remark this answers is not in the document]
josh146: done

strawberryfields/devicespecs/__init__.py
```diff
@@ -11,14 +11,58 @@
...
+
+This module implements the :class:`~.DeviceSpecs` class, an abstract base
+data class used to store details and specifications of the Strawberry Fields
+backend and devices.
+
+These details are used by the :class:`~.Program` class when validating and
+compiling quantum programs. By querying the data class corresponding to the
+requested device/backend, the :class:`~.Program` will be able to:
+
+1. **Validate** that the program has the correct number of modes, valid
```
co9olguy: "and consists of valid quantum operations for that device"

co9olguy: sorry, can't suggest

strawberryfields/devicespecs/__init__.py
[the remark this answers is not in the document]
co9olguy: I think that is preferable to puttingin the wrong specs. Alternatively, you could just look at which gates are explicitly covered by BaseBackend (even though it doesn't do any simulation)

strawberryfields/devicespecs/device_specs.py
co9olguy: thumb

strawberryfields/devicespecs/device_specs.py
co9olguy: Sure, as long as they are mutually exclusive. It just wasn't obvious to me at first glance that that was the case

strawberryfields/devicespecs/device_specs.py
```diff
@@ -93,26 +93,26 @@ def parameter_ranges(self) -> Dict[str, List[List[float]]]:
...
 
         This property is optional; if arbitrary topologies are allowed by the device,
         this will simply return ``None``.
 
         Returns:
-            networkx.DAGGraph: a directed acyclic
+            networkx.DiGraph: a directed acyclic graph
         """
-        if self.blackbird_template is None:
+        if self.circuit is None:
```
co9olguy: I didn't catch where `self.circuit` is actually declared. Did I miss something?

strawberryfields/program.py
```diff
@@ -696,11 +696,16 @@ def compile(self, backend='fock', **kwargs):
...
-            db = specs.backend_databases[backend]()
+        if backend in specs.backend_specs:
+            db = specs.backend_specs[backend]()
         else:
             raise ValueError("Could not find backend {} in Strawberry Fields database".format(backend))
 
+        if db.modes is not None:
+            if self.num_subsystems > db.modes:
+                raise CircuitError("This program requires {} modes, but the {} backend "
+                                   "only supports an {} mode program".format(self.num_subsystems, backend, db.modes))
```
co9olguy: an->a

strawberryfields/devicespecs/device_specs.py
josh146: Okay, I'll make this more clear

strawberryfields/program.py
josh146: done

strawberryfields/devicespecs/__init__.py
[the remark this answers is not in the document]
josh146: Done (did as you suggested, made `BaseSpecs` which simply allows what's covered by `BaseBackend`).

strawberryfields/program.py
co9olguy: Ok, so this code checks that the graph topology matches, and that at each node, the ops have the same name. Does it check that these ops have the same parameters?

strawberryfields/devicespecs/device_specs.py
co9olguy: There are two options:

- Have `DeviceSpec`s registered separately for local and remote versions of backends that are otherwise the same (e.g., Gaussian backend running locally or remotely). 

- Have a single `DeviceSpec` for a backend type, and extra properties `local` and `remote` which indicate what operation types are supported (could be one or the other, or both).

I think second is much easier

tests/frontend/test_devicespecs.py
co9olguy: ok so the parameters that are programmable should all get converted to zeros, but others will stay as specified?

strawberryfields/devicespecs/device_specs.py
co9olguy: I don't see any changes. Was this made clearer in another location/file?

tests/frontend/test_devicespecs.py
josh146: Yes. I'm not sure I like this (using zero/non-zero to distinguish free/fixed parameters), but it needs more thinking.

strawberryfields/devicespecs/device_specs.py
josh146: Line 134 and below:

https://github.com/XanaduAI/strawberryfields/blob/0e010d373f11866f8321a385e097b78de47ba4dc/strawberryfields/devicespecs/device_specs.py#L134

strawberryfields/program.py
josh146: Not yet. Currently, it just checks the name.

I tried to code this in, but it sort of bleeds into the compilation problem as well, so I removed that code. Do you want me to add it back/finish it?

co9olguy: So currently two graphs are considered isomorphic if they have all the same gates in the same places, but whether arguments match or not does not influence it?

strawberryfields/devicespecs/device_specs.py
josh146: Added the second option :+1:

strawberryfields/program.py
josh146: Yes, that is not currently done - this will be done in a new ticket.

Issue #70, opened 2019-04-25
Add sf.about() and sf.cite()


smite:
#### Issue description

Stealing the idea from https://github.com/XanaduAI/pennylane/issues/181 and @quantshah,
SF could use two new top-level functions `sf.about()` and `sf.cite()`.

* `sf.about()` prints all the relevant system information that should be included in a bug report.
* `sf.cite()` prints a Bibtex citation that can be easily copypasted to cite the Strawberry Fields white paper.

`sf.about()` would allow us to replace all the stuff below ↓

* **Strawberry Fields version:**
  This can be found by running
  python -c "import strawberryfields as sf; print(sf.version())"

* **Python version:**
  This can be found by running: python --version

* **NumPy, SciPy, and Tensorflow versions:**
  These can be found by running
  python -c "import numpy as np; import scipy as sp; import tensorflow as tf; print(np.__version__,sp.__version__,tf.__version__)"

PR #85, opened 2019-05-20
Add about() and cite() functions

smite:

Implements issue #70.

Adds two top-level functions:

* `about()`, which prints human-readable system info including installed versions of various Python packages
* `cite()`, which prints a bibtex citation for SF

I also added the SF citation to the `references.bib` file.

**Benefits:**

Complete, informative bug reports are easier for the users to produce.  Possibly more citations :)

### DIFF
--- a/strawberryfields/__init__.py
+++ b/strawberryfields/__init__.py
@@ -92,8 +92,8 @@ def about():
     import scipy
 
     # a QuTiP-style infobox
-    print('\nStrawberry Fields: a Python library for continuous variable quantum circuits.')
-    print('Copyright 2019 Xanadu Quantum Technologies Inc.\n')
+    print('\nStrawberry Fields: a Python library for continuous-variable quantum circuits.')
+    print('Copyright 2018-2019 Xanadu Quantum Technologies Inc.\n')
 
     print('Python version:            {}.{}.{}'.format(*sys.version_info[0:3]))
     print('Platform info:             {}'.format(platform.platform()))
--- a/strawberryfields/__init__.py
+++ b/strawberryfields/__init__.py
@@ -103,9 +103,10 @@ def about():
     print('Scipy version:             {}'.format(scipy.__version__))
     try:
         import tensorflow
-        print('TensorFlow version:        {}'.format(tensorflow.__version__))
-    except:
-        pass
+        tf_version = tensorflow.__version__
+    except ModuleNotFoundError:
+        tf_version = None
+    print('TensorFlow version:        {}'.format(tf_version))
 
 
 def cite():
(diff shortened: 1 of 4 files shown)


strawberryfields/__init__.py
```diff
@@ -75,5 +79,48 @@ def version():
...
+    and some system info. Please include this information in bug reports.
+    """
+    import sys
+    import platform
+    import os
+    import numpy
+    import scipy
+
+    # a QuTiP-style infobox
+    print('\nStrawberry Fields: a Python library for continuous variable quantum circuits.')
```
co9olguy:
```suggestion
    print('\nStrawberry Fields: a Python library for continuous-variable quantum circuits.')
```

strawberryfields/__init__.py
```diff
@@ -75,5 +79,48 @@ def version():
...
+    print('Copyright 2019 Xanadu Quantum Technologies Inc.\n')
```
co9olguy:
```suggestion
    print('Copyright 2018-2019 Xanadu Quantum Technologies Inc.\n')
```

josh146:
> Nice, thanks. Any thoughts on simple tests for these?
> Probably sufficient to just check that the two commands run without failing

Yep, should be sufficient to just replicate [`test_about.py`](https://github.com/XanaduAI/pennylane/blob/master/tests/test_about.py) from the PennyLane repo @smite

smite:
> Yep, should be sufficient to just replicate [`test_about.py`](https://github.com/XanaduAI/pennylane/blob/master/tests/test_about.py) from the PennyLane repo @smite

Added the tests. BTW, it seems to me the asserts in the PennyLane version do not actually test anything, since the `readouterr` method always returns two strings.

strawberryfields/__init__.py
```diff
@@ -75,5 +79,48 @@ def version():
...
 
-#: float: numerical value of hbar for the frontend (in the implicit units of position * momentum)
-hbar = 2
+def about():
+    """About box for Strawberry Fields.
+
+    Prints the installed version numbers for SF and its dependencies,
+    and some system info. Please include this information in bug reports.
+    """
+    import sys
```
jswinarton: Any reason these are imported in the function and not at the top of the file?

josh146: This will avoid these superfluous imports from occurring on standard Python import - they will only be imported when `about()` or `cite()` is called.

On repeated calls, they will not be re-imported.

I assume this is the reason, @smite?

smite: That's the idea, although I think most of them are imported anyway in the other modules.

PR #86, opened 2019-05-24
Decompositions module docstrings updated.

smite:

Docstrings updated in the `decompositions` module.
Also fixes a (latent) bug in `graph_embed`.

### DIFF
--- a/doc/conventions/gates.rst
+++ b/doc/conventions/gates.rst
@@ -135,7 +135,7 @@ Two important special cases of the last formula are obtained when :math:`r \to 0
 
   to deduce that :math:`f_{n,m}(r,\phi,0)` is zero if :math:`n` is even and :math:`m` is odd or vice versa.
 
-When writing the Bloch Messiah reduction :cite:`gianfranco2016`:cite:`gianfranco2016b` of a Gaussian state in the Fock basis one often needs the following matrix element
+When writing the Bloch-Messiah reduction :cite:`cariolaro2016`:cite:`cariolaro2016b` of a Gaussian state in the Fock basis one often needs the following matrix element
 
 .. math::
    \bra{k} D(\alpha) R(\theta) S(r) \ket{l}  = e^{i \theta l }  \bra{k} D(\alpha) S(r e^{2i \theta}) \ket{l} = e^{i \theta l} f^*_{l,k}(-r,-2\theta,-\alpha)
--- a/doc/conventions/states.rst
+++ b/doc/conventions/states.rst
@@ -101,7 +101,7 @@ In the limit :math:`r \to \infty`, the squeezed state's variance in the :math:`\
 
 We can also consider the case :math:`r \to -\infty`, where we find a state proportional to the eigenstate :math:`\ket{p}` of the :math:`\p` quadrature with eigenvalue 0. In the laboratory and in numerical simulation we must approximate every quadrature eigenstate using a finitely squeezed state (being careful that the variance of the relevant quadrature is much smaller than any other uncertainty relevant to the system). Any other quadrature eigenstate can be obtained from the :math:`x=0` eigenstate by applying suitable displacement and rotation operators.
 
-Finally, note that Gaussian operations will transform the vector of means via an affine transformation and the covariance matrix via similarity transformation; for a detailed discussion of these transformation, see Sec. 2 of *Gaussian quantum information* by Weedbrook et al. in the :ref:`further reading <futher_reading>` section.
+Finally, note that Gaussian operations will transform the vector of means via an affine transformation and the covariance matrix via similarity transformation; for a detailed discussion of these transformation, see Sec. 2 of *Gaussian quantum information* by Weedbrook et al. in the :ref:`further reading <further_reading>` section.
 
 Given a :math:`2N \times 2N` real symmetric matrix, how can we check that it is a valid covariance matrix? And if it is valid, which operations (displacement, squeezing, multiport interferometers) should be performed to prepare the corresponding Gaussian state?
 
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -11,7 +11,36 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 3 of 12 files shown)


josh146:
Added @nquesada (the main decompositions developer) as a reviewer

doc/conventions/gates.rst
```diff
@@ -135,7 +135,7 @@ Two important special cases of the last formula are obtained when :math:`r \to 0
 
   to deduce that :math:`f_{n,m}(r,\phi,0)` is zero if :math:`n` is even and :math:`m` is odd or vice versa.
 
-When writing the Bloch Messiah reduction :cite:`gianfranco2016`:cite:`gianfranco2016b` of a Gaussian state in the Fock basis one often needs the following matrix element
+When writing the Bloch-Messiah reduction :cite:`cariolaro2016`:cite:`cariolaro2016b` of a Gaussian state in the Fock basis one often needs the following matrix element
```
josh146: were these incorrect citations?

doc/conventions/states.rst
```diff
@@ -101,7 +101,7 @@ In the limit :math:`r \to \infty`, the squeezed state's variance in the :math:`\
...
 
-Finally, note that Gaussian operations will transform the vector of means via an affine transformation and the covariance matrix via similarity transformation; for a detailed discussion of these transformation, see Sec. 2 of *Gaussian quantum information* by Weedbrook et al. in the :ref:`further reading <futher_reading>` section.
+Finally, note that Gaussian operations will transform the vector of means via an affine transformation and the covariance matrix via similarity transformation; for a detailed discussion of these transformation, see Sec. 2 of *Gaussian quantum information* by Weedbrook et al. in the :ref:`further reading <further_reading>` section.
```
josh146: Nice catch

doc/conventions/gates.rst
josh146: Nevermind, I see that you updated the bibtex key to make it better fit convention

strawberryfields/decompositions.py
```diff
@@ -79,36 +107,40 @@ def takagi(N, tol=1e-13, rounding=13):
...
+        max_mean_photon (float): Threshold value. It guarantees that the mode with
             the largest squeezing has ``max_mean_photon`` as the mean photon number
-            i.e., :math:`sinh(r_{max})^2 == max_mean_photon`
-        make_traceless (boolean): removes the trace of the input matrix, by performing the transformation
+            i.e., :math:`sinh(r_{max})^2 ==` max_mean_photon.
+        make_traceless (bool): Removes the trace of the input matrix, by performing the transformation
             :math:`\tilde{A} = A-\mathrm{tr}(A) \I/n`. This may reduce the amount of squeezing needed to encode
             the graph.
-        tol (float): the tolerance used when checking if the input matrix is symmetric: :math:`|mat-mat^T| < tol`
+        tol (float): Tolerance used when checking if the input matrix is symmetric: :math:`|A-A^T| <` tol.
```
josh146: Following Nathan's rule, this is a sentence fragment, so doesn't need capitalization and periods (I think I have this right)
```suggestion
        tol (float): tolerance used when checking if the input matrix is symmetric: :math:`|A-A^T| <` tol
```

strawberryfields/decompositions.py
```diff
@@ -79,36 +107,40 @@ def takagi(N, tol=1e-13, rounding=13):
...
+        raise ValueError("The matrix is not square.")
 
-    if np.linalg.norm(mat-np.transpose(mat)) >= tol:
-        raise ValueError("The input matrix is not symmetric")
+    if np.linalg.norm(A-np.transpose(A)) >= tol:
+        raise ValueError("The matrix is not symmetric.")
 
     if make_traceless:
-        A = mat - np.trace(mat)*np.identity(n)/n
+        A = A - np.trace(A)*np.identity(n)/n
```
nquesada: I understand now. I didn't know that in python numpy arrays are not passed by reference. The way it was written before was assuming that they were passed by reference and thus I was trying to prevent the matrix the user passes from being overwritten. I understand this is sub optimal, but fail to see why it is a bug.

strawberryfields/decompositions.py
```diff
@@ -22,21 +51,20 @@
 
 
 def takagi(N, tol=1e-13, rounding=13):
-    r"""Computes the Autonne-Takagi decomposition of a complex symmetric (not Hermitian!) matrix.
+    r"""Autonne-Takagi decomposition of a complex symmetric (not Hermitian!) matrix.
```
co9olguy:
```suggestion
    r"""Autonne-Takagi decomposition of a complex symmetric (not necessarily Hermitian) matrix.
```
@nquesada can you confirm this suggestion is ok?

strawberryfields/decompositions.py
```diff
@@ -79,36 +107,40 @@ def takagi(N, tol=1e-13, rounding=13):
...
+
     Args:
-        mat (array): square symmetric complex (or real or integer) array representing a (weighted) adjacency matrix of a graph
-        max_mean_photon (float): threshold value. It guarantees that the mode with
+        A (array[complex]): square, symmetric (weighted) adjacency matrix of the graph
+        max_mean_photon (float): Threshold value. It guarantees that the mode with
             the largest squeezing has ``max_mean_photon`` as the mean photon number
-            i.e., :math:`sinh(r_{max})^2 == max_mean_photon`
-        make_traceless (boolean): removes the trace of the input matrix, by performing the transformation
+            i.e., :math:`sinh(r_{max})^2 ==` max_mean_photon.
```
co9olguy:
```suggestion
            i.e., :math:`sinh(r_{max})^2 ==` ``max_mean_photon``.
```

strawberryfields/decompositions.py
co9olguy: thumb

strawberryfields/decompositions.py
smite: Yeah, Python can be occasionally surprising if you come from a C/C++ background. What happens here (as far as I know :) is that the subtraction creates a new array object that then gets bound to the local symbol A by the assignment, which releases the previous binding. 

The bug: previously, if make_traceless was False, nothing would be assigned to A, and the takagi call below would throw an error.

PR #87, opened 2019-05-27
Addition of rectangular_symmetric_decomposition to decompositions.py

lneuhaus:

Addition of a new, low-level variant of Clements decomposition that extends the function `strawberryfields.decompositions.clements_phase_end`. This variant, for which I propose the name 'rectangular_symmetric_decomposition', maps the unitary on a rectangular array of 2x2 symmetric Mach-Zehnder interferometers with one external phase shift to be applied to the upper input mode, and one internal phase shift to be applied to the upper mode between the two 50:50 beamsplitters. This function uses the convention that both beamsplitters are described by the beamsplitter `BSGate(pi/4, pi/2)`. The function returns the same data structure as `clements_phase_end` with the difference that the phases `theta` and `phi` of the original version are replaced by the new phases `internal_phase` and `external_phase`.

**Benefits:**
The new decomposition yields numbers that can be directly applied to calibrated phase-shifters in a hardware implementation of the above mentioned circuit. 

**Possible Drawbacks:**
Possibly confusing name of the new function. Alternative name propositions are welcome. 

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -216,6 +216,7 @@ def clements(V, tol=1e-11):
 
     return tilist, tlist, np.diag(localV)
 
+
 def clements_phase_end(V, tol=1e-11):
     r"""Clements decomposition of unitary matrix
 
@@ -257,6 +258,74 @@ def clements_phase_end(V, tol=1e-11):
         new_tlist = new_tlist + [new_i]
 
     return (new_tlist, new_diags)
+
+
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Rectangular decomposition of a unitary into symmetric beamsplitters.
+
+    This decomposition starts with the output from :func:`clements_phase_end`
+    and further decomposes each of the T unitaries into two phase-shifters and
+    two symmetric (50:50) beamsplitters.
+
+    The two beamsplitters in this decomposition of T are modeled by :class:`ops.BSgate`
+    with arguments (pi/4, pi/2), and the two phase-shifters (see :class:`ops.Rgate`)
+    act on the input mode with the lower index of the two. The phase imposed
+    by the first phaseshifter (before the first beamsplitter) is named
+    `external_phase`, while we call the phase shift between the beamsplitters
+    `internal_phase`.
+
+    The algorithm applied in this function makes use of the following identity:
+
+    ::
+
+        Rgate(alpha) | 1
+        Rgate(beta) | 2
+        Rgate(phi) | 1
+        BSgate(theta, 0) | 1, 2
+
+        equals
+
+        Rgate(phi+alpha-beta) | 1
+        BSgate(pi/4, pi/2) | 1, 2
+        Rgate(2*theta+pi) | 1, 2
+        BSgate(pi/4, pi/2) | 1, 2
+        Rgate(beta-theta+pi) | 1
+        Rgate(beta-theta) | 2
+
+    The phase-shifts by alpha and beta are thus pushed consecutively through
+    all the T unitaries of the interferometer and these unitaries are converted
+    into pairs of symmetric beamsplitters with two phase shifts. The phase
+    shifts at the end of the interferometer are added to the ones from the
+    diagonal unitary at the end of the interferometer obtained from :func:`clements_phase_end`.
+
+    Args:
+        V (array): Unitary matrix of size n_size
+        tol (int): the number of decimal places to use when determining
+          whether the matrix is unitary
+
+    Returns:
+        tuple[array]: returns a tuple of the form ``(tlist,np.diag(localV))``
+            where:
+
+            * ``tlist``: list containing ``[n,m,internal_phase,external_phase,n_size]`` of the T unitaries needed
+            * ``localV``: Diagonal unitary matrix to be applied at the end of circuit
+    """
+    tlist, diags = clements_phase_end(V, tol)
+    new_tlist, new_diags = [], np.ones(len(diags), dtype=diags.dtype)
+    for i in tlist:
+        em, en = int(i[0]), int(i[1])
+        alpha, beta = np.angle(new_diags[em]), np.angle(new_diags[en])
+        theta, phi = i[2], i[3]
+        external_phase = np.fmod((phi + alpha - beta), 2 * np.pi)
+        internal_phase = np.fmod((np.pi + 2.0 * theta), 2 * np.pi)
+        new_alpha = beta - theta + np.pi
+        new_beta = 0*np.pi - theta + beta
+        new_i = [i[0], i[1], internal_phase, external_phase, i[4]]
+        new_diags[em], new_diags[en] = np.exp(1j*new_alpha), np.exp(1j*new_beta)
+        new_tlist = new_tlist + [new_i]
+    new_diags = diags * new_diags
+    return (new_tlist, new_diags)
+
 
 def triangular_decomposition(V, tol=1e-11):
     r"""Triangular decomposition of unitary due to Reck et al.
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -260,12 +260,33 @@ def clements_phase_end(V, tol=1e-11):
     return (new_tlist, new_diags)
 
 
+def mach_zehnder(m, n, internal_phase, external_phase, nmax):
+    r"""A two-mode Mach-Zehnder interferometer section.
+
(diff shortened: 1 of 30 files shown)


strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
     return (new_tlist, new_diags)
 
 
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Rectangular decomposition of a unitary into symmetric beamsplitters.
+
+    This decomposition starts with the output from :func:`clements_phase_end`
```
```suggestion
    This decomposition starts with the output from :func:`~.clements_phase_end`
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+    and further decomposes each of the T unitaries into two phase-shifters and
+    two symmetric (50:50) beamsplitters.
+
+    The two beamsplitters in this decomposition of T are modeled by :class:`ops.BSgate`
+    with arguments (pi/4, pi/2), and the two phase-shifters (see :class:`ops.Rgate`)
```
```suggestion
    with arguments :math:`(\pi/4, \pi/2)`, and the two phase-shifters (see :class:`~.ops.Rgate`)
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
 
 
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Rectangular decomposition of a unitary into symmetric beamsplitters.
+
+    This decomposition starts with the output from :func:`clements_phase_end`
+    and further decomposes each of the T unitaries into two phase-shifters and
+    two symmetric (50:50) beamsplitters.
+
+    The two beamsplitters in this decomposition of T are modeled by :class:`ops.BSgate`
```
```suggestion
    The two beamsplitters in this decomposition of T are modeled by :class:`~.ops.BSgate`
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+    with arguments (pi/4, pi/2), and the two phase-shifters (see :class:`ops.Rgate`)
+    act on the input mode with the lower index of the two. The phase imposed
+    by the first phaseshifter (before the first beamsplitter) is named
+    `external_phase`, while we call the phase shift between the beamsplitters
```
```suggestion
    ``external_phase``, while we call the phase shift between the beamsplitters
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+    `internal_phase`.
```
```suggestion
    ``internal_phase``.
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+
+    The algorithm applied in this function makes use of the following identity:
+
+    ::
```
```suggestion
    .. code-block:: python
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+        Rgate(2*theta+pi) | 1, 2
+        BSgate(pi/4, pi/2) | 1, 2
+        Rgate(beta-theta+pi) | 1
+        Rgate(beta-theta) | 2
+
+    The phase-shifts by alpha and beta are thus pushed consecutively through
+    all the T unitaries of the interferometer and these unitaries are converted
+    into pairs of symmetric beamsplitters with two phase shifts. The phase
+    shifts at the end of the interferometer are added to the ones from the
+    diagonal unitary at the end of the interferometer obtained from :func:`clements_phase_end`.
```
```suggestion
    diagonal unitary at the end of the interferometer obtained from :func:`~.clements_phase_end`.
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+        equals
+
+        Rgate(phi+alpha-beta) | 1
+        BSgate(pi/4, pi/2) | 1, 2
+        Rgate(2*theta+pi) | 1, 2
+        BSgate(pi/4, pi/2) | 1, 2
+        Rgate(beta-theta+pi) | 1
+        Rgate(beta-theta) | 2
+
+    The phase-shifts by alpha and beta are thus pushed consecutively through
```
```suggestion
    The phase-shifts by ``alpha`` and ``beta`` are thus pushed consecutively through
```

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
...
+    all the T unitaries of the interferometer and these unitaries are converted
+    into pairs of symmetric beamsplitters with two phase shifts. The phase
+    shifts at the end of the interferometer are added to the ones from the
+    diagonal unitary at the end of the interferometer obtained from :func:`clements_phase_end`.
+
+    Args:
+        V (array): Unitary matrix of size n_size
```
```suggestion
        V (array): unitary matrix of size n_size
```

tests/frontend/test_decompositions.py
```diff
@@ -213,6 +213,72 @@ def test_random_unitary_phase_end(self, tol):
...
+        with pytest.raises(ValueError, match="matrix is not unitary"):
+            dec.rectangular_symmetric(A)
+
+    @classmethod
+    def S(cls, m, n, internal_phase, external_phase, nmax):
+        r"""a section composed by external phase shift, BS, internal phase shift, BS
+
+        .. todo::
+
+           Move this function into decompositions and maybe rename and / or optimize.
```
josh146: :+1:

lneuhaus: you have a name suggestion? Maybe 'MachZehnder'?

co9olguy: I like that name, it seems to be the term I hear most often (assuming it is accurate and won't mislead people, of course)

josh146: How about `mach_ zehnder` to match the python convention

strawberryfields/decompositions.py
```diff
@@ -291,6 +291,73 @@ def clements_phase_end(V, tol=1e-11):
     return (new_tlist, new_diags)
 
 
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Rectangular decomposition of a unitary into symmetric beamsplitters.
```
nquesada: ... into an array of symmetric beamsplitters?

tests/frontend/test_decompositions.py
nquesada: I would much prefer to not use names. Symmetric is quite descriptive.

lneuhaus: @nquesada, the question is whether this test function, which essentially is a modified BSGate with the rectangular_symmetric phase convention (i.e. taking two phases and outputting a matrix), should go into decompositions.py, akin to the function T that represents a BSGate in the Clements decomposition. If it does, it needs a name. Currently I randomly chose S, but it could also be mach_zehner, or even MZGate. What option would you suggest here (S, mach_zehnder, MZGate, not moving into decompositions.py at all, or something else)?

strawberryfields/decompositions.py
lneuhaus: Implemented this in commit c230be7.

tests/frontend/test_decompositions.py
lneuhaus: I chose the option `mach_zehnder` and added a commit for this now. But I can undo it if you want.

josh146: > the question is whether this test function, which essentially is a modified BSGate with the rectangular_symmetric phase convention (i.e. taking two phases and outputting a matrix), should go into decompositions.py

IMO it should definitely go into `decompositions.py` - it is useful to have the decomposition separate from Strawberry Fields classes and objects.

In terms of accessing this operation from the SF frontend, this should be done in a future PR, but I can think of two options:

* Introduce an `MZgate(theta, phi)` operation in Strawberry Fields that uses the above decomposition, compiling to a combination of `Rgate`'s and `BSgate(pi/4, pi/2)`.

* Add a keyword argument to `Interferometer` allowing you to change the geometry of the unitary decomposition, i.e.,

  ```python
  Interferometer(U, geometry='rectangular')
  Interferometer(U, geometry='triangular')
  Interferometer(U, geometry='mach_zehnder') # or maybe 'hardware'?
  ```

As both have valid use cases, I'm leaning towards implementing both approaches.

PR #88, opened 2019-05-31
Remove TF dependency from Strawberry Fields

josh146:

* Removes TF from `setup.py` (it is install in requirements.txt for our continuous integration).

* Adds a try-except for tensorflow import in `parameters.py`. If there is an `ImportError`, the module `tf` is simply mocked out.

* Added TensorFlow import logic and a useful exception message to the `TFBackend`, that checks whether TensorFlow is installed and is the correct version. If not, information is given to the user on how to install SF with TensorFlow support.

**Benefits:**

* Strawberry Fields will now install and run correctly without TensorFlow dependence. The exception message above will only be triggered if the user specifically requests the `'tf'` backend.

**Possible Drawbacks:**

* This is a stopgap solution; at some point, we should aim to spin the TF simulator to its own package.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -77,6 +77,7 @@ mock_fns.update({"pi": MagicMock(),
                  "Number": int,
                  "Tensor": list,
                  "Variable": list,
+                 "__version__": "1.3",
                  "ndarray": MagicMock})
 
 mock = Mock(**mock_fns)
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -63,7 +63,7 @@ Python version:
     python --version
 
 If the above prints out 3.7, and you still want to use TensorFlow, you will need to install Python 3.6.
-The recommended method is to install `Miniconda3 <https://docs.conda.io/en/latest/miniconda.html>`_.
+The recommended method is to install `Anaconda for Python 3 <https://www.anaconda.com/download/>`_.
 
 Once installed, you can then create a Python 3.6 Conda environment:
 
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -20,38 +20,58 @@ as well as the following Python packages:
 * `NumPy <http://numpy.org/>`_  >=1.13.3
 * `SciPy <http://scipy.org/>`_  >=1.0.0
 * `NetworkX <http://networkx.github.io/>`_ >=2.0
-* `Tensorflow <https://www.tensorflow.org/>`_ >=1.3,<1.7
+* `Blackbird <https://quantum-blackbird.readthedocs.io>`_ >=0.2.0
 
 
 If you currently do not have Python 3 installed, we recommend `Anaconda for Python 3 <https://www.anaconda.com/download/>`_, a distributed version of Python packaged for scientific computation.
-
-.. note::
-
-    As Tensorflow does not currently support Python 3.7 and above, we are unable to
-    also support Strawberry Fields on Python 3.7.
 
 
 Installation
 ============
 
 Installation of Strawberry Fields, as well as all required Python packages mentioned above, can be installed via ``pip``:
-::
 
-    $ python -m pip install strawberryfields
+.. code-block:: console
 
-
-If you are using the ``tensorflow-gpu`` module for TensorFlow GPU support, you can install the following package for GPU support in Strawberry Fields:
-::
-
-    $ python -m pip install strawberryfields-gpu
+    pip install strawberryfields
 
 
-Make sure you are using the Python 3 version of pip.
+TensorFlow support
+==================
 
-Alternatively, you can install Strawberry Fields from the source code by navigating to the top directory and running
-::
+To use Strawberry Fields with TensorFlow, version 1.3 of
+TensorFlow is required. This can be installed alongside Strawberry Fields
+as follows:
 
-    $ python setup.py install
+.. code-block:: console
+
+    pip install strawberryfields tensorflow==1.3
+
+Or, to install Strawberry Fields and TensorFlow with GPU and CUDA support:
+
+.. code-block:: console
+
+    pip install strawberryfields tensorflow-gpu==1.3
+
+
+Note that TensorFlow version 1.3 is only support on Python version
+less than 3.7. You can use the following command to check your
+Python version:
+
+.. code-block:: console
+
+    python --version
+
+If the above prints out 3.7, you will need to install Python 3.6.
+The recommended method is to install `Miniconda3 <https://docs.conda.io/en/latest/miniconda.html>`_.
+
+Once installed, you can then create a Python 3.6 Conda environment:
+
+.. code-block:: console
+
+    conda create --name new_env python=3.6
+    conda activate new_env
+    pip install strawberryfields tensorflow==1.3
 
 
 Notebook downloads
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -54,7 +54,7 @@ Or, to install Strawberry Fields and TensorFlow with GPU and CUDA support:
     pip install strawberryfields tensorflow-gpu==1.3
 
 
-Note that TensorFlow version 1.3 is only support on Python version
+Note that TensorFlow version 1.3 is only supported on Python versions
 less than 3.7. You can use the following command to check your
 Python version:
 
@@ -62,15 +62,15 @@ Python version:
 
     python --version
 
-If the above prints out 3.7, you will need to install Python 3.6.
+If the above prints out 3.7, and you still want to use TensorFlow, you will need to install Python 3.6.
 The recommended method is to install `Miniconda3 <https://docs.conda.io/en/latest/miniconda.html>`_.
 
 Once installed, you can then create a Python 3.6 Conda environment:
 
 .. code-block:: console
 
-    conda create --name new_env python=3.6
-    conda activate new_env
+    conda create --name sf_tensorflow_env python=3.6
+    conda activate sf_tensorflow_env
     pip install strawberryfields tensorflow==1.3
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 networkx>=2.0
 numpy>=1.16.3
 scipy>=1.0.0
-tensorflow>=1.3.0,<1.7
+tensorflow==1.3
 tensorflow-tensorboard>=0.1.8
 quantum-blackbird
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,6 @@ requirements = [
     "numpy>=1.16.3",
     "scipy>=1.0.0",
     "networkx>=2.0",
-    "tensorflow>=1.3.0,<1.7",
     "quantum-blackbird"
 ]
 
--- a/strawberryfields/backends/__init__.py
+++ b/strawberryfields/backends/__init__.py
@@ -14,16 +14,20 @@
 """This module loads the required backend classes"""
 
 from .base import BaseBackend, BaseFock, BaseGaussian, ModeMap
-from .tfbackend import TFBackend
 from .gaussianbackend import GaussianBackend
 from .fockbackend import FockBackend
 
-__all__ = ['BaseBackend', 'BaseFock', 'BaseGaussian', 'FockBackend', 'GaussianBackend', 'TFBackend']
+__all__ = [
+    "BaseBackend",
+    "BaseFock",
+    "BaseGaussian",
+    "FockBackend",
+    "GaussianBackend",
+    "TFBackend",
+]
 
-supported_backends = {"base": BaseBackend,
-                      "tf": TFBackend,
-                      "gaussian": GaussianBackend,
-                      "fock": FockBackend}
+supported_backends = {"base": BaseBackend, "gaussian": GaussianBackend, "fock": FockBackend}
+
 
 def load_backend(name):
     """Loads the specified backend by mapping a string
@@ -31,8 +35,15 @@ def load_backend(name):
     dictionary. Note that this function is used by the
     frontend only, and should not be user-facing.
     """
+    if name == "tf":
+        # treat the tensorflow backend differently, to
+        # isolate the import of tensorflow
+        from .tfbackend import TFBackend
+
+        return TFBackend
+
     if name in supported_backends:
         backend = supported_backends[name]()
         return backend
-    else:
-        raise ValueError("Backend '{}' is not supported.".format(name))
+
+    raise ValueError("Backend '{}' is not supported.".format(name))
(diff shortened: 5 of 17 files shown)


strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,54 @@
...
+    tf_available = True
+    tf_version = tf.__version__
+
+
+tf_info = """\
+ImportError:
+To use Strawberry Fields with TensorFlow support, version 1.3 of
+TensorFlow is required. This can be installed as follows:
+
+pip install tensorflow==1.3
```
jswinarton: Have you thought about having this installed using `extras_require` (i.e. pip install `strawberryfields[tf]`? That seems more standards-based.

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,54 @@
...
+
+Note that TensorFlow version 1.3 is only support on Python version
```
jswinarton: This stuff seems like it would be a better fit for the documentation. Then perhaps a short error message that directs the user to the docs? Then you don't have to specify in both places.

strawberryfields/parameters.py
```diff
@@ -90,17 +90,20 @@
...
 from .program import (RegRef, RegRefTransform)
 
 
-# supported TF classes
-_tf_classes = (tf.Tensor, tf.Variable)
+try:
+    import tensorflow as tf
+    _tf_classes = (tf.Tensor, tf.Variable)
+except ImportError:
+    tf = mock.MagicMock()
```
jswinarton: What's the point of the mock?

josh146: There are (module level) calls to TensorFlow classes, types, and functions, which are only used if the user is using the `'tf'` backend. If the user is using the `'tf'` backend, the `ImportError` exception in `tfbackend/__init__.py` will be raised before the `parameter.py` file is used.

This branch takes care of the case where TensorFlow is not installed, and the user wants to use the `'fock'` or `'gaussian'` backends. The alternative to mocking would be a significant refactor to change the logic required; since the `Parameter` class still needs to be refactor/replaced anyway this cycle, it doesn't make sense to do this now.

strawberryfields/backends/tfbackend/__init__.py
josh146: The only disadvantage would be that tensorflow installation errors would look like they are coming from Strawberry Fields (as it does currently), and I would like to avoid that.

This way, if something goes wrong, it is explicit to the user that the

```python
pip install tensorflow==1.3
```

is failing.

strawberryfields/backends/tfbackend/__init__.py
josh146: True, it should be added to the documentation as well. My preference is to have both, and keep the verbose error message (with commands to run), even though it is a bit of duplication.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 networkx>=2.0
 numpy>=1.16.3
 scipy>=1.0.0
-tensorflow>=1.3.0,<1.7
+tensorflow==1.3
```
josh146: By pinning this to 1.3, we can ensure that everyone on the simulator is now using the same TF version.

doc/conf.py
```diff
@@ -77,6 +77,7 @@ class TypeMock(type):
                  "Number": int,
                  "Tensor": list,
                  "Variable": list,
+                 "__version__": "1.3",
```
josh146: This is done to 'trick' Strawberry Fields into letting it import the `tf` backend without TensorFlow being installed.

setup.py
```diff
@@ -27,7 +27,6 @@
     "numpy>=1.16.3",
     "scipy>=1.0.0",
     "networkx>=2.0",
-    "tensorflow>=1.3.0,<1.7",
     "quantum-blackbird"
```
smite: No minimum version number for quantum-blackbird?

doc/conf.py
co9olguy: might as well write that directly as a comment in file

doc/installing.rst
```diff
@@ -20,38 +20,58 @@ as well as the following Python packages:
...
+    pip install strawberryfields tensorflow==1.3
+
+Or, to install Strawberry Fields and TensorFlow with GPU and CUDA support:
+
+.. code-block:: console
+
+    pip install strawberryfields tensorflow-gpu==1.3
+
+
+Note that TensorFlow version 1.3 is only support on Python version
```
co9olguy:
```suggestion
Note that TensorFlow version 1.3 is only supported on Python versions
```

doc/installing.rst
```diff
@@ -20,38 +20,58 @@ as well as the following Python packages:
...
+less than 3.7. You can use the following command to check your
+Python version:
+
+.. code-block:: console
+
+    python --version
+
+If the above prints out 3.7, you will need to install Python 3.6.
```
co9olguy:
```suggestion
If the above prints out 3.7, and you still want to use TensorFlow, you will need to install Python 3.6.
```

doc/installing.rst
```diff
@@ -20,38 +20,58 @@ as well as the following Python packages:
...
+The recommended method is to install `Miniconda3 <https://docs.conda.io/en/latest/miniconda.html>`_.
```
co9olguy: elsewhere we recommend and link to Anaconda. Should we keep same recommendation throughout?

doc/installing.rst
```diff
@@ -20,38 +20,58 @@ as well as the following Python packages:
...
+
+Once installed, you can then create a Python 3.6 Conda environment:
+
+.. code-block:: console
 
-    $ python setup.py install
+    conda create --name new_env python=3.6
```
co9olguy:
```suggestion
    conda create --name sf_tensorflow_env python=3.6
```

doc/installing.rst
```diff
@@ -20,38 +20,58 @@ as well as the following Python packages:
...
+    conda activate new_env
```
co9olguy:
```suggestion
    conda activate sf_tensorflow_env
```

setup.py
co9olguy: is it because the api is not set yet?

strawberryfields/backends/__init__.py
```diff
@@ -14,25 +14,36 @@
...
 def load_backend(name):
     """Loads the specified backend by mapping a string
     to the backend type, via the ``supported_backends``
     dictionary. Note that this function is used by the
     frontend only, and should not be user-facing.
     """
+    if name == "tf":
+        # treat the tensorflow backend differently, to
+        # isolate the import of tensorflow
+        from .tfbackend import TFBackend
```
co9olguy: no try/except statement in case `TFBackend` fails to import?

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,52 @@
...
```
co9olguy:
```suggestion
Note that TensorFlow version 1.3 is only supported on Python versions
```

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,52 @@
...
+less than 3.7. You can use the following command to check your
+Python version:
+
+python --version
+
+If the above prints out 3.7, you will need to install Python 3.6.
```
co9olguy:
```suggestion
If the above prints out 3.7, and you still want to use TensorFlow, you will need to install Python 3.6.
```

doc/installing.rst
```diff
@@ -20,38 +20,58 @@ as well as the following Python packages:
...
+    pip install strawberryfields
 
-If you are using the ``tensorflow-gpu`` module for TensorFlow GPU support, you can install the following package for GPU support in Strawberry Fields:
-::
 
-    $ python -m pip install strawberryfields-gpu
+TensorFlow support
+==================
 
+To use Strawberry Fields with TensorFlow, version 1.3 of
```
co9olguy: A lot of this is duplicated in the TFBackend file itself. Can we not get sphinx to read that in here?

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,52 @@
...
+The recommended method is to install Miniconda3:
```
co9olguy: same question as above about miniconda

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,52 @@
...
+"""
+
+
+def excepthook(type, value, traceback):
+    """Exception hook to suppress superfluous exceptions"""
+    #pylint: disable=unused-argument
+    print(value)
+
+
+if not (tf_available and tf_version[:3] == "1.3"):
```
co9olguy: Could this cause issues if TF changes how they write their version string in the future?

doc/installing.rst
josh146: Not that I am aware. If the file was a `rst` file we could, using `.. include::`, but it wouldn't work with autodoc.

doc/installing.rst
josh146: Oh, I missed that. Should I update it to miniconda everywhere, or stick with anaconda?

setup.py
josh146: Yep, essentially. I can preliminarily make it `>=0.2.0`?

strawberryfields/backends/__init__.py
josh146: The try-except occurs during `TFBackend/__init__.py` import (or at least, it should)

strawberryfields/backends/tfbackend/__init__.py
josh146: It shouldn't, since they have pledged to stick to semantic versioning for all future releases

strawberryfields/backends/tfbackend/__init__.py
josh146: Changed the above to anaconda

doc/installing.rst
josh146: I've updated to miniconda everywhere

doc/conf.py
josh146:
```suggestion
                 # The following line is to 'trick' Strawberry Fields into
                 # letting Sphinx import the tf backend without TensorFlow being installed.
                 "__version__": "1.3",
```

setup.py
josh146:
```suggestion
    "quantum-blackbird>=0.2.0"
```

josh146:
Note: we should start doing testing with Tox --- this would allow us to easily test different Python versions/TensorFlow versions etc., and is compatible with most CI services (including Travis and Circle).

tests/conftest.py
```diff
@@ -24,7 +24,37 @@
...
+eng_backend_params = [
+    pytest.param("fock", marks=pytest.mark.fock),
+    pytest.param("gaussian", marks=pytest.mark.gaussian),
+]
+
+
+if tf_available and tf.__version__[:3] == "1.3":
+    from strawberryfields.backends.tfbackend import TFBackend
+
+    backend_params.append(pytest.param(TFBackend, marks=pytest.mark.tf))
```
co9olguy: what are these lines doing?

tests/conftest.py
```diff
@@ -198,7 +212,10 @@ def _setup_eng(num_subsystems, **kwargs):
 
 def pytest_runtest_setup(item):
     """Automatically skip tests if they are marked for only certain backends"""
-    allowed_backends = {"gaussian", "tf", "fock"}
+    if tf_available:
```
co9olguy: it seems to have been a lucky accident that this fixture was already here :dango:

tests/integration/test_parameters_integration.py
```diff
@@ -20,15 +20,17 @@
...
-import tensorflow as tf
+# import tensorflow as tf
 
 import strawberryfields as sf
 from strawberryfields import ops
 from strawberryfields.parameters import Parameter
-from strawberryfields.backends.tfbackend import TFBackend
+# from strawberryfields.backends.tfbackend import TFBackend
 
+TFBackend = int # hack to avoid removing TFBackend from test
```
co9olguy: I know it's a hack, but is there anything special about `int`?

tests/integration/test_tf_symbolic.py
```diff
@@ -23,10 +23,18 @@
...
 from scipy.special import factorial
-import tensorflow as tf
+
+try:
+    import tensorflow as tf
+except (ImportError, ModuleNotFoundError):
+    pytestmark = pytest.mark.skip("TensorFlow not installed")
+else:
+    if tf.__version__[:3] != "1.3":
+        pytestmark = pytest.mark.skip("Test only runs with TensorFlow 1.3")
```
co9olguy: Does this cause all tests to then be skipped?

josh146: Yep, since this is a module level attribute, this will skip the entire `test_tf_symbolic` module

tests/integration/test_parameters_integration.py
josh146: Nope. The only place `TFBackend` is being used is in `isinstance()` calls, which requires the arguments to be a `type`.

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,60 @@
...
+less than or equal to 3.6. To continue using TensorFlow with Strawberry Fields,
+you will need to install Python 3.6.
+
+The recommended method is to install Anaconda3:
+
+https://www.anaconda.com/download
+
+Once installed, you can then create a Python 3.6 Conda environment:
+
+conda create --name new_env python=3.6
```
co9olguy:
```suggestion
conda create --name sf_tensorflow_env python=3.6
```

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -141,6 +141,60 @@
...
+conda activate new_env
```
co9olguy:
```suggestion
conda activate sf_tensorflow_env
```

strawberryfields/parameters.py
```diff
@@ -90,17 +90,20 @@
 ~~~~~~~~~~~~
 
 """
-
 import numbers
 
+import mock
```
co9olguy: This forces us to make `mock` a requirement, doesn't it?

josh146:
```suggestion
from unittest import mock
```

PR #90, opened 2019-06-11
Fixing broken or skipped tests

smite:

This PR

* fixes a number of tests that were skipped or marked as broken
* simplifies hbar handling in the `ops.Gaussian` class
* makes `ops.Gaussian` always a Preparation
* improves some docstrings
* introduces a new general tolerance constant for Decompositions

### DIFF
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1817,7 +1817,6 @@ class Gaussian(Preparation, Decomposition):
             self.nbar = 0.5 * (np.diag(th)[:self.ns] - 1.0)
 
         self.decomp = decomp  #: bool: if False, use the backend API call instead of decomposition
-        # FIXME merge() probably does not work for Gaussians if r is not zero?
 
     def _apply(self, reg, backend, **kwargs):
         p = _unwrap(self.p)
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -305,6 +305,7 @@ from .decompositions import clements, bloch_messiah, williamson, graph_embed
 
 # numerical tolerances
 _decomposition_merge_tol = 1e-13
+_decomposition_tol = 1e-13  # TODO this tolerance is used for various purposes and is not well-defined
 
 
 def warning_on_one_line(message, category, filename, lineno, file=None, line=None):
@@ -1596,12 +1597,12 @@ class All(MetaOperation):
 class Interferometer(Decomposition):
     r"""Apply a linear interferometer to the specified qumodes.
 
-    This operation uses the Clements decomposition to decompose
+    This operation uses the :ref:`Clements decomposition <clements>` to decompose
     a linear interferometer into a sequence of beamsplitters and
     rotation gates.
 
     Args:
-        U (array): an :math:`N\times N` complex unitary matrix.
+        U (array[complex]): an :math:`N\times N` unitary matrix
         tol (float): the tolerance used when checking if the matrix is unitary:
             :math:`|UU^\dagger-I| \leq` tol
     """
@@ -1622,20 +1623,20 @@ class Interferometer(Decomposition):
 
         if not self.identity:
             for n, m, theta, phi, _ in self.BS1:
-                if np.abs(phi) >= _decomposition_merge_tol:
+                if np.abs(phi) >= _decomposition_tol:
                     cmds.append(Command(Rgate(phi), reg[n]))
-                if np.abs(theta) >= _decomposition_merge_tol:
+                if np.abs(theta) >= _decomposition_tol:
                     cmds.append(Command(BSgate(theta, 0), (reg[n], reg[m])))
 
             for n, expphi in enumerate(self.R):
-                if np.abs(expphi - 1) >= _decomposition_merge_tol:
+                if np.abs(expphi - 1) >= _decomposition_tol:
                     q = log(expphi).imag
                     cmds.append(Command(Rgate(q), reg[n]))
 
             for n, m, theta, phi, _ in reversed(self.BS2):
-                if np.abs(theta) >= _decomposition_merge_tol:
+                if np.abs(theta) >= _decomposition_tol:
                     cmds.append(Command(BSgate(-theta, 0), (reg[n], reg[m])))
-                if np.abs(phi) >= _decomposition_merge_tol:
+                if np.abs(phi) >= _decomposition_tol:
                     cmds.append(Command(Rgate(-phi), reg[n]))
 
         return cmds
@@ -1675,10 +1676,10 @@ class GraphEmbed(Decomposition):
 
         if not self.identity:
             for n, s in enumerate(self.sq):
-                if np.abs(s) >= _decomposition_merge_tol:
+                if np.abs(s) >= _decomposition_tol:
                     cmds.append(Command(Sgate(s), reg[n]))
 
-            if np.all(np.abs(self.U - np.identity(len(self.U))) >= _decomposition_merge_tol):
+            if np.all(np.abs(self.U - np.identity(len(self.U))) >= _decomposition_tol):
                 cmds.append(Command(Interferometer(self.U), reg))
 
         return cmds
@@ -1687,8 +1688,8 @@ class GraphEmbed(Decomposition):
 class GaussianTransform(Decomposition):
     r"""Apply a Gaussian symplectic transformation to the specified qumodes.
 
-    This operation uses the Bloch-Messiah decomposition to decompose a symplectic
-    matrix :math:`S`:
+    This operation uses the :ref:`Bloch-Messiah decomposition <bloch_messiah>`
+    to decompose a symplectic matrix :math:`S`:
 
     .. math:: S = O_1 R O_2
 
@@ -1703,7 +1704,8 @@ class GaussianTransform(Decomposition):
     where :math:`I` is the :math:`N\times N` identity matrix, and :math:`0` is the zero matrix.
 
     The two orthogonal symplectic unitaries describing the interferometers are then further
-    decomposed via the :class:`~.Interferometer` operator and the Clements decomposition:
+    decomposed via the :class:`~.Interferometer` operator and the
+    :ref:`Clements decomposition <clements>`:
 
     .. math:: U_i = X_i + iY_i
 
@@ -1712,31 +1714,29 @@ class GaussianTransform(Decomposition):
     .. math:: O_i = \begin{bmatrix}X&-Y\\Y&X\end{bmatrix}
 
     Args:
-        S (array): a :math:`2N\times 2N` symplectic matrix describing the Gaussian transformation.
+        S (array[float]): a :math:`2N\times 2N` symplectic matrix describing the Gaussian transformation.
         vacuum (bool): set to True if acting on a vacuum state. In this case, :math:`O_2 V O_2^T = I`,
             and the unitary associated with orthogonal symplectic :math:`O_2` will be ignored.
         tol (float): the tolerance used when checking if the matrix is symplectic:
             :math:`|S^T\Omega S-\Omega| \leq` tol
     """
-    ns = None
-
     def __init__(self, S, vacuum=False, tol=1e-10):
         super().__init__([S])
 
+        self.vacuum = vacuum  #: bool: if True, ignore the first unitary matrix when applying the gate
         N = S.shape[0]//2
 
         # check if input symplectic is passive (orthogonal)
         diffn = np.linalg.norm(S @ S.T - np.identity(2*N))
+        self.active = (np.abs(diffn) > _decomposition_tol)  #: bool: S is an active symplectic transformation
 
-        if np.abs(diffn) <= _decomposition_merge_tol:
+        if not self.active:
             # The transformation is passive, do Clements
-            self.active = False
             X1 = S[:N, :N]
             P1 = S[N:, :N]
             self.U1 = X1+1j*P1
         else:
             # transformation is active, do Bloch-Messiah
-            self.active = True
             O1, smat, O2 = bloch_messiah(S, tol=tol)
             N = S.shape[0]//2
 
@@ -1745,12 +1745,9 @@ class GaussianTransform(Decomposition):
             X2 = O2[:N, :N]
             P2 = O2[N:, :N]
 
-            self.U1 = X1+1j*P1
-            self.U2 = X2+1j*P2
-            self.Sq = np.diagonal(smat)[:N]
-
-        self.ns = N
-        self.vacuum = vacuum
+            self.U1 = X1+1j*P1  #: array[complex]: unitary matrix corresponding to O_1
+            self.U2 = X2+1j*P2  #: array[complex]: unitary matrix corresponding to O_2
+            self.Sq = np.diagonal(smat)[:N]  #: array[complex]: diagonal vector of the squeezing matrix R
 
     def _decompose(self, reg):
         cmds = []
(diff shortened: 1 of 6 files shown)


strawberryfields/ops.py
```diff
@@ -1712,31 +1714,29 @@ class GaussianTransform(Decomposition):
...
 
     Args:
-        S (array): a :math:`2N\times 2N` symplectic matrix describing the Gaussian transformation.
+        S (array[float]): a :math:`2N\times 2N` symplectic matrix describing the Gaussian transformation.
         vacuum (bool): set to True if acting on a vacuum state. In this case, :math:`O_2 V O_2^T = I`,
             and the unitary associated with orthogonal symplectic :math:`O_2` will be ignored.
         tol (float): the tolerance used when checking if the matrix is symplectic:
             :math:`|S^T\Omega S-\Omega| \leq` tol
     """
-    ns = None
```
josh146: It doesn't break the Program by using an operation with `ns` not defined?

strawberryfields/ops.py
```diff
@@ -1745,12 +1745,9 @@ def __init__(self, S, vacuum=False, tol=1e-10):
             X2 = O2[:N, :N]
             P2 = O2[N:, :N]
 
-            self.U1 = X1+1j*P1
-            self.U2 = X2+1j*P2
-            self.Sq = np.diagonal(smat)[:N]
-
-        self.ns = N
```
josh146: If I recall, setting `ns` here allowed a exception to be raised if `2*len(modes) != len(S)`, but I could be wrong?

strawberryfields/ops.py
```diff
@@ -1823,21 +1817,21 @@ def __init__(self, V, r=None, decomp=True, tol=1e-6):
...
         # FIXME merge() probably does not work for Gaussians if r is not zero?
 
     def _apply(self, reg, backend, **kwargs):
         p = _unwrap(self.p)
         s = sqrt(sf.hbar / 2)  # scaling factor, since the backend API call is hbar-independent
-        backend.prepare_gaussian_state(p[1]/s, p[0]/(s*s), reg)
+        backend.prepare_gaussian_state(p[1]/s, p[0], reg)
 
     def _decompose(self, reg):
+        # FIXME this is a Preparation, are some of the decompositions below missing Vacuum preparations?
```
josh146: Good catch - I think this can be fixed simply by changing the `Sgate`'s to `Squeezed` state preparations

tests/frontend/test_ops_decompositions.py
```diff
@@ -275,7 +275,6 @@ def test_decomposition(self, hbar, tol):
         assert np.allclose(ratio, np.ones([n, n]), atol=tol, rtol=0)
 
 
-@pytest.mark.broken('FIXME hbar issue')
```
josh146: nice!

tests/frontend/test_ops_decompositions.py
```diff
@@ -285,41 +284,20 @@ def test_merge(self, hbar, tol):
...
+        G1inv = ops.GaussianTransform(np.linalg.inv(S1))
+        G2 = ops.GaussianTransform(S2)
 
         # a symplectic merged with its inverse is identity
         assert G1.merge(G1inv) is None
 
         # two merged symplectics are the same as their product
         assert np.allclose(G1.merge(G2).p[0].x, S2 @ S1, atol=tol, rtol=0)
 
-    def test_setting_hbar(self, hbar):
```
josh146: Is hbar now determined fully through `sf.hbar`? So if a user wants to use `hbar=0.5`, they use `sf.hbar = 0.5`, and then they never have to provide it to `Gaussian`?

strawberryfields/ops.py
smite: Oops, you're right.

strawberryfields/ops.py
smite: Fixed.

strawberryfields/ops.py
smite: Ok, I'll fix it.

tests/frontend/test_ops_decompositions.py
smite: Yes.

smite: Besides I think `GaussianTransform` was always hbar-independent since it's just a symplectic transform. `Gaussian` OTOH depends on hbar.

strawberryfields/ops.py
smite: Had to add Vacuum preparations as else-cases.

strawberryfields/ops.py
```diff
@@ -1803,39 +1795,34 @@ class Gaussian(Preparation, Decomposition):
...
-        #: float: value of :math:`\hbar` used in the definition of the :math:`\x` and :math:`\p` quadrature operators
-        # TODO can we just divide V by hbar/2 here and remove hbar from further expressions?
-        self.hbar = sf.hbar
-        self.ns = V.shape[0]//2
+        # internally we eliminate hbar from the covariance matrix V (or equivalently set hbar=2), but not from the means vector r
+        V = V / (sf.hbar / 2)
+        self.ns = V.shape[0] // 2
 
         if r is None:
-            # all zeroes
```
co9olguy: 😸

strawberryfields/ops.py
```diff
@@ -1803,39 +1795,34 @@ class Gaussian(Preparation, Decomposition):
...
-            self.nbar = np.diag(th)[:self.ns]/self.hbar - 0.5
-
-        super().__init__([V, r])
-
-        self.decomp = decomp
+            self.pure = np.abs(np.linalg.det(V) - 1.0) < tol
+            self.nbar = 0.5 * (np.diag(th)[:self.ns] - 1.0)
 
+        self.decomp = decomp  #: bool: if False, use the backend API call instead of decomposition
         # FIXME merge() probably does not work for Gaussians if r is not zero?
```
co9olguy: What is status of this `FIXME`? Will it impact decomposition refactoring?

strawberryfields/ops.py
```diff
@@ -1852,31 +1839,43 @@ def _decompose(self, reg):
...
             # covariance matrix consists of x/p quadrature squeezed state
-            for n, expr in enumerate(D[:self.ns]*2/self.hbar):
-                if np.abs(expr - 1) >= _decomposition_merge_tol:
+            for n, expr in enumerate(D[:self.ns]):
+                if np.abs(expr - 1) >= _decomposition_tol:
                     r = abs(log(expr)/2)
-                    cmds.append(Command(Sgate(r, 0), reg[n]))
+                    cmds.append(Command(Squeezed(r, 0), reg[n]))
+                else:
+                    cmds.append(Command(Vac, reg[n]))
```
co9olguy: This catches the case where `r=0`, correct? What was happening before?

strawberryfields/ops.py
```diff
@@ -1852,31 +1839,43 @@ def _decompose(self, reg):
...
 
         elif self.pure and is_block_diag:
             # covariance matrix consists of rotated squeezed states
             for n, v in enumerate(BD_modes):
-                if not np.all(v - np.identity(2)*self.hbar/2 < 1e-10):
-                    r = np.abs(arccosh(np.sum(np.diag(v/self.hbar)))/2)
-                    phi = arctan(2*v[0, 1] / np.sum(np.diag(v)*[1, -1]))
-                    cmds.append(Command(Sgate(r, phi), reg[n]))
+                if not np.all(v - np.identity(2) < 1e-10):   # FIXME magic tolerance
```
co9olguy: Who wrote this code? We should ask them the purpose of this tolerance. Was it @nquesada?

strawberryfields/ops.py
```diff
@@ -305,6 +305,7 @@ def square(q):
 
 # numerical tolerances
 _decomposition_merge_tol = 1e-13
```
co9olguy: is this tolerance no longer used?

strawberryfields/utils.py
```diff
@@ -736,6 +736,7 @@ def is_channel(prog):
     Returns:
         bool: True if all operations in the program are of types :class:`strawberryfields.ops.Gate` and :class:`strawberryfields.ops.Channel`
     """
+    # FIXME isn't a preparation also a quantum channel?
```
co9olguy: I'd say no. A channel has an arbitrary input, and gives some related output, whereas a preparation doesn't have any "input" slot.

tests/frontend/test_ops_decompositions.py
```diff
@@ -285,41 +284,20 @@ def test_merge(self, hbar, tol):
         S1 = random_symplectic(n)
         S2 = random_symplectic(n)
 
-        G1 = ops.GaussianTransform(S1, hbar=hbar)
```
co9olguy: Do we have tests which test the hbar-dependence of the r parameter? (if I'm reading the comments correctly, it still depends on hbar)

tests/frontend/test_ops_decompositions.py
```diff
@@ -593,50 +525,46 @@ def test_decomposition(self, hbar, tol):
...
-        """Test that an squeeze state decomposition provides correct the covariance matrix"""
+        """Test that an axially squeezed covariance matrix decomposes into Squeezed preparations."""
         n = 3
-        prog = sf.Program(n, hbar=hbar)
+        prog = sf.Program(n)
 
         sq_r = np.array([0.453, 0.23, 0.543])
         S = np.diag(np.exp(np.concatenate([-sq_r, sq_r])))
-        cov = S @ S.T
+        cov = S @ S.T * (hbar / 2)
```
co9olguy: Interesting 🤔 
Why did hbar need to be inserted here, while it was largely dropped everywhere else?

tests/frontend/test_ops_decompositions.py
```diff
@@ -646,17 +574,36 @@ def test_rotated_squeezed_decomposition(self, hbar, tol):
...
         # calculating the resulting decomposed symplectic
         for i, cmd in enumerate(cmds):
-            # all operations should be Sgates
-            assert isinstance(cmd.op, ops.Sgate)
+            assert isinstance(cmd.op, ops.Squeezed)
             assert np.allclose(cmd.op.p[0].x, sq_r[i], atol=tol, rtol=0)
             assert np.allclose(cmd.op.p[1].x, sq_phi[i], atol=tol, rtol=0)
+
+    def test_degenerate_decomposition(self, hbar, tol):
+        """Test that a decomposition involving no squeezing results in a Vacuum preparation."""
```
co9olguy: Nice!

tests/frontend/test_utils.py
```diff
@@ -580,7 +580,6 @@ def dummy_func(r, theta, phi, q):
 class TestEngineUtilityFunctions:
     """Tests for some engine auxiliary functions"""
 
-    @pytest.mark.skip('FIXME is_unitary')
```
co9olguy: Perhaps I missed it, but I didn't see where else in the PR that changes were made which were connected to these failing tests 🤔

tests/integration/test_decompositions_integration.py
```diff
@@ -122,7 +122,7 @@ def test_graph_embed_identity(self, setup_eng, tol):
         """Test that nothing is done if the adjacency matrix is the identity"""
         prog = sf.Program(3)
         with prog.context as q:
-            ops.GraphEmbed(np.identity(6)) | q
```
co9olguy: purely for speed?

co9olguy:
@smite Nice work. Any more bypassed tests, or is the full test suite now running?

strawberryfields/ops.py
josh146: Currently, merging of decompositions assumes matrix multiplication, which is not the case with displacements (need to be added instead). This caused no issues previously because no compilation/optimization was being performed, but this issue might now arise.

strawberryfields/ops.py
josh146: Previously, it would simply run `Sgate(0)`, which was incorrect. Indicates that even though we have 100% test coverage here, our test suite was not sufficient as this was not flagged.

tests/frontend/test_ops_decompositions.py
nquesada: It's because unlike symplectic transformations, covariance matrices do depend on the value of the vacuum fluctuations (=hbar)

strawberryfields/ops.py
smite: Previously when r=0 it would do nothing, add no gates at all.

josh146: True :thinking: My comment was incorrect.

strawberryfields/ops.py
smite: There are lots of tolerance checks in the Decomposition code, and some of them might be hard to justify exactly. I just noticed that this one used a numerical value instead of _decomposition_tol.

strawberryfields/ops.py
smite: It is, but just for merges. I added _decomposition_tol for all the other checks.

strawberryfields/ops.py
josh146: Can we replace it with `_decomposition_tol` for consistency?

strawberryfields/utils.py
smite: Technically a quantum channel linearly maps valid states to valid states. A preparation maps any state to a fixed state, so you can represent it as a quantum channel  \rho \mapsto \trace(\rho) \xi. But this is a technicality, it all depends on what `is_channel` is meant to do.

tests/integration/test_decompositions_integration.py
smite: No, the Decomposition infers the number of modes it acts on from the size of the matrix. 6 was wrong, but it did not cause an error because of another bug in the code.

smite: Or more specifically the matrix size must match the number of modes it acts on.

strawberryfields/ops.py
co9olguy: That was my thought as well

tests/frontend/test_ops_decompositions.py
co9olguy: ok, in that case, then why wasn't the hbar used before?

nquesada: Probably because a particular value of hbar was assumed (likely hbar = 2)

josh146: Weird that this never broke when we changed hbar=1.7

strawberryfields/ops.py
smite: Surprisingly the merging of `ops.Gaussian` instances works, because its first parent class is `Preparation`, and `Preparation.merge()` just says that when two Preparations are merged, only the latter one matters. `Decomposition.merge()` is never called. I removed the FIXME and updated a test.

josh146: Oh, wow haha

strawberryfields/ops.py
smite: Ok

tests/frontend/test_ops_decompositions.py
smite: I added the r parameter to a few of the Gaussian tests.

tests/frontend/test_utils.py
smite: I think I fixed the function back when I updated utils.py, but I forgot to enable the tests.

tests/frontend/test_ops_decompositions.py
smite: Yeah that's the reason.

co9olguy:
Thanks @smite. Let us know when you're complete and we can do a final review/merge

PR #91, opened 2019-06-11
New_modes bugfix.

smite:

Replaces the `ops.New` singleton object with a similarly named function that creates a `New_modes` instance and appends it to the `Program`.

**Benefits:**

Previously, since there was just a single object, the `New_modes.n` instance variable would be overwritten each time `New.__call__` was called. `New_modes.n` is currently only used by `New_modes.__str__`, which is why this bug escaped notice.

This should be a drop-in replacement.

### DIFF
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -110,10 +110,9 @@ There are six kinds of :class:`Operation` objects:
         Dgate(RR(q[0])) | q[1]
         Dgate(RR(q[0], lambda q: q ** 2)) | q[2]
 
-* Meta-operations such as :class:`Delete` and :class:`New_modes` delete
-  and create modes during program execution.
-  In practice the user only deals with the pre-constructed
-  instance :py:data:`Del` and the function :func:`New`::
+* Modes can be created and deleted during program execution using the
+  function :func:`New` and the pre-constructed object :py:data:`Del`.
+  Behind the scenes they utilize the meta-operations :class:`_New_modes` and :class:`_Delete`::
 
     with prog.context as (alice,):
         Sgate(1)    | alice
@@ -241,8 +240,8 @@ Meta-operations
 
 .. autosummary::
    All
-   New_modes
-   Delete
+   _New_modes
+   _Delete
 
 
 Operations shortcuts
@@ -263,8 +262,8 @@ this is to provide shorthands for operations that accept no arguments, as well a
 
 ======================   =================================================================================
 **Shorthand variable**   **Operation**
-``New``                  :class:`~.New_modes`
-``Del``                  :class:`~.Delete`
+``New``                  :class:`~._New_modes`
+``Del``                  :class:`~._Delete`
 ``Vac``                  :class:`~.Vacuum`
 ``Fourier``              :class:`~.Fouriergate`
 ``Measure``              :class:`~.MeasureFock`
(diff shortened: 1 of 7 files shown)


strawberryfields/ops.py
```diff
@@ -110,10 +110,10 @@ def square(q):
         Dgate(RR(q[0])) | q[1]
         Dgate(RR(q[0], lambda q: q ** 2)) | q[2]
 
-* Meta-operations such as :class:`Delete` and :class:`New_modes` Operations delete
+* Meta-operations such as :class:`Delete` and :class:`New_modes` delete
   and create modes during program execution.
   In practice the user only deals with the pre-constructed
-  instances :py:data:`Del` and :py:data:`New`::
+  instance :py:data:`Del` and the function :func:`New`::
```
josh146: Since `Del` and `New` are user facing, perhaps they should come first in this sentence, and the meta operations mentioned after?

strawberryfields/ops.py
```diff
@@ -1525,33 +1525,40 @@ def __str__(self):
...
+    # create RegRefs for the new modes
+    refs = Program._current_context._add_subsystems(n)
+    # append the actual Operation to the Program
+    Program._current_context.append(New_modes(n), refs)
+    return refs
+
 
-    def __call__(self, n=1):
-        """Adds one or more new modes to the system in a deferred way.
+class New_modes(MetaOperation):
```
josh146: Since these are not user facing, does it make sense to use underscores so that they are 'private' i.e., `_NewModes`

co9olguy:
@smite can you clarify what you mean by "there was just a single object"?

smite:
> @smite can you clarify what you mean by "there was just a single object"?

Before `ops.New` was a pre-constructed instance of the class `ops.New_modes`. It was callable via its `__call__` method. You always used the same object, so when it stored the number of modes created in `self.n`, the previous value would always be overwritten. Now the `ops.New()` function constructs a new `ops.New_modes` instance each time it's called.

co9olguy:
@smite got it, make sense

strawberryfields/ops.py
smite: Fixed.

strawberryfields/ops.py
josh146: (I could imagine a use case where the user is in IPython or Jupyter notebook, and types `ops.N`, and `New_modes` comes up as an option)

smite: Maybe? The only reason not to use underscores that I can think of is that they ever so slightly make the names longer and clutter up the program listing.

smite: I'll add the underscores, it's probably safer that way.

co9olguy:
Ready for merging from my perspective. @josh146 any last issues?

josh146:
Looks good from my end. Weird that code factor doesn't seem to be working.

tests/frontend/test_ops_metaoperation.py
```diff
@@ -81,17 +81,17 @@ def test_dispatch_two_mode_gates(self, gate):
...
+        """_New_modes must not be called via its __or__ method"""
         with pytest.raises(ValueError):
-            ops.New.__or__(0)
+            ops._New_modes(1).__or__(0)
 
     def test_create_non_positive_integer(self, prog):
         """number of new modes must be a positive integer"""
         with pytest.raises(ValueError):
-            ops.New.__call__(-2)
+            ops.New(-2)
```
josh146: much cleaner and easier to read!

PR #92, opened 2019-06-18
Adds a Program context check to ops.New(). Updates related tests.

smite:

Now `ops.New()` raises an exception if it is called outside a Program context, instead of just failing with a cryptic error message.

Also improved some related tests.

### DIFF
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1536,6 +1536,8 @@ def New(n=1):
     Returns:
         tuple[RegRef]: tuple of the newly added subsystem references
     """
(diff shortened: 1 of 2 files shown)


strawberryfields/ops.py
```diff
@@ -1536,6 +1536,8 @@ def New(n=1):
     Returns:
         tuple[RegRef]: tuple of the newly added subsystem references
     """
+    if Program._current_context is None:
+        raise RuntimeError('New() can only be called inside a Program context.')
```
josh146: Nice!

PR #93, opened 2019-06-18
APIClient library

zzzeid:

Adds a communications client library that manages API requests and responses, as well as abstracts jobs, job results, and job circuits into resource objects.


strawberryfields/api_client.py
```diff
@@ -0,0 +1,101 @@
+from urllib.parse import urljoin
```
josh146: Don't forget to add the copyright notice to the top of this file :slightly_smiling_face:

strawberryfields/api_client.py
```diff
@@ -0,0 +1,101 @@
...
+import requests
+
+
+class APIClient:
```
josh146: Even though this is a scaffolding PR, docstrings should be included in the classes/methods/functions. Even if it's still vague, it would be useful for the engine to know the general use-case of each class/method/function. The details of the docstring can then be filled in on subsequent PRs.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,101 @@
...
+            self.HEADERS.update(kwargs['headers'])
+
+        if 'base_url' in kwargs:
+            base_url = kwargs['base_url']
+            if base_url in self.ALLOWED_BASE_URLS:
+                self.BASE_URL = base_url
+            else:
+                raise ValueError('base_url parameter not in allowed list')
+        else:
+            self.BASE_URL = self.DEFAULT_BASE_URL
```
josh146: here, you could use
```python
self.BASE_URL = kwargs.get('base_url', self.DEFAULT_BASE_URL)

if self.base_url not in self.ALLOWED_BASE_URLS:
    raise ValueError('base_url parameter not in allowed list')
```

to avoid the outer if statement. But I guess we'll be moving the base URL to the configuration file, so this is not too important to change now.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,101 @@
...
+    def get(self, path):
+        return requests.get(
+            url=self.join_path(path), headers=self.headers)
+
+    def post(self, path, payload):
+        return requests.post(
+            url=self.join_path(path), headers=self.headers, data=payload)
+
+
+class Job:
```
josh146: here it would be helpful to have docstrings as well

strawberryfields/api_client.py
```diff
@@ -0,0 +1,101 @@
...
+        if client is None:
+            client = APIClient()
+
+        self.client = client
+
+        if id is not None:
+            self.get(id)
+
+    def join_path(self, path):
+        return urljoin(self.RESOURCE_PATH, path)
```
josh146: Since this is duplicated between both classes, does it make more sense to just have this as a utility function?

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,80 @@
...
+        assert True
+
+    def test_update_job(self):
+        assert True
+
+    def test_create(self):
+        assert True
+
+    def test_join_path(self):
+        assert True
```
josh146: More of a question than anything, but does it make more sense here to not have tests, rather than `assert True`? This way the (lack of) coverage is a reminder that the tests will need to be filled in

strawberryfields/api_client.py
zzzeid: Yep will add to all the files.

strawberryfields/api_client.py
zzzeid: Noted.

tests/api_client/test_api_client.py
zzzeid: These are just place holders (could in theory contain `pass` rather than `assert True`) - I guess I use that to help me visualize the minimum number of tests I will likely have to write, but checking coverage also makes sense in lieu of placeholders. I wanted to push a commit right away after our discussion, but I will clean this up.

strawberryfields/api_client.py
zzzeid: Makes sense, I originally used `kwargs.get` but switched to this as part of a code change and didn't refactor after changing the logic.

strawberryfields/api_client.py
zzzeid: It is not duplicated exactly, as each method joins a different base string, but I will think of how to refactor this.,

zzzeid: I did a small refactor here to simplify it a little further.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,459 @@
...
+        Handles a successful response by refreshing the instance fields.
+        """
+        self.refresh_data(response.json())
+
+    def handle_error_response(self, response):
+        """
+        Handles an error response that is returned by the server.
+        """
+
+        if response.status_code == 400:
```
zzzeid: Oops, forgot to add the user warning here: will add.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,544 @@
...
+
+.. currentmodule:: strawberryfields.api_client
+
+
+This module provides a thin client that communicates with the compute-service API over the HTTP
+protocol, based on the requests module. It also provides helper classes to facilitate interacting
+with this API via the Resource subclasses, as well as the ResourceManager wrapper around APIClient
+that is available for each resource.
+
+A single APIClient instance can be used throughout one's session in the application.
```
jswinarton: Based on my understanding it looks like a default client is created in the `ResourceManager`, and then the user can configure the client using environment variables or a file. If they want more custom functionality they can create their own `APIClient` instance and pass it down to each resource. Is this the idea? It might be good to have some examples of how the user is intended to configure this in the module docs.

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,309 @@
...
+from strawberryfields import api_client
+from strawberryfields.api_client import (
+    requests,
+    Job,
+    ResourceManager,
+    ObjectAlreadyCreatedException,
+    MethodNotSupportedException,
+)
+
+from unittest.mock import MagicMock
```
jswinarton: Aren't `MagicMock` and `monkeypatch` two tools for doing the same thing? How come we're using both?

josh146: Actually had that question as well

zzzeid: monkeypatch allows you to patch an attribute on an object (including one that you imported, for example.) MagicMock is itself a mock object that you can patch however way you want. I'm using Pytest's monkeypatch here because it's easy to use to patch one or a few attributes of an object, whereas MagicMock is used as mock objects (rather than creating mock objects from scratch). Thoughts?

strawberryfields/api_client.py
zzzeid: I will add more information about this in the docs. In theory, this library is leaving it to the user to decide how they want to initialize the client, for whatever reasons they may have. There are safe fallback defaults if the user chooses not to do anything.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,475 @@
...
+
+    pass
+
+
+class APIClient:
+    """
+    An object that allows the user to connect to the compute-service API.
+    """
+
+    ALLOWED_HOSTNAMES = ["localhost"]
```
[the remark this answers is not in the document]
zzzeid: Yes -- here will go the actual list of allowed hostnames, which currently does not exist I think.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def get(self, path):
+        """
+        Sends a GET request to the provided path. Returns a response object.
+        """
+        try:
+            response = requests.get(url=self.join_path(path), headers=self.HEADERS)
+        except requests.exceptions.ConnectionError as e:
+            response = None
+            warnings.warn(f"Could not connect to server ({e})")
```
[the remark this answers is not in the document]
zzzeid: Since this isn't a core failure of the library (and rather, potentially a temporary error), I figured it might make sense to only show a warning and not a full blown exception. In theory I could do nothing here and let the exception surface. I have a Trello card for improving messaging and error handling altogether, in which I will include a better implementation of this, as well as all other client-server errors.

zzzeid:
> Looks good overall, a few comments. Also, will the `TODO`s be resolved before merge?

@jswinarton @josh146 The ones that do not get resolved will have Trello cards to be implemented in the future. Some of them will be resolved with the configuration PR. Do we generally prefer to not keep TODOs in the code base?

josh146:
In the case of SF and PennyLane, todos are generally encouraged in order to stop the scope of PRs becoming too large. They should be converted into trello and/or github issues where appropriate though

tests/api_client/test_api_client.py
josh146: Makes sense!

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def __init__(self, resource, client=None):
+        """
+        Initialize the manager with resource and client instances. A client
+        instance is used as a persistent HTTP communications object, and a
+        resource instance corresponds to a particular type of resource (e.g.
+        Job)
+        """
+        setattr(self, "resource", resource)
+        setattr(self, "client", client or APIClient())
```
josh146: How come you're using `setattr` here, as opposed to the following?
```python
self.resource = resource
```

zzzeid: I think this is a remnant of having had keys in a tuple earlier. There is no reason to keep them like this at this point and I prefer just straight assignment. I will include this change in a clean up commit.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+        with all API requests.
+
+        Args:
+            authentication_token (str): An authentication token used to access the API.
+        """
+        self.headers["Authorization"] = authentication_token
+
+    def join_path(self, path):
+        """
+        Joins a base url with an additional path (e.g. a resource name and ID)
```
co9olguy:
```suggestion
        Joins a base url with an additional path (e.g., a resource name and ID)
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+    This class handles all interactions with APIClient by the resource.
+    """
+
+    http_status_code = None
+
+    def __init__(self, resource, client=None):
+        """
+        Initialize the manager with resource and client instances. A client
+        instance is used as a persistent HTTP communications object, and a
+        resource instance corresponds to a particular type of resource (e.g.
```
co9olguy:
```suggestion
        resource instance corresponds to a particular type of resource (e.g.,
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+        Job)
+        """
+        setattr(self, "resource", resource)
+        setattr(self, "client", client or APIClient())
+
+    def join_path(self, path):
+        """
+        Joins a resource base path with an additional path (e.g. an ID)
```
co9olguy:
```suggestion
        Joins a resource base path with an additional path (e.g., an ID)
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+        return join_path(self.resource.PATH, path)
+
+    def get(self, job_id):
+        """
+        Attempts to retrieve a particular record by sending a GET
+        request to the appropriate endpoint. If successful, the resource
+        object is populated with the data in the response.
+
+        Args:
+            job_id (int): The ID of an object to be retrieved.
```
co9olguy: Does it make sense to have the name be somtething more general than `job_id`? elsewhere in the docstrings, you are more generic ("object", "record", etc.)

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+                "The server did not accept the request due to an authentication error "
+                "({}: {}).".format(response.status_code, response.text),
+                UserWarning,
+            )
+        elif response.status_code in (500, 503, 504):
+            warnings.warn(
+                "The client encountered an unexpected temporary server error "
+                "({}: {}).".format(response.status_code, response.text),
+                UserWarning,
+            )
```
co9olguy: what about other errors, e.g., 404?

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+    resource endpoint.
+    """
+
+    SUPPORTED_METHODS = ()
+    PATH = ""
+    fields = ()
+
+    def __init__(self, client=None):
+        """
+        Initialize the Resource by populating attributes based on fields and settings a manager.
```
co9olguy:
```suggestion
        Initialize the Resource by populating attributes based on fields and setting a manager.
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+
+        Args:
+            client (APIClient): An APIClient instance to use as a client.
+        """
+        self.manager = ResourceManager(self, client=client)
+        for field in self.fields:
+            setattr(self, field.name, field)
```
co9olguy: I suppose `setattr` is needed here, unlike above

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+    """
+
+    value = None
+
+    def __init__(self, name, clean=str):
+        """
+        Initialize the Field object with a name and a cleaning function.
+
+        Args:
+            name (str): A string representing the name of the field (e.g. "created_at").
```
co9olguy:
```suggestion
            name (str): A string representing the name of the field (e.g., "created_at").
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+        """
+        Return the string representation of the value.
+        """
+        return str(self.value)
+
+    def __bool__(self):
+        """
+        Use the value to determine boolean state.
+        """
+        return self.value is not None
```
co9olguy: I guess this is determining if there is a non-`None` state. 
So you could have a Field F whose value is `False`, but `__bool__` would still return `True`, correct?

strawberryfields/api_client.py
```diff
@@ -0,0 +1,546 @@
...
+    def __init__(self, job_id, client=None):
+        """
+        Initialize the JobResult resource with a pre-defined field.
+
+        Args:
+            job_id (int): The ID of the Job object corresponding to the JobResult object.
+        """
+        self.fields = (Field("result", json.loads),)
+
+        self.PATH = self.PATH.format(job_id=job_id)
```
co9olguy: nice!

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,309 @@
...
+
+class MockGETResponse(MockResponse):
+    possible_responses = {
+        200: SAMPLE_JOB_RESPONSE,
+        401: {
+            "code": "unauthenticated",
+            "detail": "Requires authentication"
+        },
+        404: {
+            "code": "",
```
co9olguy: why blank entries here?

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,309 @@
...
+        assert client.BASE_URL == 'http://localhost'
+        assert client.HEADERS == {}
+
+    def test_init_custom_token_client(self):
+        test_token = 'TEST'
+        client = api_client.APIClient(authentication_token=test_token)
+        assert client.AUTHENTICATION_TOKEN == test_token
+
+    def test_load_configuration(self, client):
+        with pytest.raises(NotImplementedError):
```
co9olguy: nice 👍

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,309 @@
...
+
+        mock_resource.SUPPORTED_METHODS = ('GET',)
+
+        manager = ResourceManager(mock_resource, mock_client)
+        monkeypatch.setattr(manager, "handle_response", MagicMock())
+
+        manager.get(1)
+
+        # TODO test that this is called with correct path
+        mock_client.get.assert_called_once()
```
co9olguy: You use this pattern a lot in the tests, but I'm not too familiar with it. Could you explain what the tests are checking with `assert_called_once()` and `assert_called_once_with`. 🤔

strawberryfields/api_client.py
zzzeid: Makes perfect sense, I forgot to rename this when I moved things to ResourceManager.

strawberryfields/api_client.py
zzzeid: I am adding 404 error handling, as well as a catchall. This section of the code will be refactored in the future as well for better error handling, it's not really doing anything right now.

strawberryfields/api_client.py
zzzeid: Yes that is correct. I probably should use a dirty flag here, or refactor the code to have a `Field.is_set` boolean value.

tests/api_client/test_api_client.py
zzzeid: Forgot to put data in here, will update.

tests/api_client/test_api_client.py
zzzeid: I should be using `assert_called_once_with` here as well, but this test is simplified for now. In this and other tests, I'm ensuring that a method is called, but I don't necessarily need that method to be called or test any further than that, as other tests would cover the rest. In this case, `get` calls `client.get` and `manager.handle_response`, and all I need to ensure is that they are being called with the correct parameters. Thoughts?

co9olguy: makes sense. I suppose richer functionality will be tested in future tests

requirements.txt
```diff
@@ -4,3 +4,4 @@ scipy>=1.0.0
 tensorflow==1.3
 tensorflow-tensorboard>=0.1.8
 quantum-blackbird
+python-dateutil==2.8.0
```
josh146: Don't forget to add this to `setup.py` as well! (without the version number pinning)

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,309 @@
...
+    def json(self):
+        return self.possible_responses[self.status_code]
+
+    def raise_for_status(self):
+        raise requests.exceptions.HTTPError()
+
+
+@pytest.mark.api_client
+class TestAPIClient:
+    def test_init_default_client(self):
```
josh146: Before merging, make sure to add a short 1-2 line docstring to each test function, just describing (1) what is being tested, and (2) the expected result. This shows up in pytest tracebacks so is useful for debugging.

tests/api_client/test_api_client.py
```diff
@@ -0,0 +1,309 @@
...
+            client.load_configuration_from_file()
+
+    def test_authenticate(self, client):
+        with pytest.raises(NotImplementedError):
+            username = 'TEST_USER'
+            password = 'TEST_PASSWORD'
+            client.authenticate(username, password)
+
+    def test_join_path(self, client):
+        assert client.join_path('jobs') == f'{client.BASE_URL}/jobs'
```
josh146: Don't forget to remove all f-strings for now (we can evaluate later whether to shift the minimum supported Python version to 3.6)

zzzeid: Done.

tests/api_client/test_api_client.py
zzzeid: Done.

requirements.txt
zzzeid: Done. I used "greater than or equal to" for the version spec.

zzzeid:
Proposed changes for usability -- I think the below makes a little more sense than the current implementation of the library API, based on having used this in the python shell and noticed some clunkyness:

(1) **ResourceManagers** should be attributes of the client, not the Resource. For example:
```
client = APIClient()
client.jobs  # This is a ResourceManager object for Job
job = client.jobs.get(id=123)  # get here returns a Job instance populated with Job with ID #123
job.reload()  # This reloads the job

job = client.jobs.create(circuit="...")  # create here also returns a Job instance.
```

(2) **ResourceManager** should keep track of requests and responses for debug purposes, and include a reference to the request and response objects. This could be enabled/disabled with a debug flag.

(3) **Field** attributes on a Resource should map to Field.value, instead of the Field instance. For example:
```
...
job = client.jobs.get(id=123)
job.status  # Should return job.fields['status'].value, currently returns job.fields['status']
```

(4) **Resource** should have a "writable_fields" attribute that specifies which fields can be included/passed in the POST request.

Thoughts? @josh146 @jswinarton @co9olguy

default_config.toml
```diff
@@ -5,7 +5,7 @@
 [api]
 # Options for the Strawberry Fields Cloud API
 # Fill in your authentication
-authentatication_token = None # example token form: 071cdcce-9241-4965-93af-4a4dbc739135
+authentication_token = None # example token form: 071cdcce-9241-4965-93af-4a4dbc739135
```
josh146: Thanks for catching this!

strawberryfields/_dev/configuration.py
```diff
@@ -54,14 +54,21 @@
...
 
-todo
+SF_API_USE_SSL:
+    Whether to use SSL or not when connecting to the API. True or False.
+SF_API_HOSTNAME:
+    The hostname of the server to connect to. Defaults to localhost. Must be one of the allowed
+    hosts.
+SF_API_AUTHENTICATION_TOKEN:
+    The authentication token to use when connecting to the API. Will be sent with every request in
+    the header.
```
josh146: Perhaps it's worth modifying this to specify the configuration file subsection (`[api]`) and lower case config keys (`use_ssl`, `hostname`, `authentication_token`), as well as the environment variables.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+            Zgate(sqrt(2)*q1) | 2
+            MeasureHeterodyne() | 2
+        '''
+        job.manager.create(circuit=circuit)
+
+        job.id  # Returns the job ID that was generated by the server
+        job.reload()  # Fetches the latest job data from the server
+        job.status  # Prints the status of this job
+        job.result  # Returns a JobResult object
+        job.circuit  # Returns a JobCircuit object
```
josh146: API looks awesome!

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+
+.. currentmodule:: strawberryfields.api_client
+
+
+This module provides a thin client that communicates with the Xanadu Platform API over the HTTP
+protocol, based on the requests module. It also provides helper classes to facilitate interacting
+with this API via the Resource subclasses, as well as the ResourceManager wrapper around APIClient
+that is available for each resource.
+
+A single APIClient instance can be used throughout one's session in the application. The application
```
josh146:
```suggestion
A single :class:`~.APIClient` instance can be used throughout one's session in the application. The application
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+will attempt to configure the APIClient instance using a configuration file or defaults, but the
```
josh146:
```suggestion
will attempt to configure the :class:`~.APIClient` instance using a configuration file or defaults, but the
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+user can choose to override various parameters of the APIClient manually.
```
josh146:
```suggestion
user can choose to override various parameters of the :class:`~.APIClient` manually.
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+    ALLOWED_HOSTNAMES = ["localhost", "localhost:8080", "platform.strawberryfields.ai"]
+
+    DEFAULT_HOSTNAME = "localhost"
+
+    ENV_KEY_PREFIX = "SF_API_"
+    ENV_AUTHENTICATION_TOKEN_KEY = "{}AUTHENTICATION_TOKEN".format(ENV_KEY_PREFIX)
+    ENV_API_HOSTNAME_KEY = "{}API_HOSTNAME".format(ENV_KEY_PREFIX)
+    ENV_USE_SSL_KEY = "{}USE_SSL".format(ENV_KEY_PREFIX)
+
+    DEFAULT_CONFIG = {"use_ssl": True, "hostname": DEFAULT_HOSTNAME, "authentication_token": None}
```
josh146: Now that the `ENV_KEY_PREFIX` is defined in the `Configuration` class, and the default config defined in `configuration.DEFAULT_CONFIG`, I'm wondering if it makes sense to simply let the `Configuration()` class properly determine the defaults to avoid code duplication.

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+
+    def __init__(self, **kwargs):
+        """
+        Initialize the API client with various parameters.
+        """
+        # TODO: Load username, password, or authentication token from
+        # configuration file
```
josh146: Can remove this now

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        return configuration.Configuration().api
+
+    def authenticate(self, username, password):
+        """
+        Retrieve an authentication token from the server via username
+        and password authentication and calls set_authorization_header.
+
+        Args:
+            username (str): A user name.
```
josh146:
```suggestion
            username (str): a user name
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+            password (str): A password.
```
josh146:
```suggestion
            password (str): password
```

josh146: Although these arg docstrings might be a bit redundant for now 🤔

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        raise NotImplementedError()
+
+    def set_authorization_header(self, authentication_token):
+        """
+        Adds the authorization header to the headers dictionary to be included
+        with all API requests.
+
+        Args:
+            authentication_token (str): An authentication token used to access the API.
```
josh146:
```suggestion
            authentication_token (str): an authentication token used to access the API
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        self.HEADERS["Authorization"] = authentication_token
+
+    def join_path(self, path):
+        """
+        Joins a base url with an additional path (e.g., a resource name and ID)
```
josh146:
```suggestion
        Joins a base url with an additional path (e.g., a resource name and ID).
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+
+        Args:
+            path (str): A path to be joined with BASE_URL.
```
josh146:
```suggestion
            path (str): A path to be joined with ``BASE_URL``
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+
+        Returns:
+            str: A joined path.
```
josh146:
```suggestion
            str: resulting joined path
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        return join_path(self.BASE_URL, path)
+
+    def get(self, path):
+        """
+        Sends a GET request to the provided path. Returns a response object.
+
+        Args:
+            path (str): A path to send the GET request to.
```
josh146:
```suggestion
            path (str): path to send the GET request to
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+
+        Returns:
+            requests.Response: A response object, or None if no response could be fetched from the
+            server.
+        """
+        try:
+            response = requests.get(url=self.join_path(path), headers=self.HEADERS)
+        except requests.exceptions.ConnectionError as e:
+            response = None
+            warnings.warn("Could not connect to server ({})".format(e))
```
josh146: Rather than a warning, should this be a `ConnectionError`?

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        return response
+
+    def post(self, path, payload):
+        """
+        Converts payload to a JSON string. Sends a POST request to the provided
+        path. Returns a response object.
+
+        Args:
+            path (str): A path to send the GET request to.
```
josh146:
```suggestion
            path (str): path to send the GET request to
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+            payload: A JSON serializable object to be sent to the server.
```
josh146:
```suggestion
            payload: JSON serializable object to be sent to the server
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        return join_path(self.resource.PATH, path)
+
+    def get(self, resource_id=None):
+        """
+        Attempts to retrieve a particular record by sending a GET
+        request to the appropriate endpoint. If successful, the resource
+        object is populated with the data in the response.
+
+        Args:
+            resource_id (int): The ID of an object to be retrieved.
```
josh146:
```suggestion
            resource_id (int): the ID of an object to be retrieved
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+            response = self.client.get(self.resource.PATH)
+        self.handle_response(response)
+
+    def create(self, **params):
+        """
+        Attempts to create a new instance of a resource by sending a POST
+        request to the appropriate endpoint.
+
+        Args:
+            **params: Arbitrary parameters to be passed on to the POST request.
```
josh146:
```suggestion
            **params: arbitrary parameters to be passed on to the POST request
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+
+        self.handle_response(response)
+
+    def handle_response(self, response):
+        """
+        Store the status code on the manager object and handle the response
+        based on the status code.
+
+        Args:
+            response (requests.Response): A response object to be parsed.
```
josh146:
```suggestion
            response (requests.Response): a response object to be parsed
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        Placeholder method to handle an unsuccessful request (e.g. due to no network connection).
+        """
+        warnings.warn("Your request could not be completed")
+
+    def handle_success_response(self, response):
+        """
+        Handles a successful response by refreshing the instance fields.
+
+        Args:
+            response (requests.Response): A response object to be parsed.
```
josh146:
```suggestion
            response (requests.Response): a response object to be parsed
```

strawberryfields/api_client.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        self.refresh_data(response.json())
+
+    def handle_error_response(self, response):
+        """
+        Handles an error response that is returned by the server.
+
+        Args:
+            response (requests.Response): A response object to be parsed.
```
josh146:
```suggestion
            response (requests.Response): a response object to be parsed
```

strawberryfields/api_client.py
zzzeid: There's a slight fix to this doc that I will push shortly. There are also a few changes to the API that I'd like to make in a follow up PR to simplify the API a little.

strawberryfields/api_client.py
zzzeid: @josh146 this has come up before -- upon more thinking, I think it does make sense to leave this as an exception and let the user decide what to do with it. Will change. However, there is a separate ticket for better handling and messaging of errors.

strawberryfields/api_client.py
zzzeid: Makes perfect sense.

strawberryfields/_dev/configuration.py
zzzeid: @josh146 are you referring to something other than the above (lines 55-59)

strawberryfields/api_client.py
zzzeid: Done.

strawberryfields/api_client.py
zzzeid: Done

strawberryfields/_dev/configuration.py
josh146: Oh, I meant more so that there are two ways the user can specify the options. Either as environment variables (`SF_API_AUTHENTICATION_TOKEN`) or in the config file, where it looks slightly different (no `SF_` prefix, `API` is the config section, and everything is lowercase):

```toml
[api]
authentication_token = 450984390845fgdsljkfsdg98324
```

So I was wondering if there was a way of including both environment variable key and config key in the glossary, maybe like

```rest
``SF_API_AUTHENTICATION_TOKEN`` (env variable) or ``authentication_token`` (file):
```
^ this looks a bit unwieldy, there is probably a better way of writing this.

zzzeid: We could add a blurb explaining the conventions, e.g.:

```
...
The names of the environment variables are uppercased keys in the configuration file prefixed with
``SF_API_``. 

SF_API_USE_SSL:
    Whether to use SSL or not when connecting to the API. True or False.
...
```
Thoughts?

PR #94, opened 2019-06-20
Configuration file for Strawberry Fields

josh146:
**Description of the Change:** Adds a basic configuration module to SF.

**Benefits:**

* Stores default configuration

* Platform-independent search of configuration file locations

* Environment variables have the same name, prefixed with `SF_{SECTION}_`. If they exist, they take precedence over a configuration file.

**Possible Drawbacks:** n/a

**Related GitHub Issues:**

### DIFF
--- a/default_config.toml
+++ b/default_config.toml
@@ -1,0 +1,12 @@
+## =================================================================
+##      STRAWBERRY FIELDS API CONFIGURATION OPTIONS
+## =================================================================
+
+[api]
+# Options for the Strawberry Fields Cloud API
+# Fill in your authentication
+authentatication_token = "071cdcce-9241-4965-93af-4a4dbc739135"
+# Fill in the hostname of the Cloud API
+hostname = "localhost"
+# Whether Strawberry Fields should use SSL to connect to the API
+use_ssl = true
--- a/default_config.toml
+++ b/default_config.toml
@@ -5,7 +5,7 @@
 [api]
 # Options for the Strawberry Fields Cloud API
 # Fill in your authentication
-authentatication_token = "071cdcce-9241-4965-93af-4a4dbc739135"
+authentatication_token = None # example token form: 071cdcce-9241-4965-93af-4a4dbc739135
 # Fill in the hostname of the Cloud API
 hostname = "localhost"
 # Whether Strawberry Fields should use SSL to connect to the API
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,5 @@ scipy>=1.0.0
 tensorflow==1.3
 tensorflow-tensorboard>=0.1.8
 quantum-blackbird
+toml
+user_config_dir
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ tensorflow==1.3
 tensorflow-tensorboard>=0.1.8
 quantum-blackbird
 toml
-user_config_dir
+appdirs
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -174,7 +174,7 @@ class Configuration:
         """Load a configuration file.
 
         Args:
-            filepath (str): path to the configuration file.
+            filepath (str): path to the configuration file
         """
         with open(filepath, "r") as f:
             self._config_file = toml.load(f)
@@ -183,7 +183,7 @@ class Configuration:
         """Save a configuration file.
 
         Args:
-            filepath (str): path to the configuration file.
+            filepath (str): path to the configuration file
         """
         with open(filepath, "w") as f:
             toml.dump(self._config, f)
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -23,7 +23,7 @@ This module contains the :class:`Configuration` class, which is used to
 load, store, save, and modify configuration options for Strawberry Fields.
 
 Behaviour
----------
+--------
 
 On first import, Strawberry Fields attempts to load the configuration file `config.toml`, by
 scanning the following three directories in order of preference:
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -52,7 +52,7 @@ and has the following format:
 
 .. code-block:: toml
 
-    [API]
+    [api]
     # Options for the Strawberry Fields Cloud API
     authentatication_token = "071cdcce-9241-4965-93af-4a4dbc739135"
     hostname = "localhost"
@@ -113,7 +113,7 @@ class Configuration:
     def __repr__(self):
         return "Strawberry Fields Configuration <{}>".format(self._filepath)
 
-    def __init__(self, name):
+    def __init__(self, name="config.toml"):
         # Look for an existing configuration file
         self._config = DEFAULT_CONFIG
         self._config_file = {}
@@ -142,10 +142,10 @@ class Configuration:
 
         The environment variable takes precedence."""
         for section, section_config in self._config.items():
-            env_prefix = "SF_" + section.upper()
+            env_prefix = "SF_{}_".format(section.upper())
 
             for key, value in section_config.items():
-                # environment variables take precendence
+                # environment variables take precedence
                 env = env_prefix + key.upper()
                 if env in os.environ:
                     self._config[section][key] = os.environ[env]
@@ -160,7 +160,7 @@ class Configuration:
         if key in self._config:
             return self._config[key]
 
-        raise ConfigurationError("Unkown Strawberry Fields configuration section.")
+        raise ConfigurationError("Unknown Strawberry Fields configuration section.")
 
     @property
     def path(self):
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -126,16 +126,16 @@ class Configuration:
         # variable SF_CONF, and default user config directory, in that order.
         directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
         for idx, directory in enumerate(directories):
+            self._filepath = os.path.join(directory, self._name)
             try:
-                self._filepath = os.path.join(directory, self._name)
                 self.load(self._filepath)
-                break
             except FileNotFoundError:
                 # check if we have searched all directories
                 if idx == len(directories) - 1:
                     log.info("No Strawberry Fields configuration file found.")
-
-        self.update_config()
+            else:
+                self.update_config()
+                break
 
     def update_config(self):
         """Updates the configuration from either a loaded configuration
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -144,7 +144,7 @@ class Configuration:
         for section, section_config in self._config.items():
             env_prefix = "SF_{}_".format(section.upper())
 
-            for key, value in section_config.items():
+            for key, _ in section_config.items():
                 # environment variables take precedence
                 env = env_prefix + key.upper()
                 if env in os.environ:
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -142,11 +142,11 @@ class Configuration:
 
         The environment variable takes precedence."""
         for section, section_config in self._config.items():
-            env_prefix = "SF_" + section
+            env_prefix = "SF_" + section.upper()
 
             for key, value in section_config.items():
                 # environment variables take precendence
-                env = env_prefix + key
+                env = env_prefix + key.upper()
                 if env in os.environ:
                     self._config[section][key] = os.environ[env]
                     continue
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -127,15 +127,12 @@ class Configuration:
         directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
         for directory in directories:
             self._filepath = os.path.join(directory, self._name)
-            try:
-                config = self.load(self._filepath)
+            config = self.load(self._filepath)
+            if config:
+                self.update_config()
                 break
-            except FileNotFoundError:
-                pass
 
-        if config:
-            self.update_config()
-        else:
+        if config is None:
             log.info("No Strawberry Fields configuration file found.")
 
     def update_config(self):
@@ -177,9 +174,12 @@ class Configuration:
         Args:
             filepath (str): path to the configuration file
         """
-        with open(filepath, "r") as f:
-            self._config_file = toml.load(f)
-        return self._config_file
+        try:
+            with open(filepath, "r") as f:
+                self._config_file = toml.load(f)
+            return self._config_file
+        except FileNotFoundError:
+            return None
 
     def save(self, filepath):
         """Save a configuration file.
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -122,7 +122,7 @@ class Configuration:
         self._user_config_dir = user_config_dir("strawberryfields", "Xanadu")
         self._env_config_dir = os.environ.get("SF_CONF", "")
 
-        # search the current directory, the directory under environment
+        # Search the current directory, the directory under environment
         # variable SF_CONF, and default user config directory, in that order.
         directories = [os.curdir, self._env_config_dir, self._user_config_dir, ""]
         for idx, directory in enumerate(directories):
@@ -131,6 +131,7 @@ class Configuration:
                 self.load(self._filepath)
                 break
             except FileNotFoundError:
+                # check if we have searched all directories
                 if idx == len(directories) - 1:
                     log.info("No Strawberry Fields configuration file found.")
 
@@ -144,21 +145,20 @@ class Configuration:
         for section, section_config in self._config.items():
             env_prefix = "SF_{}_".format(section.upper())
 
-            for key, _ in section_config.items():
-                # environment variables take precedence
+            for key in section_config:
+                # Environment variables take precedence
                 env = env_prefix + key.upper()
-                if env in os.environ:
-                    self._config[section][key] = os.environ[env]
-                    continue
 
-                # check if a configuration file exists
-                if self._config_file:
-                    # update from configuration file
+                if env in os.environ:
+                    # Update from environment variable
+                    self._config[section][key] = os.environ[env]
+                elif self._config_file:
+                    # Update from configuration file
                     self._config[section][key] = self._config_file[section][key]
 
-    def __getattr__(self, key):
-        if key in self._config:
-            return self._config[key]
+    def __getattr__(self, section):
+        if section in self._config:
+            return self._config[section]
 
         raise ConfigurationError("Unknown Strawberry Fields configuration section.")
 
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -124,7 +124,7 @@ class Configuration:
 
         # Search the current directory, the directory under environment
         # variable SF_CONF, and default user config directory, in that order.
-        directories = [os.curdir, self._env_config_dir, self._user_config_dir, ""]
+        directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
         for idx, directory in enumerate(directories):
             try:
                 self._filepath = os.path.join(directory, self._name)
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -1,0 +1,189 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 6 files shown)


strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+    and plugin/device configurations.
+
+    Args:
+        name (str): filename of the configuration file.
+        This should be a valid TOML file. You may also pass an absolute
+        or a relative file path to the configuration file.
+    """
+
+    def __str__(self):
+        return "{}".format(self._config)
```
zzzeid: I've been using f-strings lately and find them more readable -- e.g. `return f'{self._config}'`.

josh146: Unfortunately we still have SF targeting Python 3.5, which means no f strings for now (as nice as they are)

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+load, store, save, and modify configuration options for Strawberry Fields.
+
+Behaviour
+---------
+
+On first import, Strawberry Fields attempts to load the configuration file `config.toml`, by
+scanning the following three directories in order of preference:
+
+1. The current directory
+2. The path stored in the environment variable ``SF_CONF``
```
trbromley: Where is ``SF_CONF`` defined?

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+loading the device.
+
+The user can access the initialized configuration via `strawberryfields.config`, view the
+loaded configuration filepath, print the configurations options, access and modify
+them via keys, and save/load new configuration files.
+
+Configuration files
+-------------------
+
+The configuration file `config.toml` uses the `TOML standard <https://github.com/toml-lang/toml>`_,
```
trbromley: I never got to the bottom of single or double underscores for links - once you gave a suggestion to change from single to double. Is the single underscore here ok?

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+and has the following format:
```
trbromley: Could mention ``default_config.toml`` as an example?

trbromley: Also how does the `default_config.toml` file differ from the dict of dicts `DEFAULT_CONFIG` you use below?

tests/frontend/test_about.py
```diff
@@ -14,9 +14,14 @@
 """
 Unit tests for top-level Strawberry Fields functions.
 """
+import pytest
```
trbromley: Why did these changes effect this test?

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+=============
+
+**Module name:** :mod:`strawberryfields.configuration`
+
+.. currentmodule:: strawberryfields.configuration
+
+This module contains the :class:`Configuration` class, which is used to
+load, store, save, and modify configuration options for Strawberry Fields.
+
+Behaviour
```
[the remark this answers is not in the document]
josh146: @co9olguy, how is this spelt in Canadian?

strawberryfields/configuration.py
josh146: it's an environment variable, i.e.,
```bash
export SF_CONF=/some/path
```

strawberryfields/configuration.py
josh146: Yep.  A double underscore just refers to an anonymous hyperlink reference. ReST keeps reference to all hyperlinks even if you intend for them to be anonymous, which can cause errors if you have the same link repeated in the same document. If the same link is not repeated, it's fine to use a single underscore.

strawberryfields/configuration.py
josh146: It's the same :slightly_smiling_face: The dictionary is what `default_config.toml` looks like when it's parsed. Just good practice to ship the code with a configuration template the user can fill in if they want.

tests/frontend/test_about.py
josh146: I realised this test wasn't being marked as a frontend test, so it was running numerous times.

strawberryfields/configuration.py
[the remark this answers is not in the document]
trbromley: :canada:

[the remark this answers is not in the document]
co9olguy: our

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+            filepath (str): path to the configuration file.
+        """
+        with open(filepath, "r") as f:
+            self._config_file = toml.load(f)
+
+    def save(self, filepath):
+        """Save a configuration file.
+
+        Args:
+            filepath (str): path to the configuration file.
```
co9olguy:
```suggestion
            filepath (str): path to the configuration file
```

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+            str: If no configuration is loaded, this returns ``None``."""
+        return self._filepath
+
+    def load(self, filepath):
+        """Load a configuration file.
+
+        Args:
+            filepath (str): path to the configuration file.
```
co9olguy:
```suggestion
            filepath (str): path to the configuration file
```

tests/frontend/test_configuration.py
```diff
@@ -0,0 +1,134 @@
...
+        conf.Configuration(name="noconfig")
+        assert "No Strawberry Fields configuration file found." in caplog.text
+
+    def test_save(self, tmpdir):
+        """Test saving a configuration file."""
+        filename = str(tmpdir.join("test_config.toml"))
+        config = conf.Configuration()
+
+        # make a change
+        config._config['api']['hostname'] = "https:/6.4.2.4"
```
co9olguy:
```suggestion
        config._config['api']['hostname'] = "https://6.4.2.4"
```

tests/frontend/test_configuration.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_attribute_loading(self):
+        """Test attributes automatically get the correct section key"""
+        config = conf.Configuration()
+        assert config.api == config._config['api']
+
+    def test_failed_attribute_loading(self):
+        """Test an exception is raised if key does not exist"""
+        config = conf.Configuration()
+        with pytest.raises(conf.ConfigurationError, match="Unknown"):
```
co9olguy: maybe have slightly longer match to avoid false positives?

default_config.toml
```diff
@@ -0,0 +1,12 @@
+## =================================================================
+##      STRAWBERRY FIELDS API CONFIGURATION OPTIONS
+## =================================================================
+
+[api]
+# Options for the Strawberry Fields Cloud API
+# Fill in your authentication
+authentatication_token = "071cdcce-9241-4965-93af-4a4dbc739135"
```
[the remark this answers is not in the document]
josh146: Just a fake stand-in (I generated it via `uuid.uuid4()`)

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+    [api]
+    # Options for the Strawberry Fields Cloud API
+    authentatication_token = "071cdcce-9241-4965-93af-4a4dbc739135"
+    hostname = "localhost"
+    use_ssl = true
+
+Summary of options
+------------------
+
+todo
```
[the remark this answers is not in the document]
josh146: Details about the configuration options --- I left this blank as this is still a WIP, and best filled in by @zzzeid

default_config.toml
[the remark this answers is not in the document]
josh146: Should I remove it?

josh146:
All suggested changes applied

tests/frontend/test_configuration.py
josh146: done

default_config.toml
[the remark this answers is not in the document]
co9olguy: Probably best to put in a blank token (so we can catch errors when users haven't gotten and configured token) and leave a comment indicating what a token looks like:
```suggestion
authentatication_token = None # example token form: 071cdcce-9241-4965-93af-4a4dbc739135
```

strawberryfields/configuration.py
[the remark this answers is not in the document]
co9olguy: ok I'll leave an issue so we don't forget

co9olguy:
I'd like @zzzeid to take a quick look over before merging

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        # search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.curdir, self._env_config_dir, self._user_config_dir, ""]
+        for idx, directory in enumerate(directories):
+            try:
+                self._filepath = os.path.join(directory, self._name)
+                self.load(self._filepath)
+                break
+            except FileNotFoundError:
+                if idx == len(directories) - 1:
```
zzzeid: Checking the index here is a little cryptic. Perhaps it's better to be more explicit about what we're checking?

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+    def __init__(self, name="config.toml"):
+        # Look for an existing configuration file
+        self._config = DEFAULT_CONFIG
+        self._config_file = {}
+        self._filepath = None
+        self._name = name
+        self._user_config_dir = user_config_dir("strawberryfields", "Xanadu")
+        self._env_config_dir = os.environ.get("SF_CONF", "")
+
+        # search the current directory, the directory under environment
```
zzzeid: Various occurrences of this, but PEP8 suggests that the first word should always be capitalized in comments (https://www.python.org/dev/peps/pep-0008/#id30)

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        The environment variable takes precedence."""
+        for section, section_config in self._config.items():
+            env_prefix = "SF_{}_".format(section.upper())
+
+            for key, _ in section_config.items():
+                # environment variables take precedence
+                env = env_prefix + key.upper()
+                if env in os.environ:
+                    self._config[section][key] = os.environ[env]
+                    continue
```
zzzeid: I think it would be a little more readable if instead of continue here, we used `elif` to combine with the logic below.

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.curdir, self._env_config_dir, self._user_config_dir, ""]
+        for idx, directory in enumerate(directories):
+            try:
+                self._filepath = os.path.join(directory, self._name)
+                self.load(self._filepath)
+                break
+            except FileNotFoundError:
```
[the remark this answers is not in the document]
josh146: My understanding is that the try-except block (with `FileNotFoundError`) is faster and more robust (see https://stackoverflow.com/questions/82831/how-do-i-check-whether-a-file-exists-without-exceptions). It also tends to be reasonably Pythonic to use try-except patterns.

strawberryfields/configuration.py
josh146: True, I've added a comment clarifying (let me know if you have other suggestions)

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def update_config(self):
+        """Updates the configuration from either a loaded configuration
+        file, or from an environment variable.
+
+        The environment variable takes precedence."""
+        for section, section_config in self._config.items():
+            env_prefix = "SF_{}_".format(section.upper())
+
+            for key, _ in section_config.items():
```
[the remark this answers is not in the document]
josh146: Good catch! Fixed

strawberryfields/configuration.py
josh146: done

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+                if env in os.environ:
+                    self._config[section][key] = os.environ[env]
+                    continue
+
+                # check if a configuration file exists
+                if self._config_file:
+                    # update from configuration file
+                    self._config[section][key] = self._config_file[section][key]
+
+    def __getattr__(self, key):
```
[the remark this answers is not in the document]
josh146: Fixed!

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+            str: If no configuration is loaded, this returns ``None``."""
+        return self._filepath
+
+    def load(self, filepath):
+        """Load a configuration file.
+
+        Args:
+            filepath (str): path to the configuration file
+        """
+        with open(filepath, "r") as f:
```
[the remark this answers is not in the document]
josh146: I'm a bit on the fence. We would just be replicating the default Python 3 error message,
```python
FileNotFoundError: No such file or directory: 'nonexistent.txt'
```
which I feel is pretty self-explanatory for the user.

strawberryfields/configuration.py
josh146: Done

strawberryfields/configuration.py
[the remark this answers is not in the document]
zzzeid: That's fair, but I think this could still be refactored a little to be more readable and reduce some of the code in the try block, for example:
```
        for idx, directory in enumerate(directories):
            self._filepath = os.path.join(directory, self._name)
            try:
                self.load(self._filepath)
            except FileNotFoundError:
                # Do stuff...
            else:
                self.update_config()
                break
```

Thoughts?

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        self._config = DEFAULT_CONFIG
+        self._config_file = {}
+        self._filepath = None
+        self._name = name
+        self._user_config_dir = user_config_dir("strawberryfields", "Xanadu")
+        self._env_config_dir = os.environ.get("SF_CONF", "")
+
+        # search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.curdir, self._env_config_dir, self._user_config_dir, ""]
```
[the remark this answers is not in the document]
josh146: Fixed

strawberryfields/configuration.py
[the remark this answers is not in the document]
zzzeid: Fair enough -- I made a different suggestion where `load` is used above in relation to this. What you have here makes sense on second thought, especially that we are setting `self.config_file` and not actually returning anything here.

strawberryfields/configuration.py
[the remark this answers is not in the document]
josh146: Yeah that looks good 🙂

default_config.toml
```diff
@@ -0,0 +1,12 @@
+## =================================================================
```
jswinarton: Why TOML? Any benefits to using this over python's built-in config language?

requirements.txt
```diff
@@ -4,3 +4,5 @@ scipy>=1.0.0
 tensorflow==1.3
 tensorflow-tensorboard>=0.1.8
 quantum-blackbird
+toml
```
jswinarton: Should these have fixed versions?

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        # Search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
+        for idx, directory in enumerate(directories):
+            self._filepath = os.path.join(directory, self._name)
+            try:
+                self.load(self._filepath)
+            except FileNotFoundError:
+                # check if we have searched all directories
+                if idx == len(directories) - 1:
```
zzzeid:
```suggestion
                if directory == directories[-1]:
```

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
+        for idx, directory in enumerate(directories):
+            self._filepath = os.path.join(directory, self._name)
+            try:
+                self.load(self._filepath)
+            except FileNotFoundError:
+                # check if we have searched all directories
```
zzzeid:
```suggestion
                # Check if we have searched all directories
```

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        self._config_file = {}
+        self._filepath = None
+        self._name = name
+        self._user_config_dir = user_config_dir("strawberryfields", "Xanadu")
+        self._env_config_dir = os.environ.get("SF_CONF", "")
+
+        # Search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
+        for idx, directory in enumerate(directories):
```
zzzeid:
```suggestion
        for directory in directories:
```

strawberryfields/configuration.py
jswinarton: I think this check could be avoided entirely if you modified the `load` method to return the config object at the end (which I think is good practice).

```python
directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
for directory in directories:
    self._filepath = os.path.join(directory, self._name)
    try:
        config = self.load(self._filepath)
        break
    except FileNotFoundError:
        pass

if config:
    self.update_config()
else:
    log.info("No Strawberry Fields configuration file found.")
```

strawberryfields/configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        # Search the current directory, the directory under environment
+        # variable SF_CONF, and default user config directory, in that order.
+        directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
+        for directory in directories:
+            self._filepath = os.path.join(directory, self._name)
+            try:
+                self.load(self._filepath)
+            except FileNotFoundError:
+                # Check if we have searched all directories
+                if directory == directories[-1]:
```
jswinarton: I think this check could be avoided entirely if you modified the `load` method to return the config object at the end (which I think is good practice).

```python
directories = [os.getcwd(), self._env_config_dir, self._user_config_dir]
for directory in directories:
    self._filepath = os.path.join(directory, self._name)
    try:
        config = self.load(self._filepath)
        break
    except FileNotFoundError:
        pass

if config:
    self.update_config()
else:
    log.info("No Strawberry Fields configuration file found.")
```

default_config.toml
josh146: Partly because PennyLane uses it, and I want to maintain consistency. Also, TOML is generally easier to hand-edit than JSON, not as overcomplicated as YAML, and better suited to structured data than INI files (the format used by Python's in-built ConfigParser library).

It also seems that the Python ecosystem is moving towards TOML, e.g., PEP518 and pyproject.toml files.

requirements.txt
josh146: Yes, probably. We should fix this in a new PR

strawberryfields/configuration.py
josh146: This looks way better! Sad you can't yet do multi-line GitHub suggestions 😂

zzzeid: In this case, I would suggest that we throw and catch the exception in `load`, and return None if the file does not exist, to make this code cleaner. I.e. 

```
for directory in directories:
    self._filepath = os.path.join(directory, self._name)
    config = self.load(self._filepath)
    if config:
        self._config_file = config
        self.update_config()

```

default_config.toml
jswinarton: Neat, didn't know about the PEP.

strawberryfields/configuration.py
josh146: The one disadvantage is that `load()` no longer will raise a `FileNotFoundError`, instead, it will silently fail. A developer building an app against SF is using `Configuration.load()` to manually load a configuration file will have to manually check that the function returned `None`, and manually implement error handling.

jswinarton: I implemented this as @zzzeid suggested but I agree with this, it depends on what the use case is. Do we intend to expose `load` as a method for developers to build against? If so, we should probably change this back. Or, we could change the method name to `_load`.

josh146: I think it's still an open question - to early to say how we intend for it to be used/how people might use it. My preference would be to change it back, as it could lead to unexpected or hard to debug behavior down the line.

zzzeid: My original comment was in relation to catching the exception and passing, as that would suppress the exception and not provide any indication as to what happened to the file loading. I think earlier we were catching the exception and showing a warning message. Essentially I guess we are still doing the same thing. I'm good with either implementation.

josh146: We'll always need to catch the exception and pass, this is more a question about where to do it. I feel it's better to do that in __init__ and not in load(), as (1) load could be developer/user facing, and (2) needing to pass on the exception is only required during initialization.

zzzeid: @josh146 This makes sense, and reduces the responsibility of the `load` method.

josh146: :+1: I'll make a PR to revert this behaviour.

PR #96, opened 2019-06-21
Removing double hats

nquesada:
The macros \x and \p already put hats, no need to add them by hand. Now it should render with only a single hat on top of the canonical operators instead of two.


josh146:
image

Oops. Good catch @nquesada

co9olguy:
The double hats look cool at least

PR #101, opened 2019-06-26
StarshipEngine

zzzeid:

- New class called StarshipEngine
- Compiles and validates quantum programs
- Pass on jobs to communication backend (APIClient)
- Receives job results from communication backend and returns them

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/engine.py
```diff
@@ -340,4 +346,127 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+            program (strawberryfields.program.Program): A program instance to be executed remotely.
+
+        Returns:
+            (list): A list representing the result samples
+
+        Raises:
+            Exception: In case a job could not be submitted or completed.
+            TypeError: In case a job is already queued and a user is trying to submit a new job.
+        """
+        blackbird_code = program.get_blackbird_syntax()
```
josh146: Note: rather than modifying `program.py` and `ops.py` to print out the Blackbird command, it would be better to use the `strawberryfields.io` module:

```python
from strawberryfields.io import to_blackbird
blackbird_code = to_blackbird(program).serialize()
```

Mainly because it avoids code duplication, is already well tested, and takes into account lots of edge cases (such as array parameters, measurements, register transforms, etc.)

Aside: should we make a `Program.to_blackbird` method that calls the above function?

zzzeid: @josh146 Thanks! Did not look hard enough for this `to_blackbird` method, I will switch this.

zzzeid: @josh146 updated.

strawberryfields/engine.py
```diff
@@ -74,11 +74,16 @@
...
 
 import abc
+import uuid
 from collections.abc import Sequence
 from numpy import stack, shape
+from time import sleep
 
 from .backends import load_backend
-from .backends.base import (NotApplicableError, BaseBackend)
+from .backends.base import NotApplicableError, BaseBackend
```
josh146: How come `BaseBackend` and `NotApplicableError` are being imported here? With the backends, we are currently trying to keep it as separate from the frontend as possible, with the aim of spinning them out into separate Python packages in the future (hence the use of the `load_backend` helper function).

strawberryfields/engine.py
```diff
@@ -209,7 +218,7 @@ def _run(self, program, *, shots=1, compile_options={}, **kwargs):
             Result: results of the computation
         """
 
-        def _broadcast_nones(val, shots):
+        def _normalize_sample(val):
```
josh146: I prefer `_broadcast_nones` in this context, as it is (a) slightly more explicit (the `None`'s are being broadcast a la NumPy terminology), and (b) 'normalize samples' has slightly different connotations in quantum physics.

But this is likely being removed in #111 anyway, so I would suggest changing it back just to avoid any annoying merge conflicts arising.

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+    Executes :class:`.Program` instances on the chosen remote backend, and makes
+    the results available via :class:`.Result`.
+
+    Args:
+        backend (str, BaseBackend): name of the backend, or a pre-constructed backend instance
+    """
+
+    API_DEFAULT_REFRESH_SECONDS = 1
+    HARDWARE_BACKENDS = ("chip0",)
```
josh146: I think this is fine for now, but for future PRs, I'm a bit wary about hardcoding available hardware backends directly into the StarshipEngine, as the hardware availability might fluctuate independently of SF releases.

Once the corresponding API endpoints are available, we should add support for

* `StrarshipEngine.backend_info()`: show a list of available devices/simulators/backends

* `StarshipEngine.backend_info('chip0')`: shows more detailed information and stats about a specific backend

and use the results to provide the logic for which backends are available.

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+    def __init__(self, api_client_params=None):
+        # Only chip0 backend supported initially.
+        backend = "chip0"
+        super().__init__(backend)
+
+        api_client_params = api_client_params or {}
```
josh146: Nice! Clear and avoids the additional if statement.

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
 
+class StarshipEngine(BaseEngine):
+    """
+    Starship quantum program executor engine.
+
+    Executes :class:`.Program` instances on the chosen remote backend, and makes
+    the results available via :class:`.Result`.
+
+    Args:
+        backend (str, BaseBackend): name of the backend, or a pre-constructed backend instance
```
josh146: Could we add support for also initializing the engine with API client arguments via keyword arguments, cf. ADR STRAW-92:

```python
eng = sf.HTTPEngine(backend="chip0", host="api.strawberryfields.ai", api_key="XXXXX")
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+    def generate_job_content(self, name, shots, blackbird_code):
+        """
+        Generates a string representing the Blackbird code that will be sent to the server.
+        Assumes the current backend as the target.
+
+        Args:
+            name (str): The name of the job to be created (e.g. StateTeleportation).
+            shots (int): The number of shots.
+            blackbird_code: The blackbird code of the job.
```
josh146: Due to being sentence fragments, no first word capitalization/ending period is required. 
```suggestion
            blackbird_code: the blackbird code of the job
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+        job.manager.create(circuit=job_content)
+        self.jobs.append(job)
+        return job
+
+    def _run_program(self, program, **kwargs):
+        """
+        Given a compiled program, gets the blackbird circuit code and creates (or resumes) a job
+        via the API. If the job is completed, returns the job result.
+
+        A queued job can be interrupted by a KeyboardInterrupt event, at which point if the job ID
```
josh146:
```suggestion
        A queued job can be interrupted by a ``KeyboardInterrupt`` event, at which point if the job ID
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+        was retrieved from the server, the job will be accessible via engine.jobs.
```
josh146:
```suggestion
        was retrieved from the server, the job will be accessible via :meth:`~.Starship.jobs`.
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+            (list): A list representing the result samples
+
+        Raises:
+            Exception: In case a job could not be submitted or completed.
+            TypeError: In case a job is already queued and a user is trying to submit a new job.
+        """
+        if self.jobs:
+            raise TypeError("A job is already queued. Please reset the engine and try again.")
+
+        blackbird_code = to_blackbird(program)
```
josh146: You can pass the target blackbird version number here:
```suggestion
        blackbird_code = to_blackbird(program, version=1.0)
```
Future versions of the Blackbird library will allow you to generate Blackbird targeting multiple versions. Currently, there is only one pre-release version, hence why it just says `1.0` everywhere :laughing:

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+        target = self.backend_name
+        return "\n".join(
+            [
+                "name {name}",
+                "version 1.0",
+                "target {target} (shots={shots})",
+                "",
+                "{blackbird_code}",
+            ]
+        ).format(name=name, target=target, shots=str(shots), blackbird_code=blackbird_code)
```
josh146: Note that the blackbird program is a Python class representation of the Blackbird circuit, and calling `str(blackbird_code)` simply returns the result of `blackbird_code.serialize()`. In this case, this method could be simplified by setting the circuit metadata first before serializing:

```python
blackbird_code._name = ... # we can also implement a proper property setter here if needed
blackbird_code._target['options'] = {'shots': shots}
return blackbird_code.serialize()
```

This will also ensure all syntax/metadata matches the requested Blackbird version.

Note that the target should already match the backend name, due to the compilation step performed in `BaseEngine`.

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+        Returns:
+            (list): A list representing the result samples
+
+        Raises:
+            Exception: In case a job could not be submitted or completed.
+            TypeError: In case a job is already queued and a user is trying to submit a new job.
+        """
+        if self.jobs:
+            raise TypeError("A job is already queued. Please reset the engine and try again.")
```
josh146: Should this be a type error? It feels like it should be a custom API exception class.

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+        try:
+            while not job.is_failed and not job.is_complete:
+                job.reload()
+                sleep(self.API_DEFAULT_REFRESH_SECONDS)
+        except KeyboardInterrupt:
+            if job.id:
+                print("Job {} is queued in the background.".format(job.id.value))
+            else:
+                raise Exception("Job was not sent to server. Please try again.")
```
josh146: Another good place for a custom API exception

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+        if job.is_failed:
+            # TODO: Add failure details here, and use a better exception.
+            raise Exception("Job execution failed. Please try again.")
+        elif job.is_complete:
+            job.result.manager.get()
+            return job.result.result.value
```
josh146: Should this return the `job` instance, so that the user can manually check the status/reload the job/get the results when complete? See http://platform-docs/research/STRAW-92-engine-design.html#id1

tests/frontend/test_engine.py
```diff
@@ -19,13 +19,18 @@
 import strawberryfields as sf
 from strawberryfields import ops
 from strawberryfields.backends.base import BaseBackend
+from unittest.mock import MagicMock
+from strawberryfields import StarshipEngine
+from strawberryfields.api_client import APIClient
+from strawberryfields.ops import S2gate, MeasureFock, Rgate, BSgate
```
josh146: Note that `ops` is already imported above, so you can simply call `ops.S2gate`.

josh146: P.S. might be good to run `isort` on this test file

tests/frontend/test_engine.py
```diff
@@ -146,13 +159,207 @@ def inspect():
...
+            Rgate(3) | 2
+            Rgate(3) | 3
+            MeasureFock() | [0]
+            MeasureFock() | [1]
+            MeasureFock() | [2]
+            MeasureFock() | [3]
+
+        engine.run(prog)
+
+        mock_api_client_post.assert_called_once()
```
josh146: Tests look awesome!

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+    Starship quantum program executor engine.
+
+    Executes :class:`.Program` instances on the chosen remote backend, and makes
+    the results available via :class:`.Result`.
+
+    Args:
+        backend (str, BaseBackend): name of the backend, or a pre-constructed backend instance
+    """
+
+    API_DEFAULT_REFRESH_SECONDS = 1
```
josh146: Should this be an `APIClient` argument?

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+        self.client = APIClient(**api_client_params)
+        self.jobs = []
+
+    def __str__(self):
+        return self.__class__.__name__ + "({})".format(self.backend_name)
+
+    def reset(self, backend_options=None):
+        """
+        Reset must be called in order to submit a new job. This clears the job queue as well as
+        any ran Programs.
```
josh146: Needs to add `backend_options` argument description (can likely just copy from elsewhere).

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+        # Do nothing for now...
+        pass
+
+    def generate_job_content(self, name, shots, blackbird_code):
+        """
+        Generates a string representing the Blackbird code that will be sent to the server.
+        Assumes the current backend as the target.
+
+        Args:
+            name (str): The name of the job to be created (e.g. StateTeleportation).
```
josh146:
```suggestion
            name (str): the name of the job to be created (e.g. StateTeleportation)
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+            shots (int): The number of shots.
```
josh146:
```suggestion
            shots (int): the number of shots
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+            blackbird_code: The blackbird code of the job.
+
+        Returns:
+            str: A string containing the job content to be sent to the server.
```
josh146:
```suggestion
            str: job content to be sent to the server
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+            ]
+        ).format(name=name, target=target, shots=str(shots), blackbird_code=blackbird_code)
+
+    def _queue_job(self, job_content):
+        """
+        Create a Job instance based on job_content, and send the job to the API. Append to list
+        of jobs.
+
+        Args:
+            job_content (str): The Blackbird code to execute
```
josh146:
```suggestion
            job_content (str): the Blackbird code to execute
```
(Aside - we should work out if Blackbird should be capitalized or not, I think I've seen both across the codebase).

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+        Returns:
+            (strawberryfields.api_client.Job): A Job instance referencing the queued job.
```
josh146:
```suggestion
            (strawberryfields.api_client.Job): a Job instance referencing the queued job
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+    def _run_program(self, program, **kwargs):
+        """
+        Given a compiled program, gets the blackbird circuit code and creates (or resumes) a job
+        via the API. If the job is completed, returns the job result.
+
+        A queued job can be interrupted by a KeyboardInterrupt event, at which point if the job ID
+        was retrieved from the server, the job will be accessible via engine.jobs.
+
+        Args:
+            program (strawberryfields.program.Program): A program instance to be executed remotely.
```
josh146:
```suggestion
            program (strawberryfields.program.Program): program to be executed remotely
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+        Returns:
+            (list): A list representing the result samples
```
josh146:
```suggestion
            (list): a list representing the result samples
```

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+            # TODO: Add failure details here, and use a better exception.
+            raise Exception("Job execution failed. Please try again.")
+        elif job.is_complete:
+            job.result.manager.get()
+            return job.result.result.value
+
+    def run(self, program, shots=1, name=None, **kwargs):
+        """
+        Compile a given program and queue a job in the Starship.
+        """
```
josh146: Since this is a not an inherited method, need to provide details of arguments/return type (can be mostly copied from `BaseEngine._run`).

strawberryfields/engine.py
zzzeid: Looks like these are used in `LocalEngine` right now, but they lit up after running `black`. I'll look into this though.

strawberryfields/engine.py
zzzeid: That's fair -- I actually patched this as part of a commit from the `compile_in_stages` branch and just renamed it here. I'll revert the changes I made on top to avoid conflicts.

strawberryfields/engine.py
zzzeid: This makes sense, and will make this more modular. I will keep a note of this.

strawberryfields/engine.py
zzzeid: Thanks, will update this.

strawberryfields/engine.py
josh146: Ah, of course! Must have missed it in the previous PR.

strawberryfields/engine.py
zzzeid: I think this is specific to the engine, as it sets the polling delay that the engine would wait before fetching the job status. Perhaps renaming it to `POLLING_DELAY_SECONDS` would be better?

josh146: Make sense :+1: It was the `API` that threw me 😆 

Would this be better as a instance attribute, so that different engines can have different polling delays?

strawberryfields/engine.py
```diff
@@ -336,8 +360,144 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+        super().reset(backend_options)
+        self.jobs.clear()
+
+    def _init_backend(self, *args):
+        """
+        TODO: This does not do anything right now.
+        """
+        # Do nothing for now...
+        pass
```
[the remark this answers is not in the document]
zzzeid: Yes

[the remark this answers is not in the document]
josh146: Strange. I was going to say, it doesn't seem needed for a remote backend, but perhaps this will be the place where https://github.com/XanaduAI/strawberryfields/pull/101#discussion_r300222618 goes.

strawberryfields/engine.py
zzzeid: Noted.

strawberryfields/engine.py
zzzeid: We should potentially rethink this in terms of how it applies to all engines. `_run_program` in `LocalEngine` and `BaseEngine` in fact, both return a list of applied commands, which is not applicable to this class. Currently, `StarshipEngine.jobs` has a list of jobs that are running, and can be checked or reloaded there. Async implementation, for the time being, is implemented as a `KeyboardInterrupt`, although I should add a flag for this (or possibly a different method, as the research doc suggests) so that a user can intentionally queue a job asynchronously without input.

josh146: Tbh, I'm not entirely sure why `_run_program` needs to return a list of applied commands; it looks like `LocalEngine` doesn't even store the return value/do anything with it (unless I'm missing something).

Yep, it would be good to add a `run_async` in a later PR. Although, I quite like your `KeyboardInterrupt` idea, and even when we implement `run_async`, I still think we should keep both.

strawberryfields/engine.py
zzzeid: Seems like this may be fixed in the future by #111 , so will leave it as is for now.

strawberryfields/engine.py
zzzeid: Added parameters that match APIClient parameters. Let me know what you think.

tests/frontend/test_engine.py
zzzeid: Thanks!

strawberryfields/engine.py
smite: Looks like `BaseBackend` is only needed for isinstance checks. `_run_program` still catches `NotApplicableErrors` just in case, although the validation process during program compilation should prevent them.

strawberryfields/engine.py
josh146: :+1:

strawberryfields/engine.py
zzzeid: Changed to instance attribute, and agreed that having different polling delays could come in handy.

strawberryfields/engine.py
```diff
@@ -218,45 +227,51 @@ def _broadcast_nones(val, shots):
...
+        # signatures of methods in Operations to remain cleaner, since only
+        # Measurements need to know about shots
+
+        if self.backend_name in getattr(self, "HARDWARE_BACKENDS", []):
+            p = program[0]
+            p = p.compile(self.backend_name)  # TODO: does compile need to know about shots?
+            p.lock()
+            self.run_progs.append(p)
+            samples = self._run_program(p, **kwargs)
+            return Result(samples)
```
smite: I guess the idea here is that hardware backends can only run a single program before they have to be reset? If so, maybe `DeviceSpecs.interactive` should be checked instead.

Also, it would be possible to combine the Programs on the client side before execution.

strawberryfields/engine.py
zzzeid: I removed this argument for now as it is not used anywhere in StarshipEngine.

strawberryfields/engine.py
josh146: Using `DeviceSpecs.interactive` during compilation sounds like the right approach.

zzzeid: Ok let me look into this.

strawberryfields/engine.py
smite: If the child class only slightly changes the behavior of the method, maybe it'd be enough to describe the different behavior and then just add a link to the original method with the original docstring?
```
  See :meth:`BaseBackend.reset`
```

strawberryfields/engine.py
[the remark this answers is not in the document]
smite: The method could be removed, it's just a one-liner in LocalEngine.

strawberryfields/engine.py
zzzeid: @smite the only way to combine programs with the current hardware support, as I understand it, would be to concatenate all the programs and ensure that there is one measurement at the end. This is out of scope for this PR but it is something I discussed with @co9olguy briefly.

strawberryfields/engine.py
```diff
@@ -336,8 +357,159 @@ def run(self, program, *, shots=1, compile_options={}, modes=None, state_options
...
+
+    def _get_blackbird(self, name, shots, program):
+        """
+        Returns a Blackbird object to be sent later to the server when creating a job.
+        Assumes the current backend as the target.
+
+        Args:
+            name (str): the name of the job to be created (e.g. StateTeleportation)
+            shots (int): the number of shots
+            blackbird_code: the blackbird code of the job
```
smite:
```suggestion
            program (Program): program to be converted to Blackbird code
```

strawberryfields/engine.py
smite: `_run_program` could be removed from the base class and made a LocalEngine method only. `BaseEngine` is a very preliminary design, so if it makes sense to change the @abc.abstractmethods it defines, go for it.

strawberryfields/engine.py
zzzeid: @smite @josh146 using `DeviceSpecs.interactive` here breaks some existing tests (e.g. `test_print_applied`) because the backend fixture in those tests uses the `base` backend which is set to interactive. I assume this is intentional, but tests that run two programs will have to be updated. E.g.
```
        # ... line 162 in test_engine.py
        # reapply history
        eng.reset()
        eng.run([p1, p2])
        assert inspect() == expected2
```

strawberryfields/program_utils.py
```diff
@@ -108,12 +109,18 @@ def __init__(self, op, reg):
         self.reg = reg
 
     def __str__(self):
-        """Print the command using Blackbird syntax."""
-        temp = str(self.op)
+        """
+        Return a string containing the Blackbird syntax.
```
smite:
```suggestion
        Return a string containing the command in Blackbird syntax.
```

strawberryfields/engine.py
smite: Should be version="1.0" :)

strawberryfields/engine.py
smite: It may be that local and remote engines don't really have that much in common.

strawberryfields/engine.py
smite: Checking the interactivity of the backend during compilation probably does not make much sense, since the Programs are compiled individually and don't know about each other...

strawberryfields/engine.py
zzzeid: Thanks, this was stale, left over from the old implementation.

strawberryfields/engine.py
zzzeid: @smite Yes true, originally I was trying to ensure a good fit between the two. They do have certain things in common obviously -- I guess this will get more interesting when/if we introduce even more engines and backends for each engine.

strawberryfields/engine.py
zzzeid: That check fails if it's in the `_run` method, before any compiling happens, but because those tests are passing multiple programs but the base backend is set to `interactive`. Likely if these tests expect the backend to not be interactive, they should initialize one with that setting.

strawberryfields/engine.py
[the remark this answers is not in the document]
zzzeid: @smite I wonder if the call to `_init_backend` in `_run` should be moved out of the for loop (line 278 in `engine.py`). I don't think it makes sense to remove this method from the abstract base class until we refactor how it is used in `_run`. Thoughts?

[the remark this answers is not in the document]
smite: Yeah, that's probably a good idea. But most likely `StarshipEngine.run()` should not call `BaseEngine._run()` at all, since what it needs to do is so different from what a local engine needs to do.

It seems to me StarshipEngine (at least for now) only needs to be able to run a single Program at a time, so the loop isn't necessary. No backend init is needed on the client side. 
All it needs from `BaseEngine._run()` is the Program compilation/locking part and the assignment of `self.samples`, so it's probably easiest to add these directly to `StarshipEngine.run()`.

tests/frontend/test_engine.py
```diff
@@ -146,13 +157,211 @@ def inspect():
...
+        """
+        mock_api_client = MagicMock()
+        monkeypatch.setattr("strawberryfields.engine.APIClient", mock_api_client)
+        engine = StarshipEngine()
+        assert engine.client == mock_api_client()
+        assert engine.jobs == []
+
+    def test_reset(self, starship_engine):
+        """
+        Tests that StarshipEngine.jobs is correctly cleared when callling StarshipEngine.reset.
```
co9olguy:
```suggestion
        Tests that StarshipEngine.jobs is correctly cleared when calling StarshipEngine.reset.
```

tests/frontend/test_engine.py
```diff
@@ -146,13 +157,211 @@ def inspect():
...
+        monkeypatch.setattr("strawberryfields.engine.to_blackbird", mock_to_blackbird)
+        monkeypatch.setattr(starship_engine, "_get_blackbird", mock__get_blackbird)
+        monkeypatch.setattr(starship_engine, "_queue_job", lambda job_content: mock_job)
+
+        some_params = {"param": MagicMock()}
+        result = starship_engine._run_program(program, **some_params)
+
+        mock__get_blackbird.assert_called_once_with(program=program, param=some_params["param"])
+
+        assert result == mock_job.result.result.value
```
co9olguy: how come `result` appears twice?

tests/frontend/test_engine.py
```diff
@@ -146,13 +157,211 @@ def inspect():
...
+
+        result = starship_engine._queue_job(mock_job_content)
+        mock_job.assert_called_once_with(client=starship_engine.client)
+        result.manager.create.assert_called_once_with(circuit=mock_job_content)
+        assert starship_engine.jobs == [result]
+
+    def test__run_program(self, starship_engine, monkeypatch):
+        """
+        Tests StarshipEngine._run_program. Asserts that a program is converted to blackbird code,
+        compiled into a job content string and that the job is queued. Also asserts that a
```
co9olguy: the mocking makes it a bit hard to parse, but I couldn't see where it is asserted that the program is compiled into a job content string and queued.

tests/frontend/test_engine.py
```diff
@@ -146,13 +157,211 @@ def inspect():
...
+    def test__run(self, starship_engine, monkeypatch):
+        """
+        Tests StarshipEngine._run, with the assumption that the backend is a hardware backend
+        that supports running only a single program. This test ensures that a program is compiled
+        for the hardware backend, is locked, is added to self.run_progs, that it is run and that
+        a Result object is returned populated with the result samples.
+        """
+
+        inputs = MagicMock()
+        inputs.shots = 1
```
co9olguy: would the test look any different if we had a value other than 1 here?

tests/frontend/test_engine.py
```diff
@@ -146,13 +157,211 @@ def inspect():
...
+        inputs.params = {"param": MagicMock()}
+
+        starship_engine.run(inputs.program, inputs.name, inputs.shots, **inputs.params)
+        mock__run.assert_called_once_with(
+            inputs.program, shots=inputs.shots, name=inputs.name, param=inputs.params["param"]
+        )
+
+    def test_engine_with_mocked_api_client_sample_job(self, monkeypatch):
+        """
+        This is an integration test that tests and actual program being submitted to a mock API, and
```
co9olguy:
```suggestion
        This is an integration test that tests an actual program being submitted to a mock API, and
```

tests/frontend/test_engine.py
```diff
@@ -146,13 +157,211 @@ def inspect():
...
+        how the engine handles a successful response from the server (first by queuing a job then by
+        fetching the result.)
```
co9olguy:
```suggestion
        fetching the result).
```

co9olguy:
@zzzeid @josh146 @smite What is status of this PR? Is it functional enough to merge into `demo` branch?

tests/frontend/test_engine.py
zzzeid: `mock_job.result` is the resource (i.e. `job/<id>/result`). We have chosen to use `"result"` as a key in the response, so a `JobResult` resource (which can be accessed via `job.result` has a "result" attribute that is in fact the result. The response looks like `{"result": "..."}`.

Potentially a better key name could be "content".

tests/frontend/test_engine.py
zzzeid: It is expecting an integer. This was added to support a change in #111 -- we could use a `FuzzyInteger` which could be more appropriate.

strawberryfields/engine.py
```diff
@@ -385,4 +420,154 @@ def run(self, program, *, compile_options=None, run_options=None):
...
+            )
+
+        job_content = self._get_blackbird(program=program, **kwargs).serialize()
+        job = self._queue_job(job_content)
+
+        try:
+            while not job.is_failed and not job.is_complete:
+                job.reload()
+                sleep(self.polling_delay_seconds)
+        except KeyboardInterrupt:
```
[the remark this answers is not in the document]
zzzeid: Great question -- I think that some refactoring is needed here. A job without an ID probably should never be added to `self.jobs`. I believe with the current setup this will occur (as an exception will be raised if the server does not respond with a `201`). I do agree that this can be coded better though.

tests/frontend/test_engine.py
co9olguy: No plan to merge #111 immediately.

Probably enough to parameterize the test with two values, 1, and any other positive integer (say 1000)

josh146:
🎆

PR #105, opened 2019-06-28
install on win10

XueshiGuo:
I get a problem similar to this one when setup on win10

https://github.com/rkern/line_profiler/issues/37


josh146:
Hi @XueshiGuo, thanks for the PR! It seems that this is an issue caused by a non-standard default system encoding, hence why we haven't caught it before.

PR #111, opened 2019-06-30
Multi-shot sampling

### LINKED ISSUE #2: Sampling
It would be nice to specify 'sampling' as well as the number of shots to sample - if this is specified, the simulation should then sample the results of the measurement operators.

The fastest way to do this is to 'cheat' and simply run the circuit once *without* the measurement operators (as these reset the modes to the vacuum state), and then sample the Fock state probabilities/quadrature probabilities of the qumodes that are to be measured.

This feature should also ideally be supported in Blackbird in a way that doesn't require the user to modify the engine queue to remove existing measurement gates. Perhaps as a keyword argument when running the engine?

`eng.run(n_subsystems=3, backend=b, sampling=50)`

In this case, the frontend would:
1. 'hide' the measurement operators from the queue, but remember the qumodes to be measured
2. extract the required probability distribution and sample respective qumodes 50 times
3. return the **list** of sampled values to the user, perhaps using the attribute `reg.val` (or defining `reg.sample`?)

### LINKED ISSUE #107: Unify the format of returned measurement data
We should make the format (type, shape) of returned measurement results more uniform. Currently we sometimes return scalars, arrays, lists, tuples, etc. 

With the new `Results` object being created, now is the time to refactor the return types for various backends, and have a unified standard expressed in `Results`. 

This may also have relevance for register values.

Things to take into account:
1. shots
1. batch_size
1. number of measured modes
1. whether the user entered measured modes out of order -- should we return them in same order requested?

co9olguy:
Work in progress.

* Implements [multi-shot sampling](https://github.com/XanaduAI/platform-docs/blob/master/src/research/0030-sf-multi-shot-sampling.rst) in the `LocalEngine` interface as well as the backend API.
* Changes the way the measurement samples are stored in the `RegRef`s and the `Result` objects:
  For each measured mode, the samples are stored in an `array` of shape `([batch_size,] shots)`. When the measurement samples of multiple modes are stored together, the array shape is `(modes, [batch_size,] shots)`. This is also the format the `measure_*` methods of the backend API use for their return values.
* Currently all the `measure_*` methods implement the `shots` argument, except for the TF backend.

Related issues:

Closes #2, closes #107


co9olguy:
See #107

tests/integration/test_engine_integration.py
```diff
@@ -322,3 +308,435 @@ def test_measurefock_shots_exception(self, setup_eng):
                            match=r"""(Measure|MeasureFock) has not been implemented in {} """
                                   """for the arguments {{'shots': {}}}""".format(backend_name, shots)):
             eng.run(p1, shots=shots).samples
+
+class TestResults:
```
co9olguy: Note: I have a feeling some of these tests could be simplified/unified, perhaps with a `pytest.mark.parameterize`? Suggestions welcome

tests/integration/test_engine_integration.py
```diff
@@ -19,7 +19,7 @@
```
co9olguy: Note: I did not add any tests that check if the object returned from the backend has a particular shape, but we could add one for `gaussianbackend.measure_fock` I suppose?

strawberryfields/engine.py
```diff
@@ -86,15 +87,40 @@ class Result:
 
     Represents the results of the execution of a quantum program.
     Returned by :meth:`.BaseEngine.run`.
+
+    Args:
+        samples (list[list[int, float, None]]): nested list of
+            measurement samples, with shape ``(modes, shots)``
```
smite: How about storing the samples already in `BaseEngine` as a Dict[int, List[float,int]], and passing a copy of that to Result? I thought we agreed that storing Nones isn't the way to go, right?

smite: If the measurement samples obtained from the backend API call are stored in RegRefs first, transferring them into the samples dictionary only requires
`BaseEngine.samples = {r.ind: r.val for r in prog.reg_refs.values() if r.val is not None}`

I don't know how RegRefs would store batched data though.

co9olguy: Yes, that's right about `None`s. I think @josh146 just forgot that changing that behaviour was in-scope for this issue as well

co9olguy: > If the measurement samples obtained from the backend API call are stored in RegRefs first, transferring them into the samples dictionary only requires
> `BaseEngine.samples = {r.ind: r.val for r in prog.reg_refs.values() if r.val is not None}`
> 
> I don't know how RegRefs would store batched data though.

Yep, seems like this is the easiest way to do it. Presuming everything is returned from backend in good order, should be simple to convert to a dict and wrap in `Results`

co9olguy: I believe batched data was just stored as an array in each RegRef

co9olguy:
I simplified the logic in the `Results` class, and fock and gaussian backends pass all tests except the ones which check for `NotImplementedError`. It seems the engine catches the `NotImplementedError`s given by the backends and replaces them with its own. I tried to fix the tests to sync with this (see `test_results_measure_homodyne_shots` in `test_engine_integration.py`), but for some reason couldn't get the regex matches to work :(. Someone will have to pick up the torch if they can see why the regex is not working.

I also shored up some of the failing TF tests by restructuring how the TF Backend returns its measurement results. There is some potential ambiguity between `shots` axis and `batch_size` axis. I decided to remedy this by having TF Backend always return measurements with shape `(modes, [batch_size,] shots)`. The makes things unambiguous to the frontend. Unfortunately, this has led to a propagation of other tests failing, which will need to be fixed now as well.

strawberryfields/engine.py
```diff
@@ -86,16 +87,45 @@ class Result:
...
+            arrs = [np.array(v) for v in samples.values()]
+            samples_shapes = [v.shape for v in arrs]
+            ref_shape = samples_shapes[0]
+
+            if not np.all([s == ref_shape for s in samples_shapes]):
+                raise ValueError("The returned samples do not all have the same shape.")
+
+            if len(ref_shape) == 0:
+                #  scalars, single shot
+                samples_array = np.expand_dims(np.stack(arrs), axis=1)  # shape==(modes, 1)
```
smite: Isn't the sample data already always in a list, even if shots==1? This case should not be necessary then.

strawberryfields/ops.py
```diff
@@ -569,15 +569,14 @@ def apply(self, reg, backend, **kwargs):
         """
         values = super().apply(reg, backend, **kwargs)
         # convert the returned values into an iterable with the measured modes indexed along
-        # the first axis and shots along second axis (if larger than 1), so that we can assign
+        # the first axis and shots along second axis, so that we can assign
         # register values
         shots = kwargs.get("shots", 1)
         if self.ns == 1:
             values = [values]  # values is either a scalar, or has shape (shots,)
```
smite: This check (and the transpose below) should be unnecessary too, if the backend API funcs always return measurement results in a 2d array with shape == (modes, shots) or shape == (modes, batch_size, shots).

strawberryfields/ops.py
```diff
@@ -569,15 +569,14 @@ def apply(self, reg, backend, **kwargs):
...
         else:
-            if shots > 1:
-                values = values.T  # shape of values would be (shots, num_meas,)
-
+            if shots <= 0:
+                raise ValueError("Number of shots must be greater than zero.")
```
smite: `Measurement.apply()` does not need to know about shots, it just needs to slice the 2d or 3d array it receives into the RegRefs. It's good to test that shots > 0, but it should happen already in `BaseEngine._run()`.

tests/integration/test_engine_integration.py
```diff
@@ -531,10 +531,10 @@ def test_results_measure_heterodyne_with_shots(self, setup_eng):
         shots = 5
         with p.context as q:
             ops.MeasureHeterodyne() | q[1]
-        name = eng.backend._short_name.capitalize()
+        name = eng.backend.__class__.__name__
         with pytest.raises(NotImplementedError,
-                           match="{} backend currently does not support "
-                                 "shots != 1 for heterodyne measurement".format(name)):
+                           match="The operation (MeasureHeterodyne|MeasureHD) has not been "
+                                 "implemented in {} for the arguments {{'shots': {}}}".format(name, shots)):
```
josh146: nice!

co9olguy: thanks for the tip :man_facepalming:

josh146:
>  I decided to remedy this by having TF Backend always return measurements with shape (modes, [batch_size,] shots).
:+1:

co9olguy:
@smite I uncovered an issue with using `RegRef`s of measurement results in later gates, since the return shape now always has an extra axis for `shots`. Any ideas how to remedy? My feeling is we should disallow conditional preparations when `shots != 1` and adopt the `RegRef`s to take into account `shots==1`

smite:
@co9olguy This problem is analogous to the question what the register state should be after a multi-shot measurement. Since the measurement was performed several times, there is not a single unique history following it. Hence anything that happens after the multishot measurement has to follow one of the possible histories, that is, choose a single measurement result (first? last?) and stick with it. This is not perfect, but it's the best we can do.

I think there's basically two options for us:
1. Forbid any Operations in the circuit after the first multi-shot measurement, refuse to return a state object if the user requests it.
2. Allow further Operations, define the state after the multi-shot measurement as outlined above.

co9olguy:
There are too many open issues/questions in this refactor to confidently put it in next release. I'm going to pull it out of v0.11 milestone

smite:
The tests seem to run now.

There's one major hack remaining, in `Measurement.apply()`. It takes care of converting results returned by `measure_homodyne()` and `measure_heterodyne()` into the correct format. It can be removed once the aforementioned API functions are updated in every backend so that they always return a 2d/3d-array with `shape==(n_modes, [batch_size,] shots)`. As a result some backend tests may also need to be updated.

Finally, the TF backend can also return unevaluated `tf.Tensors` with `shape==(batch_size, shots)`, one per measured mode. They could be returned in a list, only slightly bending the return format, and everything should work.

tests/integration/test_engine_integration.py
smite: Added in the measure_* backend tests.

smite:
Now only the TFBackend `measure_*` methods are missing multishot capability.

josh146:
Closing for now, as this has been made outdated after #300 was merged

PR #119, opened 2019-07-05
Fixes the bug found by Tom in mean photon per mode

nquesada:
Fixes the bug in decompositions/graph_embed and adds a test that catches it (test_embed_graph)
**Benefits:**
One less bug one more tests
**Possible Drawbacks:**
~~The test takes about 6 minutes on my pc, because a lot of samples (O(10000)) are needed to catch it~~
**Related GitHub Issues:**

### DIFF
--- a/tests/integration/test_measurement_integration.py
+++ b/tests/integration/test_measurement_integration.py
@@ -104,7 +104,7 @@ class TestPostselection:
                 ops.Fock(n) | q[0]
                 ops.Fock(total_photons - n) | q[1]
                 ops.BSgate() | q
-                ops.MeasureFock(select=n//2) | q[0]
+                ops.MeasureFock(select=n // 2) | q[0]
                 ops.MeasureFock() | q[1]
 
             eng.run(prog)  # FIXME measurements above commute, but they should not since the postselection may fail if the other one is performed first!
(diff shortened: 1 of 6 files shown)


co9olguy:
Thanks @nquesada for a quick fix

co9olguy:
@nquesada new test seems to be failing. What is it testing for?

tests/integration/test_measurement_integration.py
```diff
@@ -114,3 +114,25 @@ def test_measure_fock(self, setup_eng, cutoff, batch_size):
...
+    @pytest.mark.fixture('pure')
+    def test_embed_graph(self, setup_eng, cutoff, batch_size):
+        """Test that a graph embeded with the right photon number gives the right statistics. """
+        n_samples = 10000
+        eng, prog = setup_eng(2)
+        A = np.array([[0.0, 1.0], [1.0, 0.0]])
+        n_mean_per_mode = 1
+        with prog.context as q:
+            ops.GraphEmbed(A, n_mean_per_mode) | (q[0], q[1])
+            ops.MeasureFock() | (q[0], q[1])
```
co9olguy: If the goal here is just to test if `GraphEmbed` has the correct mean photon numbers, can we not look at the mean photon numbers directly? e.g., 
```
state = eng.run(prog).state
cov = state.cov()
mean_photon_numbers = some_function(state)
```

nquesada: yes. That should work much better since it is not stochastic.

nquesada: The latest commit should do exactly what you requested.

co9olguy:
:thinking: the tests that are failing on travis seem to be passing fine on my local branch

PR #121, opened 2019-07-07
[WIP] Renaming DeviceSpecs and refactoring state_options

co9olguy:

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


co9olguy:
* [ ]  put the `modes` argument from `eng.run` into `state_options` dictionary

* [ ]  rename `state_options` dictionary to `result_options`, `return_options`, or `run_options` (last one is my favourite, since TFBackend actually needs these options during execution)

* [ ]  rename `DeviceSpecs` to `CircuitSpecs`

co9olguy:
`DeviceSpecs` name change done.

Any help on other two items would be appreciated, if possible

josh146:
Since we are moving away from having a 'spec' per backend, towards instead having a `circuitspec` that can be used by more than one backend (e.g., `GBSSpec`), the `TFSpec` is probably now redundant, and it can simply use the `FockSpec` class?

smite:
> Since we are moving away from having a 'spec' per backend, towards instead having a `circuitspec` that can be used by more than one backend (e.g., `GBSSpec`), the `TFSpec` is probably now redundant, and it can simply use the `FockSpec` class?

In principle yes, but currently the spec corresponding to a backend is identified using the backend's short_name.

smite:
Putting the `modes` argument to `state_options` is a bit more complicated than I first thought.

* User wants no return state? Now: `eng.run(prog, modes=[])`
* User wants a return state with all modes? Now: `eng.run(prog)`
* User wants a return state with some modes? Now: `eng.run(prog, modes=[3,5,8])`

How should these cases look in the new system?

josh146:
> Putting the `modes` argument to `state_options` is a bit more complicated than I first thought.
> 
>* User wants no return state? Now: `eng.run(prog, modes=[])`
>* User wants a return state with all modes? Now: `eng.run(prog)` 
>* User wants a return state with some modes? Now: `eng.run(prog, modes=[3,5,8])`
>
> How should these cases look in the new system?

This currently applies to the returned `state`, will this also apply to `Result.samples`? It probably should for consistency. In this case, `results_options` would need to be passed to the `Results` object, rather than directly to `State`?

co9olguy:
> Since we are moving away from having a 'spec' per backend, towards instead having a `circuitspec` that can be used by more than one backend (e.g., `GBSSpec`), the `TFSpec` is probably now redundant, and it can simply use the `FockSpec` class?

At the moment, as far as circuitspecs is concerned, Fock and TF Backends are same class. I suppose we could also have additional attributes that apply only on TF Backend (e.g., `differentiable`), but not urgent to distinguish at the moment.

co9olguy:
> Putting the `modes` argument to `state_options` is a bit more complicated than I first thought.
> 
>     * User wants no return state? Now: `eng.run(prog, modes=[])`
> 
>     * User wants a return state with all modes? Now: `eng.run(prog)`
> 
>     * User wants a return state with some modes? Now: `eng.run(prog, modes=[3,5,8])`
> 
> 
> How should these cases look in the new system?

:laughing: I tried to tackle last night and came to the same conclusion, that it is not *quite* as straightforward as expected.

Regarding the options for the new `state_options`. It makes have a blank dictionary as default. I would have "return state with all modes" as the default when nothing is specified (as it is already). Alternatively, the user can pass a list:
```
eng.run(prog, state_options={"modes": []}  # return no modes. We could also replace this with `None` since `None`
eng.run(prog, state_options={"modes": [3,5,8]}  # return subset
```

doc/glossary.rst
```diff
@@ -21,21 +22,22 @@ Expressions that are grouped together are used synonymously in the Strawberry Fi
...
+      The process of converting a `source` quantum :term:`program` into an :term:`equivalent program`
+      that belongs in the `target` :term:`circuit class`.
+      The compilation can fail if the program cannot be made equivalent to the target circuit class.
 
    equivalent circuit
    equivalent program
       Two quantum :term:`circuits <circuit>` are equivalent iff they produce the same output
-      probability distributions. Equivalent circuits are necessarily in the same :term:`circuit class`.
+      probability distributions. Two equivalent circuits are not necessarily in the same
+      :term:`circuit class`.
```
co9olguy: good catch

strawberryfields/circuitspecs/__init__.py
```diff
@@ -68,7 +85,7 @@
 
 specs = (BaseSpecs, Chip0Specs, FockSpecs, GaussianSpecs, GBSSpecs, TFSpecs)
 
-backend_specs = {c.short_name: c for c in specs}
-"""dict[str, CircuitSpecs]: dictionary mapping circuit family short_name to the corresponding class."""
+circuit_db = {c.short_name: c for c in specs}
```
co9olguy: Should we add to autosummary?

smite:
> > Putting the `modes` argument to `state_options` is a bit more complicated than I first thought.
> 
> This currently applies to the returned `state`, will this also apply to `Result.samples`? It probably should for consistency. In this case, `results_options` would need to be passed to the `Results` object, rather than directly to `State`?

I would not have `modes` affect `Result.samples`. Unlike returning the state, returning all the samples is easy and cheap, and it's also easy for the user to only pick samples from the modes they're interested in.

smite:
Now `LocalEngine.run()` takes three keyword arg dictionaries: compile_options (consumed by `Program.compile()`), state_options (consumed by `backend.state()`) and run_options (consumed by other backend API methods that are called via `Operation.apply()`).

I made it so that the contents of run_options are automatically updated into state_options (overwriting similarly named kwargs if they exist!), to make it easier to use session and feed_dict with the TF backend. We can change this behavior if it's not good.

strawberryfields/circuitspecs/__init__.py
smite: Added

co9olguy:
After debugging failed tests, realized that passing an empty dictionary as a default argument is a bad idea (since these are created once and are stateful, rather than being created on call). 

Everything should pass now, but I had to change the behaviour so that `state_options={"return": False}` is the way to prevent state objects being returned.

strawberryfields/engine.py
```diff
@@ -337,33 +340,50 @@ def _run_program(self, prog, **kwargs):
...
             shots (int): number of times the program measurement evaluation is repeated
-            compile_options (Dict[str, Any]): keyword arguments for :meth:`.Program.compile`
-            modes (None, Sequence[int]): Modes to be returned in the ``Result.state`` :class:`.BaseState` object.
-                An empty sequence [] means no state object is returned. None returns all the modes.
-            state_options (Dict[str, Any]): keyword arguments for :meth:`.BaseBackend.state`
-
-        The ``kwargs`` keyword arguments are passed to :meth:`_run_program`.
+            compile_options (None, Dict[str, Any]): keyword arguments for :meth:`.Program.compile`
+            run_options (None, Dict[str, Any]): keyword arguments passed to the backend API calls via :meth:`Operation.apply`
+            state_options (None, Dict[str, Any]): Keyword arguments for :meth:`.BaseBackend.state`.
```
co9olguy:
```suggestion
            state_options (None, Dict[str, Any]): keyword arguments for :meth:`.BaseBackend.state`;
```

strawberryfields/engine.py
```diff
@@ -337,33 +340,50 @@ def _run_program(self, prog, **kwargs):
...
+                ``None`` means no state object is returned in :attr:`Result.state`.
```
co9olguy:
```suggestion
                passing ``{'results': None}`` means no state object is returned in :attr:`Result.state`.
```

strawberryfields/engine.py
```diff
@@ -337,33 +340,50 @@ def _run_program(self, prog, **kwargs):
...
-        if isinstance(modes, Sequence) and not modes:
-            # empty sequence
-            pass
-        else:
-            result._state = self.backend.state(modes, **state_options)  # tfbackend.state can use kwargs
+        ``run_options`` can contain the following:
+
+        Keyword Args:
+            eval (bool): TF backend only. If False, the backend returns unevaluated tf.Tensors instead of
+                arrays as measurement results. (default: True)
```
co9olguy:
```suggestion
                numbers/arrays as measurement results (default: True). TF backend only. 
```

strawberryfields/engine.py
```diff
@@ -337,33 +340,50 @@ def _run_program(self, prog, **kwargs):
...
-        result = super()._run(program, shots=shots, compile_options=compile_options, **kwargs)
-        if isinstance(modes, Sequence) and not modes:
-            # empty sequence
-            pass
-        else:
-            result._state = self.backend.state(modes, **state_options)  # tfbackend.state can use kwargs
+        ``run_options`` can contain the following:
+
+        Keyword Args:
+            eval (bool): TF backend only. If False, the backend returns unevaluated tf.Tensors instead of
```
co9olguy:
```suggestion
            eval (bool): If False, the backend returns unevaluated tf.Tensors instead of
```

strawberryfields/engine.py
```diff
@@ -337,33 +340,50 @@ def _run_program(self, prog, **kwargs):
...
+                arrays as measurement results. (default: True)
+            session (tf.Session): TensorFlow session. TF backend only.
+            feed_dict (dict[str, Any]): TF backend only.
```
co9olguy:
```suggestion
            feed_dict (dict[str, Any]): TensorFlow feed dictionary. TF backend only.
```

.github/CHANGELOG.md
```diff
@@ -7,13 +7,17 @@
 - The Engine API has been changed slightly:
     - `LocalEngine.run()` returns a `Result` object that contains both a state object and measurement samples.
     - The way kwargs were used has been simplified by introducing the new kwargs-like arguments `backend_options` and `state_options`. `LocalEngine.run()` passes the actual kwargs only to `Operation.apply()`.
```
co9olguy:
```suggestion
    - The way kwargs were used has been simplified by introducing the new kwargs-like arguments:
        - `backend_options` are provided upon initialization of the `LocalEngine`, and passed on to the corresponding backend when it is created.
        - `compile_options` can be provided when calling `LocalEngine.run()`. These are passed to the `compile()` method of the program before execution.
        - `run_options` can be provided when calling `LocalEngine.run()`. These are passed to the backend API calls via :meth:`Operation.apply`
        - `state_options` can be provided when calling `LocalEngine.run()`. These are used to determine the characteristic of the `Results` object returned after the program is finished executing.
```

.github/CHANGELOG.md
```diff
@@ -6,14 +6,22 @@
...
-- Added the `sf.io` module, which is used to save/load standalone Blackbird scripts from/into Strawberry Fields. Note that the Blackbird DSL has been spun off as an independent package and is now a dependency of Strawberry Fields.
-- Added a new decomposition `mach_zehnder` to the decompositions module. 
+    - The way kwargs were used has been simplified by introducing the new kwargs-like arguments:
+        - `backend_options` are provided upon initialization of the `LocalEngine`, and passed on to the corresponding backend when it is created.
+        - `compile_options` can be provided when calling `LocalEngine.run()`. These are passed to the `compile()` method of the program before execution.
+        - `run_options` can be provided when calling `LocalEngine.run()`. These are passed to the backend API calls via :meth:`Operation.apply`
+        - `state_options` can be provided when calling `LocalEngine.run()`. These are used to determine the characteristic of the `Results` object returned after the program is finished executing.
```
co9olguy: Feeling like `state_options` should be combined with `run_options`. The only things that `run_options` accepts are various TF parameters, which are needed in order to determine what types to return for measurement results and final states.

josh146: I support all of the above :+1: So this will result in two dicts, `compile_options` and `run_options`? I'm in favour of calling it `run_options`, simply because some of the keys are used during the engine run.

josh146:
> Edit: We could add the circuit_spec attribute to BaseBackend, which is either None or a str containing the short_name of the CircuitSpec to use when compiling for that backend.

I'm in favour of this approach, as it decouples the `CircuitSpecs` from the backends, and allows more flexibility (and also removes the need for redundant spec classes such as `TFSpecs`).

.github/CHANGELOG.md
co9olguy: Done in 60221ad

strawberryfields/engine.py
```diff
@@ -337,33 +337,51 @@ def _run_program(self, prog, **kwargs):
...
+        compile_options = compile_options or {}
+        run_options = run_options or {}
+        run_options.setdefault("shots", 1)
+        run_options.setdefault('modes', None)
+
+        # avoid unexpected keys being sent to Operations
+        eng_run_keys = ["eval", "session", "feed_dict", "shots"]
+        eng_run_options = {key: run_options[key] for key in run_options.keys() & eng_run_keys}
+
+        result = super()._run(program, compile_options=compile_options, **eng_run_options)
```
josh146: This might need a future rethink --- this will preclude future backends from allowing their own run options, without also needing to modify the frontend to whitelist it here.

Ideally, the idea behind a run_options dict rather than frontend defined kwargs is to support the behavior where run options can be defined by the backend.

However, since at the moment `_run` accepts `**kwargs`, this would be a significant change/refactor.

strawberryfields/ops.py
```diff
@@ -738,13 +730,15 @@ def apply(self, reg, backend, **kwargs):
             return
         if self.dagger:
             z = -z
-        temp = self.p  # store the original Parameters
+        original_p = self.p  # store the original Parameters
```
josh146: thumb

tests/frontend/test_about.py
```diff
@@ -30,6 +30,16 @@ def test_about(capfd):
...
     assert len(out) > 300
 
+    assert "Strawberry Fields version" in out
+    sf_version_match = re.search(r"Strawberry Fields version:\s+([\S]+)\n", out).group(1)
+    assert sf_version_match == sf.version()
+
+    assert "Numpy version" in out
+    assert "Scipy version" in out
+    assert "Hafnian version" in out
+    assert "Blackbird version" in out
```
josh146: Nice, this is a much better test now

.github/CHANGELOG.md
```diff
@@ -6,14 +6,21 @@
...
-- Added a new decomposition `mach_zehnder` to the decompositions module. 
+    - The way kwargs were used has been simplified by introducing the kwargs-like arguments:
+        - `backend_options` are provided upon initialization of the `LocalEngine`, and passed on to the corresponding backend when it is created.
+        - `compile_options` can be provided when calling `LocalEngine.run()`. These are passed to the `compile()` method of the program before execution.
+        - `run_options` can be provided when calling `LocalEngine.run()`. These are used to determine the characteristics of the measurements and state contained in the `Results` object returned after the program is finished executing.
+ - The Gaussian backend now officially supports Fock-basis measurements (`MeasureFock`/`Measure`/
+  `measure_fock`), but does not update the quantum state after a Fock measurement.
+- `shots` keyword argument added to `Engine.run()`, enabling multi-shot sampling. Supported only
```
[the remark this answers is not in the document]
josh146: I suppose it still _technically_ is.

.github/CHANGELOG.md
```diff
@@ -6,14 +6,21 @@
...
-- Added the `sf.io` module, which is used to save/load standalone Blackbird scripts from/into Strawberry Fields. Note that the Blackbird DSL has been spun off as an independent package and is now a dependency of Strawberry Fields.
-- Added a new decomposition `mach_zehnder` to the decompositions module. 
+    - The way kwargs were used has been simplified by introducing the kwargs-like arguments:
+        - `backend_options` are provided upon initialization of the `LocalEngine`, and passed on to the corresponding backend when it is created.
+        - `compile_options` can be provided when calling `LocalEngine.run()`. These are passed to the `compile()` method of the program before execution.
+        - `run_options` can be provided when calling `LocalEngine.run()`. These are used to determine the characteristics of the measurements and state contained in the `Results` object returned after the program is finished executing.
+ - The Gaussian backend now officially supports Fock-basis measurements (`MeasureFock`/`Measure`/
```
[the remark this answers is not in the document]
smite: I agree, MeasureFock isn't too long to type.

PR #124, opened 2019-07-11
Remove `ModuleNotFoundError` from the codebase and update chip0 spec

josh146:

* Replaces all instances of `ModuleNotFoundError` with `ImportError`.

* Updates the `Chip0` template to use `MeasureFock() | [0, 1, 2, 3]`, which will allow correct fock measurement behaviour when simulated on the Gaussian backend. Also updates the gate topology of beamsplitters and phase shifters.

* Incremented the version number to 0.11.1, allowing for a small bugfix release once merged.

* Updated the changelog, brought the v0.11.0 entry up to date with the GitHub release notes.

**Benefits:**

* `ModuleNotFoundError` is a subclass of `ImportError`, that was introduced with Python 3.6+. So this should have no functionality change. However, this will now allow SF to be importable on Python 3.5; required since TF==1.3 is only supported on Python 3.5 on Windows.

**Possible Drawbacks:** None.

**Related GitHub Issues:** n/a/

### DIFF
--- a/strawberryfields/circuitspecs/chip0.py
+++ b/strawberryfields/circuitspecs/chip0.py
@@ -25,35 +25,33 @@ class Chip0Specs(CircuitSpecs):
     remote = True
     local = True
     interactive = True
-
+    
     primitives = {"S2gate", "Interferometer", "MeasureFock", "Rgate", "BSgate"}
-
-    # TODO: update the below to specify the rectangular_symmetric
-    # mapping for the interferometer when #87 is merged
     decompositions = {"Interferometer": {}}
-
-    # TODO: update the below to specify the rectangular_symmetric
-    # mapping for the interferometer when #87 is merged.
-    # The current topology defined below is just for demonstration
+    
     circuit = textwrap.dedent(
         """\
-        name chip0_template
-        version 0.0
-        target chip0 (shots=1)
+        name template_2x2_chip0
+        version 1.0
+        target chip0 (shots=10)
 
-        S2gate({sq0}, 0.0) | [0, 2]
-        S2gate({sq1}, 0.0) | [1, 3]
+        # for n spatial degrees, first n signal modes, then n idler modes, phase zero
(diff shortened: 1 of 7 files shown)


strawberryfields/circuitspecs/chip0.py
```diff
@@ -54,9 +54,6 @@ class Chip0Specs(CircuitSpecs):
         Rgate({phase}) | 2
         Rgate({phase}) | 3
 
-        MeasureFock() | [0]
-        MeasureFock() | [1]
-        MeasureFock() | [2]
-        MeasureFock() | [3]
+        MeasureFock() | [0, 1, 2, 3]
```
smite: Is this just beautification, or did it not work before? `TFBackend.measure_fock()` requires that multiple modes are always given in ascending order, does GaussianBackend too? FockBackend doesn't.

josh146: Without using a single MeasureFock statement, this will return independent samples for each mode, rather than joint statistics

josh146: So this is more an issue with how PNR measurements work in the gaussian backend, I suppose

co9olguy: yep exactly

smite: For `shots==1` the effect should be exactly the same. When `shots>1`, it'd be surprising if the behavior was different. I think this is a bug in the gaussian backend.

An easy fix would be to implement the block we were talking about last week, i.e. no additional operations allowed after a multi-shot measurement.

co9olguy: It's not a bug in Gaussian backend, more of a limitation that we have to accommodate. Gaussian backend doesn't update the conditional state, so subsequent calls to `measure_fock` on remaining wires are sampling from the wrong distributions (i.e., local statistics instead of joint). Effect is independent of number of shots.

Issue #122, opened 2019-07-10
Add the `circuit_spec` attribute to BaseBackend, which is either None or a str containing the `short_name` of the `CircuitSpec` to use when compiling for that backend.


josh146:
> Edit: We could add the circuit_spec attribute to BaseBackend, which is either None or a str containing the short_name of the CircuitSpec to use when compiling for that backend.

I'm in favour of this approach, as it decouples the `CircuitSpecs` from the backends, and allows more flexibility (and also removes the need for redundant spec classes such as `TFSpecs`).

For example, a future high performance Fock backend could simply specify that `circuit_spec = "fock"`, without needing to create a new redundant copy of `FockSpec`.

Perhaps we could even provide a list of allowed circuit specs, along with a default. For example, the Gaussian backend could have

```python
default_circuit_spec = "gaussian"
circuit_specs = ["gaussian", "GBS"]
```

_Originally posted by @josh146 in https://github.com/XanaduAI/strawberryfields/pull/121#issuecomment-509894537_

josh146:
This is related to #108

PR #125, opened 2019-07-11
Decouple backends and CircuitSpecs

smite:

* Added the `circuit_spec` attribute to `BaseBackend`. It's redefined by the child classes, and contains the short name of the CircuitSpecs instance to use to validate Programs for that backend.
* Make `BaseBackend._short_name` a public attribute by removing the initial underscore.

**Benefits:**

Backends and CircuitSpecs are now independent.

**Related GitHub Issues:**

#122


smite:
Should we also remove `circuitspecs/base.py`? I don't think it's needed.

josh146:
> Should we also remove circuitspecs/base.py? I don't think it's needed.

It may only be needed for the test suite to run, but otherwise there isn't need for it.

strawberryfields/backends/__init__.py
```diff
@@ -26,7 +26,7 @@
     "TFBackend",
 ]
 
-supported_backends = {"base": BaseBackend, "gaussian": GaussianBackend, "fock": FockBackend}
+supported_backends = {b.short_name: b for b in (BaseBackend, GaussianBackend, FockBackend)}
```
josh146: thumb

strawberryfields/backends/base.py
```diff
@@ -222,13 +222,16 @@ def add(self, num_modes):
 class BaseBackend:
     """Abstract base class for backends."""
 
+    #: str: short name of the backend
+    short_name = 'base'
+    #: str, None: Short name of the CircuitSpecs class used to validate Programs for this backend. None if no validation is required.
+    circuit_spec = None
```
josh146: Awesome

josh146: With the new `GBSSpec`, the backends now support multiple CircuitSpecs for validation. I suppose `fock` and `gaussian` are the _backend specific_ specifications - even if the user compiles againsts `GBSSpec`, it will still need to be validated by the backend against `Backend.circuit_spec` to ensure it can be correctly executed.

strawberryfields/engine.py
```diff
@@ -270,8 +270,9 @@ def _broadcast_nones(val, dim):
                     p.reg_refs[k].val = v
 
             # if the program hasn't been compiled for this backend, do it now
-            if p.target != self.backend_name:
-                p = p.compile(self.backend_name, **compile_options) # TODO: shots might be relevant for compilation?
+            target = self.backend.circuit_spec
+            if target is not None and p.target != target:
+                p = p.compile(target, **compile_options)
```
josh146: Do we want to make `Backend.circuit_spec` an abstract property that _must_ be overwritten? Or is there a use case for having backends with `Backend.circuit_spec = None`?

josh146:
@smite, don't forget to add a line to the `.github/CHANGELOG.md` file (probably under `### API changes`), summarizing the change, with a link back to this PR.

strawberryfields/engine.py
smite: `BaseBackend` :)

It's not an actual abc.ABC right now, so it can be instantiated.

josh146: Oh, of course 🙂

strawberryfields/backends/__init__.py
co9olguy: I guess TFBackend is not here because TF is no longer a requirement?

strawberryfields/engine.py
```diff
@@ -270,8 +270,9 @@ def _broadcast_nones(val, dim):
                     p.reg_refs[k].val = v
 
             # if the program hasn't been compiled for this backend, do it now
-            if p.target != self.backend_name:
-                p = p.compile(self.backend_name, **compile_options) # TODO: shots might be relevant for compilation?
+            target = self.backend.circuit_spec
+            if target is not None and p.target != target:
```
co9olguy: is it possible that `p.target` could be set to `target`, but `p` has not actually (yet) been compiled for that target? I'm thinking if someone has blackbird code loaded from an `xbb` file with `target=gaussian` (or whatever), that information will propagate to `p.target`. If the engine is attached to Gaussian backend, then the above line will effectively skip compilation. 

In this case, is it possible that the `.xbb` file will contain incompatible ops (e.g., fock state preparation), but it never gets validated?

co9olguy: Looks like compilation is carried out during `sf.io.load()`, so that is ok.

But we could still have that a user does the following:
```
prog = sf.Program(1)
with prog.context as q:
    SomeUnsupportedOp() | q

eng = sf.Engine("gaussian")
prog.target = "gaussian"
eng.run(prog)
```

co9olguy: but I think that any user that knew enough to switch `prog.target` like this should not be confused if they receive an error of some kind from `eng.run()` (just not a compilation error)

strawberryfields/backends/__init__.py
smite: Yeah, `load_backend()` takes care of `TFBackend` by importing the `tfbackend` submodule and tensorflow on-demand.

PR #126, opened 2019-07-11
Engine.run: return_state removed.

smite:

* Removed the `LocalEngine.run()` argument `return_state`. Now a state object is returned iff modes is not an empty sequence.

**Benefits:**

One less argument. I don't see any uses for state objects with no modes.

### DIFF
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -356,9 +356,7 @@ class LocalEngine(BaseEngine):
         Keyword Args:
             shots (int): number of times the program measurement evaluation is repeated
             modes (None, Sequence[int]): Modes to be returned in the ``Result.state`` :class:`.BaseState` object.
-                ``None`` returns all the modes (default).
-            return_state (bool): Whether to return a :class:`.BaseState` object object within ``Result.state``.
-                Default is True.
+                ``None`` returns all the modes (default). An empty sequence means no state object is returned.
             eval (bool): If False, the backend returns unevaluated tf.Tensors instead of
                 numbers/arrays for returned measurement results and state (default: True). TF backend only.
             session (tf.Session): TensorFlow session, used when evaluating returned measurement results and state.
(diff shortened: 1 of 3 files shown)


strawberryfields/engine.py
```diff
@@ -377,7 +375,8 @@ def run(self, program, *, compile_options=None, run_options=None):
```
smite: Is the `run_options` filtering above necessary? Currently the `Operation` subclasses do not use any kwargs, they're all passed through to the backend API functions. I think it'd be easier to agree that `Operations` and backend API funcs have a shared kwarg namespace.

josh146: I agree, it would make it cleaner and easier to maintain without filtering.

Furthermore, this way backends could introduce new run options, without having to 'whitelist' them here.

Although, it looks like the above might have been added to avoid failing tests, @co9olguy might be able to fill us in with more details.

josh146:
@smite, don't forget to add a line to the `.github/CHANGELOG.md` file, summarizing the change, with a link back to this PR.

strawberryfields/engine.py
co9olguy: Yes, without dict comprehension filtering, tests were failing because the `Operations` were complaining about receiving unexpected keyword arguments.

co9olguy: I tend to agree that, since everything at this stage is not user-facing, then it is probably fine to include `**kwargs` in the signatures for the Operations class rather than having to whitelist everything

co9olguy:
> **Description of the Change:**
> 
> * Removed the `LocalEngine.run()` argument `return_state`. Now a state object is returned iff modes is not an empty sequence.
> 
> **Benefits:**
> 
> One less argument. I don't see any uses for state objects with no modes.

One place where it might be useful have an option "please don't return a state object" is with remote simulators. The user might not care about the state object, but just the measurement results. In that case, explicitly passing `modes=[]` rather than `return_state=False` might be less intuitive. But I think that we don't need to design around that use-case yet, and when we have it implemented, we can find a clean way to do it by default.

PR #128, opened 2019-07-12
Update ops.py

nquesada:
The previous version was checking that all the elements of the interferometer were different from the identity. What you want to change is that it is *not* the identity, which implies checking that there is at least one element that is different, this can be done more elegantly by using np.allclose .

Fixes bug in ops GraphEmbed
**Benefits:**
Kills a bug
**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1698,7 +1698,7 @@ class GraphEmbed(Decomposition):
                 if np.abs(s) >= _decomposition_tol:
                     cmds.append(Command(Sgate(s), reg[n]))
 
-            if not np.allclose(self.U, np.identity(len(self.U)), atol = _decomposition_tol):
+            if not np.allclose(self.U, np.identity(len(self.U)), atol=_decomposition_tol):
                 cmds.append(Command(Interferometer(self.U), reg))
         return cmds
 
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1698,7 +1698,7 @@ class GraphEmbed(Decomposition):
                 if np.abs(s) >= _decomposition_tol:
                     cmds.append(Command(Sgate(s), reg[n]))
 
-            if not np.allclose(self.U,  np.identity(len(self.U)), atol = _decomposition_tol):
+            if not np.allclose(self.U, np.identity(len(self.U)), atol = _decomposition_tol):
                 cmds.append(Command(Interferometer(self.U), reg))
         return cmds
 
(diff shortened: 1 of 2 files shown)


strawberryfields/ops.py
```diff
@@ -1698,7 +1698,7 @@ def _decompose(self, reg):
                 if np.abs(s) >= _decomposition_tol:
                     cmds.append(Command(Sgate(s), reg[n]))
 
-            if np.all(np.abs(self.U - np.identity(len(self.U))) >= _decomposition_tol):
+            if not np.allclose(self.U,  np.identity(len(self.U)), atol = _decomposition_tol):
```
smite: Oops :)
I wonder if there's some metric for matrices that's more justified than others from a theoretical POV here?
Maybe some norm of the difference, `|A-B|`?
`np.allclose(A,B)` is almost but not quite equivalent to the elementwise max norm `|A-B|_max` (rtol makes it a bit different when B is not the zero matrix).

nquesada: I suspect that is fine since the second matrix is guaranteed to not be the zero matrix.

strawberryfields/ops.py
```diff
@@ -1698,7 +1698,7 @@ def _decompose(self, reg):
                 if np.abs(s) >= _decomposition_tol:
                     cmds.append(Command(Sgate(s), reg[n]))
 
-            if np.all(np.abs(self.U - np.identity(len(self.U))) >= _decomposition_tol):
+            if not np.allclose(self.U, np.identity(len(self.U)), atol=_decomposition_tol):
```
josh146: Should we add `rtol=0` here as well, so that it is an identical check to the previous version?

tests/frontend/test_ops_decompositions.py
```diff
@@ -275,6 +275,30 @@ def test_decomposition(self, hbar, tol):
...
+        [1, 1, 1, 1, 0, 1],
+        [1, 1, 0, 0, 1, 0],
+        ]
+        )
+        sq, U = dec.graph_embed(A)
+        assert not np.allclose(U, np.identity(n))
+        G = ops.GraphEmbed(A)
+        cmds = G.decompose(prog.register)
+
+        assert str(cmds[-1].op)[:4] == "Interferometer"[:4]
```
co9olguy: This assert didn't make much sense to me (checked that the first four letters of the string-converted op were "Inte"), so I updated it to do something which makes the intention more clear

Issue #129, opened 2019-07-12
Shots option from blackbird script is lost during `io.to_program()`


co9olguy:
When loading a blackbird script from file and converting it to a program, `to_program` uses the `target` attribute of the `BlackbirdProgram` object during a compilation step, but seems to ignore other options. Unfortunately, this is where the `shots` parameter lives, thus this information never makes it to the final program object. During execution with `StarshipEngine`, it is replaced by the default `shots=1`.

josh146:
On it.

josh146:
I've looked through the code, what is appears to be happening is this:

1. `to_program` correctly passes all target options as `**kwargs` to `program.compile`.

2. Unfortunately, `program.compile` ignores all `**kwargs` options not to do with compilation, so at this point `shots` gets lost.

What I can do is modify `Program` to store a list of `run_options` for the compiled program, which can then be accessed by the `Engine`.

This will work, I guess the question is, is this a good idea? Do we want the `Program` to be able to specify `run_options` for later execution?

I remember this came up previously, and we decided we did not want this option.

josh146:
PR #130 should solve this issue, and now treats `Program.run_options` as the _default_ run options when `eng.run()` is called, but still allows them to be overwritten.

josh146:
Note: the equivalent change in `LocalEngine.run()` will also need to be made in `StarshipEngine.run()`.

PR #130, opened 2019-07-13
Added target options pass-through

josh146:
**Context:** Currently, target options defined in a loaded Blackbird script are passed to `program.compile`, where they are ignored.

* `Program` now contains the attribute `Program.run_options`, storing any default run options for a program. During compilation, target options from the blackbird scripts are now stored in this attribute.

* During `eng.run`, if no run options are provided as a keyword argument, the engine will fall back on the run options stored within the program.

* Otherwise, if run options are provided by the user, they 'overwrite' the program's default run options.

**Benefits:**

* Blackbird `target (shots=100)` directives are now stored in the compiled SF Program, and used as the default run options in `eng.run()`. The user may still overwrite these default by passing in `eng.run(prog, run_options={"shots": 24})`.

**Possible Drawbacks:** None that I can think of.

**Related GitHub Issues:** #129


strawberryfields/program.py
```diff
@@ -552,6 +559,12 @@ def compile_sequence(seq):
         compiled = self._linked_copy()
         compiled.circuit = seq
         compiled.target = target
+
+        # get run options of compiled program
+        # for the moment, shots is the only supported run option.
+        if "shots" in kwargs:
+            compiled.run_options["shots"] = kwargs["shots"]
```
josh146: any others we should add here?

tests/frontend/test_io.py
```diff
@@ -453,6 +459,65 @@ def test_invalid_compilation(self):
...
+    """A dummy run function, that when called returns a dummy
+    results object, with run options available as an attribute.
+    This allows run_options to be returned and inspected after calling eng.run
+    """
+    results = DummyResults()
+    results.run_options = eng_run_options
+    return results
+
+
+class TestEngineIntegration:
```
co9olguy: I guess this is a "frontend integration" test?

strawberryfields/engine.py
```diff
@@ -367,21 +367,34 @@ def run(self, program, *, compile_options=None, run_options=None):
                 and state. TF backend only.
         """
         compile_options = compile_options or {}
-        run_options = run_options or {}
-        run_options.setdefault("shots", 1)
-        run_options.setdefault('modes', None)
+        temp_run_options = {}
+
+        if isinstance(program, Sequence):
+            # succesively update all run option defaults.
```
co9olguy: Feels like this is something that warrants a dedicated test

strawberryfields/program.py
co9olguy: for now, probably not

tests/frontend/test_io.py
josh146: I guess so. Frontend integration in that it is testing multiple parts of the frontend (io.load, Program, eng.run()). But everything is still contained within the frontend.

strawberryfields/engine.py
josh146: Will add.

josh146: Added in 8fe10dd2eb61bfb02b4c20942f347e8aa3493ca6

PR #131, opened 2019-07-13
remove name

josh146:
**Context:** Currently, ``name`` is set as a required positional argument of `StarshipEngine.run()`. However, this is not needed, as all loaded blackbird programs, as well as all SF programs, already have names.

* Removes the `name` positional argument from `StarshipEngine`

* Leaves a TODO comment, as a reminder to modify `StarshipEngine` to support shots pass-through once #130 is merged into master.

**Benefits:** name no longer needs to be passed to `eng.run()`, this is extracted automatically from the `Program.name` attribute.

**Possible Drawbacks:** n/a

### DIFF
--- a/tests/frontend/test_engine.py
+++ b/tests/frontend/test_engine.py
@@ -201,10 +201,9 @@ class TestStarshipEngine:
 
         monkeypatch.setattr("strawberryfields.engine.to_blackbird", methods.to_blackbird)
 
-        output = starship_engine._get_blackbird(inputs.name, inputs.shots, inputs.program)
+        output = starship_engine._get_blackbird(inputs.shots, inputs.program)
 
         methods.to_blackbird.assert_called_once_with(inputs.program, version="1.0")
-        assert output._name == inputs.name
         assert len(output._target.__setitem__.call_args_list) == 2
         assert output._target.__setitem__.call_args_list[0] == call(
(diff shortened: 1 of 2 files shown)


tests/frontend/test_engine.py
```diff
@@ -347,21 +346,18 @@ def test_engine_with_mocked_api_client_sample_job(self, monkeypatch):
...
+            ops.BSgate() | (q[0], q[1])
+            ops.Rgate(3) | q[0]
+            ops.BSgate(3) | (q[0], q[1])
+            ops.Rgate(3) | q[2]
+            ops.BSgate() | (q[2], q[3])
+            ops.Rgate(3) | q[2]
+            ops.BSgate() | (q[2], q[3])
+            ops.MeasureFock() | q
+
+        engine.run(prog)
```
co9olguy: Not sure why this was passing before the UI change, but not after :thinking: 
In any case, it has been updated to have correct topology now.

co9olguy:
For simplicity, even though tests are currently failing, merging into `demo` branch (merge commit has least failures). Fix will be applied in base branch

josh146:
It looks like the decoupling circuit specs from backend PR (#125) broke the StarshipEngine once it was merged in. Probably good to fix this separately in `demo`.

PR #136, opened 2019-07-15
Configuration build fix

zzzeid:
Added `_dev` to packages in `setup.py`. This is so that the configuration import does not fail when using a built package.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


zzzeid:
> For the `demo` branch, it makes more sense to move `configuration` back to top level and out of `_dev`, right? Having it in `_dev` was just a temporary move for the 0.11.0 release

Done!

PR #137, opened 2019-07-15
Fix field value reference bug

zzzeid:

The result and circuit fields were being initialized with the wrong variable. This bug fix corrects this issue.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
Oh I suppose CodeFactor is getting confused because the PR was originally against master.

PR #141, opened 2019-07-23
Configuration Fixes

zzzeid:

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/configuration.py
```diff
@@ -34,7 +34,7 @@
 
    * On Linux: ``~/.config/strawberryfields``
    * On Windows: ``~C:\Users\USERNAME\AppData\Local\Xanadu\strawberryfields``
-   * On MacOS: ``~/Library/Preferences/strawberryfields``
+   * On MacOS: ``~/Library/Application\ Support/strawberryfields``
```
josh146: Thanks for fixing this @zzzeid :+1:

strawberryfields/configuration.py
```diff
@@ -160,7 +160,7 @@ def update_config(self):
                 if env in os.environ:
                     # Update from environment variable
                     self._config[section][key] = os.environ[env]
-                elif self._config_file:
+                elif self._config_file and key in self._config_file[section]:
```
co9olguy: this feels like something that should lead to a new test maybe?

strawberryfields/api_client.py
```diff
@@ -141,37 +141,36 @@ class APIClient:
...
+        "localhost:8080",
+        "platform.strawberryfields.ai",
+    ]
 
     DEFAULT_HOSTNAME = "localhost"
 
-    ENV_KEY_PREFIX = "SF_API_"
-    ENV_AUTHENTICATION_TOKEN_KEY = "{}AUTHENTICATION_TOKEN".format(ENV_KEY_PREFIX)
-    ENV_API_HOSTNAME_KEY = "{}API_HOSTNAME".format(ENV_KEY_PREFIX)
-    ENV_USE_SSL_KEY = "{}USE_SSL".format(ENV_KEY_PREFIX)
```
co9olguy: Were these not used anywhere else?

strawberryfields/configuration.py
zzzeid: Yes - this line is to fix the `KeyError` issue that we got when we excluded a parameter from the config file (i.e. the file was behaving as though all parameters were required). I can add a test for this.

strawberryfields/api_client.py
zzzeid: No they are not. These were left behind after a previous refactor.

strawberryfields/configuration.py
zzzeid: Test added.

PR #142, opened 2019-07-23
APIClient fixes

zzzeid:

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/api_client.py
```diff
@@ -154,6 +150,7 @@ def __init__(self, **kwargs):
         Initialize the API client with various parameters.
         """
         self._config = self.get_configuration_from_config()
+        self._config["debug"] = self._config["debug"] in (True, "True", "true", "TRUE", 1)
```
josh146: How come the casting of `self._config["debug"]` happens here? Perhaps this check should go in the `Configuration` class, so that it is applied to all boolean configuration options from hereon

strawberryfields/api_client.py
```diff
@@ -216,6 +218,35 @@ def join_path(self, path):
...
+        ``self.responses`` if a response is returned from the server.
+
+        Args:
+            method: one of ``requests.get`` or ``requests.post``
+            **params: the parameters to pass on to the method (e.g. ``url``, ``data``, etc.)
+
+        Returns:
+            requests.Response: a response object, or None if no response could be fetched
+        """
+        assert method in (requests.get, requests.post)
```
josh146: Under what circumstances could this assertion fail?

Also, is this assertion only performed in debug mode?

josh146: (similarly to how running Python with the `-O` flag ignores assertions)

strawberryfields/engine.py
```diff
@@ -308,7 +308,8 @@ def _broadcast_nones(val, dim):
             self.run_progs.append(p) 
             prev = p
 
-        return Result(self.samples.copy())
+        if self.samples:
+            return Result(self.samples.copy())
```
josh146: What is the reason behind this change here? The `Result` class stores both the quantum state as well as the samples for the `LocalEngine`, so likely still needs to be returned even if no samples are requested

zzzeid: In the case when a user aborts the engine (e.g. with KeyboardInterrupt and the job goes to the background / is async), `self.samples` is `None` and this fails. I did not recognize the use case you mentioned with LocalEngine (I assume samples in that case could be an empty list?) I can change this to `if self.samples is not None` to resolve this issue.

NB: In the case of a KeyboardInterrupt/async, the current behaviour requires the user to continue or pick up the job manually (i.e. reload/fetch results, and wrap them in a `Result` object) but in the future we can add a helper method to resume the loop.

strawberryfields/api_client.py
zzzeid: This is mostly superfluous, although I suppose that if a developer was using this method and passing something other than `requests.get` or `requests.post`, a more ambiguous error might arise, or this could fail silently and give an unexpected result (i.e. by calling whatever method is passed with `params`, which could return a very unexpected result and still succeed.) It's mostly for future developers. So to answer your question: it would not play a part or be important or significant if we are ignoring assertions.

Though more suitably perhaps this should raise a `TypeError`, so I will change.

strawberryfields/api_client.py
zzzeid: @josh146 Good idea. I did initially put this in `Configuration` but it seemed a little overkill -- however I can see how this would be helpful in the future for any boolean configuration option. Will change.

zzzeid: Fixed.

strawberryfields/engine.py
josh146: With `LocalEngine`, samples will always be a list of size the `(shots, num_modes)` (iirc) if no shots are requested. Each value in the list will simply be `None`.

This line creates the `Result` object (even if the samples are all `None`), and later on the `State` object is attached to `result.state`. See https://github.com/XanaduAI/strawberryfields/blob/14f91bf5f2e0dc60b2a431c23c8edf6ce5dd1850/strawberryfields/engine.py#L397.

My understanding is that this line will raise an exception if `result` is now `None`.

strawberryfields/configuration.py
```diff
@@ -95,7 +95,26 @@
...
+
+    if key in BOOLEAN_KEYS:
+        if value in trues:
+            return True
+        elif value in falses:
+            return False
+        else:
+            raise ValueError("Boolean could not be parsed")
+    else:
+        return value
```
josh146: thumb

strawberryfields/engine.py
josh146: Ah, this is actually very subtle. Because the simulator backends return a list of the form `[None, None, None]` if there is no sampling, the `if self.samples:` check passes.

I think this is fine for now, but we should leave a comment here explaining exactly what is happening, so that when we return to finish #111 there isn't any confusion/bugs

PR #144, opened 2019-07-24
Add a `sorted()` call to `GBSSpec.compile()`

josh146:
**Context:** During GBS compilation, the call to `networkx.lexicographical_topological_sort()` permutes the returned order of the measured registers, since all permutations are equivalent.

**Description of the Change:** After the topological sort, the measured registers are sorted in ascending order in the compiled program. A test to confirm this behaviour is also added.

**Benefits:** This avoids unexpected results, as programs containing `MeasureFock() | q` previously could compile to `MeasureFock() | (q[0], q[2], q[1], q[3])`.

**Possible Drawbacks:** It would be nicer to preserve the order the user specifies the modes in. This is probably not possible, however, due to the topological sort. Further, the user should probably expect differences like this to occur when compiling.

**Related GitHub Issues:** #127

### DIFF
--- a/tests/frontend/test_program.py
+++ b/tests/frontend/test_program.py
@@ -628,7 +628,7 @@ class TestGBS:
         prog = prog.compile('gbs')
         assert len(prog) == 4
         last_cmd = prog.circuit[-1]
-        assert last_cmd.op.__class__ == ops.MeasureFock
+        assert isinstance(last_cmd.op, ops.MeasureFock)
         assert [x.ind for x in last_cmd.reg] == list(range(3))
 
     def test_GBS_measure_fock_register_order(self):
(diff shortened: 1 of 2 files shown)


tests/frontend/test_program.py
```diff
@@ -630,3 +630,19 @@ def test_GBS_success(self):
...
+
+        with prog.context as q:
+            ops.S2gate(0.45) | (q[0], q[2])
+            ops.S2gate(0.45) | (q[0], q[2])
+            ops.BSgate(0.54, -0.12) | (q[0], q[1])
+            ops.MeasureFock() | (q[0], q[3], q[2], q[1])
+
+        prog = prog.compile("gbs")
+        last_cmd = prog.circuit[-1]
+        assert last_cmd.op.__class__ == ops.MeasureFock
```
co9olguy: how come not `isinstance`? seems to read better with an `assert`

josh146: Good point :thinking: This was just a copy paste from the existing tests. I can change this.

Issue #123, opened 2019-07-10
Deprecate the `Measure` shortcut


josh146:
We should deprecate `Measure`. It doesn't really serve the same purpose as `MeasureX` and `MeasureP` (which are shortcuts to `MeasureHomodyne` with specific values of `phi`), except for typing 4 less characters.

At the same time, it makes the documentation confusing, and goes against the zen of Python:
> There should be one-- and preferably only one --obvious way to do it.

_Originally posted by @josh146 in https://github.com/XanaduAI/strawberryfields/pull/121_

smite:
What's the deprecation policy in Strawberry Fields? Do we use `DeprecationWarning`?

josh146:
We haven't officially deprecated anything in SF yet, but when we deprecated `qml.expval.Observable` in PennyLane, we used `DeprecationWarning`. So let's use it here as well.

PR #145, opened 2019-07-24
Deprecate the Measure shortcut in favour of MeasureFock

josh146:
**Context:**

Currently, we have two ways of specifying a Fock measurement. Instantiating a `MeasureFock()` operation class, or using the pre-initialized `Measure` object (which is simply identical to `MeasureFock()`.

This is confusing, doesn't really serve much purpose, and makes the documentation and code examples more confusing.

Deprecates `Measure` in favour of `MeasureFock()`. Using `ops.Measure` now raises a warning, instructing the user that this shortcut is now deprecated, and instead to use `MeasureFock()`.

Documentation and tests have also been cleaned up to remove reference to `Measure`.

**Benefits:**

Cleaner documentation, tests, and serialization of Blackbird code.

**Possible Drawbacks:**

The line `from strawberryfields.ops import *` will raise the deprecation warning, as the user is importing the `Measure` object implicitly. I think this is unavoidable, due to our use of module level attributes. There isn't anywhere to place the deprecation message apart from 

**Related GitHub Issues:** #123


strawberryfields/ops.py
```diff
@@ -2032,6 +2026,43 @@ def _decompose(self, reg, **kwargs):
...
+        ``__getattr__`` function, which should enable this functionality
+        without needing to modify ``sys.modules``.
+    """
+    deprecation_map = {"Measure": "MeasureFock"}
+
+    def __init__(self, mod):
+        self.mod = mod
+
+    def __getattr__(self, name):
+        if name in self.deprecation_map:
```
zzzeid: Is it conceivable that a shorthand (or other name) would be deprecated without a replacement? If so, perhaps the deprecation message belongs in `deprecation_map` as the value, that way it can be a little more versatile.

josh146: Good point. I can't forsee that happening tbh, more so that we rename existing operations. @co9olguy?

co9olguy: Seems unlikely that a shorthand would ever be deprecated without a replacement. I also don't foresee removing existing ops (though I guess it could happen). As @josh146 says, most likely thing is just a renaming (which is a form of deprecation I suppose; so far, we've just been doing it with breaking changes, but deprecating things when we change name is probably better option going forward).

josh146: Okay, let's keep it simple for now, and if we ever have a more complicated deprecation we can change this class.

josh146:
> Seems to be still one reference to Measure in examples/GaussianBosonSampling.ipynb

Fixed.

PR #146, opened 2019-07-24
Add starship helper command

zzzeid:

This command helps execute .xbb files from the command line, as well as
to automatically reconfigure the api parameters and write them to a
config file, if needed.

**Benefits:**

Users who do not wish to interact with the Python shell can still execute remote commands.

**Possible Drawbacks:**

**Related GitHub Issues:**


starship
```diff
@@ -0,0 +1,77 @@
...
+from strawberryfields.io import load
+from strawberryfields import configuration
+
+DEFAULT_HOSTNAME = "platform.strawberryfields.ai"
+PROMPTS = {
+    "hostname": "Please enter the hostname of the server to connect to: [{}] ",
+    "authentication_token": "Please enter the authentication token to use when connecting: [{}] ",
+    "save": "Would you like to save these settings to a local cofiguration file in the current "
+    "directory? [Y/n] ",
+}
```
josh146: nice!

starship
```diff
@@ -0,0 +1,76 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="run Strawberry Fields code on StarshipEngine")
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument("--input", "-i", help="the XBB file to run")
```
josh146: Is this required? What happens if you run the script without this argument, you just get the help messages?

starship
```diff
@@ -0,0 +1,76 @@
...
+                sys.stdout.write("Updating configuration in current working directory...\n")
+
+            config.api["hostname"] = hostname
+            config.api["authentication_token"] = authentication_token
+            config.save("config.toml")
+        sys.exit()
+
+    eng = StarshipEngine()
+    program = load(args.input)
+    eng.run(program)
```
josh146: Wow, after setting up the configuration, this is really succinct :boom:

starship
```diff
@@ -0,0 +1,76 @@
...
+
+    if args.output_path:
+        with open(args.output_path, "w") as file:
+            file.write(str(eng.samples))
+            file.close()
+    else:
+        sys.stdout.write(str(eng.samples))
```
josh146: Just curious the reason why you have decided to access `eng.samples` directly, as opposed to the
```python
results = eng.run(program)
print(results.samples)
```

Mostly because I've realised that `eng.samples` is not supposed to be user-facing, and should probably be a 'private' attribute.

starship
zzzeid: Yes the output is:

```
usage: starship [-h] [--input INPUT] [--output OUTPUT] [--debug]
                [--reconfigure]
starship: error: one of the arguments --input/-i --reconfigure is required
```

in that case.

starship
zzzeid: Thanks for noting this --- I will change. I think I saw someone using it (or perhaps it was in an example somewhere) and defaulted to that.

josh146: Strange - do you remember where you saw it?

zzzeid: I can't recall -- but it's possible that I am not remembering correctly. In any case, this is now fixed!

starship
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python
```
co9olguy: still needed after @josh146's comment below?

starship
```diff
@@ -0,0 +1,90 @@
...
+    "hostname": "Please enter the hostname of the server to connect to: [{}] ",
+    "authentication_token": "Please enter the authentication token to use when connecting: [{}] ",
+    "save": "Would you like to save these settings to a local cofiguration file in the current "
+    "directory? [Y/n] ",
+}
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="run Strawberry Fields code on StarshipEngine")
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument("--input", "-i", help="the XBB file to run")
```
co9olguy:
```suggestion
    group.add_argument("--input", "-i", help="the xbb file to run")
```

starship
zzzeid: Yes - had a conversation about this with @josh146 offline, but basically pip will replace this with the path based on the installation environment. `scripts` works OK in this respect and no change is needed immediately.

This line is necessary for pip to translate it correctly.

zzzeid:
> Just left 1-2 suggestions.
> Does it make sense to have tests for a shell script? (my opinion is _yes_, but it depends on whether testing via pytest would be straightforward or needlessly artificial)

Yes it does. This script should probably be rewritten as a python module and added to setup.py via `entry_points`. Technically, the functionality it is doing does not have to be exclusively in a shell script. At that point, testing would be much easier. It would probably need a bit of a rewrite before it can be tested properly.

Issue #143, opened 2019-07-24
Create `BipartiteGraphEmbed` to support embedding bipartite graphs using two-mode squeezing


josh146:
Now that #127 is merged, it is possible to use `Interferometer` to decompose an arbitrary unitary to match the topology of a hardware device using Mach-Zehnder interferometers.

If the device uses squeezed initial states, then `GraphEmbed` will also now work for such hardware devices, due to its use of `Interferometer`.

However, if the device uses two-mode squeezed initial states (`S2gate`), then we cannot support embedding arbitrary graphs using `GraphEmbed`; we are limited to embedding bipartite graphs.

This is relatively easy to do:

0. Perform input validation to ensure that (a) the adjacency matrix represents a bipartite graph, and (b) that the number of nodes in the graph is even

1. We perform the existing `strawberryfields.decomposition.graph_embed` decomposition to determine the initial squeezing values `Sq`, and the interferometer `U`.

2. We then perform some slight manipulation in order to convert the initial `Sgate` operations to `S2gate` operations. This uses the fact that, _starting from a vacuum state_,

   ```python
   Sgate(s1) | q[0]
   Sgate(s1) | q[2]
   ```

   is equivalent to

   ```python
   S2gate(s1) | (q[0], q[2])
   BSgate(pi/4, 0) | (q[0], q[2])
   Rgate(pi/2) | q[2]
   ```

3. The additional beamsplitters and rotation gates introduced are then simply subsumed into the original unitary `U` to provide `U2`, leading to

   ```python
   S2gate(s1) | (q[0], q[2])
   S2gate(s2) | (q[1], q[3])
   Interferometer(U2, mesh='rectangular_symmetric') | (q[0], q[1], q[2], q[3])
   ```

---

**Open questions:**

* Some hardware devices might restrict the allowed values of initial squeezing. How do we scale the adjacency matrix to ensure the singular value decompositions match the allowed squeezing values?

* Ideally, we should also be able to manipulate the unitary `U2` such that it is block diagonal (`U2 = sp.linalg.block_diag(Ua, Ua)`), allowing us to decompose the above even further:

  
  ```python
  S2gate(s1) | (q[0], q[2])
  S2gate(s2) | (q[1], q[3])
  Interferometer(Ua, mesh='rectangular_symmetric') | (q[0], q[1])
  Interferometer(Ua, mesh='rectangular_symmetric') | (q[2], q[3])
  ```

PR #150, opened 2019-07-25
Bipartite graph embed

josh146:

* Adds a new `decomposition.bipartite_graph_embed`, to embed a bipartite graph into a Gaussian state using two-mode squeezed states.

* Adds the new `ops.BipartiteGraphEmbed` user-facing quantum operation. This operation decomposes the graph into the set `{S2gate, Interferometer}`.

**Benefits:** Provides a more structured decomposition than `GraphEmbed` for bipartite graphs

**Possible Drawbacks:** n/a

**Related GitHub Issues:** #143


strawberryfields/ops.py
```diff
@@ -1816,8 +1815,8 @@ class BipartiteGraphEmbed(Decomposition):
     rotation gates.
 
     Args:
-        A (array): An :math:`N\times N` complex or real symmetric matrix representing
-            a bipartite graph, where the number of vertices in the two vertex sets are even.
+        A (array): an :math:`N\times N` complex or real symmetric adjacency matrix representing
```
nquesada: It would be a good idea to have an option to pass either B = [[0,A],[A^T,0]] or A directly and infer which by how many qmodes are available....

josh146: Yep I will make this change now

josh146: Okay, I've added a keyword argument

strawberryfields/ops.py
```diff
@@ -1808,6 +1808,98 @@ def _decompose(self, reg, **kwargs):
...
+        else:
+            self.ns = A.shape[0]
+
+            # check if A is a bipartite graph
+            N = A.shape[0]//2
+            A00 = A[:N, :N]
+            A11 = A[N:, N:]
+
+            diag_zeros = np.allclose(A00, np.zeros_like(A00), atol=tol, rtol=0)
+            diag_zeros = diag_zeros and np.allclose(A11, np.zeros_like(A11), atol=tol, rtol=0)
```
zzzeid: (minor) Is there a reason this assignment is separate from the line above it? E.g. it could be rewritten as:

```python
diag_zeros = np.allclose(A00, np.zeros_like(A00), atol=tol, rtol=0) and np.allclose(
    A11, np.zeros_like(A11), atol=tol, rtol=0)
```

strawberryfields/ops.py
```diff
@@ -1808,6 +1808,98 @@ def _decompose(self, reg, **kwargs):
...
+
+            if (not diag_zeros) or (not np.allclose(A, A.T, atol=tol, rtol=0)):
+                raise ValueError("Adjacency matrix {} does not represent a bipartite graph".format(A))
+
+            B = A[:N, N:]
+
+        super().__init__([B])
+
+    def _decompose(self, reg, **kwargs):
```
zzzeid: Can a docstring be added here?

strawberryfields/ops.py
```diff
@@ -1808,6 +1808,98 @@ def _decompose(self, reg, **kwargs):
...
+
+        sq, U, V = dec.bipartite_graph_embed(B, mean_photon_per_mode=mean_photon, atol=tol, rtol=0)
+
+        if not self.identity or not drop_identity:
+            for m, s in enumerate(sq):
+                s = s if np.abs(s) >= _decomposition_tol else 0
+
+                if not (drop_identity and s == 0):
+                    cmds.append(Command(S2gate(-s), (reg[m], reg[m+N])))
+
```
zzzeid: The below could possibly be refactored to reduce code repetition, e.g.:
```python
for X, _reg in ((U, reg[:N]), (V, reg[N:]):
    if np.allclose(X, np.identity(len(X)), atol=_decomposition_tol, rtol=0):
        X = np.identity(len(X))
    if not (drop_identity and np.all(X == np.identity(len(X)))):
        cmds.append(Command(Interferometer(X, mesh=mesh, drop_identity=drop_identity, tol=tol), _reg))
```

strawberryfields/ops.py
josh146: Ah, it was originally one line, but it started to confuse me every time I re-read it :) I split it into two lines, and found it easier to read/work with.

strawberryfields/ops.py
josh146: Since this is an inherited method of an abstract method, my preference is not to, as it will automatically inherit the docstring of the parent method. This avoids duplication, and the issue of updating one `_decompose()` method, forgetting others, and having them go out-of-sync.

We used to have docstrings for each of these methods, but there was a PR to remove them as we kept running into the sync issue. (I think this was when a particular python/sphinx version was released that supported inheriting docstrings?)

strawberryfields/ops.py
josh146: Can change this back to one line if you want?

strawberryfields/ops.py
zzzeid: Ah makes sense, thanks!

strawberryfields/ops.py
zzzeid: Was it the assignment itself or splitting it into two lines that alleviated the confusion? (we could still split it into two lines in this fashion without re-assigning, but either way this is very minor and go with whatever your preference is)

josh146: Oh it was just the length of the line, not the assignment!

strawberryfields/ops.py
josh146: Thanks! Made this change.

strawberryfields/ops.py
josh146: Changed it to
```python
diag_zeros = np.allclose(A00, np.zeros_like(A00), atol=tol, rtol=0) \
    and np.allclose(A11, np.zeros_like(A11), atol=tol, rtol=0)
```

josh146:
> technically we mostly use the SVD. Only call Takagi if the B is symmetric.

Thanks @nquesada, have removed this reference

strawberryfields/ops.py
```diff
@@ -1771,7 +1771,7 @@ class GraphEmbed(Decomposition):
 
     Args:
         A (array): an :math:`N\times N` complex or real symmetric matrix
-        mean_photon_per_mode (float): guarantees that the mean photon number in the pure Gaussian state
+        mean_photon (float): guarantees that the mean photon number in the pure Gaussian state
```
co9olguy: @josh146 just curious why this was changed?

josh146: I don't remember anymore... I think I was bringing it inline with other frontend components, that all used `mean_photon` as opposed to `mean_photon_per_mode`

co9olguy: From what I recall, this was specifically changed to be `mean_photon_per_mode` during the graph embed bugfix, since previously it was looking at total mean photon number, and doing something physically different

@nquesada can you comment?

nquesada: The name of the variable should be ```mean_photon_per_mode```.  The docstring and the usage when passing to ```dec.graph_embed``` are correct.

PR #152, opened 2019-07-30
Adds Torontonian sampling support to the Gaussian backend

nquesada:
Adds threshold (torontonian) sampling to the capabilities of the Gaussian backend.
**Benefits:**
Threshold samples can be generated directly instead of doing PNR and then threshold by hand. Should give significant saving in time and it is useful for apps upstream.
**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -17,6 +17,8 @@
   allowing a bipartite graph to be embedded on a device that allows for
   initial two-mode squeezed states, and block diagonal unitaries.
 
+* `measure_threshold` is now available in the Gaussian backend. [#152](https://github.com/XanaduAI/strawberryfields/pull/152)
+
 ### API Changes
 
 * The `strawberryfields.ops.Measure` shorthand has been deprecated in favour
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -17,7 +17,9 @@
   allowing a bipartite graph to be embedded on a device that allows for
   initial two-mode squeezed states, and block diagonal unitaries.
 
-* `measure_threshold` is now available in the Gaussian backend. [#152](https://github.com/XanaduAI/strawberryfields/pull/152)
+* Added support for threshold measurement in the Gaussian backend, via the new backend API
+  method `measure_threshold`.
+  [#152](https://github.com/XanaduAI/strawberryfields/pull/152)
 
 ### API Changes
 
--- a/strawberryfields/__init__.py
+++ b/strawberryfields/__init__.py
@@ -98,7 +98,7 @@ def about():
     import os
     import numpy
     import scipy
-    import hafnian
+    import thewalrus
     import blackbird
 
     # a QuTiP-style infobox
@@ -111,7 +111,7 @@ def about():
     print('Strawberry Fields version: {}'.format(__version__))
     print('Numpy version:             {}'.format(numpy.__version__))
     print('Scipy version:             {}'.format(scipy.__version__))
-    print('Hafnian version:           {}'.format(hafnian.__version__))
+    print('The Walrus version:           {}'.format(thewalrus.__version__))
     print('Blackbird version:         {}'.format(blackbird.__version__))
 
     try:
--- a/strawberryfields/backends/base.py
+++ b/strawberryfields/backends/base.py
@@ -507,7 +507,7 @@ class BaseBackend:
         raise NotImplementedError
 
     def measure_threshold(self, modes, shots=1, select=None, **kwargs):
-        """Measure the given modes in the thresholded Fock basis, i.e., 0 or non zero photons).
+        """Measure the given modes in the thresholded Fock basis, i.e., zero or nonzero photons).
 
         ..note::
           When :code:``shots == 1``, updates the current system state to the conditional state of that
--- a/strawberryfields/backends/base.py
+++ b/strawberryfields/backends/base.py
@@ -509,9 +509,10 @@ class BaseBackend:
     def measure_threshold(self, modes, shots=1, select=None, **kwargs):
         """Measure the given modes in the thresholded Fock basis, i.e., zero or nonzero photons).
 
-        ..note::
-          When :code:``shots == 1``, updates the current system state to the conditional state of that
-          measurement result. When :code:``shots > 1``, the system state is not updated.
+        .. note::
+
+            When :code:``shots == 1``, updates the current system state to the conditional state of that
+            measurement result. When :code:``shots > 1``, the system state is not updated.
 
         Args:
             modes (Sequence[int]): which modes to measure
--- a/strawberryfields/backends/base.py
+++ b/strawberryfields/backends/base.py
@@ -506,6 +506,24 @@ class BaseBackend:
         """
         raise NotImplementedError
 
+    def measure_threshold(self, modes, shots=1, select=None, **kwargs):
+        """Measure the given modes in the thresholded Fock basis, i.e., 0 or non zero photons).
+
+        ..note::
+          When :code:``shots == 1``, updates the current system state to the conditional state of that
+          measurement result. When :code:``shots > 1``, the system state is not updated.
+
+        Args:
+            modes (Sequence[int]): which modes to measure
+            shots (int): number of measurement samples to obtain
+            select (None or Sequence[int]): If not None: desired values of the measurement results.
+                Enables post-selection on specific measurement results instead of random sampling.
+                ``len(select) == len(modes)`` is required.
+        Returns:
+            tuple[int]: measurement results
+        """
+        raise NotImplementedError
+
     def is_vacuum(self, tol=0.0, **kwargs):
         r"""Test whether the current circuit state is vacuum (up to given tolerance).
 
--- a/strawberryfields/backends/gaussianbackend/backend.py
+++ b/strawberryfields/backends/gaussianbackend/backend.py
@@ -230,7 +230,7 @@ class GaussianBackend(BaseGaussian):
         cov = self.circuit.scovmatxp()
         # check we are sampling from a gaussian state with zero mean
         if not allclose(mu, zeros_like(mu)):
-            raise NotImplementedError("PNR measurement is only supported for "
+            raise NotImplementedError("Threshold measurement is only supported for "
                                       "Gaussian states with zero mean")
         x_idxs = array(modes)
         p_idxs = x_idxs + len(mu)
--- a/strawberryfields/backends/gaussianbackend/backend.py
+++ b/strawberryfields/backends/gaussianbackend/backend.py
@@ -222,9 +222,9 @@ class GaussianBackend(BaseGaussian):
         if shots != 1:
             if select is not None:
                 raise NotImplementedError("Gaussian backend currently does not support "
-                                          "postselection if shots != 1 for Fock measurement")
+                                          "postselection if shots != 1 for Threshold measurement")
             warnings.warn("Cannot simulate non-Gaussian states. "
-                          "Conditional state after Fock measurement has not been updated.")
+                          "Conditional state after Threshold measurement has not been updated.")
 
         mu = self.circuit.mean
         cov = self.circuit.scovmatxp()
--- a/strawberryfields/backends/gaussianbackend/backend.py
+++ b/strawberryfields/backends/gaussianbackend/backend.py
@@ -196,7 +196,7 @@ class GaussianBackend(BaseGaussian):
         if shots != 1:
             if select is not None:
                 raise NotImplementedError("Gaussian backend currently does not support "
-                                          "postselection if shots != 1 for Fock measurement")
+                                          "postselection")
             warnings.warn("Cannot simulate non-Gaussian states. "
                           "Conditional state after Fock measurement has not been updated.")
 
@@ -222,7 +222,7 @@ class GaussianBackend(BaseGaussian):
         if shots != 1:
             if select is not None:
                 raise NotImplementedError("Gaussian backend currently does not support "
-                                          "postselection if shots != 1 for Threshold measurement")
+                                          "postselection")
             warnings.warn("Cannot simulate non-Gaussian states. "
                           "Conditional state after Threshold measurement has not been updated.")
 
--- a/strawberryfields/backends/gaussianbackend/backend.py
+++ b/strawberryfields/backends/gaussianbackend/backend.py
@@ -30,7 +30,7 @@ from numpy import (
     ix_,
 )
 from numpy.linalg import inv
-from hafnian.samples import hafnian_sample_state, torontonian_sample_state
+from thewalrus.samples import hafnian_sample_state, torontonian_sample_state
 
 from strawberryfields.backends import BaseGaussian
 from strawberryfields.backends.shared_ops import changebasis
--- a/strawberryfields/backends/gaussianbackend/backend.py
+++ b/strawberryfields/backends/gaussianbackend/backend.py
@@ -30,7 +30,7 @@ from numpy import (
     ix_,
 )
 from numpy.linalg import inv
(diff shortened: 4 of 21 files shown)


tests/backend/test_threshold_measurement.py
```diff
@@ -0,0 +1,66 @@
...
+
+NUM_REPEATS = 50
+
+
+@pytest.mark.backends("gaussian")
+class TestGaussianRepresentation:
+    """Tests that make use of the Fock basis representation."""
+
+    def measure_threshold_gaussian_warning(self, setup_backend):
+        """Tests that Threshold measurements are not implemented when shots != 1.
```
trbromley: Is this the right docstring?

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -217,6 +217,33 @@ def measure_fock(self, modes, shots=1, select=None):
             samples = samples.reshape((len(modes),))
         return samples
 
+
+    def measure_threshold(self, modes, shots=1, select=None):
+        if shots != 1:
+            if select is not None:
+                raise NotImplementedError("Gaussian backend currently does not support "
```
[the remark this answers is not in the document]
nquesada: No. But this is also what is done in MeasureFock, for which there is now way to support conditional states in the Gaussian Backend. @josh146 any idea why it is done like that for the MeasureFock case?

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -217,6 +217,33 @@ def measure_fock(self, modes, shots=1, select=None):
...
+            raise NotImplementedError("Threshold measurement is only supported for "
+                                      "Gaussian states with zero mean")
+        x_idxs = array(modes)
+        p_idxs = x_idxs + len(mu)
+        modes_idxs = concatenate([x_idxs, p_idxs])
+        reduced_cov = cov[ix_(modes_idxs, modes_idxs)]
+        samples = torontonian_sample_state(reduced_cov, shots)
+        # for backward compatibility with previous measurement behaviour,
+        # if only one shot, then we drop the shots axis
+        if shots == 1:
```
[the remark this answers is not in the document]
nquesada: Same as above.

strawberryfields/circuitspecs/gaussian.py
```diff
@@ -40,6 +40,7 @@ class GaussianSpecs(CircuitSpecs):
         "MeasureHomodyne",
         "MeasureHeterodyne",
         "MeasureFock",
+        "MeasureThreshold",
```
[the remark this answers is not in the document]
nquesada: That is a great idea.

strawberryfields/ops.py
```diff
@@ -1052,6 +1053,23 @@ def _apply(self, reg, backend, shots=1, **kwargs):
         return backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
 
 
+class MeasureThreshold(Measurement):
+    """:ref:`photo_detection`: measures a set of modes in the thresholded Fock basis, i.e.
+    0 or more than 0 photons.
+
+    After measurement, the modes are reset to the vacuum state.
```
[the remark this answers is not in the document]
nquesada: Same as with MeasureFock.

tests/backend/test_threshold_measurement.py
nquesada: Nop

strawberryfields/backends/gaussianbackend/backend.py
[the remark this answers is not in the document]
nquesada: I updated the messages for both measure_fock and measure_threshold

strawberryfields/ops.py
```diff
@@ -1052,6 +1053,23 @@ def _apply(self, reg, backend, shots=1, **kwargs):
...
+    zero or nonzero photons.
+
+    After measurement, the modes are reset to the vacuum state.
+    """
+    ns = None
+
+    def __init__(self, select=None):
+        if select is not None and not isinstance(select, Sequence):
+            select = [select]
+        super().__init__([], select)
```
josh146: In this case, would `select` be a list of integers, i.e., `[0, 0, 1, 1, 0, 1]`?

tests/backend/test_threshold_measurement.py
```diff
@@ -0,0 +1,65 @@
...
+            backend.reset(pure=pure)
+
+            r = 0.25
+            # Circuit to prepare two mode squeezed vacuum
+            backend.squeeze(-r, 0)
+            backend.squeeze(r, 1)
+            backend.beamsplitter(np.sqrt(0.5), -np.sqrt(0.5), 0, 1)
+            meas_modes = [0, 1]
+            meas_results = backend.measure_threshold(meas_modes)
+            assert np.all(meas_results[0] == meas_results[1])
```
josh146: How come the loop is needed here? Looks like it can be removed without affecting anything

tests/backend/test_threshold_measurement.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_vacuum_measurements(self, setup_backend, pure):
+        """Tests Threshold measurement on the vacuum state."""
+        backend = setup_backend(3)
+
+        for _ in range(NUM_REPEATS):
+            backend.reset(pure=pure)
+
+            meas = backend.measure_threshold([0, 1, 2])[0]
+            assert np.all(np.array(meas) == 0)
```
josh146: How come the loop is needed here? Looks like it can be removed without affecting anything

strawberryfields/backends/gaussianbackend/backend.py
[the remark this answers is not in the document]
nquesada: There is more to it than that. Even if there is a single shot you cannot update a Gaussian state after measurement in the Fock basis *within* the Gaussian formalism

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -217,6 +217,33 @@ def measure_fock(self, modes, shots=1, select=None):
...
+                                          "postselection")
+            warnings.warn("Cannot simulate non-Gaussian states. "
+                          "Conditional state after Threshold measurement has not been updated.")
+
+        mu = self.circuit.mean
+        cov = self.circuit.scovmatxp()
+        # check we are sampling from a gaussian state with zero mean
+        if not allclose(mu, zeros_like(mu)):
+            raise NotImplementedError("Threshold measurement is only supported for "
+                                      "Gaussian states with zero mean")
```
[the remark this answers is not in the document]
nquesada: I am working on it :)

strawberryfields/ops.py
nquesada: I guess it would. This would be useful once `MeasureThreshold` is implemented in the Fock backends. In the gaussian backend if you try to postselect you will get an error message.

tests/backend/test_threshold_measurement.py
nquesada: Because you want to make sure that you *always* get zero.

nquesada:
@josh146  any idea what this 
image means?

strawberryfields/backends/gaussianbackend/backend.py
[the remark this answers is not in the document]
josh146: Ah, yeah. My answer was with respect to the Fock backends.

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -196,7 +196,7 @@ def measure_fock(self, modes, shots=1, select=None):
         if shots != 1:
             if select is not None:
```
trbromley: Looks ok now, but shouldn't this be placed outside of `if shots != 1:` since we don't support postselection generally (yet)? (also for `measure_threshold`)

trbromley: Also, is postselection something that we don't support _yet_ in the Gaussian backend, or will never support? If never, then should we not remove `select` as an argument in `measure_fock` and `measure_threshold`, and also remove the corresponding sentences in docstrings?

strawberryfields/backends/base.py
```diff
@@ -506,6 +506,25 @@ def measure_fock(self, modes, shots=1, select=None, **kwargs):
         """
         raise NotImplementedError
 
+    def measure_threshold(self, modes, shots=1, select=None, **kwargs):
+        """Measure the given modes in the thresholded Fock basis, i.e., zero or nonzero photons).
+
+        .. note::
+
+            When :code:``shots == 1``, updates the current system state to the conditional state of that
```
trbromley: Should we remove these lines (also for `measure_fock`)?

strawberryfields/ops.py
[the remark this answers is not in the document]
trbromley: So just so I understand, regardless of the value of `shots`, currently the modes are definitely reset to the vacuum state? Or are they left as is?

strawberryfields/backends/gaussianbackend/backend.py
josh146: The reason we need to keep the `select` argument is that this class inherits from `BaseBackend`, as does the Fock and TF backends. Therefore the method docstrings must match that of the base class.

strawberryfields/backends/base.py
josh146: These should remain, as this is the abstract base class of the backend API, and serves as instructions for new developers on how this method should act, _if_ they have the ability to calculate the conditional state of the system after measurement.

strawberryfields/backends/gaussianbackend/backend.py
nquesada: For Fock measurements I don't see how. For Threshold measurements it is possible in principle, but I am not there is any interest in doing it. A reason why select is there is just to keep consistency with the other backends, but not sure if it is worth or not keeping it. I suspect that should be dealt with in a different PR. This PR only tries to include threshold measurements at the same level as Fock measurements

strawberryfields/ops.py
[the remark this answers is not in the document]
nquesada: In the backend nothing happens. In the frontend I think you are simply not allowed to ask *anything* about a state in the gaussian backend once you do a non gaussian measurement.

strawberryfields/ops.py
```diff
@@ -1052,6 +1053,23 @@ def _apply(self, reg, backend, shots=1, **kwargs):
         return backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
 
 
+class MeasureThreshold(Measurement):
+    """:ref:`photo_detection`: measures a set of modes in the thresholded Fock basis, i.e.
```
```suggestion
    """:ref:`photo_detection`: measures a set of modes in the thresholded Fock basis, i.e.,
```

co9olguy: just curious, where does `photo_detection` resolve to in docs?

co9olguy: "thresholded Fock basis" may be confusing (it's not a basis)

strawberryfields/ops.py
```diff
@@ -1052,6 +1053,23 @@ def _apply(self, reg, backend, shots=1, **kwargs):
...
+    zero or nonzero photons.
+
+    After measurement, the modes are reset to the vacuum state.
+    """
+    ns = None
+
+    def __init__(self, select=None):
+        if select is not None and not isinstance(select, Sequence):
+            select = [select]
```
co9olguy: How come this extra logic was needed on top of calling the parent `__init__` method?

tests/backend/test_threshold_measurement.py
```diff
@@ -0,0 +1,65 @@
...
+
+        backend = setup_backend(3)
+
+        with pytest.warns(Warning, match="Cannot simulate non-Gaussian states. Conditional state after "
+                                         "Threshold measurement has not been updated."):
+            backend.measure_threshold([0, 1], shots=5)
+
+
+@pytest.mark.backends("gaussian")
+class TestRepresentationIndependent:
```
co9olguy: Nice, so when we add to other backends, we would just unmark here?

tests/backend/test_threshold_measurement.py
co9olguy: lol then you'd need to set `NUM_REPEATS` to infinity :laughing:

co9olguy: Perhaps add a test which asserts that all results returned by `measure_threshold` fall in [0,1]?

strawberryfields/ops.py
nquesada: Now that I think about it doesn't. Photodetection is however briefly mention in the sf paper in Quantum.

strawberryfields/__init__.py
```diff
@@ -98,7 +98,7 @@ def about():
     import os
     import numpy
     import scipy
-    import hafnian
```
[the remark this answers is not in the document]
josh146: This is my fault - I did a semi auto-refactor in my editor to allow this PR to pass the tests (it required a new released version of hafnian/thewalrus).

tests/backend/test_threshold_measurement.py
nquesada: Done :)

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -196,7 +196,7 @@ def measure_fock(self, modes, shots=1, select=None):
...
                 raise NotImplementedError("Gaussian backend currently does not support "
-                                          "postselection if shots != 1 for Fock measurement")
```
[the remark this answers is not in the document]
nquesada: Currently, even if shots==1 the Gaussian backend does not support postselection.

strawberryfields/ops.py
josh146: This was copied from `MeasureFock`. This is because the abstract `Measurement` class assumes `select` is always a sequence; at some point, we must have decided that select can be a single number if the measurement applies to a single mode.

josh146: We should work out how exactly we want the post-selection/measurement API to look, and then refactor this to make it cleaner.

josh146: @nquesada, even though the Gaussian backend cannot support threshold post-selection, should we add some input validation here to make sure `select` is only 0 or 1?

nquesada: I don't think it's worth it unless we at some decide to write a backend that deals with states that are linear combinations of Gaussian states (this was partially done in https://github.com/xanaduAI/torontonian-sampling).

josh146:
All looks good from my end now @nquesada

josh146:
As one last commit before merging, can you update the changelog, adding a one-to-two sentence description of the PR to the relevant section, and also link back to the PR.

.github/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
   allowing a bipartite graph to be embedded on a device that allows for
   initial two-mode squeezed states, and block diagonal unitaries.
 
+* `measure_threshold` is now available in the Gaussian backend. [#152](https://github.com/XanaduAI/strawberryfields/pull/152)
```
josh146:
```suggestion
* Adds support for threshold measurement in the Gaussian backend, via the new backend API
  method `measure_threshold`.
  [#152](https://github.com/XanaduAI/strawberryfields/pull/152)
```

strawberryfields/ops.py
```diff
@@ -1052,6 +1053,23 @@ def _apply(self, reg, backend, shots=1, **kwargs):
         return backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
 
 
+class MeasureThreshold(Measurement):
+    """:ref:`photo_detection`: measures a set of modes in the thresholded Fock basis, i.e.
+    zero or nonzero photons.
```
co9olguy: measuring whether a mode contain zero or nonzero photons.

PR #153, opened 2019-07-30
Starship hello fix

zzzeid:
Fix status code check (200)

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/starship
+++ b/starship
@@ -88,7 +88,11 @@ if __name__ == "__main__":
         except Exception as e:
             sys.stderr.write("Could not connect to server:\n{}\n".format(e))
             sys.exit(1)
-        if response.status_code in (401, 403):
+
+        if response.status_code == 200:
+            sys.stdout.write("You have successfully authenticated to the platform!\n")
+            sys.exit()
+        elif response.status_code in (401, 403):
             if not client.AUTHENTICATION_TOKEN:
                 sys.stderr.write("Could not authenticate -- no token provided\n")
             else:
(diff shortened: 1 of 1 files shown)


starship
```diff
@@ -97,9 +101,9 @@ if __name__ == "__main__":
                         client.AUTHENTICATION_TOKEN))
             sys.exit(1)
         else:
-            # Hack at the moment, but should check status code 200 against test endpoint
-            sys.stdout.write("You have successfully authenticated to the platform!\n")
-            sys.exit()
+            sys.stderr.write("Could not connect to server: {}, {}\n".format(
```
jswinarton: Can we use f-strings here?

zzzeid: Strawberry Fields has Python 3.6 as target for compatibility, so not yet.

josh146: (Python 3.5+ currently.)

josh146:
Approved, but agree with @jswinarton, would be nice to move the script into the module.

josh146:
By the way @jswinarton, @zzzeid  and I looked into using `entry_points` for the script, but it looks like this approach doesn't work with `argparse`, and in general is designed not to support command line arguments (outside of `sys.argv`).

So we made the decision to stick with `setuptools.scripts` for now, which may affect the placement of the script file.

starship
zzzeid: Sorry I meant to say 3.5+!

zzzeid:
> Looks good! One general comment, I think it would be best to consolidate this as part of the strawberryfields module in general instead of putting it at the root of the directory. You could move the file to `strawberryfields/scripts/starship.py` (or elsewhere depending on your preferences), wrap everything in a method (`def main()`), and then use a setup.py entrypoint as follows:
> 
> ```
> setup(
>     # ...
>     entry_points={
>         'console_scripts': [
>             'starship=strawberryfields.scripts.starship:main'
>         ]
>     }
> )
> ```
> 
> Then all of your python code is contained within the module. I think in general this makes it easier to share and test everything.

@jswinarton as @josh146 mentioned this does not work nicely with `argparse`, but in the future we could refactor this to use something else and have much of this functionality in the module instead (then we can test things as usual too). Much of the script relies on argparse at the moment.

PR #154, opened 2019-07-31
Removes the restriction of zero mean for PNR sampling

nquesada:
Removes the restriction of zero mean for PNR sampling. This is allowed by recently added capabilities in the hafnian library.
**Benefits:**
Can do more sophisticated sampling. Also useful in Franck-Condon factor sampling.
**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/strawberryfields/backends/gaussianbackend/backend.py
+++ b/strawberryfields/backends/gaussianbackend/backend.py
@@ -201,16 +201,19 @@ class GaussianBackend(BaseGaussian):
                           "Conditional state after Fock measurement has not been updated.")
 
         mu = self.circuit.mean
+        mean = self.circuit.smeanxp()
         cov = self.circuit.scovmatxp()
-        # check we are sampling from a gaussian state with zero mean
-        if not allclose(mu, zeros_like(mu)):
-            raise NotImplementedError("PNR measurement is only supported for "
-                                      "Gaussian states with zero mean")
+
         x_idxs = array(modes)
         p_idxs = x_idxs + len(mu)
         modes_idxs = concatenate([x_idxs, p_idxs])
         reduced_cov = cov[ix_(modes_idxs, modes_idxs)]
-        samples = hafnian_sample_state(reduced_cov, shots)
+        reduced_mean = mean[modes_idxs]
+        # check we are sampling from a gaussian state with zero mean
+        if allclose(mu, zeros_like(mu)):
+            samples = hafnian_sample_state(reduced_cov, shots)
+        else:
+            samples = hafnian_sample_state(reduced_cov, shots, mean=reduced_mean)
         # for backward compatibility with previous measurement behaviour,
         # if only one shot, then we drop the shots axis
         if shots == 1:
--- a/strawberryfields/backends/gaussianbackend/gaussiancircuit.py
+++ b/strawberryfields/backends/gaussianbackend/gaussiancircuit.py
@@ -249,7 +249,7 @@ class GaussianModes:
         nmodes = self.nlen
         r = np.empty(2*nmodes)
         r[0:nmodes] = 2*self.mean.real
-        r[nmodes:2*nmodes] = 2*self.mean.real
+        r[nmodes:2*nmodes] = 2*self.mean.imag
         return r
 
     def scovmat(self):
--- a/tests/backend/test_fock_measurement.py
+++ b/tests/backend/test_fock_measurement.py
@@ -162,15 +162,12 @@ class TestRepresentationIndependent:
         """Test that a coherent state with a mean photon number of 4 and sampled NUM_REPEATS times will produce photons"""
         backend = setup_backend(1)
         alpha = 2.0
-        nonzero = False
+        meas = np.array(backend.measure_fock([0]))
 
         for _ in range(NUM_REPEATS):
             backend.reset(pure=pure)
             backend.displacement(alpha, 0)
-            meas = backend.measure_fock([0])[0]
-            if meas > 0:
-                nonzero = True
-                break
-        assert nonzero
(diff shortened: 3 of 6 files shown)


strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -201,16 +201,19 @@ def measure_fock(self, modes, shots=1, select=None):
                           "Conditional state after Fock measurement has not been updated.")
 
         mu = self.circuit.mean
+        mean = self.circuit.smeanxp()
         cov = self.circuit.scovmatxp()
```
josh146: Out of curiosity, what is the difference between `self.circuit.mean` and `self.circuit.smeanxp`?

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -201,16 +201,19 @@ def measure_fock(self, modes, shots=1, select=None):
...
         p_idxs = x_idxs + len(mu)
         modes_idxs = concatenate([x_idxs, p_idxs])
         reduced_cov = cov[ix_(modes_idxs, modes_idxs)]
-        samples = hafnian_sample_state(reduced_cov, shots)
+        reduced_mean = mean[modes_idxs]
+        # check we are sampling from a gaussian state with zero mean
+        if allclose(mu, zeros_like(mu)):
+            samples = hafnian_sample_state(reduced_cov, shots)
+        else:
+            samples = hafnian_sample_state(reduced_cov, shots, mean=reduced_mean)
```
josh146: :fire:

strawberryfields/backends/gaussianbackend/backend.py
nquesada: self.circuit.mean returns alpha, smeanxp return x and p.

strawberryfields/backends/gaussianbackend/backend.py
nquesada: Thanks! As for the tests, should they be frontend or backend?

josh146: Ideally both; one in `tests/backend/` to test the above implementation directly, and another in `tests/integration`, to ensure that it works correctly when accessed from the frontend

nquesada: Apparently sf does not know of the latest version the walrus... and because of this the tests are failing in travis...

josh146: Ah. We will need to release a new version of TheWalrus before the tests will be able to pass here. This should be fine, I think there are more than enough new features in TheWalrus to warrant a new release.

strawberryfields/backends/gaussianbackend/gaussiancircuit.py
```diff
@@ -249,7 +249,7 @@ def smeanxp(self):
         nmodes = self.nlen
         r = np.empty(2*nmodes)
         r[0:nmodes] = 2*self.mean.real
-        r[nmodes:2*nmodes] = 2*self.mean.real
+        r[nmodes:2*nmodes] = 2*self.mean.imag
```
co9olguy: @nquesada was this a bug before?

josh146: Looks like it was; since the convention is [x1,..., xN, p1,..., pN], the second half of the vector should contain the imaginary parts of alpha.

Highlights an issue with our existing backend tests. They are mostly state-based, which ensures that all backends agree with one another, but is missing unit tests which would catch bugs like this.

josh146:
Note: this will need to be merged after #152 to allow the tests and continuous integration to pass.

strawberryfields/backends/gaussianbackend/gaussiancircuit.py
nquesada: Yup. It was a bug. It was detected by the tests introduced in this PR.

tests/backend/test_fock_measurement.py
```diff
@@ -156,3 +156,21 @@ def test_vacuum_measurements(self, setup_backend, pure):
...
+        nonzero = False
+
+        for _ in range(NUM_REPEATS):
+            backend.reset(pure=pure)
+            backend.displacement(alpha, 0)
+            meas = backend.measure_fock([0])[0]
+            if meas > 0:
+                nonzero = True
+                break
+        assert nonzero
```
co9olguy: @josh146 @nquesada any thoughts on how we might test the same thing (or something similar enough) deterministically?

nquesada: The test looks stochastic and ugly (I am not a fan of that break but it was the simpler than doing a while with two conditionals).
The probability of getting vacuum for this particular a state with \alpha = 2 is \exp(-4) ~ 0.018.
We are checking this NUM_REPEATS = 50 times so the chance that in those 50 trials we don't get at least a nonzero event is 10^-87. This test is in practice deterministic.

co9olguy:
Looks like there is still one failing test case, having to do with the new test not accounting for batched operation of TFBackend. Should be easy enough to fix

co9olguy:
Locally, I'm seeing more tests fail than on travis:
- `backend/test_fock_measurement.py:test_fock_measurements`
- `backend/test_fock_measurement.py:test_coherent_state_has_photons`

nquesada:
Which version of the hafnian are you using?

co9olguy:
@nquesada `test_coherent_state_has_photons` is now passing with latest `pip` version of hafnian library

co9olguy:
> Which version of the hafnian are you using?

Looks like it was an issue with tensorflow dependency. 
I have updated the new test in this PR to take into account batched operations.

PR #160, opened 2019-08-12
Added test for cubic phase gate

egbQuantum:
Added test for cubic phase gate in 'tests/backend/test_nongaussian_gates.py'

**Benefits:**
Allows fock and tf backend testing of the cubic phase gate.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/tests/backend/test_nongaussian_gates.py
+++ b/tests/backend/test_nongaussian_gates.py
@@ -21,7 +21,7 @@ import numpy as np
 from scipy.linalg import expm
 
 KAPPAS = np.linspace(0, 2 * np.pi, 7)
-GAMMAS = np.linspace(0, 2 * np.pi, 7)
+GAMMAS = np.linspace(0, 6, 7)
 
 @pytest.mark.backends("fock", "tf")
 class TestFockRepresentation:
@@ -48,13 +48,13 @@ class TestFockRepresentation:
         np.fill_diagonal(a[:, 1:], ladder_vals)
 
         #Construct (unnormalized) x matrix, ie a+a^dag, and it's third power
-        x = a + np.transpose(a)
-        x3 = np.matmul(x, np.matmul(x, x))
+        x = a + a.T
+        x3 = x @ x @ x
 
         gate = expm(1j * gamma * x3 / 6)
 
         #state to be transformed
-        ket = np.ones([cutoff]) / np.sqrt(cutoff)
+        ket = np.ones(cutoff) / np.sqrt(cutoff)
 
         ref_state = np.matmul(gate, ket)
         assert np.allclose(numer_state, ref_state, atol=tol, rtol=0.0)
--- a/tests/backend/test_nongaussian_gates.py
+++ b/tests/backend/test_nongaussian_gates.py
@@ -46,18 +46,18 @@ class TestFockRepresentation:
         ladder_vals = np.arange(1, cutoff)
         ladder_vals = np.sqrt(ladder_vals)
         a = np.zeros([cutoff, cutoff])
-        np.fill_diagonal(a[:,1:], ladder_vals)
+        np.fill_diagonal(a[:, 1:], ladder_vals)
 
         #Construct (unnormalized) x matrix, ie a+a^dag, and it's third power
         x = a + np.transpose(a)
-        x3 = np.matmul(x, np.matmul(x,x))
+        x3 = np.matmul(x, np.matmul(x, x))
 
         gate = expm(1j * gamma * x3 / 6)
 
         #state to be transformed
         ket1 = np.array([1.0 for n in range(cutoff)])
         ket1 /= np.linalg.norm(ket1)
-        ket = ket1.reshape(-1,1)
+        ket = ket1.reshape(-1, 1)
 
         ref_state = np.matmul(gate, ket).flatten()
         assert np.allclose(numer_state, ref_state, atol=tol, rtol=0.0)
--- a/tests/backend/test_nongaussian_gates.py
+++ b/tests/backend/test_nongaussian_gates.py
@@ -21,7 +21,7 @@ import numpy as np
 from scipy.linalg import expm
 
 KAPPAS = np.linspace(0, 2 * np.pi, 7)
-GAMMAS = np.linspace(0, 2 * np.pi, 7)
+GAMMAS = np.linspace(0, 6, 7)
 
 @pytest.mark.backends("fock", "tf")
 class TestFockRepresentation:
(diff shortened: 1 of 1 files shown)


josh146:
Thanks @egbQuantum! This looks fantastic.

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
 class TestFockRepresentation:
     """Tests that make use of the Fock basis representation."""
 
+    @pytest.mark.parametrize("gamma", GAMMAS)
+    def test_cubic_phase(self, setup_backend, gamma, cutoff, tol):
+        """Tests if the Cubic phase gate has the right effect on states in the Fock basis"""
+
+        backend = setup_backend(1)
+
+        backend.prepare_ket_state(np.ones([cutoff]) / np.sqrt(cutoff), 0)
```
josh146: Nice! I've noticed that the ket prepared for the Kerr interaction test instead is normalised by `cutoff` rather than `sqrt(cutoff)`, is this a bug in `test_kerr_interaction`?

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
+        np.fill_diagonal(a[:, 1:], ladder_vals)
+
+        #Construct (unnormalized) x matrix, ie a+a^dag, and it's third power
+        x = a + np.transpose(a)
+        x3 = np.matmul(x, np.matmul(x, x))
+
+        gate = expm(1j * gamma * x3 / 6)
+
+        #state to be transformed
+        ket1 = np.array([1.0 for n in range(cutoff)])
```
josh146: Note that this could be replaced with
```suggestion
        ket1 = np.ones([cutoff])
```
as done above when setting the backend initial state

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
+        ket1 /= np.linalg.norm(ket1)
+        ket = ket1.reshape(-1, 1)
+
+        ref_state = np.matmul(gate, ket).flatten()
+        assert np.allclose(numer_state, ref_state, atol=tol, rtol=0.0)
```
josh146: This looks great, and passes all the tests :+1:

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
+
+        #Create annihilation operator matrix
+        ladder_vals = np.arange(1, cutoff)
+        ladder_vals = np.sqrt(ladder_vals)
+        a = np.zeros([cutoff, cutoff])
+        np.fill_diagonal(a[:, 1:], ladder_vals)
+
+        #Construct (unnormalized) x matrix, ie a+a^dag, and it's third power
+        x = a + np.transpose(a)
+        x3 = np.matmul(x, np.matmul(x, x))
```
josh146: Note that this can equivalently be written
```python
x = a + a.T
x3 = x @ x @ x
```
Either style is fine, but we tend to err on the side of readability :slightly_smiling_face:

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
+        if s.is_pure:
+            numer_state = s.ket()
+        else:
+            numer_state = s.dm()
+
+        #Create annihilation operator matrix
+        ladder_vals = np.arange(1, cutoff)
+        ladder_vals = np.sqrt(ladder_vals)
+        a = np.zeros([cutoff, cutoff])
+        np.fill_diagonal(a[:, 1:], ladder_vals)
```
josh146: I'm surprised that the tests pass when constructing the ladder operators to size `cutoff x cutoff`. I would think that, due to the matrix exponentiation, there would be error in the last diagonal element of `gate`.

@co9olguy, any thoughts?

tests/backend/test_nongaussian_gates.py
egbQuantum: No, `cutoff` works fine in `test_kerr_interaction` because you also use `cutoff` instead of `sqrt(cutoff)` in the reference state `ref_state = np.exp(1j * kappa * np.arange(cutoff) ** 2) / cutoff`, so everything is consistent. 

On the other hand in `test_cubic_phase` I decided to start with the correctly normalized reference state `ket1 /= np.linalg.norm(ket1)`, so I needed to use `sqrt(cutoff)`.

tests/backend/test_nongaussian_gates.py
egbQuantum: Indeed! Thanks.

tests/backend/test_nongaussian_gates.py
egbQuantum: Cool, thanks. I'll use that next time :)

tests/backend/test_nongaussian_gates.py
egbQuantum: I believe that, almost by definition, the action of a gate on a cutoff Hilbert space is necessarily generated by the same gate Hamiltonian but just reduced to the cutoff dimension. I think the error you're thinking of is exactly the numerical error we get from using a cutoff (which for cubic phase is particularly bad due to the 3rd order of the Hamiltonian).

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
+        #Construct (unnormalized) x matrix, ie a+a^dag, and it's third power
+        x = a + np.transpose(a)
+        x3 = np.matmul(x, np.matmul(x, x))
+
+        gate = expm(1j * gamma * x3 / 6)
+
+        #state to be transformed
+        ket1 = np.ones([cutoff])
+        ket1 /= np.linalg.norm(ket1)
+        ket = ket1.reshape(-1, 1)
```
co9olguy: Just nitpicking, but you could also do this in one (still readable) line: 
```
ket = np.ones([cutoff,1]) / np.sqrt(cutoff)
```

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
...
+
+        ref_state = np.matmul(gate, ket).flatten()
```
co9olguy: `np.matmul` should support matrix-vector multiplication without needing to artificially expand the dimensions of the vector. How come you did it this way?

tests/backend/test_nongaussian_gates.py
```diff
@@ -18,13 +18,49 @@
 
 import numpy as np
 
-KAPPAS = np.linspace(0, 2 * np.pi, 7)
+from scipy.linalg import expm
 
+KAPPAS = np.linspace(0, 2 * np.pi, 7)
+GAMMAS = np.linspace(0, 2 * np.pi, 7)
```
co9olguy: :laughing:
No need to use `pi` here, and might needlessly confuse people . Could just do something like `np.linspace(0, 6, 7)

co9olguy:
```suggestion
GAMMAS = np.linspace(0, 6, 7)
```

egbQuantum: You're right, I was thinking to be consistent but it doesn't make sense to think of gamma as an angle. Thanks man!

egbQuantum: I've changed it in the new commit.

tests/backend/test_nongaussian_gates.py
egbQuantum: Yeah you're right there's no reason. I started by using copied code and didn't optimize after the fact. Thanks for catching! This is fixed in the new commit.

tests/backend/test_nongaussian_gates.py
egbQuantum: Done :)

PR #163, opened 2019-08-15
Symbolic Operation parameters using SymPy

### LINKED ISSUE #61: Refactor the Parameter class tests
Currently, the `test_parameter_operations` integration test in the test_parameters_integration.py` file this is very hard to understand. It should be split up, and separated into many individual tests that ideally test one thing only. Furthermore, it doesn't make any assertions - these should also be added in to ensure that, not only does the test run without error, but it is also doing what it is supposed to do.

Another small note: in the original test suite, this test appears to work on the Gaussian backend, even though it never explicitly ignores non-Gaussian operations - something to double check.

In addition, the parameters unit tests need to be expanded, to verify more of the Parameter class functionality.

### LINKED ISSUE #82: Refactor `p.x` into `p.val`
2. Let's mark to change this to `p.val`, as per @co9olguy's suggestion. It is more descriptive than `p.x`.

_Originally posted by @josh146 in https://github.com/XanaduAI/strawberryfields/pull/80_

### LINKED ISSUE #84: Make TFBackend.reset() accept the batch_size parameter
`TFBackend.reset()` should be able to change the batching behavior of the backend by accepting the `batch_size` argument, like `TFBackend.__init__()` does.

This would allow `TFBackend.reset()` and `TFBackend.__init__()` to take the same set of input arguments. The corresponding Fock and Gaussian backend methods already do this.

Currently `LocalEngine` has to pop `batch_size` out of the `backend_options` dictionary before it can be passed to `backend.reset()`.

smite:
**Context:**

Currently `Operation` instances can use either fixed or measured values as parameters through the `Parameter` class. Arbitrary algebraic transformations are provided by the `RegRefTransform` class.

* `Parameter` class is replaced by `MeasuredParameter` and `FreeParameter`, both inheriting from `sympy.Symbol`. Fixed numeric parameters are handled by the built-in Python numeric classes and numpy arrays.

* `Parameter`, `RegRefTransform` and `convert` are removed.

**Benefits:**

`Programs` can now have free parameters/arguments which are only bound to numerical values when the Program is executed, by supplying the actual argument values to the `Engine.run` method.

Sympy implements a wide array of algebraic manipulations and mathematical functions that act on the parameters, and handles the printing and evaluation of the symbolic expressions, making the implementation in `parameters.py` quite short.

Since RegRefTransforms are replaced by compound symbolic expressions, they now can be easily converted to strings during Program serialization into Blackbird code. **The corresponding deserialization step does not yet work for MeasuredParameters.**

Operations (gates) can now be decomposed symbolically. The only exception is `Decomposition` subclasses, since they use complicated numerical matrix decompositions.

**Possible Drawbacks:**

* Sympy comes with the occasional underdocumented/surprising feature, and also has its own set of bugs. However, all interaction with Sympy happens through the `strawberryfields.parameters` module.
* `CXgate._decompose` needs `sympy.functions.atan2`, but in Sympy 1.4 it has a bug which prevents us from using it. The bug has been fixed, and the fix should be included in Sympy 1.5. For the time being we use a workaround which cannot handle the `s=0` case (this results in division by zero).

**Related GitHub Issues:**

Resolves #61, closes #82, closes #84


co9olguy:
@smite Is this ready for review? Or just feedback/discussion?

smite:
Just discussion for now. It's a draft PR, I'll mark it ready for review once all features are working.

smite:
The CXgate test is failing because of a bug in SymPy (the atan2 function specifically). I've reported it and it's being fixed. In the meantime I'll implement a workaround in SF.

josh146:
Just curious if you have a link to the GitHub issue? :laughing:

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -48,12 +48,10 @@ class Circuit:
...
          The state of the modes is manipulated by calling the various methods."""
     # pylint: disable=too-many-instance-attributes,too-many-public-methods
-    def __init__(self, graph, num_modes, cutoff_dim, pure=True, batch_size=None):
+    def __init__(self, graph, num_modes, cutoff_dim, pure, batch_size):
         self._graph = None # will be set when reset is called below, but reset needs something to compare to
-        self._batch_size = batch_size
-        self._batched = False if batch_size is None else True
         self._hbar = 2
-        self.reset(pure, graph, num_subsystems=num_modes, cutoff_dim=cutoff_dim)
+        self.reset(graph=graph, num_subsystems=num_modes, pure=pure, cutoff_dim=cutoff_dim, batch_size=batch_size)
```
smite: This was a tangentially related issue, now `TFBackend.reset` can also set/change the batch size.

strawberryfields/program.py
```diff
@@ -574,3 +577,52 @@ def draw_circuit(self, tex_dir='./circuit_tex', write_to_file=True):
             document = drawer.compile_document(tex_dir=tex_dir)
 
         return [document, tex]
+
+
+    def args(self, *args):
```
smite: Maybe this isn't a great name for the method? Especially since the argument has the same name.

strawberryfields/backends/tfbackend/circuit.py
co9olguy: Issue #84

strawberryfields/program.py
co9olguy: Feel like ``params`` is probably the best name. That's what we usually use in pennylane, and it ties to "parameterized circuits" as well

strawberryfields/engine.py
```diff
@@ -213,7 +213,7 @@ def _run_program(self, prog, **kwargs):
             list[Command]: commands that were applied to the backend
         """
 
-    def _run(self, program, *, compile_options=None, **kwargs):
+    def _run(self, program, *, args, compile_options, **kwargs):
```
co9olguy: See comment in `program.py` about using `params` instead of `args`

strawberryfields/engine.py
```diff
@@ -338,13 +337,14 @@ def _run_program(self, prog, **kwargs):
                 raise NotImplementedError('The operation {} has not been implemented in {} for the arguments {}.'.format(cmd.op, self.backend, kwargs)) from None
         return applied
 
-    def run(self, program, *, compile_options=None, run_options=None):
+    def run(self, program, *, args=None, compile_options=None, run_options=None):
```
co9olguy: ditto

strawberryfields/ops.py
```diff
@@ -584,6 +556,7 @@ class Decomposition(Operation):
     into a sequence of gates and state preparations.
 
     The first parameter p[0] of the Decomposition is always a square matrix.
+    TODO p[0] cannot be a symbolic parameter yet...
```
co9olguy: should be put as a comment in code, not in docstring

strawberryfields/ops.py
```diff
@@ -640,27 +613,24 @@ class Channel(Transformation):
...
-            if np.all(self.p[1:] != other.p[1:]):
-                raise MergeFailure('Other parameters differ.')
-
-            # determine the new loss parameter
-            T = self.p[0] * other.p[0]
+        if not self.__class__ == other.__class__:
+            raise MergeFailure('Not the same channel family.')
 
+        # channels can be merged if they are the same class and share all the other parameters
+        if self.p[1:] == other.p[1:]:
```
co9olguy: I suppose at one stage, we might consider merging gates which differ in a parameter other than the first?

strawberryfields/ops.py
```diff
@@ -640,27 +613,24 @@ class Channel(Transformation):
...
+            # determine the combined first parameter
+            T = self.p[0] * other.p[0]
             # if one, replace with the identity
             if T == 1:
                 return None
 
-            if len(self.p) == 1:
-                return self.__class__(T)
+            # return a copy
+            # NOTE deepcopy would make copies the parameters which would mess things up
```
```suggestion
            # NOTE deepcopy would make copies of the parameters which would mess things up
```

strawberryfields/ops.py
```diff
@@ -725,66 +694,44 @@ def apply(self, reg, backend, **kwargs):
...
-            # can only merge them if they are each others' inverses
-            if self.p == other.p and self.dagger != other.dagger:
-                return None
+        # gates can be merged if they are the same class and share all the other parameters
+        if self.p[1:] == other.p[1:]:
+            # make sure the gates have the same dagger flag, if not, invert the second p[0]
+            if self.dagger == other.dagger:
+                temp = other.p[0]
+            else:
+                temp = -other.p[0]
```
co9olguy: I'm a bit confused by the situation here. Is this because this method gets called like `G.merge(H)`?

strawberryfields/ops.py
```diff
@@ -1349,14 +1302,15 @@ class BSgate(Gate):
...
+    def __init__(self, theta=np.pi/4, phi=0.):
         # default: 50% beamsplitter
         super().__init__([theta, phi])
 
     def _apply(self, reg, backend, **kwargs):
-        t = cos(self.p[0])
-        r = sin(self.p[0]) * exp(1j * self.p[1])
-        backend.beamsplitter(t.x, r.x, *reg)
+        p = par_evaluate(self.p)
+        tr = (np.cos(p[0]),
```
co9olguy: I don't really see the benefit of making the tuple here. Suggest reverting back to two variables

strawberryfields/ops.py
```diff
@@ -1435,16 +1389,17 @@ def __init__(self, s=1):
...
-        r = arcsinh(-s/2)
-        theta = 0.5*arctan2(-1.0/cosh(r), -tanh(r))
-
-        BS1 = BSgate(theta, 0)
-        BS2 = BSgate(theta+pi/2, 0)
+        r = pf.asinh(-s/2)
+        #theta = 0.5*pf.atan2(-1.0/pf.cosh(r), -pf.tanh(r))
+        # FIXME in sympy 1.4 atan2._eval_evalf() has a bug, it does not work with Symbol._eval_evalf().
+        # This is a partial workaround for s>0. When sympy is fixed (in version 1.5?), go back to using pf.atan2.
+        # If s<0 we would need to add pi/2 to theta, but since s is symbolic at this point, it'd be complicated.
```
co9olguy: So what is our course of action in the case that s<0?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
-* value measured from the quantum register (:class:`.RegRefTransform`)
-* TensorFlow object (:class:`tf.Variable` or :code:`tf.placeholder`)
-
-
-The normal lifecycle of an Operation object and its associated Parameter instances is as follows:
+There are three basic types of parameters:
+
+1. Numerical parameters (bound and fixed): An immediate, immutable numerical object
+   (float, complex, int, array).
+   NOTE: currently implemented as-is, not encapsulated in a class. This can be changed if necessary,
```
co9olguy: More appropriate as a comment/TODO, rather than putting in docstring?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+   numerical values before the circuit can be executed on a hardware quantum device or a numeric
+   simulator. Represented by :class:`FreeParameter` instances.
+   Simulators with symbolic capability can accept a parametrized circuit as input (and should
+   return symbolic expressions representing the measurement results, with the same free parameters,
+   as output).
+   Free parameters belong to a single :class:`Program` instance, are constructed using the
+   :meth:`Program.args` method, and are bound using :meth:`Program.bind_args`.
+
+The Operations can accept parameters that are functions or algebraic combinations of any number of
+these basic types of parameters, made possible by the measured and free parameters inheriting
```
co9olguy:
```suggestion
these basic types of parameters, made possible by the measured and free parameters inheriting from
```

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+  the Operation and the RegRefs it acts on, which is appended to :attr:`.Program.circuit`.
 
 * Before the Program is run, it is compiled and optimized for a specific backend. This involves
   checking that the Program only contains valid Operations, decomposing non-elementary Operations
   using :meth:`~ops.Operation.decompose`, and finally merging and commuting Commands inside
   the graph representing the quantum circuit.
   The circuit graph is built using the knowledge of which subsystems the Commands act and depend on.
 
-* Merging two :class:`.Gate` instances of the same subclass involves
+* Decompositions, merges and commutations often involve creation new Operations with algebraically
```
co9olguy:
```suggestion
* Decompositions, merges, and commutations often involve the creation of new Operations with algebraically
```

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
 import numpy as np
+import sympy
+import sympy.functions as sf
+
+
+def wrap_mathfunc(func):
+    """Applies the wrapped sympy function elementwise to numpy arrays.
+
+    Required because the sympy math functions cannot deal with arrays.
+    We implement no broadcasting; if the first argument is a numpy array, we assume
```
co9olguy: Any possible edge-case bugs you can foresee?

co9olguy: presumably `zip` would just give an error?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+    def wrapper(*args):
+        if isinstance(args[0], np.ndarray):
+            # apply func elementwise, recursively, on the args
+            return np.array([wrapper(*k) for k in zip(*args)])
+        return func(*args)
+    return wrapper
 
-from .program_utils import (RegRef, RegRefTransform)
+# namespace of mathematical functions for manipulating Parameters
+parfuncs = types.SimpleNamespace(**{name: wrap_mathfunc(getattr(sf, name)) for name in dir(sf) if name[0] != '_'})
```
co9olguy: Will this effect how various objects in the sf namespace are printed??

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+    """Returns True iff p is an object array."""
+    return isinstance(p, np.ndarray) and p.dtype == object
 
-def _unwrap(params):
-    """Unwrap a parameter sequence.
+
+def par_evaluate(params):
+    """Evaluate a parameter sequence.
+
+    Any parameters descending from sympy.Basic are evaluated, others are returned as is.
```
co9olguy:
```suggestion
    Any parameters descending from sympy.Basic are evaluated, others are returned as-is.
```

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+
+    def do_evaluate(p):
+        if is_object_array(p):
+            return np.array([do_evaluate(k) for k in p])
+
+        if not par_is_symbolic(p):
+            return p
+
+        #p = p.evalf()  # TODO sympy 1.4 has some bugs which prevent using this for now, workaround below
+        if 0:
```
co9olguy: this block never gets executed?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+            # TODO the float() conversion below prevents symbolic params from being passed through, maybe the backend should do float() conversion?
+            if not p.is_real:
+                return complex(p)
+            if p.is_integer:
+                return int(p)
+            return float(p)
+
+        # using lambdify we can also substitute np.ndarrays and tf.Tensors for the atoms
+        atoms = list(p.atoms(MeasuredParameter, FreeParameter))
+        func = sympy.lambdify(atoms, p, 'numpy')
```
co9olguy: How does this work if `p` contains tensorflow variables?

co9olguy: Or is this just a recognition that when we are evaluating, the output should be a numpy object?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+    """
+    ret = set()
+    if is_object_array(p):
+        # p is an object array, possibly containing symbols
+        for k in p:
+            ret.update(par_regref_deps(k))
+    elif isinstance(p, sympy.Basic):
+        # p is a Sympy expression, possibly containing measured parameters
+        for k in p.atoms(MeasuredParameter):
+            ret.add(k.regref)
```
co9olguy: how come `add()` is used here, but `update` is used above?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
-        """
-        if isinstance(self.x, RegRefTransform):
-            return Parameter(self.x.evaluate())
-        return self
 
-    def _unwrap_and_cast(self, other):
-        """Unwrap Parameters and cast TensorFlow-type parameters to other dtypes during arithmetic.
 
-        The main reason we need this is that TensorFlow does not automatically promote int to float or float to complex but requires an explicit cast.
+#class MeasuredParameter(sympy.AtomicExpr):  # something is messed up in Sympy caching, maybe, frontend tests fail depending on their execution order
```
co9olguy: What is current status of this line?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
-    # comparisons
-    def __eq__(self, other):
-        """Equality comparison.
-
-        .. note:: This method may be too permissive, maybe it should return False if either parameter is not a numbers.Number or a np.ndarray?
+    Used for deferred measurements, i.e., using a measurement's value
+    symbolically in defining a gate before the numeric value of that
+    measurement is available.
+
+    Former RegRefTransform functionality is provided by the sympy.Symbol base class.
```
co9olguy: maybe make a note about version, e.g., `RegRefTransform (SF<=0.11)`

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+
+    Args:
+        regref (RegRef): register reference responsible for storing the measurement result
+    """
+
+    def __new__(cls, regref):
+        # sympy.Basic.__new__ wants a name, other arguments must not end up in self._args
+        return super().__new__(cls, 'q'+str(regref.ind))
```
co9olguy: I guess we're forced to pick a convention here, even though the user is not forced to use ``q``?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+
+    def __init__(self, regref):
+        if not regref.active:
+            # TODO: Maybe we want to delete a mode right after measurement to save comp effort?
+            # The measurement result would still be preserved in the RegRef...
+            raise ValueError('Trying to use an inactive RegRef.')
+        #: RegRef: the parameter value depends on this RegRef, it can only be evaluated after the corresponding subsystem has been measured
```
co9olguy: something off about this comment?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+        self.regref = regref
+
+    def _sympystr(self, printer):
+        """"The Sympy printing system uses this method instead of __str__."""
+        return str(self.regref) + '.par'
+
+    def _eval_evalf(self, prec):
```
co9olguy: convention from `sympy`? Irksome name

strawberryfields/backends/tfbackend/backend.py
```diff
@@ -128,14 +128,14 @@ def reset(self, pure=True, **kwargs):
                 If False, then the circuit is reset to its initial state, but ops that
                 have already been declared are still accessible.
         """
-        hard = kwargs.get('hard', True)
+        hard = kwargs.pop('hard', True)
```
josh146: Just out of curiosity, what is `hard`? Does it refer to a hard reset?

strawberryfields/circuitspecs/chip0.py
```diff
@@ -19,6 +19,7 @@
 from scipy.linalg import block_diag
 
 from strawberryfields.program_utils import CircuitError, Command, group_operations
+from strawberryfields.parameters import par_evaluate
```
josh146: `par_evaluate` is a function which evaluates symbolic parameters, but in a 'vectorized' fashion? So it works with single parameters, as well as sequences of parameters?

strawberryfields/engine.py
josh146: So `args` always needs to be passed here? i.e., `_run(prog, args={}, compile_options={})`? Previously, `compile_options` had a default value `None`. Curious what the reason for this change is.

strawberryfields/io.py
```diff
@@ -70,35 +71,15 @@ def to_blackbird(prog, version="1.0"):
...
-                            "The RegRefTransform in operation {} "
-                            "is not supported by Blackbird.".format(cmd.op)
-                        )
-                    # else:
-                    op["args"].append(a.x)
+            for a in cmd.op.p:
+                if par_is_symbolic(a):
+                    # SymPy object, convert to string
+                    a = str(a)
+                op["args"].append(a)
```
josh146: Nice! :+1:

strawberryfields/ops.py
```diff
@@ -87,28 +87,15 @@
   i.e., we may symbolically use the measurement result before it exists::
 
     with prog.context as (alice, bob):
-        MeasureFock()| alice
-        Dgate(alice) | bob
+        MeasureFock()    | alice
+        Dgate(alice.par) | bob
```
josh146: This is a pretty big UI change (requiring `alice.par` rather than `alice` for measurement values) - is it possible to retain the previous behaviour? It would also require changes across the tutorials and blackbird

strawberryfields/ops.py
```diff
@@ -87,28 +87,15 @@
...
-  an optional lambda function can be used to achieve the same functionality::
+  One may also use classical algebraic post-processing on the measurement result, to be applied
+  before using it as the argument to another :class:`Operation`::
 
     with prog.context as q:
-        MeasureFock()   | q[0]
-        Dgate(RR(q[0])) | q[1]
-        Dgate(RR(q[0], lambda q: q ** 2)) | q[2]
+        MeasureFock()        | q[0]
+        Dgate(q[0].par ** 2) | q[1]
```
josh146: Much nicer not needing to use the previous @convert or RR functions though!

strawberryfields/ops.py
```diff
@@ -435,14 +416,13 @@ def decompose(self, reg, **kwargs):
...
-        # todo: For now decompose() works on unevaluated Parameters.
-        # This causes an error if a :class:`.RegRefTransform`-based Parameter is used, and
-        # decompose() tries to do arithmetic on it.
         return self._decompose(reg, **kwargs)
 
     def _decompose(self, reg, **kwargs):
         """Internal decomposition method defined by subclasses.
 
+        NOTE: Does not evaluate Operation parameters, symbolic parameters remain symbolic.
+
```
josh146: 😮

strawberryfields/ops.py
```diff
@@ -601,7 +574,7 @@ def merge(self, other):
             # decompositions, which cannot be merged.
             U1 = self.p[0]
             U2 = other.p[0]
-            U = matmul(U2, U1).x
+            U = np.matmul(U2, U1)
```
josh146: We could even go back to 
```suggestion
            U = U2 @ U1
```
now that we no longer dynamically replace the matmul function, right?

strawberryfields/ops.py
```diff
@@ -725,66 +694,44 @@ def apply(self, reg, backend, **kwargs):
...
         if np.all(z == 0):
             # identity, no need to apply
             return
         if self.dagger:
             z = -z
-        original_p = self.p  # store the original Parameters
-        # evaluate the rest of the Parameters, restore the originals later
-        self.p = [z] + [x.evaluate() for x in self.p[1:]]
+        original_p0 = self.p[0]  # store the original Parameter
+        self.p[0] = z
```
josh146: The parameter code is definitely cleaner than it was before

strawberryfields/ops.py
josh146: Yep

strawberryfields/ops.py
josh146: Concur. It adds a layer of extra complexity when reading the code.

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
-
-
-The normal lifecycle of an Operation object and its associated Parameter instances is as follows:
+There are three basic types of parameters:
+
+1. Numerical parameters (bound and fixed): An immediate, immutable numerical object
+   (float, complex, int, array).
+   NOTE: currently implemented as-is, not encapsulated in a class. This can be changed if necessary,
+   in which case most of the top-level functions in this module would become Parameter methods.
+2. Measured parameters (bound but not fixed): Certain quantum circuits/protocols require that
```
josh146:
```suggestion

2. Measured parameters (bound but not fixed): Certain quantum circuits/protocols require that
```

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+   Operations can be conditioned on measurement results obtained during the execution of the
+   circuit. In this case the parameter value is not known/fixed until the measurement is made
+   (or simulated). Represented by :class:`MeasuredParameter` instances.
+   Constructed from the :class:`RegRef` instance storing the measurement
+   result using the :meth:`RegRef.par` method.
+3. Free parameters (not bound nor fixed): A *parametrized circuit template* is a circuit that
```
josh146:
```suggestion

3. Free parameters (not bound nor fixed): A *parametrized circuit template* is a circuit that
```

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+   depends on a number of unbound (free) parameters. These parameters need to be bound to fixed
+   numerical values before the circuit can be executed on a hardware quantum device or a numeric
+   simulator. Represented by :class:`FreeParameter` instances.
+   Simulators with symbolic capability can accept a parametrized circuit as input (and should
+   return symbolic expressions representing the measurement results, with the same free parameters,
+   as output).
+   Free parameters belong to a single :class:`Program` instance, are constructed using the
+   :meth:`Program.args` method, and are bound using :meth:`Program.bind_args`.
```
josh146: This is a completely new feature in this PR, right? Are there details on how it is accessible from the frontend once there are compatible backends? Or even, what a compatible backend must do?

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+Functions
+---------
+
+.. currentmodule:: strawberryfields.parameters
+
+.. autosummary::
+   par_evaluate
+   par_is_symbolic
+   par_regref_deps
+   par_str
```
josh146: Very minor comment, but do these need to all start with `par`? Being in the parameter module/namespace, it should be implicit

strawberryfields/parameters.py
josh146: Actually, what change does this make to the namespace? It might be worth adding a lot more info to the comment

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+# namespace of mathematical functions for manipulating Parameters
+parfuncs = types.SimpleNamespace(**{name: wrap_mathfunc(getattr(sf, name)) for name in dir(sf) if name[0] != '_'})
 
 
-try:
-    import tensorflow as tf
-    _tf_classes = (tf.Tensor, tf.Variable)
-except ImportError:
-    tf = mock.MagicMock()
-    _tf_classes = tuple()
```
josh146: So the parameter class no longer has any TF dependency?? 😀

strawberryfields/parameters.py
```diff
@@ -19,335 +19,334 @@
...
+        params = [params]
+
+    def do_evaluate(p):
+        if is_object_array(p):
+            return np.array([do_evaluate(k) for k in p])
+
+        if not par_is_symbolic(p):
+            return p
+
+        #p = p.evalf()  # TODO sympy 1.4 has some bugs which prevent using this for now, workaround below
```
josh146: Could you add the link to the issue here, so that we know when it can be uncommented?

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,156 @@
...
+        """String representations of parameters."""
+
+        a = 0.1234567
+        b = np.array([0, 0.987654])
+        c = MeasuredParameter(RegRef(1))
+        d = FreeParameter('x')
+
+        assert par_str(a) == '0.1235'  # rounded to 4 decimals
+        assert par_str(b) == '[0.     0.9877]'
+        assert par_str(c) == 'q[1].par'
```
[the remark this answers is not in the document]
josh146: At the same time, I don't think the users variable name will ever be accessible. I think this is acceptable, especially because we use `q` in almost all documentation.

Note: In 'pure' blackbird, we force printed regref names to be q0, q1, etc. Might be worth doing the same in SF?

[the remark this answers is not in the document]
co9olguy: Might make sense to not make it look like a python object. Just stick to `q0`,`q1`, etc

strawberryfields/backends/tfbackend/backend.py
co9olguy: yes

co9olguy: otherwise, it will make a separate circuit simulation, but in the same underlying graph (i.e., the old circuit objects are still in tensorflow graph)

strawberryfields/ops.py
co9olguy: I don't see it as a backbreaker. It's 3 extra lines and still pretty clear what's happening. Bigger question is the second one: what tutorials will we have to update?

strawberryfields/parameters.py
co9olguy: :+1:

strawberryfields/__init__.py
```diff
@@ -66,11 +66,10 @@
 """
 from .engine import (Engine, LocalEngine)
 from .io import save, load
-from .program_utils import _convert as convert
 from .program import Program
 from ._version import __version__
 
-__all__ = ["Engine", "LocalEngine", "Program", "convert", "version", "save", "load", "about", "cite"]
```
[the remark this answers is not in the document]
smite: Yeah, RegRefTransforms no longer exist so the `convert` decorator is also unnecessary.

strawberryfields/circuitspecs/chip0.py
smite: Right.

strawberryfields/engine.py
smite: I think default values are not very useful for "internal use only" methods, especially here since `BaseEngine._run` is only called by `LocalEngine.run`, which provides its own default values anyway.

Depending on how the remote engines work it might also be a good idea to merge the BaseEngine and LocalEngine classes.

strawberryfields/io.py
smite: What should `op["args"]` contain?

The operation parameters in SF can now be (1) symbolic (incl. arrays with symbols in them), (2) numeric arrays, or (3) python numeric scalars. The code here converts case (1) to strings but stores other types as is, is that correct?

strawberryfields/ops.py
smite: Making `Dgate(alice)` work would be easy (just have `Operation.__init__` convert the RegRef into a MeasuredParameter). The hard case is something like `Dgate(2 * alice.par**2)`. Here we cannot drop `.par` because `alice` alone is a RegRef instance and does not implement arithmetic operations. I think it's clearer if we always require the `.par`.

strawberryfields/ops.py
smite: It's a restriction of the Decomposition class which I think cannot be easily overcome since we use various numerical matrix decompositions on `p[0]`, the user needs to be aware of it. I'll remove the words TODO and yet :)

strawberryfields/ops.py
smite: I think so, yes.

strawberryfields/ops.py
smite: Maybe, but there's no general rule for it. `Gate` subclasses are guaranteed to be one-parameter subgroups w.r.t. `p[0]` but nothing more, and for `Channel` subclasses we have not even properly defined how `p[0]` should be treated...

strawberryfields/ops.py
smite: Right

strawberryfields/ops.py
smite: Hmm, maybe add `sympy.Heaviside(-s)*np.pi/2` to theta? But Heaviside is undefined at zero, which could also cause some problems.

Another option would be to evaluate the parameters right here until sympy is fixed (should happen when the next version is released this fall), and leave a note to warn the user of the unexpected behavior.

strawberryfields/parameters.py
smite: I'll reword it better.

strawberryfields/parameters.py
smite: This is a frontend-only feature, `Engine.run` binds all the free parameters. However, they could be bound to e.g. TensorFlow objects which are then passed to the backend, in which case the backend must be able to deal with them.

strawberryfields/parameters.py
smite: I wanted to make the context explicit so it's easy to search the function names from the codebase (evaluate and str would otherwise show up a lot)

strawberryfields/parameters.py
smite: Here `sf` is `sympy.functions`, I'm wrapping all the public members of that module and making them available to the user in the `strawberryfields.parameters.parfuncs` namespace. I added a docstring which explains this.

strawberryfields/parameters.py
smite: None!

strawberryfields/parameters.py
smite: See josh's comment above

strawberryfields/parameters.py
smite: I left this comment and the `if 0:` block below to show the approaches that did not work, for discussion.
I'll remove them before this PR is merged.

smite: It seems `sympy.lambdify` works best, it allows us to substitute anything for the atoms in the symbolic expressions.

strawberryfields/parameters.py
smite: Good question. Roughly speaking `lambdify` converts the symbolic expression `p` into a string of python code, and then evaluates that code in a specific namespace (which provides definitions for the various functions) to create a lambda function that substitutes the passed argument values for `atoms`.

Here that namespace is `numpy`. This means that you probably cannot apply functions such as `sin` (which is translated to `np.sin`) to TF objects, but basic arithmetic operations should work. 

lambdify could also use `tensorflow` as the namespace if needed, but I'm not sure what would be the best way to choose the namespace.

https://docs.sympy.org/latest/modules/utilities/lambdify.html

strawberryfields/parameters.py
smite: `add` adds a single element to the set, `update` adds all the elements of another set.

smite: We handle object arrays recursively.

strawberryfields/parameters.py
smite: Should be removed, inheriting from sympy.Symbol seems to work fine.

strawberryfields/parameters.py
smite: Ahhh, not exactly. As far as I can tell this is just an internal name used by sympy, which could be anything. The name  used when doing a string conversion is defined in the `_sympystr` method below.

strawberryfields/parameters.py
smite: Yeah. `sympy.Symbol.evalf()` eventually calls this method. Technically we do not have to use it since we call `_eval_evalf` directly, so I can change the name if you like.

tests/frontend/test_ops_gate.py
```diff
@@ -159,24 +159,32 @@ def dummy_apply(self, reg, backend, **kwargs):
...
         ops.MeasureX | q[0]
-        D = ops.Dgate(q[0])
-        F = ops.Dgate(q[0], 0.1)
+        mpar = q[0].par  # measured parameter
+        D = ops.Dgate(mpar)
+        F = ops.Dgate(1.0)
+        G = ops.Dgate(mpar, 0.1)  # different p[1]
 
-    # gates that are the inverse of each other
+    # mp gates that are the inverse of each other
```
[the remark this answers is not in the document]
smite: measured parameter

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,156 @@
...
 
+import pytest
 import numpy as np
 
-import strawberryfields as sf
-from strawberryfields import ops
-from strawberryfields.parameters import Parameter
-
-# make test deterministic
-np.random.random(32)
```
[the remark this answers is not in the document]
smite: At least pylint warns about these, we'll root them out eventually :D

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,156 @@
...
-TEST_VALUES = [3, 0.14, 4.2 + 0.5j, np.random.random(3)]
-
-
-try:
-    import tensorflow as tf
-except (ImportError, ModuleNotFoundError) as e:
-    tf_available = False
-else:
-    tf_available = True
-    TEST_VALUES.extend([tf.Variable(2), tf.Variable(0.4), tf.Variable(0.8 + 1.1j)])
```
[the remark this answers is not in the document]
smite: The TF tests were moved to `tests/integration/test_tf_symbolic.py`, where we use one of these try: import blocks.

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,156 @@
...
+        c = MeasuredParameter(RegRef(1))
+        d = FreeParameter('x')
+
+        assert str(c) == 'q[1].par'
+        assert str(d) == 'x'
+        assert str(0.1234567 * d) == '0.1234567*x'
+        assert str(np.array([0, 1, -3, 0.987654]) * d) == '[0 1.0*x -3.0*x 0.987654*x]'
+        assert str(pf.exp(1 + c) / d ** 2) == 'exp(q[1].par + 1)/x**2'
+
+    @pytest.mark.parametrize("p", TEST_VALUES)
```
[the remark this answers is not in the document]
smite: `binary_arithmetic` has some asserts in it.

tests/integration/test_parameters_integration.py
```diff
@@ -13,154 +13,124 @@
...
-    # Fock requires an integer parameter
-    testset = set(
-        ops.one_args_gates + ops.two_args_gates + ops.channels + scalar_arg_preparations
-    ) - {ops.ThermalLossChannel}
-    print(testset)
+        prog.circuit = []
+        eng.reset()
 
     for G in testset:
+        print(G)
```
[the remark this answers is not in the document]
smite: pytest only shows stdout when the test fails, this is pretty useful in determining which gate caused the failure.

strawberryfields/program.py
smite: I'll change it.

strawberryfields/parameters.py
smite: Yeah. Most of the math funcs only take a single argument anyway.

tests/integration/test_parameters_integration.py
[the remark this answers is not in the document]
smite: I guess you could also remove the for loop from the test by using `pytest.mark.parametrize` with a cleverly designed fixture.

strawberryfields/parameters.py
smite: Is it better now?

co9olguy:
@smite let us know when this PR is ready for final review :+1:

tests/integration/test_parameters_integration.py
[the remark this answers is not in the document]
smite: I parametrized the test and removed the explicit loop.

strawberryfields/parameters.py
smite: Actually zip will not give an error, it'll just stop when the first iterable is exhausted. I added some checks to fix this.

tests/frontend/test_parameters.py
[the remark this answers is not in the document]
smite: This is a slightly larger issue, `RegRef.__str__` also returns 'q[1]` etc at the moment.

strawberryfields/ops.py
```diff
@@ -1408,11 +1409,13 @@ def __init__(self, s=1):
...
+        #theta = 0.5 * pf.atan2(-1.0 / pf.cosh(r), -pf.tanh(r))
         # FIXME in sympy 1.4 atan2._eval_evalf() has a bug, it does not work with Symbol._eval_evalf().
-        # This is a partial workaround for s>0. When sympy is fixed (in version 1.5?), go back to using pf.atan2.
-        # If s<0 we would need to add pi/2 to theta, but since s is symbolic at this point, it'd be complicated.
-        theta = 0.5 * pf.atan(1 / pf.sinh(r))
+        # This is a workaround. When sympy is fixed (in version 1.5?), go back to using pf.atan2.
+        # If s<0 we need to add pi/2 to theta. If s==0, we need to avoid division by zero.
+        theta = pf.Piecewise((0.5 * pf.atan(1 / pf.sinh(r)), s > 0),
+                             (-np.pi/4, s == 0),
+                             (0.5 * pf.atan(1 / pf.sinh(r)) +np.pi/2, True))
```
co9olguy: Should it be `-np.pi/2` instead?
If `s<0`, then `r>0`, meaning `y<0` and `x<0`. From [wikipedia](https://en.wikipedia.org/wiki/Inverse_trigonometric_functions), when both `x` and `y` are less than zero, we get `atan2(y,x)=atan(y/x)-pi`. Am I mistaken somehow?

tests/frontend/test_parameters.py
```diff
@@ -67,7 +69,23 @@ def test_par_is_symbolic(self, r):
         assert par_is_symbolic(p * r)
         assert par_is_symbolic(p / r)
         assert par_is_symbolic(p ** r)
-        assert par_is_symbolic(p / p)  # no simplification
+        assert par_is_symbolic(p - p)  # no simplification
```
co9olguy: How come the `p/p` case was dropped? Did it actually provide a simplification? 
If not, why not keep it in tests? If so, why does `p-p` not do the same?

tests/frontend/test_parameters.py
```diff
@@ -114,6 +132,46 @@ def test_raw_parameter_printing(self):
         assert str(np.array([0, 1, -3, 0.987654]) * d) == '[0 1.0*x -3.0*x 0.987654*x]'
         assert str(pf.exp(1 + c) / d ** 2) == 'exp(q[1].par + 1)/x**2'
 
+    def test_par_functions_with_arrays(self):
+        """Parameter functions with array arguments."""
+        a = np.random.rand(2, 2)
+        b = np.random.rand(2, 3)
+        c = np.random.rand(2)
+        d = np.random.rand(2, 2)
```
co9olguy: No need for `random` here

tests/integration/test_parameters_integration.py
```diff
@@ -11,156 +11,129 @@
...
-            other_inputs.append(np.random.uniform(size=(batch_size,)))
 
-    other_pars = tuple(Parameter(k) for k in other_inputs)
+@pytest.mark.parametrize("G", testset)
+def test_parameters_with_operations(eng_prog_params, G):
+    """Test all combinations of different types of Parameters with different Operation subclasses.
+
+    This test is successful if no exceptions are raised.
+
+    Some operation/backend combinations forbidden by the CircuitSpecs instance of the backend.
```
co9olguy:
```suggestion
    Some operation/backend combinations are forbidden by the CircuitSpecs instance of the backend.
```

tests/integration/test_parameters_integration.py
```diff
@@ -120,17 +129,11 @@ def check(G, par):
         prog.circuit = []
```
co9olguy: better to put these at the start of the function?

co9olguy:
Also, a couple tests are failing with the TF backend, with `TypeError: Argument must be a Basic object, not `ndarray``

strawberryfields/io.py
josh146: I believe it should be converting case (1) to a 'templated' blackbird variable, i.e., for `Symbol(s) -> "{s}"`

tests/frontend/test_parameters.py
smite: `p/p` should maybe not be simplified since `p` might be zero, but `p-p` always simplifies to zero for finite `p`.

strawberryfields/parameters.py
```diff
@@ -270,8 +289,6 @@ def par_str(p):
     return '{:.4g}'.format(p)  # scalar parameters
 
 
-
-#class MeasuredParameter(sympy.AtomicExpr):  # something is messed up in Sympy caching, maybe, frontend tests fail depending on their execution order
```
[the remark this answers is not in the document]
smite: The problem does not arise when we inherit from `sympy.Symbol`.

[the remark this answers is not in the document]
co9olguy: so ok to remove the comment now?

tests/frontend/test_parameters.py
co9olguy: good point

strawberryfields/parameters.py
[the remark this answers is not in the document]
smite: Yeah, removed.

strawberryfields/io.py
smite: This isn't happening at the moment. Free parameters are replaced with their name, measured parameters by "q[k].par".

strawberryfields/ops.py
smite: `theta` is only defined up to `n \pi`, that should account for the difference.

tests/frontend/test_parameters.py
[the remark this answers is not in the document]
smite: I changed it, and left Regref.__str__ as it was.

strawberryfields/ops.py
```diff
@@ -556,8 +556,7 @@ class Decomposition(Operation):
     This class provides the base behaviour for decomposing various multimode operations
     into a sequence of gates and state preparations.
 
-    The first parameter p[0] of the Decomposition is always a square matrix.
-    TODO p[0] cannot be a symbolic parameter yet...
```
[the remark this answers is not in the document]
smite: Not really. The matrix decompositions depend on numeric algorithms such as `np.linalg.svd`, which do not have reasonable symbolic counterparts. I can raise an exception if the user tries to use a symbolic parameter here.

strawberryfields/io.py
smite: So free params should be converted to "{parameter_name}" and measured parameters to "qx", is that correct? Do all the free params of a program have to be listed somewhere?

josh146: Yep that's correct, and no they don't need to be listed

josh146: See here: https://quantum-blackbird.readthedocs.io/en/latest/syntax.html#templates

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,198 @@
...
+
+        assert str(c) == 'q1'
+        assert str(d) == 'x'
+        assert str(0.1234567 * d) == '0.1234567*x'
+        assert str(np.array([0, 1, -3, 0.987654]) * d) == '[0 1.0*x -3.0*x 0.987654*x]'
+        assert str(pf.exp(1 + c) / d ** 2) == 'exp(q1 + 1)/x**2'
+
+    def test_par_functions_with_arrays(self):
+        """Parameter functions with array arguments."""
+        a = np.random.rand(2, 2)
```
co9olguy:
```suggestion
        a = np.ones((2, 2))
```

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,198 @@
...
+        b = np.random.rand(2, 3)
```
co9olguy:
```suggestion
        b = np.ones((2, 3))
```

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,198 @@
...
+        c = np.random.rand(2)
```
co9olguy:
```suggestion
        c = np.ones((2,))
```

tests/frontend/test_parameters.py
```diff
@@ -11,118 +11,198 @@
...
+        d = np.random.rand(2, 2)
```
co9olguy:
```suggestion
        d = np.ones((2, 2))
```

strawberryfields/io.py
smite: I just pushed the first half of compatibility with Blackbird template parameters (which correspond to FreeParameters) and measured parameters. I left a FIXME comment in `io.py` where the conversion of the symbolic expressions should take place, and two for-now skipped tests to test it.

strawberryfields/ops.py
```diff
@@ -1453,16 +1415,20 @@ def __init__(self, s=1):
...
-        r = arcsinh(-s/2)
-        theta = 0.5*arctan2(-1.0/cosh(r), -tanh(r))
+        r = pf.asinh(-s/2)
+        #theta = 0.5 * pf.atan2(-1.0 / pf.cosh(r), -pf.tanh(r))
+        # FIXME in sympy 1.4 atan2._eval_evalf() has a bug, it does not work with Symbol._eval_evalf().
+        # This is a workaround. When sympy is fixed (in version 1.5?), go back to using pf.atan2.
+        # See https://github.com/sympy/sympy/pull/17469
+        # If s<0 we need to add pi/2 to theta. If s==0, we need to avoid division by zero.
+        temp = 0.5 * pf.atan(1 / pf.sinh(r))  # NOTE s==0 will cause a division by zero when this is evaluated
+        theta = temp +pf.Heaviside(-s) * np.pi/2
```
josh146: @smite, @co9olguy. Is the above correct? When I plot line 1419 vs line 1424-1425, I get different results for s<0:

Screenshot_20191011_171531

josh146: changing the sign of the Heaviside function leads to the correct result.
```suggestion
        theta = temp - pf.Heaviside(-s) * np.pi/2
```

josh146: Screenshot_20191011_171741

josh146: @nquesada

smite: `theta` is only defined up to `n \pi` in the CXgate decomposition formula so I think in this case there is no practical difference between the two expressions.

tests/integration/test_decompositions_integration.py
```diff
@@ -554,7 +554,7 @@ def test_Pgate_decomp_equal(self, setup_eng, s, tol):
         eng.run(prog)
         assert np.all(eng.backend.is_vacuum(tol))
 
-    @pytest.mark.parametrize('s', np.linspace(0., 0.6, 5))
+    @pytest.mark.parametrize('s', np.linspace(-0.5, 0.6, 5))  # FIXME include 0 in testset
```
co9olguy: @smite What needs to be fixed here?

tests/integration/test_decompositions_integration.py
```diff
@@ -575,7 +575,7 @@ def test_CXgate_decomp_equal(self, setup_eng, s, tol):
...
```
co9olguy: same question

tests/integration/test_decompositions_integration.py
```diff
@@ -590,21 +590,20 @@ def test_CZgate_decomp_equal(self, setup_eng, s, tol):
         eng.run(prog)
         assert np.all(eng.backend.is_vacuum(tol))
 
-    @pytest.mark.parametrize('r', np.linspace(0., 0.6, 5))
+    @pytest.mark.parametrize('r', np.linspace(-0.3, 0.3, 5))
```
co9olguy: Does this one have the same problem as the FIXMEs above?

strawberryfields/ops.py
co9olguy: How come this was switched away from `piecewise`? According to the comment in code, we will get a division by zero now when s=0, whereas before we were accounting for it

smite: Currently Piecewise is implemented in a bad way in Sympy, it evaluates every expression regardless of the truth value of the conditions, and also causes a divide by zero error. It also does not work with batched input because `parameters.wrap_mathfunc` can't handle the more complicated way it treats its arguments.

tests/integration/test_decompositions_integration.py
smite: For now CXgate causes a divide-by-zero error when `s=0`. When Sympy 1.5 is released we can fix it, and then this test should include the value zero in the testset.

tests/integration/test_decompositions_integration.py
smite: Same answer :)

tests/integration/test_decompositions_integration.py
smite: No, this one's fine. I just changed the testset so that it also includes negative values.

tests/integration/test_decompositions_integration.py
josh146: Can you add this information to the FIXME comment

tests/integration/test_decompositions_integration.py
josh146: Also here (adding this information to the FIXME comment)

strawberryfields/ops.py
co9olguy: If the divide by zero error is happening for a case where the gate would have a parameter of 0 (i.e., identity), can we not catch that earlier with simple python logic and avoid the decomposition entirely?

smite: No, because at the point where the decomposition is made the quantities are still symbolic, we do not know their numeric values. But this entire issue should go away when atan2 is fixed in the next Sympy version.

tests/integration/test_decompositions_integration.py
smite: Done.

tests/integration/test_decompositions_integration.py
smite: Done.

euxsun:
A bug. Massive changes in `utils` module are not reflected in the docs, so almost every code example crashes with `convert` or `scale` functions call (f.e. `examples/QuantumTeleportation.ipynb`). There is no advice how to overcome these issues.

co9olguy:
Thanks for the report @euxsun. I think the examples folder is probably not sync'd up correctly with the latest changes. We are planning to change the behaviour of the examples, so they are automatically checked as a test before any merges take place. But we have not yet finished that yet.

In the meantime, if you are following along with the docs, I encourage you to downgrade to v0.11.1

euxsun:
Thank you! I confirm that downgrading works.

PR #164, opened 2019-08-19
Introduce applications layer

trbromley:
**Context:** Strawberry Fields allows users to design photonic quantum circuits and simulate or run on hardware. However, designing quantum circuits can be a big learning curve, and it makes sense to offer a higher level of abstraction where a user can input problems without having to specify the underlying quantum circuit.

**Description of the Change:** Add in first version of an applications layer aimed at solving problems using Gaussian boson sampling. This layer focuses on graph-based problems and currently has algorithms for the densest ``k``-subgraph problem.

We have chosen `future.apps` as the location for the applications layer-based code, i.e.,

- Python modules are added to `strawberryfields.future.apps`
- Documentation is added to `doc.code.future.apps`
- Tests are added to `tests.future.apps`

*Hint:* The commits for this PR are broken down into elementary steps, which should help in looking through this rather large change.

**What's missing:** The documentation has not been integrated into, e.g., `index.rst`, due to an incoming planned refactor of the docs by @josh146. A tutorial is also missing as this is written using sphinx-gallery.


strawberryfields/future/apps/graph/sample.py
```diff
@@ -0,0 +1,139 @@
...
+    sample_subgraphs
+    to_subgraphs
+
+Code details
+------------
+"""
+from typing import Optional
+import networkx as nx
+import numpy as np
+from .. import sample
```
zzzeid: I generally prefer absolute imports vs. relative imports (mostly because they are more readable), and also pep8 recommends! https://www.python.org/dev/peps/pep-0008/#imports

I know this is a migration PR but might be worthwhile to change all the import paths to be absolute as part of this PR. In this case, this line could be rewritten as:
```
from strawberryfields.future.apps import sample
```

Makefile
```diff
@@ -51,7 +51,7 @@ docs:
 clean-docs:
 	make -C doc clean
 
-test: test-frontend test-gaussian test-fock test-tf batch-test-tf
+test: test-frontend test-gaussian test-fock test-tf batch-test-tf test-apps
```
zzzeid: Should also add coverage commands for apps below.

setup.py
```diff
@@ -53,6 +53,7 @@
         "strawberryfields.backends.tfbackend",
         "strawberryfields.backends.fockbackend",
         "strawberryfields.backends.gaussianbackend",
+        "strawberryfields.future.apps"
```
zzzeid: Are there any changes to `requirements` above that should be added to this file?

Makefile
trbromley: Thanks, done!

setup.py
trbromley: It depends, the `requirements.txt` for the applications layer is here:
https://github.com/XanaduAI/glassonion/blob/master/requirements.txt

It has things like `pytest`, which is not included in SF. However, I put things like `pytest` in the apps `requirements.txt` because I was following the style of exactly constructing a reproducible environment. Instead, it looks like the SF `requirements.txt` is not quite as strict as this. In this case, SF definitely has all the bare essentials needed for the applications layer (which is really `networkx` and `numpy`).

strawberryfields/future/apps/graph/sample.py
trbromley: Thanks @zzzeid, I'm on the fence for this one, I thought the relative imports would be better since things are buried relatively deep, and I saw pep8 is fine for explicit relative imports in this case. I can be persuaded though :grin:.

josh146: This is generally more a matter of library convention than anything (and can depend on how the test suite is set up/the recommended way of installing or developing). The approach we've taken so far in Strawberry Fields and PennyLane has been:

* Relative imports for files in the same subpackage
* Absolute imports for files in different subpackages

This is mostly just for readability.

setup.py
josh146: The current `requirements.txt` file in Strawberry Fields is not ideal. Ideally, we should be including pinned versions of the required packages we are using in development (i.e., `numpy==1.17.0`).

Some guides recommend using `pip freeze > requirements.txt`, but this is problematic, and can lead to broken installations (we had this issue early on with the first few SF releases).

In terms of packages required for testing rather than adding them to the existing `requirements.txt`, I would prefer something like `requirements-test.txt`, which is pretty common (but not ubiquitous).

strawberryfields/apps/__init__.py
```diff
@@ -0,0 +1 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
```
josh146: Note: since this is now part of Strawberry Fields, it makes sense to specify the license at the top of each file in `apps/`. You can just copy the Apache 2.0 license from other Python files. If we wish to change the license on the `apps/` subfolder at some point, the Apache license is permissive and will allow this.

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
+r"""
+Dense subgraph identification
+=============================
+
+**Module name:** :mod:`strawberryfields.future.apps.graph.dense`
```
josh146: Can you remind me the reasoning behind the `graph` subpackage again? Is this because `apps` could potentially include multiple applications, which are not graph-based (and will go in another submodule?)

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+
+.. currentmodule:: strawberryfields.future.apps.graph.dense
+
+The frontend module for users to find dense subgraphs. The :func:`find_dense` function
```
josh146: Probably best to remove 'frontend', as it can now be confused with the SF frontend
```suggestion
Functions for finding dense subgraphs. The :func:`find_dense` function
```

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+-------
+
+.. autosummary::
+    METHOD_DICT
+    OPTIONS_DEFAULTS
+    find_dense
+    random_search
+
+Code details
+------------
```
josh146: This wills stop Code Details from showing up in the sidebar TOC (something we are in the process of removing from SF documentation)
```suggestion
^^^^^^^^^^^^
```

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+"""
+from typing import Tuple, Optional
+import networkx as nx
+from ..sample import BACKEND_DEFAULTS
```
josh146: As this is a separate sub-package, my inclination is to make it an absolute import (so the module location is clearer), but this is nitpicking and happy to leave it as-is.

josh146: Although, for clarity, it would be nice to include line breaks between the three main groups of imports (standard imports, third party imports, relative imports)

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+        graph (graph_type): the input graph
+        nodes (int): the size of desired dense subgraph
+        iterations (int): number of iterations to use in algorithm
+        options (dict[str, dict[str, Any]]): dict-of-dicts specifying options in different parts
+            of heuristic search algorithm; defaults to :const:`OPTIONS_DEFAULTS`
+
+    Returns:
+        Tuple[float, list]: the density and list of nodes corresponding to the densest subgraph
+        found
+    """
```
josh146: I know you mentioned in another PR that we are leaving the wrapper function for future consideration, but the complexity of this docstring indicates to me that it _should_ be removed and users should simply interact with the wrapped callables.

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+    Args:
+        graph (nx.Graph): the input graph
+        nodes (int): the size of desired dense subgraph
+        iterations (int): number of iterations to use in algorithm
+        options (dict[str, dict[str, Any]]): dict-of-dicts specifying options in different parts
+            of heuristic search algorithm; defaults to :const:`OPTIONS_DEFAULTS`
+
+    Returns:
+        Tuple[float, list]: the density and list of nodes corresponding to the densest subgraph
+        found
```
josh146: (in comparison to above, this docstring it much easier to parse and understand)

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Args:
+        graph (graph_type): the input graph
+        nodes (int): the size of desired dense subgraph
+        iterations (int): number of iterations to use in algorithm
+        options (dict[str, dict[str, Any]]): dict-of-dicts specifying options in different parts
+            of heuristic search algorithm; defaults to :const:`OPTIONS_DEFAULTS`
+
+    Returns:
+        Tuple[float, list]: the density and list of nodes corresponding to the densest subgraph
```
josh146: should tuple be capitalized here? When used directly from the `Typing` module yes (to avoid confusion with the inbuilt `tuple()`), but as a Sphinx return-type, my understanding was it should be `tuple`
```suggestion
        tuple[float, list]: the density and list of nodes corresponding to the densest subgraph
```

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Args:
+        graph (nx.Graph): the input graph
+        nodes (int): the size of desired dense subgraph
+        iterations (int): number of iterations to use in algorithm
+        options (dict[str, dict[str, Any]]): dict-of-dicts specifying options in different parts
+            of heuristic search algorithm; defaults to :const:`OPTIONS_DEFAULTS`
+
+    Returns:
+        Tuple[float, list]: the density and list of nodes corresponding to the densest subgraph
```
josh146: Same, should tuple be capitalized?

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+    - ``"backend"``: specifying options used by backend quantum samplers; corresponding
+      dictionary of options explained further in :mod:`~strawberryfields.future.apps.sample`
+    - ``"resize"``: specifying options used by resizing method; corresponding dictionary of
+      options explained further in :mod:`~strawberryfields.future.apps.graph.resize`
+    - ``"sample"``: specifying options used in sampling; corresponding dictionary of options
+      explained further in :mod:`~strawberryfields.future.apps.graph.sample`
+
+    If unspecified, a default set of options is adopted for a given option type.
+
+    .. _heuristic:
```
josh146: Does this work? I thought that Sphinx references only worked above section titles

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+        backend_options=options["backend"],
+    )
+
+    samples = resize.resize_subgraphs(
+        subgraphs=samples, graph=graph, target=nodes, resize_options=options["resize"]
+    )
+
+    density_and_samples = [(nx.density(graph.subgraph(s)), s) for s in samples]
+
+    return max(density_and_samples)
```
josh146: This function is almost a wrapper function itself 😆

strawberryfields/apps/graph/resize.py
```diff
@@ -0,0 +1,234 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
+r"""
+Graph resizing
```
josh146: You may have to remind me, but is this the module that is being renamed to 'modify'?

strawberryfields/apps/graph/resize.py
```diff
@@ -0,0 +1,234 @@
...
+            by a list of nodes
+        graph (nx.Graph): the input graph
+        target (int): the target number of nodes (must be larger than 1)
+        resize_options (dict[str, Any]): dictionary specifying options used during resizing;
+            defaults to :const:`RESIZE_DEFAULTS`
+
+    Returns:
+        list[list[int]]: a list of resized subgraphs, where each subgraph is given by a list of
+        nodes
+    """
```
josh146: Similarly to the previous module, I think it makes more sense to remove the wrapper function, and simply have two top level functions: `greedy_density_resize` and `greedy_degree_resize`.

strawberryfields/apps/graph/resize.py
```diff
@@ -0,0 +1,234 @@
...
+    density. It then considers all :math:`M-2` node subgraphs of the chosen :math:`M-1` node
+    subgraph, and continues until the desired subgraph size has been reached.
+
+    When growing, the algorithm considers all :math:`M+1` node subgraphs of the input graph given by
+    combining the input :math:`M` node subgraph with another node from the remainder of the input
+    graph, and picks the one with the greatest density. It then considers all :math:`M+2` node
+    subgraphs given by combining the chosen :math:`M+1` node subgraph with nodes remaining from
+    the overall graph, and continues until the desired subgraph size has been reached.
+
+    ``resize_options`` is not currently in use in :func:`greedy_density`.
```
josh146: ^ removing the wrapper will avoid requiring notices such as this

strawberryfields/apps/graph/sample.py
```diff
@@ -0,0 +1,139 @@
...
+    sample_subgraphs
+    to_subgraphs
+
+Code details
+------------
+"""
+from typing import Optional
+import networkx as nx
+import numpy as np
+from .. import sample
```
josh146: same as before re: linebreak between import categories.

strawberryfields/apps/graph/sample.py
```diff
@@ -0,0 +1,139 @@
...
+
+SAMPLE_DEFAULTS = {"distribution": "gbs", "postselect_ratio": 0.75}
+"""Dict[str, Any]: Dictionary to specify default parameters of options in :func:`sample_subgraphs`.
```
josh146: As with the case of `Tuple` above, should it just be `dict` here?

strawberryfields/apps/graph/sample.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def to_subgraphs(graph: nx.Graph, samples: list) -> list:
+    """Converts a list of samples to a list of subgraphs.
+
+    Given a list of samples, with each sample of ``len(nodes)`` being a list of zeros and ones,
+    this function returns a list of subgraphs selected by, for each sample, picking the nodes
+    corresponding to ones and creating the induced subgraph. The subgraph is specified as a list
+    of selected nodes. For example, given an input 6-node graph, a sample :math:`[0, 1, 1, 0, 0,
+    1]` is converted to a subgraph :math:`[1, 2, 5]`.
```
josh146: Something to keep in mind for future PRs (I think @ixfoduap mentioned this as some point?) but it might be nice to have (short) code examples showing how to use the function within the docstring. This is something we can also work on in Strawberry Fields and PennyLane.

strawberryfields/apps/graph/utils.py
```diff
@@ -0,0 +1,143 @@
...
+import networkx as nx
+import numpy as np
+
+graph_type = Union[nx.Graph, np.ndarray]
+
+
+def to_networkx_graph(graph: graph_type) -> nx.Graph:
+    """Converts input into a NetworkX ``Graph`` (i.e., an undirected graph).
+
+    Given an input undirected graph of type ``graph_type = Union[nx.Graph, np.ndarray]``,
```
josh146: Minor comment, but within function docstrings, it might be more readable for less technical users to avoid the type-hint notation, and instead write something plain and simple like

```rest
    Converts an undirected NetworkX graph or real symmetric NumPy adjacency matrix
    into a NetworkX graph object.
```

The input and return types don't need to be explicitly mentioned here, as it is just duplicating the `Args:` and `Returns:` section.

josh146: A follow on thought: it seems a bit weird to have this function also act on `nx.Graph` objects, but just perform the identity. I remember that this function was originally a convenience function to avoid code duplication, is it still needed with the current state of the codebase?

Might make more sense to position this function more for input validation of graphs (and conveniently also returning a NX graph) as opposed to a casting function.

strawberryfields/apps/graph/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    this function outputs a NetworkX graph of type ``nx.Graph``. The input ``np.ndarray`` must be a
+    matrix that is symmetric (i.e., satisfy :func:`is_undirected`) and also real.
+
+    Args:
+        graph (graph_type): input graph to be processed
+
+    Returns:
+        graph: the NetworkX graph corresponding to the input
```
josh146:
```suggestion
        nx.Graph: the NetworkX graph corresponding to the input
```

strawberryfields/apps/graph/utils.py
```diff
@@ -0,0 +1,143 @@
...
+            and np.allclose(graph, graph.conj())
+            and is_undirected(graph)
+        ):
+            graph = nx.Graph(graph.real.astype("float"))
+        elif (
+            graph.dtype is np.dtype("float") or graph.dtype is np.dtype("int")
+        ) and is_undirected(graph):
+            graph = nx.Graph(graph)
+        else:
+            raise Exception(
```
josh146: This bare exception should be replaced with a `ValueError` at the least
```suggestion
            raise ValueError(
```

strawberryfields/apps/graph/utils.py
```diff
@@ -0,0 +1,143 @@
...
+                "Graphs input as a NumPy array must be real, symmetric and of dtype "
+                "float or complex"
```
josh146: This message reads a bit weirdly

strawberryfields/apps/graph/utils.py
josh146: (this just avoids ambiguity with any existing graph objects/classes/modules in the apps layer)

strawberryfields/apps/sample.py
```diff
@@ -0,0 +1,240 @@
...
+r"""
+Sampling functions
+==================
+
+**Module name:** :mod:`strawberryfields.future.apps.sample`
+
+.. currentmodule:: strawberryfields.future.apps.sample
+
+This module provides functionality for generating samples from both a quantum device and the
+uniform distribution.
```
josh146: In this case, the sampling is with respect to graph structures, so I'm not sure I follow the structure of having this outside the graph package?

strawberryfields/apps/sample.py
```diff
@@ -0,0 +1,240 @@
...
+    r"""Performs a validation check that the requested backend can be run based upon the
+    ``QUANTUM_BACKENDS`` tuple.
+
+    Args:
+        backend (str): requested backend for sampling
+
+    Returns:
+        bool: returns only ``True`` if a valid backend is selected
+    """
+    return backend in QUANTUM_BACKENDS
```
josh146: Since this function only performs a (relatively simple) one line operation, it would probably make the code clearer/easier to follow to remove this function, and have the line `backend in QUANTUM_BACKENDS` used directly where it is called

strawberryfields/apps/sample.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+# pylint: disable=expression-not-assigned,pointless-statement
```
josh146: these are generally very useful things pylint helps check for, so it might be useful to have this directly on the lines that use the SF `|` syntax

strawberryfields/apps/sample.py
```diff
@@ -0,0 +1,240 @@
...
+        n_mean (float): mean photon number
+        samples (int): number of samples; defaults to 1
+        backend_options (dict[str, Any]): dictionary specifying options used by backends during
+        sampling; defaults to :const:`BACKEND_DEFAULTS`
+
+    Returns:
+        list[list[int]]: a list of length ``samples`` whose elements are length :math:`N` lists of
+        integers indicating events (e.g., photon numbers or clicks) in each of the :math:`N`
+        modes of the detector
+    """
```
josh146: Nice! in terms of the high-level 'wrapper' functions, this one has the clearest docstring and short but essential logic.

tests/apps/test_graph_resize.py
```diff
@@ -0,0 +1,232 @@
...
+Unit tests for strawberryfields.future.apps.graph.resize
+"""
+# pylint: disable=no-self-use,unused-argument
+import networkx as nx
+import numpy as np
+import pytest
+
+from strawberryfields.apps.graph import utils, resize
+
+pytestmark = pytest.mark.apps
```
josh146: Nice!

strawberryfields/apps/graph/dense.py
smite: I think Sphinx shouldn't generate a link here either way, so it's just for the benefit of the human readers. Technically we're already using the `typing` generic syntax when we write e.g. `tuple[float]`. OTOH, so far we have not used the `Tuple[Union[x, y]]` syntax suggested by `typing`, instead we tend to write `tuple[x, y]` because it's easier and maybe a bit more readable...
Adding to the confusion, I myself have been occasionally using `dict[x->y]`, which in `typing` notation would be `Dict[x, y]`.

I think this really only matters if we started using Python 3 type annotations instead of specifying the argument types in the docstring.

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,179 @@
...
+describing the method, while the dictionary values are callable functions corresponding to the
+method."""
+
+OPTIONS_DEFAULTS = {
+    "heuristic": {"method": random_search},
+    "backend": BACKEND_DEFAULTS,
+    "resize": resize.RESIZE_DEFAULTS,
+    "sample": sample.SAMPLE_DEFAULTS,
+}
+"""Dict[str, dict[str, Any]]: Options for dense subgraph identification heuristics. Composed of a
```
smite:
```suggestion
"""Dict[str, Dict[str, Any]]: Options for dense subgraph identification heuristics. Composed of a
```

strawberryfields/apps/graph/sample.py
smite: Depends. We tend to use `[x, y]` to mean "either x or y", whereas in `typing` syntax `Dict[x, y]` means dictionary mapping from x to y.

strawberryfields/apps/graph/utils.py
trbromley: Nice, have updated now to `validate_graph` and updated docstring. We still do use this function as in `find_dense` we allow the user to put in either type of graph. If we remove the `find_dense` wrapper, we can then allow this dual input functionality in the wrapped functions.

strawberryfields/apps/graph/utils.py
trbromley: Makes sense!

strawberryfields/apps/graph/utils.py
trbromley: Have reworded

strawberryfields/apps/sample.py
trbromley: :thinking: I'd say for `quantum_sampler` the sampling is with respect to any _symmetric_ matrix. Whether this is sampling from a graph could be debated, e.g. it could be a graph with complex weights. Things like point processes, which encode a grid of distances, would access this module, but I wouldn't see a direct link to a graph. I'd prefer to keep "graph" sampling to the ones which have real and symmetric adjacency matrices - which is what's imposed in `to_networkx_graph` or now `validate_graph`.

On the other hand, this module outputs samples that are of `len(modes)` and are zeros/ones/ints. Hence, the samples output aren't graphs (although simple to convert). Instead, in `graph.sample`, the output samples are subgraphs, i.e. a list of ints selecting nodes of a subgraph.

strawberryfields/apps/sample.py
trbromley: Removed!

strawberryfields/apps/sample.py
trbromley: Ok, have moved to just above `with p.context as q:`

trbromley: (And pylint is fine with that)

strawberryfields/apps/graph/dense.py
trbromley: Thanks @smite!

strawberryfields/apps/graph/dense.py
josh146: That makes sense :+1:

strawberryfields/apps/graph/dense.py
josh146: > docstring at the top of the file?

I think the module docstring would be the _perfect_ place to put it!

strawberryfields/apps/graph/resize.py
josh146: Sounds good!

strawberryfields/apps/graph/resize.py
josh146: Yep exactly!

strawberryfields/apps/sample.py
josh146: Ah, I think I see. I'm worked a lot with complex weighted undirected graphs in my PhD, so to me, almost any square matrix is a graph!

strawberryfields/apps/graph/dense.py
trbromley: Thanks @smite, for now I went with the all lowercase version (`dict[str, dict[str, Any]]`) but happy to open up the conversation for which to use as it seems there's arguments for both.

josh146: @trbromley @smite after reading through all the comments I'm happy with either standard. I think the main takeaway is consistency though, we should try and stick to the same style through the entire codebase

strawberryfields/apps/graph/dense.py
trbromley: Great :grin: Have put in ticket.. which I should really update on Jira!

tests/apps/test_graph_dense.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.mark.parametrize("dim", [5])
+class TestRandomSearch:
+    """Tests for the function ``dense.random_search``"""
+
+    def sampler(self, *args, **kwargs):
+        """Dummy function for sampling subgraphs"""
+
+        return (samples_subgraphs ** 2).tolist()
+
+    def test_densest_outputted(self, graph, monkeypatch):
```
jswinarton: Seems like a bit of a strange name, why not 
```suggestion
    def test_returns_densest(self, graph, monkeypatch):
```

doc/references.bib
```diff
@@ -836,3 +836,25 @@ @article{schuld2019evaluating
...
+  title={Using Gaussian Boson Sampling to Find Dense Subgraphs},
+  author={Arrazola, Juan Miguel and Bromley, Thomas R},
+  journal={Physical Review Letters},
+  volume={121},
+  number={3},
+  pages={030503},
+  year={2018},
+  publisher={APS}
+}
+
```
ixfoduap: Do we mention Kamil's paper on embedding graphs into GBS anywhere in the documentation? We should

strawberryfields/apps/graph/dense.py
```diff
@@ -0,0 +1,193 @@
...
+
+    return method(
+        graph=utils.validate_graph(graph),
+        nodes=nodes,
+        iterations=iterations,
+        options=options,
+    )
+
+
+def random_search(
```
ixfoduap: This might have been discussed before, but why do we want to include random search as an algorithm? It's a terrible way to find dense subgraphs. Is it to compare with GBS?

tests/apps/test_graph_dense.py
trbromley: Thanks @jswinarton !!

strawberryfields/apps/graph/dense.py
trbromley: The idea is to provide a couple of methods, one of them being the most elementary random search. There is also all the functionality for the charikar algorithm in the other modules, it just got missed to put the final functionality into the `dense` module as we switched to max clique, but we can definitely do that time permitting.

ixfoduap: I would prefer not to even have this as an option. Instead, we just run Charikar as the default algorithm, e.g., apps.find_dense(graph, k)

doc/references.bib
trbromley: Fixed!

strawberryfields/apps/graph/utils.py
co9olguy: This conversation was resolved, but I don't see the change reflected in code?

trbromley: Thanks @co9olguy, this slipped through the net! Have updated to a `ValueError` now

strawberryfields/apps/graph/dense.py
trbromley: My current line of thinking was to get rid of the `find_dense` wrapper function and make available both `random_search` and `charikar` functions, but we could emphasise `charikar` more in the documentation. By default algorithm, you mean that we'd _only_ offer Charikar? I'm fine with that, although I could imagine random search might have some value to people who read the original paper using random search.

trbromley:
@josh146, that's right - there is still a tutorial and proper integration of the apps layer into docs, but that will be a separate PR once the SF docs refactor has been completed.

I also just merged `master` into this branch, as there were some changes like https://github.com/XanaduAI/strawberryfields/pull/170 which would break if not fixed.

PR #172, opened 2019-08-22
Provide max-clique util functions

trbromley:
This PR includes implementations of three important functions required for implementing local search max clique algorithms:

- `is_clique` : checks if an input graph is a clique
- `c_0` : generates the set of nodes that can be added to an input clique to create a larger one
- `c_1` : generates the set of nodes that can be swapped with another node in an input clique to 
generate a new clique.

Notes:
- This is a PR co-authored by myself and @ixfoduap (mostly by @ixfoduap), so it would be good to get another reviewer
- Sorry - this probably should have been a couple of elementary PRs. Will do that next time!

### DIFF
--- a/strawberryfields/apps/graph/utils.py
+++ b/strawberryfields/apps/graph/utils.py
@@ -38,7 +38,9 @@ Summary
     subgraph_adjacency
     validate_graph
     is_subgraph
-    update_options
+    is_clique
+    c_0
+    c_1
 
 Code details
 ^^^^^^^^^^^^
@@ -153,3 +155,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
         return set(subgraph).issubset(graph.nodes)
     except TypeError:
         raise TypeError("subgraph and graph.nodes must be iterable")
+
+
+def is_clique(graph: nx.Graph) -> bool:
+    """Determines if the input graph is a clique
+
+    Args:
+        graph (nx.Graph): the input graph
+
+    Returns:
+        bool: returns ``True`` if input graph is a clique and ``False`` otherwise
+    """
+    edges = graph.edges
+    nodes = graph.order()
+
+    return len(edges) == nodes * (nodes - 1) / 2  # a clique has of n nodes has n*(n-1)/2 edges
+
+
+def c_0(subgraph: list, graph: nx.Graph):
+    """Generates the set C0 of nodes that are connected to all nodes in the input subgraph
+
+    The set C0 is defined in :cite:`pullan2006phased`.
+
+    Args:
+        subgraph (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+    Returns:
+        list[int]: a list containing the C0 nodes for the subgraph
+    """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
+        raise ValueError("Input subgraph is not a clique")
+
+    c0_nodes = []
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
+
+    for i in non_clique_nodes:
+        if set(subgraph).issubset(graph.neighbors(i)):
+            c0_nodes.append(i)
+
+    return c0_nodes
+
+
+def c_1(subgraph: list, graph: nx.Graph):
+    """Generates the set C1 of nodes that are connected to all but one of the nodes in the input
+    subgraph
+
+    The set C1 is defined in :cite:`pullan2006phased`.
+
+    Args:
+        subgraph (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+   Returns:
+       list[int]: a list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
+       ``i,j,...,k`` are the nodes in C1, while ``i_clique, j_clique,...,k_clique`` are the nodes in
+       the clique they can be swapped with.
+       """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
+        raise ValueError("Input subgraph is not a clique")
+
+    c1_nodes = []
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
+
+    subgraph = set(subgraph)
+
+    for i in non_clique_nodes:
+        neighbors_in_subgraph = subgraph.intersection(graph.neighbors(i))
+
+        if len(neighbors_in_subgraph) == len(subgraph) - 1:
+            to_swap = subgraph - neighbors_in_subgraph
+            c1_nodes.append((to_swap.pop(), i))
+
+    return c1_nodes
--- a/strawberryfields/apps/graph/utils.py
+++ b/strawberryfields/apps/graph/utils.py
@@ -19,16 +19,14 @@ Graph functions
 
 .. currentmodule:: strawberryfields.apps.graph.utils
 
-This module provides some ancillary functions for dealing with graphs. This includes
-:func:`is_undirected` to check if an input matrix corresponds to an undirected graph (i.e.,
-is symmetric) and :func:`subgraph_adjacency` to return the adjacency matrix of a subgraph when an
-input graph and subset of nodes is specified.
+This module provides some ancillary functions for dealing with graphs. This includes graph
+validation functions such as :func:`is_undirected`, :func:`validate_graph` and :func:`is_subgraph`;
+clique-based utility functions :func:`is_clique`, :func:`c_0` and :func:`c_1`; and a utility
+function to return the adjacency matrix of a subgraph, :func:`subgraph_adjacency`.
 
-Furthermore, the frontend :func:`~strawberryfields.apps.graph.dense_subgraph.find_dense`
-function allows users to input graphs both as a `NumPy <https://www.numpy.org/>`__ array
-containing the adjacency matrix and as a `NetworkX <https://networkx.github.io/>`__ ``Graph``
-object. The :func:`validate_graph` function allows both inputs to be processed into a NetworkX
-Graph for ease of processing.
+Furthermore, ``graph_type = Union[nx.Graph, np.ndarray]`` is defined here as an input type that
+can be either a NetworkX graph or an adjacency matrix in the form of a NumPy array. Some
+functions in the applications layer are able to accept either input.
 
 Summary
 -------
--- a/tests/apps/test_graph_utils.py
+++ b/tests/apps/test_graph_utils.py
@@ -29,7 +29,7 @@ adj_dim_range = range(2, 6)
 
 @pytest.mark.parametrize("dim", adj_dim_range)
 class TestValidateGraph:
-    """Tests for the function ``glassonion.graphs.utils.validate_graph``"""
+    """Tests for the function ``strawberryfields.apps.graph.utils.validate_graph``"""
 
     def test_valid_adjacency(self, adj, monkeypatch):
         """Test if function returns the NetworkX Graph class corresponding to the input adjacency
@@ -92,7 +92,7 @@ class TestValidateGraph:
 
 @pytest.mark.parametrize("dim", adj_dim_range)
 class TestIsUndirected:
-    """Tests for the function ``glassonion.graphs.utils.is_undirected``"""
+    """Tests for the function ``strawberryfields.apps.graph.utils.is_undirected``"""
 
     def test_valid_input(self, adj):
         """Test if function returns ``True`` for a symmetric matrix"""
@@ -129,7 +129,7 @@ class TestIsUndirected:
 
 @pytest.mark.parametrize("dim", [4, 5, 6])
 class TestSubgraphAdjacency:
-    """Tests for the function ``glassonion.graphs.utils.subgraph_adjacency``"""
+    """Tests for the function ``strawberryfields.apps.graph.utils.subgraph_adjacency``"""
 
     def test_input(self, dim, graph, monkeypatch):
         """Test if function returns the correct adjacency matrix of a subgraph. This test
@@ -161,7 +161,7 @@ class TestSubgraphAdjacency:
 
 @pytest.mark.parametrize("dim", [4, 5, 6])
 class TestIsSubgraph:
-    """Tests for the function ``glassonion.graphs.utils.is_subgraph``"""
+    """Tests for the function ``strawberryfields.apps.graph.utils.is_subgraph``"""
 
     def test_invalid_type(self, graph):
         """Test if function raises a ``TypeError`` when fed an invalid subgraph type (i.e.,
@@ -186,3 +186,91 @@ class TestIsSubgraph:
         subgraphs = itertools.combinations(range(dim), int(dim / 2))
 
         assert not all([utils.is_subgraph(list(s), graph) for s in subgraphs])
+
+
+@pytest.mark.parametrize("dim", range(2, 10))
+class TestIsClique:
+    """Tests for the function `strawberryfields.apps.graph.utils.is_clique` """
+
+    def test_no_false_negatives(self, dim):
+        """ Tests that cliques are labelled as such"""
+        g = nx.complete_graph(dim)
+        assert utils.is_clique(g)
+
+    def test_no_false_positives(self, dim):
+        """ Tests that non-cliques are labelled as such"""
+        g = nx.empty_graph(dim)
+        assert not utils.is_clique(g)
+
+
+@pytest.mark.parametrize("dim", range(2, 10))
+class TestC0:
+    """ Tests function c_0 that generates the set of nodes connected to all nodes in a clique"""
+
+    def test_grows_to_clique(self, dim):
+        """ Tests that adding nodes from C0 to a clique results in another clique"""
+        A = nx.complete_graph(dim)
+        S = [0]
+        K = utils.c_0(S, A)
+        while K:
+            new_node = K[0]
+            S.append(new_node)
+            K = utils.c_0(S, A)
+
+        assert utils.is_clique(A.subgraph(S))
+
+    def test_c0_comp_graph(self, dim):
+        """ Tests that the set C0 for a node in a clique consists of all remaining nodes"""
+        A = nx.complete_graph(dim)
+        S = [dim - 1]
+        K = utils.c_0(S, A)
+
+        assert K == list(range(dim - 1))
+
+    def test_c0input_is_clique(self, dim):
+        """ Tests if function raises a ``ValueError`` when input is not a clique"""
+        A = np.ones((dim, dim)) - np.eye(dim)
+        A = nx.Graph(A)
+        A.remove_edge(0, 1)
+        S = [0, 1]
+
+        with pytest.raises(ValueError, match="Input subgraph is not a clique"):
+            utils.c_0(S, A)
+
+
+@pytest.mark.parametrize("dim", range(4, 10))
+class TestC1:
+    """Tests function c_1 that generates the set of nodes connected to all *but one* of the nodes
+    in a clique"""
+
+    def test_c1_comp_graph(self, dim):
+        """ Tests that c1 set is correctly generated for an almost-complete graph, where edge
+        (0, 1) is removed """
+        A = nx.complete_graph(dim)
+        A.remove_edge(0, 1)
+        S = [i for i in range(1, dim)]
+        c1 = utils.c_1(S, A)
+
+        assert c1 == [(1, 0)]
+
+    def test_c1_swap_to_clique(self, dim):
+        """ Tests that c1 set gives a valid clique after swapping """
+        A = nx.complete_graph(dim)
+        A.remove_edge(0, 1)
+        S = [i for i in range(1, dim)]
+        c1 = utils.c_1(S, A)
+        swap_nodes = c1[0]
+        S.remove(swap_nodes[0])
+        S.append(swap_nodes[1])
+
+        assert utils.is_clique(A.subgraph(S))
+
+    def test_c1input_is_clique(self, dim):
+        """ Tests if function raises a ``ValueError`` when input is not a clique"""
+        A = np.ones((dim, dim)) - np.eye(dim)
+        A = nx.Graph(A)
+        A.remove_edge(0, 1)
+        S = [0, 1]
+
+        with pytest.raises(ValueError, match="Input subgraph is not a clique"):
+            utils.c_1(S, A)
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -869,3 +869,14 @@
   year={2018},
   publisher={APS}
 }
+
+@article{pullan2006phased,
+  title={Phased local search for the maximum clique problem},
+  author={Pullan, Wayne},
+  journal={Journal of Combinatorial Optimization},
+  volume={12},
+  number={3},
+  pages={303--323},
+  year={2006},
+  publisher={Springer}
+}


tests/apps/test_graph_utils.py
```diff
@@ -186,3 +186,91 @@ def test_invalid_subgraphs(self, graph, dim):
...
+        """ Tests that non-cliques are labelled as such"""
+        g = nx.empty_graph(dim)
+        assert not utils.is_clique(g)
+
+
+@pytest.mark.parametrize("dim", range(2, 10))
+class TestC0:
+    """ Tests function c_0 that generates the set of nodes connected to all nodes in a clique"""
+
+    def test_grows_to_clique(self, dim):
```
trbromley: I'm not sure about this test, as I think it's more of a remake in test form of the `clique_grow` function we plan. The test implicitly relies on the `c_0` function but there could also be mistakes in the test (you mentioned tests of tests :grin:). Of course it's a balance, and the code in this test isn't too complicated.

However, the `test_c0_comp_graph` is more along the lines of what I was thinking. How about something like:
```python
@pytest.mark.parametrize("subdim", range(2, 10))
def test_correct_c0(self, dim, subdim):
    g = nx.lollipop_graph(dim, 1)
    s = set(range(subdim))
    res = set(c_0(s, g))

    assert res not in s
    assert {dim + 1} not in res
    assert res | s == set(range(dim))
```

josh146:
@trbromley, is this the same PR as the one in the GlassOnion repository?

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    Returns:
+        bool: returns ``True`` if input graph is a clique and ``False`` otherwise
+    """
+    edges = graph.edges
+    nodes = graph.order()
+
+    return len(edges) == nodes * (nodes - 1) / 2  # a clique has of n nodes has n*(n-1)/2 edges
+
+
+def c_0(subgraph: list, graph: nx.Graph):
```
josh146: This is very minor, but since the name is defined in the paper as C0, should the function be `c0(subgraph, graph)`?

josh146: Note that `c0_nodes` is used within the function, not `c_0_nodes`.

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
+
+    for i in non_clique_nodes:
+        if set(subgraph).issubset(graph.neighbors(i)):
+            c0_nodes.append(i)
+
+    return c0_nodes
+
+
+def c_1(subgraph: list, graph: nx.Graph):
```
josh146: Same comment here re: function name

tests/apps/test_graph_utils.py
```diff
@@ -29,7 +29,7 @@
 
 @pytest.mark.parametrize("dim", adj_dim_range)
 class TestValidateGraph:
-    """Tests for the function ``glassonion.graphs.utils.validate_graph``"""
+    """Tests for the function ``strawberryfields.apps.graph.utils.validate_graph``"""
```
josh146: :laughing:

strawberryfields/apps/graph/utils.py
josh146: After going through the rest of the PR, everything else uses the `c0` (no underscore) convention, i.e., `test_c0_comp_graph`

tests/apps/test_graph_utils.py
```diff
@@ -186,3 +186,91 @@ def test_invalid_subgraphs(self, graph, dim):
...
+
+    def test_no_false_negatives(self, dim):
+        """ Tests that cliques are labelled as such"""
+        g = nx.complete_graph(dim)
+        assert utils.is_clique(g)
+
+    def test_no_false_positives(self, dim):
+        """ Tests that non-cliques are labelled as such"""
+        g = nx.empty_graph(dim)
+        assert not utils.is_clique(g)
```
josh146: Nice!

tests/apps/test_graph_utils.py
josh146: I agree here - the unit test should be testing the existing functionality of the c0 function, and it sounds like this test is a future unit test of `clique_grow`.

But you might as well keep it here for now, until the PR that introduces `clique_grow`. You could mark it with a `# TODO:` so that it is not forgotten.

strawberryfields/apps/graph/utils.py
jswinarton:
```suggestion
def c1(subgraph: list, graph: nx.Graph):
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+   Returns:
+       list[int]: a list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
+       ``i,j,...,k`` are the nodes in C1, while ``i_clique, j_clique,...,k_clique`` are the nodes in
+       the clique they can be swapped with.
+       """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
+        raise ValueError("Input subgraph is not a clique")
+
+    c1_nodes = []
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
```
jswinarton: This comment is redundant, since the var name already explains this.
```suggestion
    non_clique_nodes = set(graph.nodes) - set(subgraph)
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    Args:
+        subgraph (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+   Returns:
+       list[int]: a list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
+       ``i,j,...,k`` are the nodes in C1, while ``i_clique, j_clique,...,k_clique`` are the nodes in
+       the clique they can be swapped with.
+       """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
```
jswinarton: Redundant comment.
```suggestion
    if not is_clique(graph.subgraph(subgraph)):
```

strawberryfields/apps/graph/utils.py
jswinarton: Why not `c0`? Since this is how you refer to it within the documentation and also with internal variables.
```suggestion
def c0(subgraph: list, graph: nx.Graph):
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    The set C0 is defined in :cite:`pullan2006phased`.
+
+    Args:
+        subgraph (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+    Returns:
+        list[int]: a list containing the C0 nodes for the subgraph
+    """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
```
jswinarton: Redundant comment
```suggestion
    if not is_clique(graph.subgraph(subgraph)):
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+        raise ValueError("Input subgraph is not a clique")
+
+    c0_nodes = []
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
```
jswinarton: Redundant comment
```suggestion
    non_clique_nodes = set(graph.nodes) - set(subgraph)
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+       ``i,j,...,k`` are the nodes in C1, while ``i_clique, j_clique,...,k_clique`` are the nodes in
+       the clique they can be swapped with.
+       """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
+        raise ValueError("Input subgraph is not a clique")
+
+    c1_nodes = []
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
+
+    subgraph = set(subgraph)
```
jswinarton: Best to move this to the top of the function, since you call this on line 217 as well.

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+        list[int]: a list containing the C0 nodes for the subgraph
+    """
+    if not is_clique(graph.subgraph(subgraph)):  # Each subgraph must be a clique
+        raise ValueError("Input subgraph is not a clique")
+
+    c0_nodes = []
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
+
+    for i in non_clique_nodes:
+        if set(subgraph).issubset(graph.neighbors(i)):
```
jswinarton: You make multiple calls to `set(subgraph)`, should be called once near the top of the function.

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    Args:
+        graph (nx.Graph): the input graph
+
+    Returns:
+        bool: returns ``True`` if input graph is a clique and ``False`` otherwise
+    """
+    edges = graph.edges
+    nodes = graph.order()
+
+    return len(edges) == nodes * (nodes - 1) / 2  # a clique has of n nodes has n*(n-1)/2 edges
```
jswinarton: I would suggest removing this comment and making it part of the function docstring. Also, is it supposed to say "a clique has n nodes" ... ?

tests/apps/test_graph_utils.py
```diff
@@ -186,3 +186,91 @@ def test_invalid_subgraphs(self, graph, dim):
...
+
+    def test_c0_comp_graph(self, dim):
+        """ Tests that the set C0 for a node in a clique consists of all remaining nodes"""
+        A = nx.complete_graph(dim)
+        S = [dim - 1]
+        K = utils.c_0(S, A)
+
+        assert K == list(range(dim - 1))
+
+    def test_c0input_is_clique(self, dim):
```
jswinarton:
```suggestion
    def test_c0_input_is_clique(self, dim):
```

tests/apps/test_graph_utils.py
```diff
@@ -186,3 +186,91 @@ def test_invalid_subgraphs(self, graph, dim):
...
+        A.remove_edge(0, 1)
+        S = [i for i in range(1, dim)]
+        c1 = utils.c_1(S, A)
+        swap_nodes = c1[0]
+        S.remove(swap_nodes[0])
+        S.append(swap_nodes[1])
+
+        assert utils.is_clique(A.subgraph(S))
+
+    def test_c1input_is_clique(self, dim):
```
jswinarton:
```suggestion
    def test_c1_input_is_clique(self, dim):
```

strawberryfields/apps/graph/utils.py
jswinarton: Also, why "each" subgraph? Isn't only one subgraph being passed in? (Just for my own understanding)

tests/apps/test_graph_utils.py
```diff
@@ -186,3 +186,91 @@ def test_invalid_subgraphs(self, graph, dim):
...
+
+    def test_no_false_positives(self, dim):
+        """ Tests that non-cliques are labelled as such"""
+        g = nx.empty_graph(dim)
+        assert not utils.is_clique(g)
+
+
+@pytest.mark.parametrize("dim", range(2, 10))
+class TestC0:
+    """ Tests function c_0 that generates the set of nodes connected to all nodes in a clique"""
```
jswinarton: I notice in a handful of places there's an extra space after the leading quotes (`""" `), is this intentional?

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    non_clique_nodes = set(graph.nodes) - set(subgraph)  # All nodes that are not in the clique
+
+    subgraph = set(subgraph)
+
+    for i in non_clique_nodes:
+        neighbors_in_subgraph = subgraph.intersection(graph.neighbors(i))
+
+        if len(neighbors_in_subgraph) == len(subgraph) - 1:
+            to_swap = subgraph - neighbors_in_subgraph
+            c1_nodes.append((to_swap.pop(), i))
```
jswinarton: Interesting. `set.pop` removes an arbitrary element from a set, which makes this non-deterministic. Is that intentional?

strawberryfields/apps/graph/utils.py
jswinarton: lol I didn't see your review before I made mine and came to the same conclusion :laughing:

tests/apps/test_graph_utils.py
josh146: I saw this but didn't comment on it, thinking it might be nitpicking :laughing: 
But since @jswinarton has noticed it as well, in the rest of the SF documentation (barring some backends) we tend to stick to no-space and no-new-line (i.e., `"""Test function...`)

strawberryfields/apps/graph/utils.py
```diff
@@ -19,16 +19,14 @@
...
+function to return the adjacency matrix of a subgraph, :func:`subgraph_adjacency`.
 
-Furthermore, the frontend :func:`~strawberryfields.apps.graph.dense_subgraph.find_dense`
-function allows users to input graphs both as a `NumPy <https://www.numpy.org/>`__ array
-containing the adjacency matrix and as a `NetworkX <https://networkx.github.io/>`__ ``Graph``
-object. The :func:`validate_graph` function allows both inputs to be processed into a NetworkX
-Graph for ease of processing.
+Furthermore, ``graph_type = Union[nx.Graph, np.ndarray]`` is defined here as an input type that
+can be either a NetworkX graph or an adjacency matrix in the form of a NumPy array. Some
+functions in the applications layer are able to accept either input.
```
co9olguy:
```suggestion
functions in this module are able to accept either input.
```
To make things easier later, probably better to not use any specific naming here

strawberryfields/apps/graph/utils.py
co9olguy: Agree with @jswinarton that this might as well be in docstring.
I'd guess it should say "a clique of n nodes has n*(n-1)/2" edges.

strawberryfields/apps/graph/utils.py
```diff
@@ -19,16 +19,14 @@
...
+clique-based utility functions :func:`is_clique`, :func:`c_0` and :func:`c_1`; and a utility
+function to return the adjacency matrix of a subgraph, :func:`subgraph_adjacency`.
 
-Furthermore, the frontend :func:`~strawberryfields.apps.graph.dense_subgraph.find_dense`
-function allows users to input graphs both as a `NumPy <https://www.numpy.org/>`__ array
-containing the adjacency matrix and as a `NetworkX <https://networkx.github.io/>`__ ``Graph``
-object. The :func:`validate_graph` function allows both inputs to be processed into a NetworkX
-Graph for ease of processing.
+Furthermore, ``graph_type = Union[nx.Graph, np.ndarray]`` is defined here as an input type that
+can be either a NetworkX graph or an adjacency matrix in the form of a NumPy array. Some
```
co9olguy: Any reason why only some functions accept both types?

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+        bool: returns ``True`` if input graph is a clique and ``False`` otherwise
+    """
+    edges = graph.edges
+    nodes = graph.order()
+
+    return len(edges) == nodes * (nodes - 1) / 2  # a clique has of n nodes has n*(n-1)/2 edges
+
+
+def c_0(subgraph: list, graph: nx.Graph):
+    """Generates the set C0 of nodes that are connected to all nodes in the input subgraph
```
co9olguy:
```suggestion
    """Generates the set :math:`C0` of nodes that are connected to all nodes in the input subgraph.
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    The set C0 is defined in :cite:`pullan2006phased`.
```
co9olguy:
```suggestion
    The set :math:`C0` is defined in :cite:`pullan2006phased`.
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    Args:
+        subgraph (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+    Returns:
+        list[int]: a list containing the C0 nodes for the subgraph
```
co9olguy:
```suggestion
        list[int]: a list containing the :math:`C0` nodes for the subgraph
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    for i in non_clique_nodes:
+        if set(subgraph).issubset(graph.neighbors(i)):
+            c0_nodes.append(i)
+
+    return c0_nodes
+
+
+def c_1(subgraph: list, graph: nx.Graph):
+    """Generates the set C1 of nodes that are connected to all but one of the nodes in the input
```
co9olguy:
```suggestion
    """Generates the set :math:`C1` of nodes that are connected to all but one of the nodes in the input
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+    subgraph
```
co9olguy:
```suggestion
    subgraph.
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    The set C1 is defined in :cite:`pullan2006phased`.
```
co9olguy:
```suggestion
    The set :math:`C1` is defined in :cite:`pullan2006phased`.
```

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,76 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    Args:
+        subgraph (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+   Returns:
+       list[int]: a list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
```
co9olguy:
```suggestion
       list[int]: A list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
```

tests/apps/test_graph_utils.py
```diff
@@ -186,3 +186,91 @@ def test_invalid_subgraphs(self, graph, dim):
...
+    def test_c0_comp_graph(self, dim):
+        """ Tests that the set C0 for a node in a clique consists of all remaining nodes"""
+        A = nx.complete_graph(dim)
+        S = [dim - 1]
+        K = utils.c_0(S, A)
+
+        assert K == list(range(dim - 1))
+
+    def test_c0input_is_clique(self, dim):
+        """ Tests if function raises a ``ValueError`` when input is not a clique"""
```
co9olguy: function name doesn't seem to match docstring

strawberryfields/apps/graph/utils.py
trbromley: We made an ADR to do this only for some "user facing" functions, but I have been thinking we can just change this to all functions accepting a graph-like input. This would be a case of adding a line in each function to call `validate_graph` and also updates to the docstrings. Have added a ticket to backlog.

strawberryfields/apps/graph/utils.py
trbromley:
```suggestion
    """Generates the set :math:`C_0` of nodes that are connected to all nodes in the input subgraph
```
Just checked the paper, they do capital C and 0 as a subscript, so I think :math:`C_0` makes sense in the docstrings. For the function name, how about `c_0` (so no caps)?

strawberryfields/apps/graph/utils.py
trbromley: :thinking:
It's a set of size 1 by construction, which takes it back to being deterministic. Although, is there a better way to access the only element of a size 1 set - would happily use it :grin:.

tests/apps/test_graph_utils.py
trbromley: :blush:

tests/apps/test_graph_utils.py
trbromley: Did we disable `TODO` checks in pylint?

strawberryfields/apps/graph/utils.py
jswinarton: ah, that makes sense. You can do tuple unpacking on sets as well:
```
s = {1}
(element,) = s
element
>>> 1 
```
up to you, but I might leave a comment here just to clarify what's happening.

strawberryfields/apps/graph/utils.py
trbromley: Still not sure on whether to rename `subgraph` to `clique` :thinking:

strawberryfields/apps/graph/utils.py
josh146: I agree with :math:`C_0` for docstrings. For function usability, my feeling is that `c0` is preferable to `c_0`, just in terms of being less complex. Also an element of user preference there, if you count @jswinarton and I as users :laughing:

strawberryfields/apps/graph/utils.py
josh146: This is slightly anecdotal, but `subgraph` to me sounds like it should be a Graph object, or perhaps related to `graph` arguments/`graph_type`. Since this is instead a list, I would definitely choose `clique` over `subgraph`.

jswinarton: Agree, `clique` is good here because you explicitly reject anything that's not. In the case of `is_clique` I would prefer `subgraph`.

strawberryfields/apps/graph/utils.py
co9olguy: Yep, exactly what I was thinking

strawberryfields/apps/graph/utils.py
trbromley:
```suggestion
            c1_nodes.append((to_swap.pop(), i))  # to_swap is a set of size 1
```

strawberryfields/apps/graph/utils.py
```diff
@@ -156,73 +156,75 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
-    """Determines if the input graph is a clique
+    """Determines if the input graph is a clique. A clique of n nodes has n*(n-1)/2 edges
 
     Args:
-        graph (nx.Graph): the input graph
+        graph (nx.Graph): The input graph
 
     Returns:
-        bool: returns ``True`` if input graph is a clique and ``False`` otherwise
+        bool: Returns ``True`` if input graph is a clique and ``False`` otherwise
```
co9olguy: Actually, we can just drop `Returns` here

strawberryfields/apps/graph/utils.py
co9olguy: As long as it is consistent throughout code/docs, I have no issues with what you decide

tests/apps/test_graph_utils.py
```diff
@@ -193,42 +193,42 @@ class TestIsClique:
...
+        """ Tests that the set :math:`c_0` for a node in a clique consists of all remaining nodes"""
         A = nx.complete_graph(dim)
         S = [dim - 1]
         K = utils.c_0(S, A)
 
         assert K == list(range(dim - 1))
 
-    def test_c0input_is_clique(self, dim):
-        """ Tests if function raises a ``ValueError`` when input is not a clique"""
+    def test_c_0_input_is_clique(self, dim):
```
co9olguy: Suggest making the name of the test more explicitly reflect the test, e.g.
`test_c_0_error_on_not_clique_input`

tests/apps/test_graph_utils.py
jswinarton: I see this is resolved, but there's no change?

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +153,78 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    c_1_nodes = []
+    non_clique_nodes = set(graph.nodes) - set_subgraph
+
+    for i in non_clique_nodes:
+        neighbors_in_subgraph = set_subgraph.intersection(graph.neighbors(i))
+
+        if len(neighbors_in_subgraph) == len(set_subgraph) - 1:
+            to_swap = set_subgraph - neighbors_in_subgraph
+            (i_clique,) = to_swap
```
jswinarton: thumb

strawberryfields/apps/graph/utils.py
trbromley: Done!

tests/apps/test_graph_utils.py
trbromley: It eventually got changed to
`test_c_1_error_on_not_clique_input`
which combines your suggestion (underscore) with @co9olguy's suggestion on clearer naming

trbromley:
@josh146, should be good to go for a final review! Thanks!

strawberryfields/apps/graph/utils.py
```diff
@@ -153,3 +148,78 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    Args:
+        clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique
+        graph (nx.Graph): the input graph
+
+    Returns:
+       list[int]: A list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
+       ``i,j,...,k`` are the nodes in :math:`C_1`, while ``i_clique, j_clique,...,k_clique`` are the
+        nodes in the clique they can be swapped with.
+       """
```
josh146: Strange indentation on the """

josh146: Strange indentation on the """

trbromley: Oops sorry missed that one 😁 Fixed now!

PR #177, opened 2019-08-28
Add example code to docstrings for utils

ixfoduap:
Adding examples of function use in small code snippets can really help the reader get a quick understanding of the function. We would like to do this for functions in the applications layer, and this PR provides code snippets for max-clique based functions in `graph.utils` and also the `graph.resize.clique_swap` function.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -110,6 +110,7 @@ extensions = [
     'sphinx.ext.viewcode',
     'sphinxcontrib.bibtex',
     'edit_on_github',
+    'sphinx_autodoc_typehints',
     'nbsphinx'
 ]
 
--- a/strawberryfields/apps/graph/resize.py
+++ b/strawberryfields/apps/graph/resize.py
@@ -310,21 +310,19 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") -> 
 
     Example usage:
 
-    .. code-block::
-
-        >>> from strawberryfields.apps.graph import resize
-        >>> import networkx as nx
-        >>> graph = nx.wheel_graph(5)
-        >>> graph.remove_edge(0, 4)
-        >>> clique = [0, 1, 2]
-        >>> resize.clique_swap(clique, graph)
-        [0, 2, 3]
-        >>> graph = nx.lollipop_graph(5, 1)
-        >>> graph.remove_edge(0, 4)
-        >>> graph.remove_edge(0, 3)
-        >>> clique = [0, 1, 2]
-        >>> resize.clique_swap(clique, graph, node_select="degree")
-        [1, 2, 4]
+    >>> from strawberryfields.apps.graph import resize
+    >>> import networkx as nx
+    >>> graph = nx.wheel_graph(5)
+    >>> graph.remove_edge(0, 4)
+    >>> clique = [0, 1, 2]
+    >>> resize.clique_swap(clique, graph)
+    [0, 2, 3]
+    >>> graph = nx.lollipop_graph(5, 1)
+    >>> graph.remove_edge(0, 4)
+    >>> graph.remove_edge(0, 3)
+    >>> clique = [0, 1, 2]
+    >>> resize.clique_swap(clique, graph, node_select="degree")
+    [1, 2, 4]
 
     Args:
         clique (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
--- a/strawberryfields/apps/graph/resize.py
+++ b/strawberryfields/apps/graph/resize.py
@@ -318,7 +318,13 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") -> 
         >>> graph.remove_edge(0, 4)
         >>> subgraph = [0, 1, 2]  # these nodes form a clique
         >>> resize.clique_swap(subgraph, graph)
-        [0, 1, 4]
+        [0, 2, 3]
+        >>> graph = nx.lollipop_graph(5, 1)
+        >>> graph.remove_edge(0, 4)
+        >>> graph.remove_edge(0, 3)
+        >>> subgraph = [0, 1, 2]
+        >>> resize.clique_swap(subgraph, graph, node_select="degree")
+        [1, 2, 4]
 
     Args:
         clique (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
--- a/strawberryfields/apps/graph/resize.py
+++ b/strawberryfields/apps/graph/resize.py
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") -> 
     supported. Degree-based node selection involves picking the node with the greatest degree,
     with ties settled by uniform random choice.
 
+    Example usage:
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import resize
+        >>> import networkx as nx
+        >>> graph = nx.wheel_graph(5)
+        >>> graph.remove_edge(0, 4)
+        >>> subgraph = [0, 1, 2]  # these nodes form a clique
+        >>> resize.clique_swap(subgraph, graph)
+        [0, 1, 4]
+
     Args:
         clique (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
         graph (nx.Graph): the input graph
--- a/strawberryfields/apps/graph/resize.py
+++ b/strawberryfields/apps/graph/resize.py
@@ -310,19 +310,11 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") -> 
 
     Example usage:
 
-    >>> from strawberryfields.apps.graph import resize
-    >>> import networkx as nx
     >>> graph = nx.wheel_graph(5)
     >>> graph.remove_edge(0, 4)
     >>> clique = [0, 1, 2]
-    >>> resize.clique_swap(clique, graph)
+    >>> clique_swap(clique, graph)
     [0, 2, 3]
-    >>> graph = nx.lollipop_graph(5, 1)
-    >>> graph.remove_edge(0, 4)
-    >>> graph.remove_edge(0, 3)
-    >>> clique = [0, 1, 2]
-    >>> resize.clique_swap(clique, graph, node_select="degree")
-    [1, 2, 4]
 
     Args:
         clique (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
--- a/strawberryfields/apps/graph/resize.py
+++ b/strawberryfields/apps/graph/resize.py
@@ -316,14 +316,14 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") -> 
         >>> import networkx as nx
         >>> graph = nx.wheel_graph(5)
         >>> graph.remove_edge(0, 4)
-        >>> subgraph = [0, 1, 2]  # these nodes form a clique
-        >>> resize.clique_swap(subgraph, graph)
+        >>> clique = [0, 1, 2]
+        >>> resize.clique_swap(clique, graph)
         [0, 2, 3]
         >>> graph = nx.lollipop_graph(5, 1)
         >>> graph.remove_edge(0, 4)
         >>> graph.remove_edge(0, 3)
-        >>> subgraph = [0, 1, 2]
-        >>> resize.clique_swap(subgraph, graph, node_select="degree")
+        >>> clique = [0, 1, 2]
+        >>> resize.clique_swap(clique, graph, node_select="degree")
         [1, 2, 4]
 
     Args:
--- a/strawberryfields/apps/graph/utils.py
+++ b/strawberryfields/apps/graph/utils.py
@@ -151,7 +151,7 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
 
 
 def is_clique(graph: nx.Graph) -> bool:
-    """Determines if the input graph is a clique. A clique of :math:`n` nodes has :math:`n(
+    """Determines if the input graph is a clique. A clique of :math:`n` nodes has exactly :math:`n(
     n-1)/2` edges.
 
     Example usage:
--- a/strawberryfields/apps/graph/utils.py
+++ b/strawberryfields/apps/graph/utils.py
@@ -193,8 +193,8 @@ def c_0(clique: list, graph: nx.Graph):
         >>> from strawberryfields.apps.graph import utils
         >>> import networkx as nx
         >>> graph = nx.complete_graph(10)
-        >>> subgraph = [0, 1, 2, 3, 4]
-        >>> utils.c_0(subgraph, graph)
+        >>> clique = [0, 1, 2, 3, 4]
+        >>> utils.c_0(clique, graph)
         [5, 6, 7, 8, 9]
 
     Args:
@@ -233,8 +233,8 @@ def c_1(clique: list, graph: nx.Graph):
         >>> from strawberryfields.apps.graph import utils
         >>> import networkx as nx
         >>> graph = nx.wheel_graph(5)
-        >>> subgraph = [0, 1, 2]  # these nodes form a clique
-        >>> utils.c_1(subgraph, graph)
+        >>> clique = [0, 1, 2]
+        >>> utils.c_1(clique, graph)
         [(1, 3), (2, 4)]
 
     Args:
--- a/strawberryfields/apps/graph/utils.py
+++ b/strawberryfields/apps/graph/utils.py
@@ -195,7 +195,7 @@ def c_0(clique: list, graph: nx.Graph):
         [5, 6, 7, 8, 9]
 
     Args:
-        clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique.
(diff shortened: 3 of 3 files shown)


strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
...
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import resize
+        >>> import networkx as nx
+        >>> graph = nx.wheel_graph(5)
+        >>> graph.remove_edge(0, 4)
+        >>> subgraph = [0, 1, 2]  # these nodes form a clique
+        >>> resize.clique_swap(subgraph, graph)
+        [0, 1, 4]
```
trbromley:
```suggestion
        [0, 2, 3]
```
When I ran this (multiple times), this is what I got 🤔 
This makes sense, since `c_1 = [(1, 3)]` so we swap out node 1 for node 3.

trbromley: Which is a great motivation to use doctest 😁

strawberryfields/apps/graph/utils.py
```diff
@@ -151,13 +151,24 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
 
 
 def is_clique(graph: nx.Graph) -> bool:
-    """Determines if the input graph is a clique. A clique of n nodes has :math:`n*(n-1)/2` edges.
+    """Determines if the input graph is a clique. A clique of :math:`n` nodes has
+     :math:`n*(n-1)/2` edges.
```
trbromley:
```suggestion
     :math:`n(n-1)/2` edges.
```
Since it renders like latex, the asterisk is probably not required

strawberryfields/apps/graph/utils.py
```diff
@@ -151,13 +151,24 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+    Example usage:
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import utils
+        >>> import networkx as nx
+        >>> graph = nx.complete_graph(10)
+        >>> utils.is_clique(graph)
+        True
```
trbromley:
```suggestion
        True
        >>> graph = nx.empty_graph(10)
        >>> utils.is_clique(graph)
        False
```

strawberryfields/apps/graph/utils.py
```diff
@@ -151,13 +151,24 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
 
     Args:
-        graph (nx.Graph): The input graph
+        graph (nx.Graph): The input graph.
 
     Returns:
-        bool: ``True`` if input graph is a clique and ``False`` otherwise
+        bool: ``True`` if input graph is a clique and ``False`` otherwise.
```
trbromley:
```suggestion
        bool: ``True`` if input graph is a clique and ``False`` otherwise
```

strawberryfields/apps/graph/utils.py
```diff
@@ -151,13 +151,24 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
+
+        >>> from strawberryfields.apps.graph import utils
+        >>> import networkx as nx
+        >>> graph = nx.complete_graph(10)
+        >>> utils.is_clique(graph)
+        True
 
     Args:
-        graph (nx.Graph): The input graph
+        graph (nx.Graph): The input graph.
```
trbromley:
```suggestion
        graph (nx.Graph): the input graph
```

strawberryfields/apps/graph/utils.py
```diff
@@ -169,14 +180,27 @@ def c_0(clique: list, graph: nx.Graph):
...
+    Example usage:
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import utils
+        >>> import networkx as nx
+        >>> graph = nx.complete_graph(10)
+        >>> subgraph = [0, 1, 2, 3, 4]
+        >>> utils.c_0(subgraph, graph)
+        [5, 6, 7, 8, 9]
```
trbromley: Nice!

strawberryfields/apps/graph/utils.py
```diff
@@ -169,14 +180,27 @@ def c_0(clique: list, graph: nx.Graph):
...
 
     Args:
-        clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique
-        graph (nx.Graph): the input graph
+        clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique.
+        graph (nx.Graph): The input graph.
```
trbromley:
```suggestion
        graph (nx.Graph): the input graph
```

strawberryfields/apps/graph/utils.py
```diff
@@ -169,14 +180,27 @@ def c_0(clique: list, graph: nx.Graph):
...
+        >>> import networkx as nx
+        >>> graph = nx.complete_graph(10)
+        >>> subgraph = [0, 1, 2, 3, 4]
+        >>> utils.c_0(subgraph, graph)
+        [5, 6, 7, 8, 9]
 
     Args:
-        clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique
-        graph (nx.Graph): the input graph
+        clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique.
```
trbromley:
```suggestion
        clique (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
```

strawberryfields/apps/graph/utils.py
```diff
@@ -169,14 +180,27 @@ def c_0(clique: list, graph: nx.Graph):
...
+        graph (nx.Graph): The input graph.
 
     Returns:
-        list[int]: A list containing the :math:`C_0` nodes for the clique
+        list[int]: A list containing the :math:`C_0` nodes for the clique.
```
trbromley:
```suggestion
        list[int]: a list containing the :math:`C_0` nodes for the clique
```

strawberryfields/apps/graph/utils.py
```diff
@@ -196,16 +220,27 @@ def c_1(clique: list, graph: nx.Graph):
...
 
     Args:
         clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique
         graph (nx.Graph): the input graph
 
     Returns:
-       list[int]: A list of tuples ``[(i_clique, i), (j_clique, j),...,(k_clique, k)]``. Here
-       ``i,j,...,k`` are the nodes in :math:`C_1`, while ``i_clique, j_clique,...,k_clique`` are the
-        nodes in the clique they can be swapped with.
+       list[int]: A list of tuples. The first node in the tuple is the node in the clique and the
```
trbromley: Nice, looks a simpler explanation!
```suggestion
       list[tuple(int)]: A list of tuples. The first node in the tuple is the node in the clique and the
```

strawberryfields/apps/graph/utils.py
```diff
@@ -196,16 +220,27 @@ def c_1(clique: list, graph: nx.Graph):
...
+
+        >>> from strawberryfields.apps.graph import utils
+        >>> import networkx as nx
+        >>> graph = nx.wheel_graph(5)
+        >>> subgraph = [0, 1, 2]  # these nodes form a clique
+        >>> utils.c_1(subgraph, graph)
+        [(1, 3), (2, 4)]
 
     Args:
         clique (list[int]): A subgraph specified by a list of nodes; the subgraph must be a clique
```
trbromley:
```suggestion
         clique (list[int]): a subgraph specified by a list of nodes; the subgraph must be a clique
```

strawberryfields/apps/graph/utils.py
```diff
@@ -196,16 +220,27 @@ def c_1(clique: list, graph: nx.Graph):
...
+    Example usage:
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import utils
+        >>> import networkx as nx
+        >>> graph = nx.wheel_graph(5)
+        >>> subgraph = [0, 1, 2]  # these nodes form a clique
+        >>> utils.c_1(subgraph, graph)
+        [(1, 3), (2, 4)]
```
trbromley: 👍

strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
...
 
+    Example usage:
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import resize
+        >>> import networkx as nx
+        >>> graph = nx.wheel_graph(5)
+        >>> graph.remove_edge(0, 4)
+        >>> subgraph = [0, 1, 2]  # these nodes form a clique
```
trbromley: What about renaming from `subgraph` to `clique` (here and in other snippets)? `subgraph` makes it clear that it's a subgraph... but `clique` matches the argument for this function.

I have used `clique` in https://github.com/XanaduAI/strawberryfields/pull/178 so reminder to self to make sure things are consistent.

strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
...
+        >>> resize.clique_swap(subgraph, graph)
+        [0, 1, 4]
+
```
trbromley:
```suggestion
        >>> graph = nx.lollipop_graph(5, 1)
        >>> graph.remove_edge(0, 4)
        >>> graph.remove_edge(0, 3)
        >>> subgraph = [0, 1, 2]
        >>> resize.clique_swap(subgraph, graph, node_select="degree")
        [1, 2, 4]

```
This would give also show an example use of the `node_select` argument.

strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
     supported. Degree-based node selection involves picking the node with the greatest degree,
     with ties settled by uniform random choice.
 
+    Example usage:
```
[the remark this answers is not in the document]
jswinarton: I don't have experience with this extension specifically, but I do have experience with doc testing in general (from other languages) and I've found it to be extremely useful for making sure documentation is up to date, so I am strongly in favour provided there's nothing wrong with the extension itself.

strawberryfields/apps/graph/utils.py
```diff
@@ -151,13 +151,24 @@ def is_subgraph(subgraph: Iterable, graph: nx.Graph):
...
 def is_clique(graph: nx.Graph) -> bool:
-    """Determines if the input graph is a clique. A clique of n nodes has :math:`n*(n-1)/2` edges.
+    """Determines if the input graph is a clique. A clique of :math:`n` nodes has
+     :math:`n*(n-1)/2` edges.
+
+    Example usage:
+
+    .. code-block::
+
+        >>> from strawberryfields.apps.graph import utils
```
jswinarton: This import signature is a bit odd, why not just 
```suggestion
        >>> from strawberryfields.apps.graph.utils import is_clique
```

jswinarton: Actually, ideally you wouldn't have this import in code examples at all, since it's kind of a given. Do you need it for doctests to work?

strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
...
+
+    .. code-block::
```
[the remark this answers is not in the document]
ixfoduap: Any specific reason why the recommended usage in sphinx is without the directive? Unless it may cause problems, it's probably best to leave `.. code-block::` for now

strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,18 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
...
+
+        >>> from strawberryfields.apps.graph import resize
```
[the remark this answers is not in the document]
ixfoduap: Yes, we should have a discussion about this, probably as a research ticket for next sprint. The simpler the syntax is, the better, so I would prefer usage like Tom is suggesting above.

strawberryfields/apps/graph/resize.py
ixfoduap: Good catch! I guess I labelled the nodes differently than networkx. Indeed, good motivation for doctest

strawberryfields/apps/graph/utils.py
ixfoduap: ^ We need to check for consistency throughout the documentation regarding upper cases and full stops

strawberryfields/apps/graph/resize.py
ixfoduap: Yes, I like it! It would also mean the comment can probably be removed

strawberryfields/apps/graph/utils.py
ixfoduap: I think it's best to make import statements explicit: this will be useful for less experienced users that can use these examples to build their own code.

In that case, isn't
`from strawberryfields.apps.graph import utils` 
preferable if it's part of code that calls other functions in `utils`? <-- Genuine question, not rhetorical

jswinarton: Maybe, but it's important to consider the user's namespace. What would happen if every library recommended importing a library called `utils`? In this case since it's just a single example function, I think it's better to import just the function you're calling.

strawberryfields/apps/graph/resize.py
trbromley: Done!

strawberryfields/apps/graph/resize.py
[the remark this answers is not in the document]
trbromley: For reference, this is how it renders:

Annotation 2019-08-30 150702

strawberryfields/apps/graph/resize.py
```diff
@@ -308,6 +308,24 @@ def clique_swap(clique: list, graph: nx.Graph, node_select: str = "uniform") ->
     supported. Degree-based node selection involves picking the node with the greatest degree,
     with ties settled by uniform random choice.
 
+    Example usage:
+
```
trbromley: I also recommend adding the `sphinx_autodoc_typehints` extension to Sphinx, any objections?

This is how things render **without** the extension:
Annotation 2019-08-30 153515

and this is how things render **with**:
Annotation 2019-08-30 153552

There is also a https://pypi.org/project/sphinx-autodoc-napoleon-typehints/ but I couldn't get it working.

jswinarton: That's weird, why does the argument to `graph` display a `MagicMock` object in the first instance?

strawberryfields/apps/graph/resize.py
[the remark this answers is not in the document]
josh146: I'm not sure it really matters too much, but for consistency, I would prefer if no `.. code-block::` was used for `>>>`, since the remaining SF documentation is written that way.

It is easy enough to add `.. doc-test::` where it is needed in a future PR (and indent)

strawberryfields/apps/graph/resize.py
josh146: I'm in favour of adding this extension. The one thing to double check is if it affects the docstrings of existing SF documentation.

strawberryfields/apps/graph/utils.py
ixfoduap: Ok, following also @josh146 's advice, I'll remove import statements

strawberryfields/apps/graph/resize.py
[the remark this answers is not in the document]
trbromley: Thanks! Done!

strawberryfields/apps/graph/resize.py
trbromley: I have added in!

So to check how it looks like with the rest of the SF documentation (TLDR: I didn't see any difference!), here is a screenshot **WITH** `sphinx_autodoc_typehints`:

Annotation 2019-09-03 123027



and **WITHOUT**:


Annotation 2019-09-03 123046

trbromley: Also I don't see any warnings during `make docs`.

strawberryfields/apps/graph/utils.py
trbromley: Have removed! Thanks!

doc/conf.py
```diff
@@ -110,6 +110,7 @@ class TypeMock(type):
     'sphinx.ext.viewcode',
     'sphinxcontrib.bibtex',
     'edit_on_github',
+    'sphinx_autodoc_typehints',
```
josh146: :+1:

strawberryfields/apps/graph/resize.py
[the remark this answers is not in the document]
co9olguy: Little late getting to this, but I definitely agree that 
```
from strawberryfields import apps
apps.some_cool_function(...)
```
is the way to go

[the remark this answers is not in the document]
trbromley: Great! Planning to sort this out in next sprint

PR #180, opened 2019-09-04
Provide applications layer in docs and polish max clique documentation

trbromley:
The purpose of this PR is to make the applications layer accessible in the SF documentation by adding in as a table of contents. This does not include a full integration (e.g., giving more details about the apps layer in the top level `index.rst` file), which can be done at a later date once the applications layer matures.

This PR also includes a check through of the documentation for the recently added maximum clique functionality.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -173,3 +173,4 @@ Strawberry Fields is **free** and **open source**, released under the Apache Lic
    code/apps/graph.resize
    code/apps/graph.sample
    code/apps/graph.utils
+   code/apps/graph.max_clique
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -161,3 +161,15 @@ Strawberry Fields is **free** and **open source**, released under the Apache Lic
    code/backend.gaussian
    code/backend.fock
    code/backend.tf
+
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Applications Layer
+   :hidden:
+
+   code/apps/sample
+   code/apps/graph.dense
+   code/apps/graph.resize
+   code/apps/graph.sample
+   code/apps/graph.utils
--- a/strawberryfields/apps/graph/max_clique.py
+++ b/strawberryfields/apps/graph/max_clique.py
@@ -49,7 +49,7 @@ from strawberryfields.apps.graph import resize
 
 
 def local_search(clique: list, graph: nx.Graph, iterations, node_select: str = "uniform") -> list:
-    """Use local search algorithm to identify large cliques.
+    """Local search algorithm for identifying large cliques.
 
     This function implements a version of the local search algorithm given in
     :cite:`pullan2006dynamic` and :cite:`pullan2006phased`. It proceeds by iteratively applying
--- a/strawberryfields/apps/graph/max_clique.py
+++ b/strawberryfields/apps/graph/max_clique.py
@@ -80,7 +80,7 @@ def local_search(clique: list, graph: nx.Graph, iterations, node_select: str = "
     of iterations has been carried out or when a dead end has been reached. The function reaches
     a dead end when it is unable to perform any new swaps or growth.
 
-    Example usage:
+    **Example usage:**
 
     >>> graph = nx.lollipop_graph(4, 1)
     >>> graph.add_edge(4, 2)
--- a/strawberryfields/apps/graph/resize.py
+++ b/strawberryfields/apps/graph/resize.py
@@ -255,7 +255,7 @@ def clique_grow(clique: list, graph: nx.Graph, node_select: str = "uniform") -> 
     supported. Degree-based node selection involves picking the node with the greatest degree,
     with ties settled by uniform random choice.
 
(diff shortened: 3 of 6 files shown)


trbromley:
Note commit https://github.com/XanaduAI/strawberryfields/pull/180/commits/758844da20d9269905286b40d2820e868f093330 made the "Example usage:" in the docstrings bold.

How it looked before:
Annotation 2019-09-09 185625

How it looks now:
Annotation 2019-09-09 190016

I think it looks better now.

doc/index.rst
```diff
@@ -161,3 +161,16 @@ Strawberry Fields is **free** and **open source**, released under the Apache Lic
...
+   :maxdepth: 2
+   :caption: Applications Layer
+   :hidden:
+
+   code/apps/sample
+   code/apps/graph.dense
+   code/apps/graph.resize
+   code/apps/graph.sample
+   code/apps/graph.utils
+   code/apps/graph.max_clique
```
josh146: It probably makes more sense to have this before the backend section (since users will not interact with backends directly). Unless this was deliberate to keep the apps layer 'out of sight'?

strawberryfields/apps/graph/max_clique.py
```diff
@@ -80,7 +80,7 @@ def local_search(clique: list, graph: nx.Graph, iterations, node_select: str = "
     of iterations has been carried out or when a dead end has been reached. The function reaches
     a dead end when it is unable to perform any new swaps or growth.
 
-    Example usage:
+    **Example usage:**
```
josh146: thumb

strawberryfields/apps/graph/resize.py
```diff
@@ -19,16 +19,15 @@
...
+algorithms for finding dense subgraphs and maximum cliques for the following elements:
 
-Resizing functionality is provided by the :func:`resize_subgraphs` function, allowing the user to
-make use of different methods for resizing. The available methods are specified in
-:func:`resize_subgraphs`.
+- Samples from GBS have a variable number of clicks, resulting in subgraph samples of variable
+  size (functionality provided by :func:`~.sample_subgraphs`). Resizing can be used to realize a
+  certain property, such as to reach a desired size (see :func:`resize_subgraphs`) or to resize
+  to a clique (see :func:`clique_shrink`).
+- Heuristic algorithms can use graph resizing to explore the search space, for example growing and
```
josh146:
```suggestion

- Heuristic algorithms can use graph resizing to explore the search space, for example growing and
```

strawberryfields/apps/graph/utils.py
```diff
@@ -227,8 +227,8 @@ def c_1(clique: list, graph: nx.Graph):
         graph (nx.Graph): the input graph
 
     Returns:
-       list[tuple(int)]: A list of tuples. The first node in the tuple is the node in the clique and the
-       second node is the outside node it can be swapped with.
+       list[tuple(int)]: A list of tuples. The first node in the tuple is the node in the clique
+       and the second node is the outside node it can be swapped with.
```
josh146: what was the reason for this change?

strawberryfields/apps/sample.py
```diff
@@ -57,22 +56,14 @@
 QUANTUM_BACKENDS = ("gaussian",)
 """tuple[str]: Available quantum backends for sampling."""
 
-BACKEND_DEFAULTS = {
-    "remote": False,
-    "backend": "gaussian",
-    "threshold": True,
-    "postselect": 0,
-}
+BACKEND_DEFAULTS = {"remote": False, "backend": "gaussian", "threshold": True, "postselect": 0}
```
josh146: new black -l setting? :laughing:

doc/index.rst
trbromley: Yes I was thinking to keep things on the down low for now and when the docs refactor is done we can think about where to place it and adding some introductory text, does that work?

strawberryfields/apps/graph/resize.py
trbromley: 😮 
I didn't know this was true...
Annotation 2019-09-11 090522

Thanks!

strawberryfields/apps/graph/utils.py
trbromley: 🤔 
It was actually longer than 100 characters but had somehow slipped through the net.
IIRC that line had been changed from a github suggestion, where it is sometimes hard to see the max line width

strawberryfields/apps/sample.py
trbromley: Yep, I had run black without a `-l` originally, the new one is `black -l 100` (although I do this through a pre-commit hook).

strawberryfields/apps/graph/resize.py
josh146: Yep, makes no difference to the rendered docs, but makes it easier to read in the code!

doc/index.rst
josh146: I'm happy with it for now!

PR #181, opened 2019-09-09
Apps layer: sample_to_event and related functions for graph similarity

trbromley:
Graph similarity involves dealing with orbits and even coarser grained events composed of multiple samples from GBS. This PR provides some of the elementary functions for graph similarity:

- `sample_to_orbit`
- `sample_to_event`
- `orbits`

### DIFF
--- a/strawberryfields/apps/graph/similarity.py
+++ b/strawberryfields/apps/graph/similarity.py
@@ -1,0 +1,32 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+r"""
+Graph similarity
+================
+
+**Module name:** :mod:`strawberryfields.apps.graph.similarity`
+
+.. currentmodule:: strawberryfields.apps.graph.similarity
+
+Functionality for calculating feature vectors of graphs using GBS
+
+Summary
+-------
+
+.. autosummary::
+
+
+Code details
+^^^^^^^^^^^^
+"""
--- a/strawberryfields/apps/graph/similarity.py
+++ b/strawberryfields/apps/graph/similarity.py
@@ -55,7 +55,7 @@ def sample_to_event(sample: list, max_count_per_mode: int) -> list:
     Args:
         sample (list[int]): a sample from GBS
         max_count_per_mode (int): the maximum number of photons counted in any given mode for a
-            sample to categorized as an event. Samples with counts exceeding this value are
+            sample to be categorized as an event. Samples with counts exceeding this value are
             attributed the event ``None``.
 
     Returns:
--- a/strawberryfields/apps/graph/similarity.py
+++ b/strawberryfields/apps/graph/similarity.py
@@ -36,7 +36,8 @@ Code details
 def sample_to_orbit(sample: list) -> list:
     """Provides the orbit corresponding to a given sample.
 
-    Orbits are simply a sorting of samples in non-increasing order.
+    Orbits are simply a sorting of samples in non-increasing order with the zeros at the end
+    removed.
 
     Args:
         sample (list[int]): a sample from GBS
@@ -44,7 +45,8 @@ def sample_to_orbit(sample: list) -> list:
     Returns:
         list[int]: the orbit of the sample
     """
-    return sorted(sample, reverse=True)
+    sorted_sample = sorted(sample, reverse=True)
+    return [c for c in sorted_sample if c != 0]
 
 
 def sample_to_event(sample: list, max_count_per_mode: int) -> list:
--- a/strawberryfields/apps/graph/similarity.py
+++ b/strawberryfields/apps/graph/similarity.py
@@ -73,7 +73,7 @@ def orbits(photon_number: int) -> Generator[list, None, None]:
 
     Provides a generator over the integer partitions of ``photon_number``.
     Code derived from `website <http://jeromekelleher.net/generating-integer-partitions.html>`__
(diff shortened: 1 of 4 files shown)


trbromley:
Opening this up to @ixfoduap to take a first look

strawberryfields/apps/graph/similarity.py
```diff
@@ -0,0 +1,125 @@
...
+    if max(sample) <= max_count_per_mode:
+        return sum(sample)
+
+    return None
+
+
+def orbits(photon_number: int) -> Generator[list, None, None]:
+    """Generate all the possible orbits for a given photon number.
+
+    Provides a generator over the integer partitions of ``photon_number``.
```
ixfoduap: Great find Tom! :muscle:

strawberryfields/apps/graph/similarity.py
```diff
@@ -0,0 +1,125 @@
...
+        l = k + 1
+        while x <= y:
+            a[k] = x
+            a[l] = y
+            yield sorted(a[: k + 2], reverse=True)
+            x += 1
+            y -= 1
+        a[k] = x + y
+        y = x + y - 1
+        yield sorted(a[: k + 1], reverse=True)
```
ixfoduap: Learnt about generators thanks to this! 
Using generators is probably overkill for the usage we expect for this function, but I'm in favour keeping as is. The code is stable and employing state-of-the-art methods.

tests/apps/test_graph_similarity.py
```diff
@@ -0,0 +1,76 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+r"""
+Unit tests for strawberryfields.apps.graph.similarity
+"""
+# pylint: disable=no-self-use,unused-argument,too-many-arguments
+import itertools
+
```
ixfoduap: Is it standard to use these spaces between import statements?

tests/apps/test_graph_similarity.py
```diff
@@ -0,0 +1,76 @@
...
+pytestmark = pytest.mark.apps
+
+
+@pytest.mark.parametrize("dim", [2, 3])
+def test_sample_to_orbit(dim):
+    """Test if function ``similarity.sample_to_orbit`` correctly returns the original orbit after
+    taking all permutations over the orbit. The starting orbit is a fixed repetition of 2's,
+    followed by a repetition of 1's, each repeated ``dim`` times, and in a system of ``3 * dim``
+    modes."""
+    orbit = [2] * dim + [1] * dim
```
ixfoduap: Should we also test different orbits?

tests/apps/test_graph_similarity.py
```diff
@@ -0,0 +1,76 @@
...
+    permutations = itertools.permutations(sorted_sample)
+    assert all([similarity.sample_to_orbit(p) == orbit for p in permutations])
+
+
+def test_orbits():
+    """Test if function ``similarity.orbits`` correctly returns the integer partitions of 5.
+    This test does not require ``similarity.orbits`` to return the orbits in any specified order."""
+    partition = [[5], [4, 1], [3, 2], [3, 1, 1], [2, 1, 1, 1], [2, 2, 1], [1, 1, 1, 1, 1]]
+    orbits = list(similarity.orbits(5))
+    try:
```
ixfoduap: Why exactly do we use `try` and `except` instead of just the for loop removing partitions? Why would a `ValueError` occur?

tests/apps/test_graph_similarity.py
```diff
@@ -0,0 +1,76 @@
...
+
+@pytest.mark.parametrize("dim", [3, 4, 5])
+@pytest.mark.parametrize("max_count_per_mode", [1, 2])
+def test_sample_to_event(dim, max_count_per_mode):
+    """Test if function ``similarity.sample_to_event`` gives the correct set of events when
+    applied to all orbits with a fixed number of photons ``dim``. This test ensures: (i) that the
+    only events returned are either ``dim`` or ``None``; (ii) that orbits exceeding the
+    ``max_count_per_mode`` value are attributed the ``None`` event; and that (iii) that orbits
+    not exceeding the ``max_count_per_mode`` are attributed the event ``dim``."""
+    orbits = list(similarity.orbits(dim))
```
ixfoduap: Genuinely asking: should this call to the `orbits()` function be monkey-patched? Will this test work properly if `orbits()` has issues?

tests/apps/test_graph_similarity.py
trbromley: Ok, have changed it now so that we check all orbits for a fixed photon number.

trbromley:
Thanks @ixfoduap, have responded to your comments and made changes. Would like to open up to reviews from others now!

tests/apps/test_graph_similarity.py
```diff
@@ -0,0 +1,96 @@
...
+        """Test if function returns all the integer partitions of 5. This test does not
+        require ``similarity.orbits`` to return the orbits in any specified order."""
+        partition = [[5], [4, 1], [3, 2], [3, 1, 1], [2, 1, 1, 1], [2, 2, 1], [1, 1, 1, 1, 1]]
+        orbits = list(similarity.orbits(5))
+        try:
+            for o in orbits:
+                partition.remove(o)
+        except ValueError:
+            pass
+        assert not partition
```
josh146: I think a better approach here is to sort partition and sort orbits, and then just assert that they are identical.

tests/apps/test_graph_similarity.py
josh146: > This test ensures: (i) that the only events returned are either ``dim`` or ``None``; (ii) that orbits exceeding the ``max_count_per_mode`` value are attributed the ``None`` event; and that (iii) that orbits not exceeding the ``max_count_per_mode`` are attributed the event ``dim``.

It sounds like this should be three separate tests :thinking: 

> Genuinely asking: should this call to the orbits() function be monkey-patched? Will this test work properly if orbits() has issues?

Rather than monkeypatching `orbits`, or using `orbits`, I'll throw another spanner into the list :laughing: 

What I would do here is hardcode in some orbits directly as test data.

tests/apps/test_graph_similarity.py
trbromley: Nice! Already in my pre-commit hook 😁

PR #182, opened 2019-09-12
Improving Apps Structure: Part 1 - Remove graph layer and consolidate clique functions

trbromley:
This PR partly carries out our plan to improve the import structure of the applications layer. I have split this up because it is already a relatively large PR.

This PR covers removing the graph folder/module from the apps layer, renaming from `apps` to `gbs` and consolidating clique-based functions into one location (e.g., before they were separated into a `max_clique` module and `resize` module), as well as renaming the max clique functions, resulting in for example: `from strawberryfields import gbs; gbs.clique.grow([1, 2, 3])`.

What still needs to be done:
- Consolidate sampling functions into a single module (e.g., currently in `sample` and `g_sample`)
- Consolidate graph-based functions into a single module (e.g., currently in `dense` and `resize`)
- Consolidate some top-level functions into another module
and also update the `__init__.py` in each case


trbromley:
Thanks @jswinarton @josh146!

PR #183, opened 2019-09-12
Provide functionality to randomly sample from events and orbits

trbromley:
This PR provides the ability to generate samples from an input orbit or event. It constitutes part of our plan to calculate feature vector probabilities through Monte Carlo sampling, which will use these functions.


trbromley:
Note that https://github.com/XanaduAI/strawberryfields/pull/182 will be merged soon and will require a bit of a restructure once updated on this branch, but the content shouldn't change.

tests/apps/test_graph_similarity.py
```diff
@@ -89,3 +92,113 @@ def test_sample_to_event(dim, max_count_per_mode):
...
+            [1, 1, 1, 1, 1],
+            [2, 1, 1, 1, 0],
+            [3, 1, 1, 0, 0],
+            [2, 2, 1, 0, 0],
+            [4, 1, 0, 0, 0],
+            [3, 2, 0, 0, 0],
+            [5, 0, 0, 0, 0],
+        ]  # padding orbits with zeros at the end for comparison to samples
+
+        counts = [Counter(similarity.orbit_to_sample(o, modes)) for o in all_orbits_cumulative]
```
ixfoduap: Every PR there is a new python function to learn! Nice use of Counter :-)

tests/apps/test_graph_similarity.py
```diff
@@ -89,3 +92,113 @@ def test_sample_to_event(dim, max_count_per_mode):
...
+            similarity.event_to_sample(2, 0, 5)
+
+    def test_high_photon(self):
+        """Test if function raises a ``ValueError`` if ``photon_number`` is so high that it
+        cannot correspond to a sample given the constraints of ``max_count_per_mode`` and
+        ``modes``"""
+        with pytest.raises(ValueError, match="No valid samples can be generated."):
+            similarity.event_to_sample(5, 1, 4)
+
+    @pytest.mark.parametrize("photon_num", [3, 4, 5, 6])
```
ixfoduap: Does this test loop over 4x2x4 = 32 different input parameters? If so, is it worth decreasing this?

tests/apps/test_graph_similarity.py
```diff
@@ -89,3 +92,113 @@ def test_sample_to_event(dim, max_count_per_mode):
...
+    @pytest.mark.parametrize("modes_dim", [10, 11])
+    @pytest.mark.parametrize("count", [3, 4, 5, 6])
+    def test_sample_length(self, photon_num, modes_dim, count):
+        """Test if function returns a sample that is of correct length ``modes_dim``."""
+        samp = similarity.event_to_sample(
+            photon_number=photon_num, max_count_per_mode=count, modes=modes_dim
+        )
+        assert len(samp) == modes_dim
+
+    @pytest.mark.parametrize("photon_num", [3, 4, 5, 6])
```
ixfoduap: Same here, maybe too many inputs

tests/apps/test_graph_similarity.py
```diff
@@ -89,3 +92,113 @@ def test_sample_to_event(dim, max_count_per_mode):
...
+    @pytest.mark.parametrize("modes_dim", [10, 11])
+    @pytest.mark.parametrize("count", [3, 4, 5, 6])
+    def test_sample_sum(self, photon_num, modes_dim, count):
+        """Test if function returns a sample that has the correct number of photons."""
+        samp = similarity.event_to_sample(
+            photon_number=photon_num, max_count_per_mode=count, modes=modes_dim
```
ixfoduap: Not a big deal but, do we have to use `photon_number=photon_num, max_count_per_mode=count, modes=modes_dim` instead of just `photon_num, count, modes_dim` ?

tests/apps/test_graph_similarity.py
```diff
@@ -89,3 +92,113 @@ def test_sample_to_event(dim, max_count_per_mode):
...
+    @pytest.mark.parametrize("modes_dim", [10, 11])
+    @pytest.mark.parametrize("count", [3, 4, 5, 6])
+    def test_sample_max_count(self, photon_num, modes_dim, count):
+        """Test if function returns a sample that has maximum element not exceeding ``count``."""
+        samp = similarity.event_to_sample(
+            photon_number=photon_num, max_count_per_mode=count, modes=modes_dim
+        )
+        assert max(samp) <= count
+
+    @pytest.mark.parametrize("photon_num", [5, 6])
```
ixfoduap: Super nice test!

tests/apps/test_graph_similarity.py
```diff
@@ -89,3 +92,113 @@ def test_sample_to_event(dim, max_count_per_mode):
...
     assert events == target_events
+
+
+class TestOrbitToSample:
+    """Tests for the function ``strawberryfields.apps.graph.similarity.orbit_to_sample``"""
+
+    def test_low_modes(self):
+        """Test if function raises a ``ValueError`` if fed an argument for ``modes`` that does
+        not exceed the length of the input orbit."""
+        with pytest.raises(ValueError, match="Number of modes cannot"):
```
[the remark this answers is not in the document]
josh146: the `match` argument accepts a regex string, so it's even more powerful, you can place any regex there that you want to match!

Rule of thumb is to make it specific enough so that you will only catch the error you intend to.

tests/apps/test_graph_similarity.py
josh146: I'm in favour of less verbose argument names, even if it means users will have to double check the docstring for queries.

trbromley: Have made shorter 😁

tests/apps/test_graph_similarity.py
josh146: tbh, if it takes an insignificant amount of time, testing 32 input parameters should be okay :slightly_smiling_face: 
 The only time it is worth removing parameters you are testing is if it is a time sink when running the tests.

tests/apps/test_graph_similarity.py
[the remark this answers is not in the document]
ixfoduap: Yes, it's enough

tests/apps/test_graph_similarity.py
ixfoduap: It's ok to do more, but I don't think we gain much by doing it, so best to keep things small

PR #184, opened 2019-09-14
Apps layer: Provide event probabilities calculated by Monte Carlo

trbromley:
For the **graph similarity** application of the `gbs` layer, we plan to provide feature vectors constructed from event probabilities.

Event probabilities can be hard to calculate, with brute force requiring a loop over all orbits within an event and then all samples within an orbit - where each sample probability is calculated through a Hafnian.

This PR provides an alternative and approximate method for calculating the event probability. The method provided involves sampling of samples within an event uniformly at random, resulting in a Monte Carlo estimate of the probability.


strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +212,155 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
             available_modes.remove(j)
 
     return sample
+
+
+def fac_prod(orbit: list) -> int:
```
trbromley: 🤔 Given that we only use this function twice, and that it can also be done very simply as
```python
import numpy as np

l = [1, 2, 3]
np.prod(factorial(l))
```
I think we should remove this function, both to lighten the tests and also to not reinvent the wheel.

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +212,155 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    """
+
+    prod = 1
+    for i in orbit:
+        prod = prod * factorial(i)
+
+    return prod
+
+
+def compress_sample(sample: list) -> list:
```
trbromley: A very similar function `sample_to_subgraph` was actually used in `gbs.g_sample` to convert from samples to subgraphs. I decided to combine that function with this one and place in the `gbs.sample` module.

strawberryfields/gbs/sample.py
```diff
@@ -227,3 +228,33 @@ def random_seed(seed: int = None) -> None:
...
     """
     np.random.seed(seed)
+
+
+def modes_from_counts(sample: list) -> list:
+    r"""Convert a sample of photon counts to a list of modes where photons are detected.
+
+    There are two main methods of representing a sample. In the first, the number of photons
+    detected in each mode is specified. In the second, the modes in which each photon was detected
+    are listed. Since there are typically fewer photons than modes, the second method gives a
```
trbromley: How about we remove the "since there are..." sentence?

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +226,151 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    Args:
+        orbit (list[int]): orbit for which to estimate the probability
+        graph (nx.Graph): input graph encoded in the GBS device
+        n_mean (float): total mean photon number of the GBS device
+        samples (int): number of samples used in the Monte Carlo estimation
+
+    Returns:
+        float: estimated probability of the orbit with respect to the graph
+    """
+    modes = graph.order()
```
trbromley: I feel like `p_orbit_mc` and `estimate_event_prob` could harness SF rather than calling the more elementary parts of `thewalrus`. E.g., we could call `state.fock_prob([1,2,0,1])`? (@josh146)

strawberryfields/gbs/similarity.py
trbromley: Ok, so I eventually changed this function to a single underscore function, but after talking with @josh146 @ixfoduap we are going to delete this function and just have it directly in the code.

strawberryfields/gbs/similarity.py
```diff
@@ -285,20 +257,16 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
     return cardinality
 
 
-def p_orbit_mc(orbit: list, graph: nx.Graph, n_mean: float = 5, samples: int = 1000) -> float:
```
trbromley: Also need to update the ordering in the `Args:` section of the docstring.

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +209,145 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+        graph (nx.Graph): input graph encoded in the GBS device
+        n_mean (float): total mean photon number of the GBS device
+        samples (int): number of samples used in the Monte Carlo estimation
+
+    Returns:
+        float: estimated orbit probability
+    """
+
+    modes = graph.order()
+    photons = np.sum(orbit)
```
trbromley:
```suggestion
    photons = sum(orbit)
```

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +209,145 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    with p.context as q:
+        sf.ops.GraphEmbed(A, mean_photon_per_mode=mean_photon_per_mode) | q
+    eng = sf.LocalEngine(backend="gaussian")
+    result = eng.run(p)
+
+    prob = 0
+
+    for _ in range(samples):
+        sample = orbit_to_sample(orbit, modes)
+        prob += result.state.fock_prob(sample, cutoff=photons+1)
```
trbromley: What does the `cutoff` arg do?

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +209,145 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    for _ in range(samples):
+        sample = event_to_sample(photon_number, max_count_per_mode, modes)
+        prob += result.state.fock_prob(sample, cutoff=np.sum(sample)+1)
+
+    prob = prob*event_cardinality(photon_number, max_count_per_mode, modes) / samples
+
+    return prob
+
+
+graph = nx.erdos_renyi_graph(30, 0.5)
```
trbromley: sanity checks 😁

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,105 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+    def test_mean_computation_orbit(self, monkeypatch):
+        """Tests if the calculation of the sample mean is performed correctly. The test
+        monkeypatches the fock_prob function so that the probability is the same for each sample and
+         is equal to 1/5, i.e., one over the number of samples in the orbit [1,1,1,1] for 5 modes."""
+        graph = nx.complete_graph(5)
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                      lambda *args, **kwargs: 0.2)
+            tol = 1e-5
```
trbromley: Nice test 😁
Is the tolerance definitely needed? It's weird if we need it

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,105 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+class TestProbOrbitMC:
+    """Tests for the function ``strawberryfields.gbs.similarity.p_orbit_mc.``"""
+
+    @pytest.mark.parametrize("n_mean", [0, 2, 4])
+    def test_actual_prob_orbit(self, n_mean):
+        """Tests if the output of the function is a valid probability between 0 and 1."""
+        graph = nx.complete_graph(10)
+
+        assert 0 <= similarity.p_orbit_mc(graph, [], n_mean) <= 1
```
trbromley: I'm quite nervous about this non-deterministic test 😅 
Isn't it possible to exceed 1 if we hit some high probability events?

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,105 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+         is equal to 1/5, i.e., one over the number of samples in the orbit [1,1,1,1] for 5 modes."""
+        graph = nx.complete_graph(5)
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                      lambda *args, **kwargs: 0.2)
+            tol = 1e-5
+
+            assert np.abs(similarity.p_orbit_mc(graph, [1, 1, 1, 1]) - 1.0) < tol
+
+    def test_prob_vacuum_orbit(self):
```
trbromley: Nice test!

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,105 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+        assert similarity.p_event_mc(graph, 0, 0, 0) == 1.0
+
+    @pytest.mark.parametrize("n_mean", [0, 2, 4])
+    def test_actual_prob_event(self, n_mean):
+        """Tests if the output of the function is a valid probability between 0 and 1."""
+
+        graph = nx.complete_graph(10)
+
+        assert 0 <= similarity.p_event_mc(graph, 0, 0, 0) <= 1
```
trbromley: I'm also not sure about this probabilistic test, but if we have it do we need to replace the zeros in the above line?

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,105 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+    def test_mean_event(self, monkeypatch):
+        """Tests if the calculation of the sample mean is performed correctly. The test
+        monkeypatches the fock_prob function so that the probability is the same for each sample and
+         is equal to 1/216, i.e., one over the number of samples in the event with 5 modes,
+         6 photons, and mac 3 photons per mode."""
```
trbromley:
```suggestion
        6 photons, and max 3 photons per mode."""
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,105 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        """Tests if the output of the function is a valid probability between 0 and 1."""
+
+        graph = nx.complete_graph(10)
+
+        assert 0 <= similarity.p_event_mc(graph, 0, 0, 0) <= 1
+
+    def test_mean_event(self, monkeypatch):
+        """Tests if the calculation of the sample mean is performed correctly. The test
+        monkeypatches the fock_prob function so that the probability is the same for each sample and
+         is equal to 1/216, i.e., one over the number of samples in the event with 5 modes,
```
trbromley:
```suggestion
        is equal to 1/216, i.e., one over the number of samples in the event with 5 modes,
```

strawberryfields/gbs/similarity.py
ixfoduap: I'm not 100% sure but I think it determines the max size of hafnians that enter the calculations. Regardless, it will throw an error if it is set to any number smaller than `photons + 1`

strawberryfields/gbs/similarity.py
```diff
@@ -156,7 +165,7 @@ def orbit_to_sample(orbit: list, modes: int) -> list:
     if modes < len(orbit):
         raise ValueError("Number of modes cannot be smaller than length of orbit")
 
-    sample = orbit + [0] * (modes - len(orbit))
+    sample = list(orbit) + [0] * (modes - len(orbit))
```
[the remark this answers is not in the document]
ixfoduap: Checking that the input is a valid orbit may be a good idea. However, I believe that adding zeros here won't result in an issue since the added zeros `[0]* (modes - len(orbit))` takes the length of the orbit into account.

If the input `orbit` is a list, why do we need `list(orbit)`?

strawberryfields/gbs/similarity.py
```diff
@@ -182,8 +191,8 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
     Returns:
         list[int]: a sample in the event
     """
-    if max_count_per_mode < 1:
-        raise ValueError("Maximum number of photons per mode must be equal or greater than 1")
+    if max_count_per_mode < 0:
+        raise ValueError("Maximum number of photons must be non-negative")
```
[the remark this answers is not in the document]
ixfoduap: I think it's good to keep a clear distinction between total number of photons and maximum number of photons per mode

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    counts = list(Counter(sample).values())
+    return int(factorial(modes) / np.prod(factorial(counts)))
+
+
+def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -> int:
+    """Gives the number of samples belonging to the input event.
+
+    **Example usage**:
+
+    >>> event_cardinality(5, 2, 6)
```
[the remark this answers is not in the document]
ixfoduap: Yes! That's better

tests/gbs/test_similarity.py
ixfoduap: Hmmm, for the orbit `[]` the output will be the probability of vacuum, which will not exceed 1. However, you still have a good point: maybe it is a good idea to monkeypatch all relevant non-deterministic parts of the function?

strawberryfields/gbs/similarity.py
ixfoduap: Yes! Will take care of that

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+        sf.ops.GraphEmbed(A, mean_photon_per_mode=mean_photon_per_mode) | q
+
+    eng = sf.LocalEngine(backend="gaussian")
+    result = eng.run(p)
+
+    prob = 0
+
+    for _ in range(samples):
+        sample = orbit_to_sample(orbit, modes)
+        prob += result.state.fock_prob(sample, cutoff=photons + 1)
```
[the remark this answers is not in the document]
ixfoduap: Not 100% percent sure, but I think it limits the size of hafnians used to calculate probabilities. There is a default value, but any value below `photons+1` will raise an error

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    To make this estimate, several samples from the event are drawn uniformly at random. For each
+    sample, we calculate the probability of observing that sample from a GBS programmed according to
+    the input graph and mean photon number. These probabilities are then rescaled according to the
+    cardinality of the event. The estimate is the sample mean of the rescaled probabilities.
+
+    **Example usage**:
+
+    >>> graph = nx.complete_graph(8)
+    >>> p_event_mc(graph, 4, 2)
+    0.1395
```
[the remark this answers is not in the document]
ixfoduap: 1.8 secs

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+    Args:
+        graph (nx.Graph): the input graph encoded in the GBS device
+        photon_number (int): number of photons in the event
+        max_count_per_mode (int): maximum number of photons per mode in the event
+        n_mean (float): the total mean photon number of the GBS device
+        samples (int): the number of samples used in the Monte Carlo estimation
```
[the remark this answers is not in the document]
ixfoduap: Prefer without "the"

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+        sf.ops.GraphEmbed(A, mean_photon_per_mode=mean_photon_per_mode) | q
+
+    eng = sf.LocalEngine(backend="gaussian")
+    result = eng.run(p)
+
+    prob = 0
+
+    for _ in range(samples):
+        sample = event_to_sample(photon_number, max_count_per_mode, modes)
+        prob += result.state.fock_prob(sample, cutoff=np.sum(sample) + 1)
```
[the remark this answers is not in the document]
ixfoduap: `cutoff` must be an int, while `sample` is a list. However, I've changed `np.sum(sample)` to `photon_number`, which is already an input to the function

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+    prob = prob * event_cardinality(photon_number, max_count_per_mode, modes) / samples
+
+    return prob
```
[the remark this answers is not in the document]
ixfoduap: The orbit functions may not end up being used, so I prefer to keep as is

tests/gbs/test_similarity.py
ixfoduap: Yes, it's needed, I got an output of 0.999999999986 or something

tests/gbs/test_similarity.py
ixfoduap: Removing probabilistic tests

strawberryfields/gbs/similarity.py
trbromley: Oh 🤔 And it's not something we need to be e.g. 2 * photons?
@nquesada?

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley:
```suggestion
    sample = orbit + [0] * (modes - len(orbit))
```
@ixfoduap helped answer my questions in person 😁
And agreed, shall we remove the `list` part then?

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,161 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    12
+
+    Args:
+        orbit (list[int]): orbit to count number of samples
+        modes (int): number of modes in the sample
+
+    Returns:
+        int: number of samples in the orbit
+    """
+    sample = list(orbit) + [0] * (modes - len(orbit))
```
trbromley:
```suggestion
    sample = orbit + [0] * (modes - len(orbit))
```

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: > I think it's good to keep a clear distinction between total number of photons and maximum number of photons per mode

^ Agreed! Does this mean you agree with the suggestion or not? 🤔

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+        photon_number (int): number of photons in the event
+        max_count_per_mode (int): maximum number of photons per mode in the event
+        modes (int): number of modes in the samples
+
+    Returns:
+        int: number of samples in the event
+    """
+    cardinality = 0
+
+    for orb in orbits(photon_number):
```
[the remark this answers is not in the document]
trbromley: Either way, we agreed to just iterate through the whole list for now?

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,157 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    For each sample, this function calculates the probability of observing that sample from a GBS
+    device programmed according to the input graph and mean photon number. The sum of the
+    probabilities is then rescaled according to the cardinality of the orbit and the total number of
+     samples. The estimate is the sample mean of the rescaled probabilities.
+
+    **Example usage**:
+
+    >>> graph = nx.complete_graph(8)
+    >>> p_orbit_mc(graph, [2, 1, 1])
+    0.03744
```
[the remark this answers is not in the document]
trbromley: Awesome!

E.g. could be like this:
Annotation 2019-09-18 082509
(found in https://strawberryfields.readthedocs.io/en/latest/tutorials/tutorial_post_selection.html)

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: Nice! I guess we'll have to update when we change `event_to_sample`?

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: Ahh sorry I made a bad suggestion, I meant `np.sum` -> `sum`, but I like your change even more 😁

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: Agreed after in person discussion to keep as is!

tests/gbs/test_similarity.py
trbromley: Yep, we discussed in person and agreed it's fine!

trbromley:
Thanks @ixfoduap I have read through and resolved your comments, should just be minor things for you to read through now.

Shall we make ready for review and open up to others now? Good to go from my side!

trbromley:
For reference for future reviewers - although I took out the initial PR, @ixfoduap is the assignee for this one!

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
ixfoduap: Agree with suggestion

strawberryfields/gbs/similarity.py
ixfoduap: No, the error raised said something like "cutoff must be larger than number of photons", so `photons+1` is enough

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,161 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
             available_modes.remove(j)
 
     return sample
+
+
+def orbit_cardinality(orbit: list, modes: int) -> int:
+    """Gives the number of samples belonging to the input orbit.
```
josh146: This confused me momentarily, until I worked out what it meant. Would be worth adding a short hand-written example, which is then verified via the code example below. Would also be nice to say 'number of unique permutations', maybe?

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,161 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    Returns:
+        int: number of samples in the orbit
+    """
+    sample = orbit + [0] * (modes - len(orbit))
+    counts = list(Counter(sample).values())
+    return int(factorial(modes) / np.prod(factorial(counts)))
+
+
+def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -> int:
+    """Gives the number of samples belonging to the input event.
```
josh146: probably same comment as above, but now that I understand the above one, I understand this one :laughing:

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
josh146: @nquesada will be able to answer this one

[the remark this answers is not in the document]
josh146: Note that this approach (`state.fock_prob`) only works for local simulators, since hardware will return samples. We will be working on standardizing this interface for hardware devices as well, but it may change.

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,161 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    cardinality = 0
+
+    for orb in orbits(photon_number):
+        if max(orb) <= max_count_per_mode:
+            cardinality += orbit_cardinality(orb, modes)
+
+    return cardinality
+
+
+def p_orbit_mc(graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int = 1000) -> float:
```
josh146: Instead of `p_orbit_mc`, how about `prob_orbit_mc`? Makes it a little easier to tell what the function does

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,161 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    for _ in range(samples):
+        sample = orbit_to_sample(orbit, modes)
+        prob += result.state.fock_prob(sample, cutoff=photons + 1)
+
+    prob = prob * orbit_cardinality(orbit, modes) / samples
+
+    return prob
+
+
+def p_event_mc(
```
josh146: Same comment here, `prob_event_mc`?

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,94 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        ((1, 1, 1, 1), 5): 5,
+        ((1, 1, 2), 5): 30,
+        ((1, 2, 3), 5): 60,
+    }
+
+    calc_cardinalities = {}
+    for o, _ in orbits.items():
+        calc_cardinalities[o] = similarity.orbit_cardinality(list(o[0]), o[1])
+
+    assert calc_cardinalities == orbits
```
josh146: This is a minor design comment, but you could parametrize this test like so:
```python
@pytest.mark.parametrize("orbit, max_photon, expected", orbits.items())
def test_orbit_cardinality(orbit, max_photon, expected):
```
with very minimal changes. This would just lead to slightly nicer test output --- the failing case would be much more easily identified.

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,94 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        is equal to 1/5, i.e., one over the number of samples in the orbit [1,1,1,1] for 5 modes."""
+        graph = nx.complete_graph(5)
+        with monkeypatch.context() as m:
+            m.setattr(
+                "strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                lambda *args, **kwargs: 0.2,
+            )
+            tol = 1e-5
+
+            assert np.abs(similarity.p_orbit_mc(graph, [1, 1, 1, 1]) - 1.0) < tol
```
josh146: If the tolerance is due to numerical floating point error (and not stochastic error), then it is better to use something like
```suggestion
            assert np.allclose(similarity.p_orbit_mc(graph, [1, 1, 1, 1]), 1.0)
```
which by default is set up to take into account FP precision error. this avoids magic numbers entering the code.

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,94 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        6 photons, and max 3 photons per mode."""
+        graph = nx.complete_graph(6)
+        with monkeypatch.context() as m:
+            m.setattr(
+                "strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                lambda *args, **kwargs: 1.0 / 336,
+            )
+            tol = 1e-5
+
+            assert np.abs(similarity.p_event_mc(graph, 6, 3) - 1.0) < tol
```
josh146: same comment here re: magic tolerances

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,94 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+class TestPEventMC:
+    """Tests for the function ``strawberryfields.gbs.similarity.p_event_mc.``"""
+
+    def test_prob_vacuum_event(self):
+        """Tests if the function gives the right probability for an event with zero photons when
+        the GBS device has been configured to have zero mean photon number."""
+        graph = nx.complete_graph(10)
+
+        assert similarity.p_event_mc(graph, 0, 0, 0) == 1.0
```
josh146: nice test!

strawberryfields/gbs/g_sample.py
```diff
@@ -132,18 +131,7 @@ def to_subgraphs(graph: nx.Graph, samples: list) -> list:
...
-        Args:
-           s (list): a binary sample of ``len(nodes)``
-
-        Returns:
-            list: a subgraph specified by its nodes
-        """
-        return list(np.nonzero(s)[0])
-
-    subgraph_samples = [to_subgraph(s) for s in samples]
+    subgraph_samples = [sample.modes_from_counts(s) for s in samples]
```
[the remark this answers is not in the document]
trbromley: `to_subgraph` was an internal function of `to_subgraphs` which could take in one sample and return a subgraph specified by its nodes. E.g.: [1, 0, 1, 0, 1, 1] -> [0, 2, 4, 5]

In parallel we needed something in the `similarity` module that could take in one sample and return a list of modes where photons were counted. E.g., [1, 0, 2, 0, 3, 1] -> [0, 2, 2, 4, 4, 4, 5]

Since the functionality in the `similarity` module also encompassed `to_subgraph`, we decided to make this a single function sitting in the `sample` module (`sample.modes_from_counts`). This means that `to_subgraph` could be removed and replaced with the (tested and documented) `modes_from_counts`.

tests/gbs/test_similarity.py
ixfoduap: No magic :sob:

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: It's ok that it only works for simulators, since we have another function which accepts samples and will be able to work with samples from hardware.

We haven't thought about what things will look like for remote simulators for now.

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+@pytest.mark.parametrize("photons, max_count, modes, expected", events)
+def test_event_cardinality(photons, max_count, modes, expected):
+    """Test if function ``strawberryfields.gbs.similarity.event_cardinality`` returns the
+    correct number of samples for some hard-coded examples."""
+
+    assert similarity.event_cardinality(photons, max_count, modes) == expected
+
+
+class TestPOrbitMC:
```
trbromley:
```suggestion
class TestProbOrbitMC:
```
🤣

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,175 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    device programmed according to the input graph and mean photon number. The sum of the
+    probabilities is then rescaled according to the cardinality of the orbit and the total number of
+    samples. The estimate is the sample mean of the rescaled probabilities. To make
+    this estimate, several samples from the orbit are drawn uniformly at random using
+    :func:orbit_to_sample.
+
+    **Example usage**:
+
+    >>> graph = nx.complete_graph(8)
+    >>> p_orbit_mc(graph, [2, 1, 1])
```
trbromley:
```suggestion
    >>> prob_orbit_mc(graph, [2, 1, 1])
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+    """Tests for the function ``strawberryfields.gbs.similarity.p_orbit_mc.``"""
```
trbromley:
```suggestion
    """Tests for the function ``strawberryfields.gbs.similarity.prob_orbit_mc.``"""
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        monkeypatches the fock_prob function so that the probability is the same for each sample and
+        is equal to 1/5, i.e., one over the number of samples in the orbit [1,1,1,1] for 5 modes."""
+        graph = nx.complete_graph(5)
+        with monkeypatch.context() as m:
+            m.setattr(
+                "strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                lambda *args, **kwargs: 0.2,
+            )
+
+            assert np.allclose(similarity.p_orbit_mc(graph, [1, 1, 1, 1]) - 1.0)
```
trbromley:
```suggestion
            assert np.allclose(similarity.prob_orbit_mc(graph, [1, 1, 1, 1]) - 1.0)
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+    def test_prob_vacuum_orbit(self):
+        """Tests if the function gives the right probability for the empty orbit when the GBS
+        device has been configured to have zero mean photon number."""
+        graph = nx.complete_graph(10)
+
+        assert similarity.p_orbit_mc(graph, [], 0) == 1.0
```
trbromley:
```suggestion
        assert similarity.prob_orbit_mc(graph, [], 0) == 1.0
```

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,175 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    sample, we calculate the probability of observing that sample from a GBS programmed according to
+    the input graph and mean photon number. These probabilities are then rescaled according to the
+    cardinality of the event. The estimate is the sample mean of the rescaled probabilities. To make
+    this estimate, several samples from the event are drawn uniformly at random using
+    :func:event_to_sample.
+
+    **Example usage**:
+
+    >>> graph = nx.complete_graph(8)
+    >>> p_event_mc(graph, 4, 2)
```
trbromley:
```suggestion
    >>> prob_event_mc(graph, 4, 2)
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+
+class TestPEventMC:
+    """Tests for the function ``strawberryfields.gbs.similarity.p_event_mc.``"""
```
trbromley:
```suggestion
    """Tests for the function ``strawberryfields.gbs.similarity.prob_event_mc.``"""
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+
+    def test_prob_vacuum_orbit(self):
+        """Tests if the function gives the right probability for the empty orbit when the GBS
+        device has been configured to have zero mean photon number."""
+        graph = nx.complete_graph(10)
+
+        assert similarity.p_orbit_mc(graph, [], 0) == 1.0
+
+
+class TestPEventMC:
```
trbromley:
```suggestion
class TestProbEventMC:
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+    """Tests for the function ``strawberryfields.gbs.similarity.p_event_mc.``"""
+
+    def test_prob_vacuum_event(self):
+        """Tests if the function gives the right probability for an event with zero photons when
+        the GBS device has been configured to have zero mean photon number."""
+        graph = nx.complete_graph(10)
+
+        assert similarity.p_event_mc(graph, 0, 0, 0) == 1.0
```
trbromley:
```suggestion
        assert similarity.prob_event_mc(graph, 0, 0, 0) == 1.0
```

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        is equal to 1/216, i.e., one over the number of samples in the event with 5 modes,
+        6 photons, and max 3 photons per mode."""
+        graph = nx.complete_graph(6)
+        with monkeypatch.context() as m:
+            m.setattr(
+                "strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                lambda *args, **kwargs: 1.0 / 336,
+            )
+
+            assert np.allclose(similarity.p_event_mc(graph, 6, 3) - 1.0)
```
trbromley:
```suggestion
            assert np.allclose(similarity.prob_event_mc(graph, 6, 3) - 1.0)
```

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,175 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+    for _ in range(samples):
+        sample = event_to_sample(photon_number, max_count_per_mode, modes)
+        prob += result.state.fock_prob(sample, cutoff=photon_number + 1)
+
+    prob = prob * event_cardinality(photon_number, max_count_per_mode, modes) / samples
+
+    return prob
+
+print(event_cardinality(2, 2, 3))
```
trbromley: Need to remove

tests/gbs/test_similarity.py
trbromley: Is this the correct use of `allclose`? https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html
Same for line 286

josh146: No, should be
```suggestion
            assert np.allclose(similarity.p_orbit_mc(graph, [1, 1, 1, 1]), 1.0)
```

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +210,173 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+
+def orbit_cardinality(orbit: list, modes: int) -> int:
+    """Gives the number of samples belonging to the input orbit.
+
+    An orbit has a number of constituting samples, which are given by taking all permutations
+    over the orbit. This function counts how many samples of the given number of modes there are
+    in a given orbit. For example, there are three possible samples in the orbit [2,1,
+    1] with three modes: [1,1,2], [1,2,1], and [2,1,1]. With four modes, there are 12 samples in
+    total.
```
josh146: :+1:

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+]
+
+
+@pytest.mark.parametrize("orbit, max_photon, expected", orbits)
+def test_orbit_cardinality(orbit, max_photon, expected):
+    """Test if function ``strawberryfields.gbs.similarity.orbit_cardinality`` returns the
+    correct number of samples for some hard-coded examples."""
+    print('orbit = ', orbit)
+    print('max_photon = ', max_photon)
+    print('expected = ', expected)
```
josh146: print statements need to be removed

tests/gbs/test_similarity.py
```diff
@@ -194,3 +196,91 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        is equal to 1/216, i.e., one over the number of samples in the event with 5 modes,
+        6 photons, and max 3 photons per mode."""
+        graph = nx.complete_graph(6)
+        with monkeypatch.context() as m:
+            m.setattr(
+                "strawberryfields.backends.gaussianbackend.GaussianState.fock_prob",
+                lambda *args, **kwargs: 1.0 / 336,
+            )
+
+            assert np.allclose(similarity.prob_event_mc(graph, 6, 3) - 1.0)
```
josh146:
```suggestion
            assert np.allclose(similarity.prob_event_mc(graph, 6, 3), 1.0)
```

tests/gbs/test_similarity.py
trbromley: Done!

tests/gbs/test_similarity.py
trbromley: Yep this has been done too. thanks Josh!

PR #187, opened 2019-09-16
Improving Apps Structure: Part 2 - Consolidate graph density based functions into dense module

trbromley:
This PR carries out the second part of the plan to improve the import structure of the applications layer.

Remaining steps are:

- Consolidate sampling functions into a single module (e.g., currently in sample and g_sample)
- Consolidate some top-level functions into another module and also update the `__init__.py` in each case


trbromley:
I'd like to suggest that we rename the module from `dense` to `subgraph`.
This matches better with the focus of the module. For example:

```python
from strawberryfields import gbs

gbs.subgraph.search(...)
```

It also means we can move `is_subgraph` and `subgraph_adjacency` utility functions into this module.

What do people think?

trbromley:
> I'd like to suggest that we rename the module from `dense` to `subgraph`.
> This matches better with the focus of the module. For example:
> 
> ```python
> from strawberryfields import gbs
> 
> gbs.subgraph.search(...)
> ```
> 
> It also means we can move `is_subgraph` and `subgraph_adjacency` utility functions into this module.
> 
> What do people think?

Ok, let's leave this out of this PR for now and revisit in https://github.com/XanaduAI/strawberryfields/pull/190 😁

PR #188, opened 2019-09-16
Apps layer: Provide feature_vector_sampling

trbromley:
The purpose of this PR is to provide functionality to reconstruct a feature vector of a graph with respect to samples taken from GBS of that graph.


tests/gbs/test_similarity.py
```diff
@@ -194,3 +194,46 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+            (1, 1, 1, 0, 0): 3,
+            (1, 1, 1, 1, 0): 4,
+            (1, 1, 1, 1, 1): 5,
+            (2, 0, 0, 0, 0): None,
+            (3, 0, 0, 0, 0): None,
+            (4, 0, 0, 0, 0): None,
+            (5, 0, 0, 0, 0): None,
+            (0, 1, 1, 0, 0): 2,
+        }
+        samples = list(samples_events_mapping.keys()) + [(1, 1, 1, 1, 1)]  # add a repetition
```
ixfoduap: Does this serve a specific purpose? Or just having fun with dictionaries?

ixfoduap: Oh, it's to have 10 samples?

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,42 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
             available_modes.remove(j)
 
     return sample
+
+
+def feature_vector_sampling(
+    samples: list, event_photon_numbers: list, max_count_per_mode: int = 2
+) -> np.ndarray:
```
ixfoduap: I'm still not convinced we want a numpy array as output. Josh mentioned previously it would make normalizing vectors easier, but I'm not sure that is actually relevant. Certainly the whole GBS layer has previously operated with lists, with this function being the only exception. For example, already during the tests below we need to convert back to list

tests/gbs/test_similarity.py
```diff
@@ -194,3 +194,46 @@ def test_sample_max_count_deterministic(self, photon_num, count, monkeypatch):
...
+        with pytest.raises(ValueError, match="Cannot request events with photon number below zero"):
+            similarity.feature_vector_sampling([[1, 1, 0], [1, 0, 1]], [-1, 4], 1)
+
+    def test_bad_max_count(self):
+        """Test if function raises a ``ValueError`` when input a non-positive value for the
+        maximum photon count per mode."""
+        with pytest.raises(ValueError, match="Maximum number of photons per mode must be at least"):
+            similarity.feature_vector_sampling([[1, 1, 0], [1, 0, 1]], [2, 4], 0)
+
+    def test_correct_distribution(self, monkeypatch):
```
ixfoduap: Super nice test!

tests/gbs/test_similarity.py
trbromley: Yes, to have 10 samples overall with 1 repetition - the dictionary lets us easily get the event from the sample (used for monkeypatching in line 236), but dictionaries don't support repeated elements so I thought to make a list from the keys and add on a repetition.

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,42 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    Returns:
+        array: a feature vector of event probabilities in the same order as ``event_photon_numbers``
+    """
+    if min(event_photon_numbers) < 0:
+        raise ValueError("Cannot request events with photon number below zero")
+
+    if max_count_per_mode < 1:
+        raise ValueError("Maximum number of photons per mode must be at least one")
+
+    n_samples, _ = np.array(samples).shape
```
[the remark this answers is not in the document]
trbromley: I like it! 😁

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,42 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+    ``event_photon_numbers``. Probabilities are reconstructed by measuring the occurrence of
+    events in the input ``samples``.
+
+    **Example usage**:
+
+    >>> feature_vector_sampling(samples, [2, 4, 6], 1)
+    (0.34, 0.08, 0.01)
+
+    Args:
+        samples (list[list[int]]): a list of samples generated from :func:`~.quantum_sampler`
```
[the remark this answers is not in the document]
trbromley: Yep good point, have changed to not specifying

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,42 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+    The feature vector is composed of event probabilities, with all events having a maximum
+    photon count in each mode of ``max_count_per_mode``. Events are then described by their total
+    photon number, and those chosen as part of the feature vector can be specified through
+    ``event_photon_numbers``. Probabilities are reconstructed by measuring the occurrence of
+    events in the input ``samples``.
+
+    **Example usage**:
+
+    >>> feature_vector_sampling(samples, [2, 4, 6], 1)
```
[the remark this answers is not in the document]
trbromley: Yes :(
Was thinking we can do this really nicely if we get some pre-made samples.
But for now not sure what's best, felt like leaving it as made up was ok since `samples` is not visible.

strawberryfields/gbs/similarity.py
trbromley: I'm ok to change the output to be a list. Was thinking about what impact this would have if we do a `gram_matrix` function and I think it should be fine. We could do something like
```python
vecs = [feature_vector_sampling(samples, [2, 4, 6, 8]) for samples in composite_samples]
gram = gram_matrix(vecs)
```

trbromley: As a side note, I assumed we were going with unnormalized feature vectors - is this the advice we got?

trbromley: @josh146 - on board to switch to list instead of numpy array?

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: Have manually added this in

strawberryfields/gbs/similarity.py
josh146: Yep, happy to switch back to a list. I admit it's slight personal bias --- I tend to prefer NumPy arrays being returned!

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
josh146: is this still WIP? In case it's not, ``samples`` should be defined in the example usage. Ideally, the example usage should be able to be copy and pasted by a user, and assuming they have their imports correct, everything should work.

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,42 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+def feature_vector_sampling(
+    samples: list, event_photon_numbers: list, max_count_per_mode: int = 2
+) -> np.ndarray:
+    """Calculates feature vector with respect to input samples.
+
+    The feature vector is composed of event probabilities, with all events having a maximum
+    photon count in each mode of ``max_count_per_mode``. Events are then described by their total
+    photon number, and those chosen as part of the feature vector can be specified through
+    ``event_photon_numbers``. Probabilities are reconstructed by measuring the occurrence of
+    events in the input ``samples``.
```
josh146: Is it worth adding an equation here?

josh146: Might be easier to parse than sentences :thinking:

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,42 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+    **Example usage**:
+
+    >>> feature_vector_sampling(samples, [2, 4, 6], 1)
+    (0.34, 0.08, 0.01)
+
+    Args:
+        samples (list[list[int]]): a list of samples
+        event_photon_numbers (list[int]): a list of events described by their total photon number
+        max_count_per_mode (int): maximum number of photons per mode in every event
```
josh146: Does this read better?
```suggestion
        max_count_per_mode (int): maximum number of photons per mode for all events
```

strawberryfields/gbs/similarity.py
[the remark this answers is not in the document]
trbromley: Good points, I've now improved things and the code is now executable (it's a bit slower though, maybe 30 seconds to run)

strawberryfields/gbs/similarity.py
trbromley: I added equations 😁

trbromley:
Should be good for a second round! 🎈

strawberryfields/gbs/similarity.py
```diff
@@ -201,3 +203,50 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+
+    Probabilities are reconstructed by measuring the occurrence of events in the input ``samples``.
+
+    **Example usage**:
+
+    >>> sample.random_seed(1967)
+    >>> adj = np.ones((4, 4))
+    >>> samples = sample.quantum_sampler(adj, 6, 10, backend_options={"threshold": False})
+    >>> feature_vector_sampling(samples, [2, 4, 6])
+    [0.1, 0.2, 0.0]
```
josh146: :+1:

PR #190, opened 2019-09-17
Improving Apps Structure: Part 4

trbromley:
This PR carries out the fourth part of the plan to improve the import structure of the applications layer.

**Other parts:**

- https://github.com/XanaduAI/strawberryfields/pull/182 (1)
- https://github.com/XanaduAI/strawberryfields/pull/187 (2)
- https://github.com/XanaduAI/strawberryfields/pull/189 (3)

**Remaining steps are:**

Everything should be done. @co9olguy you mentioned one final thing we could do is import some frequently used functions at top level. Right now I'm not sure which functions we would do this for, so I was thinking to see how people find the library and choose based on their feedback. Does this work for you?

**Changes in this PR:**

- moving utility functions from `utils` module to `clique` and `dense` modules
- remove `graph_type` and functionality for accepting graphs of numpy and networkx input types, after our decision to go with just networkx
- rename `dense` to `subgraph`
- remove remaining functions in `utils` module as they were not needed


trbromley:
Would like to get people's thoughts: the `utils` module is getting pretty empty and I think could be gotten rid of (see https://github.com/XanaduAI/strawberryfields/blob/apps_to_gbs_4/strawberryfields/gbs/utils.py).

There are three functions remaining there:
- `is_undirected` (used once in `quantum_sampler`)
- `subgraph_adjacency` (not used)
- `is_subgraph` (used a few times in `dense` and `clique`)

I propose to:
- **remove** `is_undirected` and replace with the one-liner `np.allclose(mat, mat.T)` when used in `quantum_sampler` (the function also checked that an input was correctly shaped as a numpy array, but this is maybe overkill)
- **remove** `subgraph_adjacency` since the user can anyway do the one-liner `nx.to_numpy_array(g.subgraph(s))` where required - this could be shown in tutorials if needed.

The remaining question is what to do with `is_subgraph` - my preference is to move it to `dense` and rename that module to `subgraph`.

Thoughts @ixfoduap @josh146?

strawberryfields/gbs/sample.py
```diff
@@ -111,7 +110,7 @@ def quantum_sampler(
     """
     backend_options = {**BACKEND_DEFAULTS, **(backend_options or {})}
 
-    if not utils.is_undirected(A):
+    if not np.allclose(A, A.T):
```
josh146: nice!

strawberryfields/gbs/subgraph.py
```diff
@@ -76,13 +82,12 @@
 
 import networkx as nx
 
-from strawberryfields.gbs import g_sample, utils
+from strawberryfields.gbs import g_sample
```
josh146: I might be getting confused, but has `g_sample` not been merged as of this PR?

strawberryfields/gbs/subgraph.py
```diff
@@ -76,13 +82,12 @@
...
 from strawberryfields.gbs.sample import BACKEND_DEFAULTS
-from strawberryfields.gbs.utils import graph_type
 
 
 def search(
-    graph: graph_type, nodes: int, iterations: int = 1, options: Optional[dict] = None
+    graph: nx.Graph, nodes: int, iterations: int = 1, options: Optional[dict] = None
```
josh146: Much cleaner!

strawberryfields/gbs/subgraph.py
```diff
@@ -386,3 +389,21 @@ def greedy_degree(
...
+        subgraph (iterable): a collection of nodes
+        graph (nx.Graph): the input graph
+
+    Returns:
+        bool: returns ``True`` only if input subgraph is valid
+    """
+    try:
+        return set(subgraph).issubset(graph.nodes)
+    except TypeError:
+        raise TypeError("subgraph and graph.nodes must be iterable")
```
josh146: If `graph` is not an `nx.Graph` object, then it will raise an obscure sounding `AttributeError`, and it won't be caught by the try-except. Also, could the `TypeError` be raised by an edge case even when both subgraph and graph.nodes are iterable?

The more I think about it, does this need a try-except at all? Why try and catch this exception, but not any others?

strawberryfields/gbs/subgraph.py
trbromley: Nope it's coming out in the semi-orthogonal PR https://github.com/XanaduAI/strawberryfields/pull/189

The merging isn't so bad for these PR's, there are conflicts but it mainly involves sorting out the imports. How about I only ask you to look at this once 189 gets merged in?

strawberryfields/gbs/subgraph.py
trbromley: Hmmm 🤔 

I'm torn between:
- just directly returning `set(subgraph).issubset(graph.nodes)` in `is_subgraph`
- OR, just delete `is_subgraph` and place the resultant one-liner in the code.

Tempted with the latter, but I'm worried about getting too crazy with deleting functions. E.g., could `is_subgraph` be useful for the user?

josh146: I would go with the latter as well, unless you have a user story for 'user can determine if a graph contains a specific subgraph'.

I guess it comes down to, is `set(subgraph).issubset(graph.nodes)` simple enough for new users to be able to work out on their own, or not.

josh146: One of those things where it's easier to make a decision once you have user feedback.

trbromley: Have deleted it, thanks Josh!

PR #191, opened 2019-09-18
Circuit drawer fix

nquesada:
**Context:**
Adds a minor tweak in the circuit drawer so that the pdfs generated do not have page numbering. That way they are easier to crop.

Adds two lines to circuit drawer.
**Benefits:**
The pdf generated from circuit drawer can be easily cropped  with eg pdfcrop.


josh146:
Looks good to be merged in when the tests pass, thanks @nquesada

josh146:
@nquesada, looks like the tests will also need to be updated to take into account this change!

co9olguy:
I think line 47 [here](https://github.com/XanaduAI/strawberryfields/blob/master/tests/frontend/test_circuitdrawer.py) will have to be updated

nquesada:
It never crossed my mind that there were tests for this... Anyway they pass in my local machine and should hopefully pass on travis...

PR #194, opened 2019-09-19
Apps Layer: Fix event_to_sample

ixfoduap:
The previous implementation of the `event_to_sample` function was creating correct samples belonging to the event, but these were not distributed uniformly at random. This PR fixes this issue so that samples are generated uniformly at random.


josh146:
This could be a sign that the GBS module requires more integration tests to test the library from the user point of view.

strawberryfields/gbs/similarity.py
```diff
@@ -270,7 +273,9 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
     return cardinality
 
 
-def prob_orbit_mc(graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int = 1000) -> float:
+def prob_orbit_mc(
+    graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int = 1000
```
trbromley: 🤔 was this due to a run of `black`? Did we have the right line length before or after?

josh146: I asked the same question over in #190 :laughing:

ixfoduap:
Upon closer look, the test `test_sample_max_count_deterministic` is no longer relevant in the new implementation of `event_to_sample`. This test was checking whether, in the previous method, no more photons were added to a mode that already had a `max_count_per_mode` number of photons. In the new method, we don't populate modes with photons, we instead sample orbits according to their size, then call `orbit_to_sample`. 

I have therefore deleted the test. I could not think of any other test to add, but I'm open to suggestions :)

trbromley:
Actually, one thing I was thinking about:
`event_to_sample` might be a fairly heavy computation just to get one sample, since we have to iterate through all orbits each time. I think it makes more sense for this function to be `event_to_samples`, so that it generates many samples. However this can be a subsequent PR.

Issue #192, opened 2019-09-19
rectangular_symmetric docstring missing last return value


lneuhaus:
#### Issue description

The docstring of `strawberryfields.decompositions.rectangular_symmetric` does not list the last return value `None` that is returned by the function: https://github.com/XanaduAI/strawberryfields/blob/65a517ebc7eac23d51364e5f3c096d1a918a43a0/strawberryfields/decompositions.py#L454

* *Expected behavior:* (What you expect to happen)
Either list the last return value in the docstring, or change the return signature.

* *Actual behavior:* (What actually happens)
Last return value is not listed in the docstring, but returned, leading to confusion.

* *Reproduces how often:* (What percentage of the time does it reproduce?)
Anytime the return value of the function is fed into a tuple.

Issue #193, opened 2019-09-19
rectangular_symmetric yields unexpected output


lneuhaus:
* *Expected behavior:* (What you expect to happen)
`sf.decompositions.rectangular_symmetric(np.identity(4)[::-1])`
should give a decomposition with all internal and external phases equal to zero, i.e.
```
([[0, 1, 0.0, 0.0, 4],
  [2, 3, 0.0, 0.0, 4],
  [1, 2, 0.0, 0.0, 4],
  [0, 1, 0.0, 0.0, 4],
  [2, 3, 0.0, 0.0, 4],
  [1, 2, 0.0, 0.0, 4]],
 array([0, 0, 0, 0]),
 None)
```
* *Actual behavior:* (What actually happens)
I get strange values, and negative phases.
```
([[0, 1, 0.0, 0.0, 4],
  [2, 3, 0.0, 0.0, 4],
  [1, 2, 0.0, -3.141592653589793, 4],
  [0, 1, 0.0, -4.71238898038469, 4],
  [2, 3, 0.0, 4.71238898038469, 4],
  [1, 2, 0.0, 4.71238898038469, 4]],
 array([1.8369702e-16+1.j, 3.0616170e-16+1.j, 6.1232340e-17+1.j,
        1.0000000e+00+0.j]),
 None)
```
The negative phases should be transformed into positive ones in the range `[0, 2*pi)`, and the values should be all zeros.
 
* *Reproduces how often:* (What percentage of the time does it reproduce?)
For any unitary above dimension 4 that I tried.

* *System information:* (post the output of `import stawberryfields as sf; sf.about()`)
```
Strawberry Fields: a Python library for continuous-variable quantum circuits.
Copyright 2018-2019 Xanadu Quantum Technologies Inc.

Python version:            3.6.8
Platform info:             Linux-4.15.0-55-generic-x86_64-with-debian-buster-sid
Installation path:         /home/leo/xanadu/code/strawberryfields/strawberryfields
Strawberry Fields version: 0.12.0-dev
Numpy version:             1.16.3
Scipy version:             1.2.1
The Walrus version:           0.7.0
Blackbird version:         0.2.1
TensorFlow version:        1.7.0
```

lneuhaus:
I suspect the `numpy.fmod` in the implementation should simply be replaced by the regular modulus operator, and relevant tests should be added.

josh146:
[...]

PR #196, opened 2019-09-20
Fix issue #192 and part of #193 (decompositions.rectangular_*)

lneuhaus:
**Context:**
The sign of the phases returned by `strawberryfields.decompositions.rectangular_symmetric` depended on algorithm details due to the incorrect use of `numpy.fmod` instead of the standard python modulo (`%`) operator. 

Fixes bug in `strawberryfields.decompositions.rectangular_symmetric` so its returned phases are all in the interval [0, 2*pi[, adds tests for this, and corrects a bug in the function docstring. 

**Benefits:**
The user can be sure that the returned phases are in the interval from 0 to 2 pi. 

**Possible Drawbacks:**
-

**Related GitHub Issues:**
#192 and part of #193

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ### New features
 
-* Fixed bug in `strawberryfields.decompositions.rectangular_symmetric` so its
-  returned phases are all in the interval [0, 2*pi), and corrects the
-  function docstring.
-  [#196](https://github.com/XanaduAI/strawberryfields/pull/196)
-
 * Adds in first version of an applications layer aimed at solving problems using
   Gaussian boson sampling. This layer focuses on graph-based problems and
   currently has algorithms for the densest ``k``-subgraph problem.
@@ -77,6 +72,11 @@
   [#160](https://github.com/XanaduAI/strawberryfields/pull/160)
 
 ### Bug fixes
+
+* Fixed bug in `strawberryfields.decompositions.rectangular_symmetric` so its
+  returned phases are all in the interval [0, 2*pi), and corrects the
+  function docstring.
+  [#196](https://github.com/XanaduAI/strawberryfields/pull/196)
 
 * When using the `'gbs'` compilation target, the measured registers are now sorted in
   ascending order in the resulting compiled program.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### New features
 
+* Fixes bug in `strawberryfields.decompositions.rectangular_symmetric` so its
+  returned phases are all in the interval [0, 2*pi[, and corrects the
+  function docstring.
+
 * Adds in first version of an applications layer aimed at solving problems using
   Gaussian boson sampling. This layer focuses on graph-based problems and
   currently has algorithms for the densest ``k``-subgraph problem.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ### New features
 
-* Fixes bug in `strawberryfields.decompositions.rectangular_symmetric` so its
-  returned phases are all in the interval [0, 2*pi[, and corrects the
+* Fixed bug in `strawberryfields.decompositions.rectangular_symmetric` so its
+  returned phases are all in the interval [0, 2*pi), and corrects the
   function docstring.
+  [#196](https://github.com/XanaduAI/strawberryfields/pull/196)
 
 * Adds in first version of an applications layer aimed at solving problems using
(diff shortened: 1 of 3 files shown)


strawberryfields/decompositions.py
```diff
@@ -450,22 +450,29 @@ def rectangular_symmetric(V, tol=1e-11):
...
     for i in tlist:
         em, en = int(i[0]), int(i[1])
         alpha, beta = np.angle(new_diags[em]), np.angle(new_diags[en])
         theta, phi = i[2], i[3]
-        external_phase = np.fmod((phi + alpha - beta), 2 * np.pi)
-        internal_phase = np.fmod((np.pi + 2.0 * theta), 2 * np.pi)
+        external_phase = (phi + alpha - beta) % (2 * np.pi)
+        internal_phase = (np.pi + 2.0 * theta) % (2 * np.pi)
+        # repeat modulo operations , otherwise the input unitary
+        # numpy.identity(20) yields an external_phase of exactly 2 * pi
```
lneuhaus: This problem seems to be well-known: see https://docs.python.org/3.7/reference/expressions.html#id17
    
> While abs(x%y) < abs(y) is true mathematically, for floats it may not be true numerically due to roundoff. For example, and assuming a platform on which a Python float is an IEEE 754 double-precision number, in order that -1e-100 % 1e100 have the same sign as 1e100, the computed result is -1e-100 + 1e100, which is numerically exactly equal to 1e100. The function math.fmod() returns a result whose sign matches the sign of the first argument instead, and so returns -1e-100 in this case. Which approach is more appropriate depends on the application.

.github/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### New features
 
+* Fixes bug in `strawberryfields.decompositions.rectangular_symmetric` so its
+  returned phases are all in the interval [0, 2*pi[, and corrects the
```
josh146:
```suggestion
  returned phases are all in the interval `[0, 2*pi]`, and corrects the
```

.github/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  function docstring.
```
josh146:
```suggestion
  function docstring.
  [#196](https://github.com/XanaduAI/strawberryfields/pull/196)
```

.github/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### New features
 
+* Fixes bug in `strawberryfields.decompositions.rectangular_symmetric` so its
```
josh146:
```suggestion
* Fixed bug in `strawberryfields.decompositions.rectangular_symmetric` so its
```

.github/CHANGELOG.md
josh146: Also, could you please move this down into the 'bug fixes' part of the changelog. Thanks!

strawberryfields/decompositions.py
josh146: Good catch!

strawberryfields/decompositions.py
```diff
@@ -450,22 +450,29 @@ def rectangular_symmetric(V, tol=1e-11):
         tuple[array]: returns a tuple of the form ``(tlist,np.diag(localV), None)``
             where:
 
-            * ``tlist``: list containing ``[n,m,internal_phase,external_phase,n_size]`` of the T unitaries needed
+            * ``tlist``: list containing ``[n, m, internal_phase, external_phase, n_size]`` of the T unitaries needed
             * ``localV``: Diagonal unitary matrix to be applied at the end of circuit
+            * ``None``: the value ``None``, in order to make the return
+              signature identical to the one of :func:`rectangular`.
```
```suggestion
              signature identical to :func:`rectangular`
```

tests/frontend/test_decompositions.py
```diff
@@ -308,6 +308,8 @@ def test_decomposition(self, U, tol):
         tlist, diags, _ = dec.rectangular_symmetric(U)
         qrec = np.identity(nmax)
         for i in tlist:
+            assert i[2] >= 0 and i[2] < 2 * np.pi  # internal phase
+            assert i[3] >= 0 and i[3] < 2 * np.pi  # external phase
```
josh146: :+1:

.github/CHANGELOG.md
lneuhaus: Here I meant to say that the interval is open at the 2pi-end.. is there an unambiguous way to write that?

josh146: Oh right! So,
```suggestion
  returned phases are all in the interval [0, 2*pi), and corrects the
```

lneuhaus:
I believe I addressed all requested changes, so this should be good to merge from my point of view.

PR #198, opened 2019-09-25
Add GBS data interface and an initial dataset

trbromley:
The purpose of this PR is to add the ability for users to easily access pre-generated datasets of samples from GBS generated on some interesting graphs. We also provide a starting dataset.

Subsequent PRs will focus on adding more datasets.


strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+This module provides access to pre-calculated datasets of simulated GBS samples generated from a
+range of input graphs. Each dataset corresponds to a set of samples generated from a fixed graph
+with corresponding values for:
+
+- ``n_mean``: mean number of photons in the GBS device
+- ``n_max``: maximum number of photons allowed in any sample
+-  ``threshold``: flag to indicate whether samples are generated with threshold detection or
+   with photon number resolving detectors.
+- ``n_samples``: total number of samples in dataset
+- ``modes``: number of modes in GBS device or, equivalently, number of nodes in graph
```
josh146: Slight nitpick, but I think this would be easier to read with line breaks between bullet points.

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+We use the :class:`Planted` class as an example to show how to interact with the datasets. Datasets
+can be loaded by running:
+
+>>> data = Planted()
+
+Accessing samples from the dataset is then simple:
+
+>>> i = 3
+>>> sample_i = data[i]
+>>> samples = data[:10]
```
josh146: I feel the first two lines in this example is a bit confusing, why not just
```rest
>>> sample_3 = data[3]
>>> samples = data[:10]
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+
+Loading data
+^^^^^^^^^^^^
+
+We use the :class:`Planted` class as an example to show how to interact with the datasets. Datasets
+can be loaded by running:
+
+>>> data = Planted()
+
+Accessing samples from the dataset is then simple:
```
josh146:
```suggestion
Simply use indexing and slicing to access samples from the dataset:
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+>>> sample_i = data[i]
+>>> samples = data[:10]
+
+Datasets also contain metadata relevant to the GBS setup:
+
+>>> data.n_mean
+6
+
+>>> data.n_samples
+50000
```
josh146: You may as well define `__len__()`, that way you can do
```rest
>>> len(data)
50000
```
:smile:

josh146: (since it is being treated as a Python sequence anyway with indexing, using `len()` feels very natural)

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+Counting photons or clicks in each sample is available using the :meth:`Dataset.counts` method:
+
+>>> data.counts()
+[2, 0, 8, 11, ... , 6]
+
+Code details
+^^^^^^^^^^^^
+
+The :class:`Dataset` class provides the base functionality from which all datasets such as
+:class:`Planted` inherit.
```
josh146: no (auto)summary of available datasets?

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+
+DATA_PATH = pkg_resources.resource_filename("strawberryfields", "gbs/data") + "/"
+
+
+class Dataset(metaclass=ABCMeta):
+    """Base class for loading datasets of pre-generated samples.
+
+    Attributes:
+        n_mean (float): mean number of photons in the GBS device
+        n_max (float): maximum number of photons allowed in any sample. This number is set to
```
josh146:
```suggestion
        n_max (float): Maximum number of photons allowed in any sample. This number is set to
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+            limit the computation time, any sample being simulated that exceeds ``n_max`` will be
+            ignored and cause calculation to skip to the next sample.
+        threshold (bool): flag to indicate whether samples are generated with threshold detection
+            (i.e., detectors of zero or some photons) or with photon number resolving detectors.
+        adj (array): adjacency matrix of graph from which samples were generated
+        n_samples (int): total number of samples in dataset
+        modes (int): number of modes in GBS device or, equivalently, number of nodes in graph
```
josh146: Actually, I think the above can be deleted. It is already provided in both the module docstring _and_ the attribute documentation below, causing it to appear three times in the documentation. It is also at risk of being updated/changed in one place but not another.

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    _count = 0
+
+    @property
+    @abstractmethod
+    def _data_filename(self) -> str:
+        """Name of files stored in ``./data/``. Samples should be provided as a
```
josh146:
```suggestion
        """Base name of files containing the sample data stored in the ``./data/`` directory.
        
        Samples should be provided as a
```

(I like the convention of separating the short summary of the docstring from the remainder of the docstring, I find it makes it easier to skim/parse).

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+        ``scipy.sparse.csr_matrix`` saved in ``.npz`` format and the corresponding adjacency
+        matrix should be provided as a ``.npy`` binary. For ``_data_filename = "example"``,
+        the corresponding samples should be stored as ``./data/example.npz`` and the adjacency
+        matrix as ``./data/example_A.npy``."""
+        pass
+
+    def __init__(self):
+        """Instantiating ``Dataset`` causes the ``scipy.sparse.csr_matrix`` of samples and
+        ``numpy.ndarray`` adjacency matrix to be loaded."""
```
josh146: `__init__` doesn't require a docstring in this instance.

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+class Planted(Dataset):
+    """Dataset of samples generated from GBS with an embedded 30-node graph containing a dense
+    10-node subgraph planted inside :cite:`arrazola2018using`.
+
+    Attributes:
+        n_mean = 6
+        n_max = 14
+        threshold = True
+        n_samples = 50000
+        modes = 30
```
josh146: I feel this should be a bullet point list with backticks

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+        if isinstance(key, (slice, tuple)):
+            if isinstance(key, tuple):
+                key = slice(*key)
+            range_tuple = key.indices(self.n_samples)
+            return [self._elem(i) for i in range(*range_tuple)]
+
+        if key < 0:
+            key += self.n_samples
+
+        return self._elem(key)
```
josh146: Don't forget to test all these dunder methods thoroughly :laughing: Looks like there could be a lot of edge cases!

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+   with photon number resolving detectors.
+- ``n_samples``: total number of samples in dataset
+- ``modes``: number of modes in GBS device or, equivalently, number of nodes in graph
+
+Datasets are available as classes. The following graphs and datasets are provided:
+
+- **Planted graph**: A 30-node graph with a dense 10-node subgraph planted inside
+  :cite:`arrazola2018using`.
+
+  .. image:: ../../_static/planted.png
```
ixfoduap: Should probably use our graph visualization module to create this plot :grin:

strawberryfields/gbs/data.py
ixfoduap: ^ Agree

ixfoduap: Also, this is super cool!

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,199 @@
...
+
+>>> data.n_mean
+6
+
+>>> data.n_samples
+50000
+
+Counting photons or clicks in each sample is available using the :meth:`Dataset.counts` method:
+
+>>> data.counts()
```
ixfoduap: I don't understand this, what is the output of `counts`? Is it the aggregate of counts over all samples in the dataset? For instance, what does `11` refer to in the output of the example usage?

strawberryfields/gbs/data.py
ixfoduap: I don't understand dunders :sob: :sob: :sob:

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+This module provides access to pre-calculated datasets of simulated GBS samples generated from a
+range of input graphs. Each dataset corresponds to a set of samples generated from a fixed graph
+with corresponding values for:
+
+- ``n_mean``: mean number of photons in the GBS device
+
+- ``n_max``: maximum number of photons allowed in any sample
+
+-  ``threshold``: flag to indicate whether samples are generated with threshold detection or
+   with photon number resolving detectors.
```
ixfoduap:
```suggestion
   with photon-number-resolving detectors.
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+
+- ``n_samples``: total number of samples in dataset
```
ixfoduap:
```suggestion
- ``n_samples``: total number of samples in the dataset
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+
+- ``modes``: number of modes in GBS device or, equivalently, number of nodes in graph
```
ixfoduap:
```suggestion
- ``modes``: number of modes in the GBS device or, equivalently, number of nodes in the graph
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+6
+
+>>> len(data)
+50000
+
+The number of photons or clicks in each sample is available using the :meth:`Dataset.counts` method:
+
+>>> data.counts()
+[2, 0, 8, 11, ... , 6]
+
```
ixfoduap: Maybe add a sentence here saying something like: "For example, in this case the fourth sample has 11 photons"?

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+DATA_PATH = pkg_resources.resource_filename("strawberryfields", "gbs/data") + "/"
+
+
+class Dataset(metaclass=ABCMeta):
+    """Base class for loading datasets of pre-generated samples.
+
+    Attributes:
+        n_mean (float): mean number of photons in the GBS device
+        n_max (float): Maximum number of photons allowed in any sample. This number is set to
+            limit the computation time, any sample being simulated that exceeds ``n_max`` will be
```
ixfoduap:
```suggestion
            limit the computation time: any sample being simulated that exceeds ``n_max`` will be
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+            ignored and cause calculation to skip to the next sample.
```
ixfoduap:
```suggestion
            ignored and cause the calculation to skip to the next sample.
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+        threshold (bool): flag to indicate whether samples are generated with threshold detection
+            (i.e., detectors of zero or some photons) or with photon number resolving detectors.
```
ixfoduap:
```suggestion
            (i.e., detectors of zero or some photons) or with photon-number-resolving detectors.
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+        adj (array): adjacency matrix of graph from which samples were generated
```
ixfoduap:
```suggestion
        adj (array): adjacency matrix of the graph from which samples were generated
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+        n_samples (int): total number of samples in dataset
```
ixfoduap:
```suggestion
        n_samples (int): total number of samples in the dataset
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+        modes (int): number of modes in GBS device or, equivalently, number of nodes in graph
```
ixfoduap:
```suggestion
        modes (int): number of modes in the GBS device or, equivalently, number of nodes in graph
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def __init__(self):
+        self.data = scipy.sparse.load_npz(DATA_PATH + self._data_filename + ".npz")
+        self.adj = np.load(DATA_PATH + self._data_filename + "_A.npy")
+        self.n_samples, self.modes = self.data.shape
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
```
ixfoduap: For my learning process: Dunders don't need docstrings?

strawberryfields/gbs/data.py
ixfoduap: I guess Josh's comment is that it would look better with backticks? Would be nice to compare both if it's not too much trouble

tests/gbs/test_data.py
```diff
@@ -0,0 +1,131 @@
...
+        ``axis==1``. The samples are patched to the fixed ``patch_samples`` list."""
+        assert dataset_patched.counts(1) == [0, 1, 1, 2, 1, 2, 2, 3, 5, 5]
+
+    def test_iter(self, dataset_patched):
+        """Test if dataset class allows correct iteration over itself"""
+        assert [i for i in dataset_patched] == self.patch_samples
+
+    def test_slice(self, dataset_patched):
+        """Test if dataset class allows correct slicing over items"""
+        assert [i for i in dataset_patched[1, 4, 2]] == [
```
ixfoduap: Is it normal to spread `[self.patch_samples[1], self.patch_samples[3], ]` over so many lines?

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,209 @@
...
+GBS Datasets
+============
+
+**Module name:** :mod:`strawberryfields.gbs.data`
+
+.. currentmodule:: strawberryfields.gbs.data
+
+This module provides access to pre-calculated datasets of simulated GBS samples generated from a
+range of input graphs. Each dataset corresponds to a set of samples generated from a fixed graph
+with corresponding values for:
```
ixfoduap: Would it make sense to have a `description` attribute that gives more information about the dataset? For example, playing around with the `Planted` dataset I didn't find information regarding which nodes correspond to the planted subgraph. For other datasets we may want to give similar information, like location of max clique, or what class the MUTAG graphs belong to

strawberryfields/gbs/data.py
trbromley: Nice! Have added

strawberryfields/gbs/data.py
trbromley: image

^ Although maybe I need to put the backticks in some other configuration?

tests/gbs/test_data.py
trbromley: It was an automatic thing from `black` 🤔

strawberryfields/gbs/data.py
josh146: I think that's a good approach; it can then be shown via `help(Planted)`. Similarly to how it works with NumPy.

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,211 @@
...
+from abc import ABCMeta, abstractmethod
+
+import pkg_resources
+import numpy as np
+import scipy
+
+DATA_PATH = pkg_resources.resource_filename("strawberryfields", "gbs/data") + "/"
+
+
+class Dataset(metaclass=ABCMeta):
```
josh146: Note that while this was the recommended way of creating an abstract class in Python<3.4, since 3.4, the better approach has been to do
```python
from abc import ABC

class Dataset(ABC):
```

i.e., inheritance over defining the metaclass.

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,211 @@
...
+            <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html>`__.
+    """
+
+    _count = 0
+
+    @property
+    @abstractmethod
+    def _data_filename(self) -> str:
+        """Base name of files containing the sample data stored in the ``./data/`` directory.
+        Samples should be provided as a ``scipy.sparse.csr_matrix`` saved in ``.npz`` format and
```
josh146:
```suggestion

        Samples should be provided as a ``scipy.sparse.csr_matrix`` saved in ``.npz`` format and
```

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,211 @@
...
+        the corresponding adjacency matrix should be provided as a ``.npy`` binary. For
```
josh146: line break for the example?

strawberryfields/gbs/data.py
josh146: Ah, I was meaning the formatting in the actual code docstring, not the rendered page. But I didn't know you could write it like that and it would render! In which case I'm happy with it as is. Learnt something knew today 💪

strawberryfields/gbs/data.py
trbromley: Great!

I added:
"The 10-node planted clique is contained within the final 10 nodes of the graph." to the `Planted` docstring

strawberryfields/gbs/data.py
trbromley: Nice thanks Josh! Done.

strawberryfields/gbs/data.py
trbromley: Have done manually. Thanks

strawberryfields/gbs/data.py
trbromley: Done!

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,215 @@
...
+        if isinstance(key, (slice, tuple)):
+            if isinstance(key, tuple):
+                key = slice(*key)
+            range_tuple = key.indices(self.n_samples)
+            return [self._elem(i) for i in range(*range_tuple)]
+
+        if key < 0:
+            key += self.n_samples
+
+        return self._elem(key)
```
zzzeid: Since `__getitem__` can get a variety of inputs, we should probably check the input here and provide some output to the user if they provide an incorrect input (akin to the error you would get from trying to get an element in a list, etc.)

It also could be more readable to flatten the if statements a little, like this:

```python
    def __getitem__(self, key):
        if not isinstance(key, (slice, tuple, int)):
            raise TypeError("Dataset indices must be integers, slices, or tuples")

        if isinstance(key, int):
            return self._elem(key + self.n_samples)
        
        if isinstance(key, tuple):
            key = slice(*key)

        range_tuple = key.indices(self.n_samples)
        return [self._elem(i) for i in range(*range_tuple)]
```

trbromley: image

^^ That's cool!

trbromley: And thanks for this 😁 Have added!!

strawberryfields/gbs/data.py
```diff
@@ -0,0 +1,217 @@
...
+- ``n_mean``: mean number of photons in the GBS device
+
+- ``n_max``: maximum number of photons allowed in any sample
+
+-  ``threshold``: flag to indicate whether samples are generated with threshold detection or
+   with photon-number-resolving detectors.
+
+- ``n_samples``: total number of samples in the dataset
+
+- ``modes``: number of modes in the GBS device or, equivalently, number of nodes in the graph
```
josh146: Nitpick, but for the above, I would either have full-stops for all bullet point lists, or no full stops (to keep it consistent).

trbromley: Fixed 😁

PR #199, opened 2019-09-27
Add graph visualization module

ixfoduap:
The plot module enables users to create high-quality visualizations of graphs and subgraphs related to GBS applications. It uses plot.ly to create the plots and uses a custom Strawberry Fields color scheme.

This is work in progress, for example missing tests.


strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+# limitations under the License.
+r"""
+Plotting and visualization
+==========================
+
+**Module name:** :mod:`strawberryfields.gbs.plot`
+
+.. currentmodule:: strawberryfields.gbs.plot
+
+This module provides functionality for visualizing GBS outputs as graphs and point processes.
```
trbromley: Do we want to link it specifically to GBS, e.g., could be just "functionality for visualizing graphs, subgraphs and point processes"?

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+.. autosummary::
+    _node_coords
+    edge_coords
```
trbromley: Should we keep `edge_coords` private too?
Or we could make them both public, but given that they are very Plotly specific and I can't see users needing them - how about both private?

In which case, probably remove them from autosummary?

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
+    """ Provides the coordinates for the graph nodes when given an input graph layout.
+
+    Args:
+        graph (nx.Graph): input graph
+        l (dict): dictionary of nodes and their respective coordinates
+
+    Returns:
+         Tuple: x and y coordinates for each node
```
trbromley:
```suggestion
         Dict[str, list]: lists of x and y coordinates accessed as keys of a dictionary
```
:thinking: still not sure on best wording

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+^^^^^^^^^^^^
+"""
+from typing import Tuple, Optional
+
+import networkx as nx
+import plotly.graph_objects as go
+
+
+def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
+    """ Provides the coordinates for the graph nodes when given an input graph layout.
```
trbromley: I see it as more of a converter, i.e., `l` already has the coordinates, this just separates them out into x and y coordinate arrays that are keys of a dict, for easy passing to some plotly functions.

trbromley:
```suggestion
    """Provides the coordinates for the graph nodes when given an input graph layout.
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+    for n in graph.nodes():
+        n_x.append(l[n][0])
+        n_y.append(l[n][1])
+
+    return {"x": n_x, "y": n_y}
+
+
+def edge_coords(graph: nx.Graph, l: dict) -> dict:
+    """ Provides the coordinates for the graph edges when given an input graph layout.
```
trbromley:
```suggestion
    """Provides the coordinates for the graph edges when given an input graph layout.
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+Code details
+^^^^^^^^^^^^
+"""
+from typing import Tuple, Optional
+
+import networkx as nx
+import plotly.graph_objects as go
+
+
+def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
```
trbromley:
```suggestion
def _node_coords(graph: nx.Graph, l: dict) -> dict:
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+    return {"x": n_x, "y": n_y}
+
+
+def edge_coords(graph: nx.Graph, l: dict) -> dict:
+    """ Provides the coordinates for the graph edges when given an input graph layout.
+
+        Args:
+            graph (nx.Graph): input graph
+            l (dict): dictionary of edges and their respective coordinates
```
trbromley:
```suggestion
            l (dict): dictionary of nodes and their respective coordinates
```
It's also node-based from NetworkX

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+        Returns:
+             dict: x and y coordinates for beginning and end of each edge
```
trbromley: Do we want to mention that `None` is artificially placed to act as a separator between pairs of nodes/edges

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+        e_y.append(start_y)
+        e_y.append(end_y)
+
+        e_x.append(None)
+        e_y.append(None)
+
+    return {"x": e_x, "y": e_y}
+
+
+web_green1 = '#84ba5b'
```
trbromley: Will these be in the final version or just for prototyping? :grin:

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+sf_red = '#d35e60'
+sf_grey = '#efefef'
+light_grey = '#CDCDCD'
+
+graph_node_colour = web_green2
+graph_edge_colour = light_grey
+subgraph_node_colour = web_red2
+subgraph_edge_colour = web_red2
+
+graph_node_size = 14
```
trbromley: The things we do keep, they can be all capitalized to follow the module level constant style: https://www.programiz.com/python-programming/variables-constants-literals

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+subgraph_node_size = 16
+
+
+def plot(graph: nx.Graph, subgraph: Optional[list] = None, size: float = 500) -> None:
+    """ Creates a plot.ly plot of the input graph.
```
trbromley:
```suggestion
    """Creates a `Plotly <https://plot.ly/>`__ plot of the input graph.
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+    The graph layout is fixed to be the Kamada-Kawai layout with an aspect ratio of 1:1. The
```
trbromley: :+1:

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+    function can plot just the input graph or the graph with a specified subgraph highlighted.
+    The function uses the standard colour theme of green nodes, grey edges, and red highlighted
```
trbromley: Should we define the standard colour scheme somewhere, maybe module docstring?

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+    subgraph.
+
+    **Example usage**:
+        >>> graph = nx.complete_graph(10)
+        >>> fig = plot(graph, [0, 1, 2, 3])
+        >>> fig.show()
+
```
trbromley: Is the Kamada Kawaii layout deterministic? (which I like!)

We can save to html and embed the result in the docstring, which will look really nice!

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+        Args:
+            graph (nx.Graph): input graph
+            subgraph (list): list of nodes comprising the subgraph to highlight
```
trbromley:
```suggestion
            subgraph (list): optional list of nodes comprising the subgraph to highlight
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+            size (dict): size of the plot
```
trbromley:
```suggestion
            size (float): size of the plot in pixels, rendered in a square layout
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+        Returns:
+             plot.ly graph object
```
trbromley:
```suggestion
             Figure: Plotly figure with graph and subgraph plotted
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+        )
+
+        s_nodes.text = [str(i) for i in s.nodes()]
+
+        f = go.Figure(data=[g_edges, s_edges, g_nodes, s_nodes], layout=layout)
+
+    else:
+        f = go.Figure(data=[g_edges, g_nodes], layout=layout)
+
+    return f
```
trbromley: Code looks good! Although I think it would be good to be able to input a go.Figure for the user if they want to add to a current graph

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def plot_subgraph(subgraph: nx.Graph, size: float = 500) -> None:
+    """ Creates a plot.ly plot of the input subgraph.
```
trbromley:
```suggestion
    """Creates a plot.ly plot of the input subgraph.
```

trbromley: We should decide on Plotly vs plot.ly (which I am kind of neutral aside from desire for consistency)

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+        The graph layout is fixed to be the Kamada-Kawai layout with an aspect ratio of 1:1.
+        This function uses the subgraph colour theme of red nodes and edges.
+
+        **Example usage**:
+        >>> graph = nx.complete_graph(10)
+        >>> subgraph = graph.subgraph([0, 1, 2, 3])
+        >>> fig = plot_subgraph(subgraph)
+        >>> fig.show()
+
+            Args:
```
trbromley: Is this over-indented?

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+                subgraph (nx.Graph): input subgraph
+                size (dict): size of the plot
+
+            Returns:
+                 plot.ly graph object
+            """
+
+    l = nx.kamada_kawai_layout(subgraph)
+
+    g_edges = go.Scatter(
```
trbromley: Would it make sense for a `_add_graph_to_plot` private function to do these lines, which I think are more or less repeated above

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+
+        f = go.Figure(data=[g_edges, s_edges, g_nodes, s_nodes], layout=layout)
+
+    else:
+        f = go.Figure(data=[g_edges, g_nodes], layout=layout)
+
+    return f
+
+
+def plot_subgraph(subgraph: nx.Graph, size: float = 500) -> None:
```
trbromley: I think we have artistic disagreement on this bit 😆 But for me it's really cool if we can plot the subgraph in the same layout as it would appear in the original graph. Could this be an option (can default to false)?

strawberryfields/gbs/plot.py
ixfoduap: You're right, best not to link specifically to GBS. In fact, would be cool if people use it outside of GBS because it looks so good!

strawberryfields/gbs/plot.py
ixfoduap: Thanks for spotting this. Both should be private and removed from autosummary

josh146: imo, I would keep it consistent (so both public, or both private). I lean towards public.

You can also have two summary sections:

```rest
Visualization functions
^^^^^^^^^^^^^^^^^^^^

The following functions produce plot.ly visualisations of graphs/subgraphs.

.. autosummary::    plot
    plot_subgraph

Auxiliary functions
^^^^^^^^^^^^^^^^

The following functions are used to generate the graph layout and data points
for the visualization functions above.

.. autosummary::
    node_coords
    edge_coords
```

strawberryfields/gbs/plot.py
ixfoduap: You're right, `l` already has the coordinates. I'll change the docstring to clarify this

josh146: Yeah, I agree --- this function already takes coordinates as input, so the name/description doesn't seem to align.

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+import networkx as nx
+import plotly.graph_objects as go
+
+
+def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
+    """ Provides the coordinates for the graph nodes when given an input graph layout.
+
+    Args:
+        graph (nx.Graph): input graph
+        l (dict): dictionary of nodes and their respective coordinates
```
josh146: How is `l` being generated?

strawberryfields/gbs/plot.py
ixfoduap: Great idea!

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+             plot.ly graph object
+        """
+
+    s = graph.subgraph(subgraph)
+    l = nx.kamada_kawai_layout(graph)
+
+    g_nodes = go.Scatter(
+        **_node_coords(graph, l),
+        mode='markers',
+        hoverinfo='text',
```
josh146: Don't forget to run `black -l 100`!

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,235 @@
...
+    )
+
+    g_edges = go.Scatter(
+        **edge_coords(graph, l),
+        line=dict(width=1, color=graph_edge_colour),
+        hoverinfo='none',
+        mode='lines'
+    )
+
+    g_nodes.text = [str(i) for i in graph.nodes()]
```
josh146: One thought --- does the gbs apps layer ever 'rename' or 're-index' the nodes, or are the original node names kept throughout the entire process?

josh146: In fact, it might be nice to allow the user to (optionally) specify which node attributes show up in the tool tips, or even a list of node attributes to display.

strawberryfields/gbs/plot.py
josh146: Is that a common use-case?

strawberryfields/gbs/plot.py
ixfoduap: As far as I understand, yes, it's deterministic. 
Embedding the result in the docstring sounds awesome, but I'll leave it for future commits.

strawberryfields/gbs/plot.py
ixfoduap: For now, I would rather avoid having a figure as input, since to me this is a very unnatural thing to do. For example, it's not clear to me that we can't also add to the graph in some other method. Also, I agree with Josh, not sure how common this is

strawberryfields/gbs/plot.py
ixfoduap: I actually prefer just plotly, so let's go for that?

strawberryfields/gbs/plot.py
ixfoduap: Good catch!

strawberryfields/gbs/plot.py
ixfoduap: I don't like plotting the subgraph in the same layout for two main reasons:
1. The subgraph already stands out a lot when highlighted in the original graph, so effectively removing the background graph doesn't much change things.
2. This is the most important reason: the subgraph layout within the graph is often doing very little to visually provide information about the subgraph. When we plot the subgraph on its own accord, it is much easier to analyze it. Consider the example below, it's much more obvious to see that there is a connecting path when we plot the subgraph with its own layout.
newplot (4)
newplot (5)

strawberryfields/gbs/plot.py
josh146: Note that this is a very common use case with matplotlib, but I'm not sure if it carries across to plot.ly

strawberryfields/gbs/plot.py
ixfoduap: From `nx.kamada_kawai_layout(graph)`. Do you think this should be mentioned in the docstring?

strawberryfields/gbs/plot.py
ixfoduap: Yes, thanks! Still WIP :wink:

strawberryfields/gbs/plot.py
ixfoduap: I believe the gbs apps layer maintains the original node names throughout the entire process. I believe that displaying node attributes can be customized in plotly, so your suggestion should be possible. I'd prefer to keep it in mind but not implement it for now

strawberryfields/gbs/plot.py
trbromley:
```suggestion
        l (dict[int, float]): dictionary of nodes and their respective coordinates, can be
        generated using a NetworkX `layout <https://networkx.github.io/documentation/latest/
        reference/drawing.html#module-networkx.drawing.layout>`__
```

strawberryfields/gbs/plot.py
trbromley: Future commits or future PRs? I think we can fit it in here, as it's minimal overhead (I can do this?)

strawberryfields/gbs/plot.py
trbromley: @josh146 no we follow the input node labelling throughout (i.e., one could start with a graph that has nodes [1, 3, 65, 777]).

The only exception is when one takes raw samples from the device, as `sample.sample` returns samples generated from an adjacency matrix, but then `sample.subgraphs` gives subgraph samples that follow the node labelling of the input graph.

strawberryfields/gbs/plot.py
trbromley: So my understanding was that the use case in matplotlib is:

- User already has some axes, potentially with stuff already plotted on
- User calls e.g. `nx.draw` and has the axes as an argument.
- The result is that their axes have been added onto with the specified graph

Is that what you are thinking @josh146? And then I guess we should decide if that makes sense and is something we can do in Plotly.

strawberryfields/gbs/plot.py
ixfoduap: I tried to do this but in the end concluded it's not worth it for two main reasons:
1. The format is different for plotting a graph and a subgraph, so the `_add_graph_to_plot` function needs to acommodate both options. When doing so, we don't save that many lines of codes
2. Somehow the formatting of the plot is different and not as nice. For example, the edges of the graph now appear on top of the nodes, instead of hidden behind.

All these things could be fixed with more work, but I don't think it's worth it

ixfoduap:
**Note:** Still missing tests

strawberryfields/gbs/plot.py
trbromley: Done, took me ages 😆

strawberryfields/gbs/plot.py
trbromley: In person discussion with @ixfoduap: decided _not_ to have plotly figures as an input argument, and let's wait and see based on future feedback. @josh146 let us know if you agree with that!

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+graph_node_colour = GREEN
+graph_edge_colour = LIGHT_GREY
+subgraph_node_colour = RED
+subgraph_edge_colour = RED
+
+graph_node_size = 14
+subgraph_node_size = 16
+
+
+def plot(graph: nx.Graph, subgraph: Optional[list] = None, size: float = 500) -> None:
```
trbromley: image

I think we need to look at the best naming 🤔 

Actually, would it be better for the function to just be `strawberryfields.gbs.plot`? How would we do that, have the code in the `__init__` of `gbs`?

trbromley: Or, if we follow our more standard style, would this function be `gbs.plot.graph`?

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+==========================
+
+**Module name:** :mod:`strawberryfields.gbs.plot`
+
+.. currentmodule:: strawberryfields.gbs.plot
+
+This module provides functionality for visualizing graphs, subgraphs, and point processes. Graphs
+are plotted using the Kamada-Kawai layout with an aspect ratio of 1:1. The module uses a custom
+Strawberry Fields colour scheme. The standard scheme for graphs uses green nodes and grey edges,
+whereas the scheme for subgraphs uses red nodes and edges.
```
trbromley: Might be good here to mention that we need plotly for plotting, which is not a dependency of SF

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+Code details
+^^^^^^^^^^^^
+"""
+from typing import Optional, Tuple
+
+import networkx as nx
+
+
+def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
+    """Converts coordinates for the graph edges for plotting purposes.
```
trbromley:
```suggestion
    """Converts coordinates for the graph nodes for plotting purposes.
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+
+    Args:
+        graph (nx.Graph): input graph
+        l (dict[int, float]): dictionary of nodes and their respective coordinates, can be
+        generated using a NetworkX `layout <https://networkx.github.io/documentation/latest/
+        reference/drawing.html#module-networkx.drawing.layout>`__
+
+    Returns:
+         Dict[str, list]: lists of x and y coordinates accessed as keys of a dictionary
```
trbromley:
```suggestion
         dict[str, list]: lists of x and y coordinates accessed as keys of a dictionary
```
Never know whether to capitalize or not

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+import networkx as nx
+
+
+def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
+    """Converts coordinates for the graph edges for plotting purposes.
+
+    Args:
+        graph (nx.Graph): input graph
+        l (dict[int, float]): dictionary of nodes and their respective coordinates, can be
+        generated using a NetworkX `layout <https://networkx.github.io/documentation/latest/
```
trbromley:
```suggestion
            generated using a NetworkX `layout <https://networkx.github.io/documentation/latest/
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+        reference/drawing.html#module-networkx.drawing.layout>`__
```
trbromley:
```suggestion
            reference/drawing.html#module-networkx.drawing.layout>`__
```
This was probably my fault with the previous suggestion

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+
+    return {"x": n_x, "y": n_y}
+
+
+def _edge_coords(graph: nx.Graph, l: dict) -> dict:
+    """Converts coordinates for the graph edges for plotting purposes.
+
+        Args:
+            graph (nx.Graph): input graph
+            l (dict): dictionary of nodes and their respective coordinates
```
trbromley:
```suggestion
            l (dict[int, float]): dictionary of nodes and their respective coordinates, can be
                generated using a NetworkX `layout <https://networkx.github.io/documentation/latest/
                reference/drawing.html#module-networkx.drawing.layout>`__
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Returns:
+             dict: x and y coordinates for beginning and end of each edge. `None` is placed as a
```
trbromley:
```suggestion
             dict[str, list]: lists of x and y coordinates for the beginning and end of each edge.
             ``None`` is placed as a
```

strawberryfields/gbs/plot.py
trbromley:
```suggestion
def plot(graph: nx.Graph, subgraph: Optional[list] = None, size: float = 500) -> go.Figure:
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+
+graph_node_colour = GREEN
+graph_edge_colour = LIGHT_GREY
+subgraph_node_colour = RED
+subgraph_edge_colour = RED
+
+graph_node_size = 14
+subgraph_node_size = 16
+
+
```
trbromley:
```suggestion

@pytest.mark.skip(reason="Plotting functionality not currently tested")
```
Should we add this? @josh146 
Is this how to turn off testing/codecov reporting for this (and the following) function?
http://doc.pytest.org/en/latest/skipping.html

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+        size (float): size of the plot in pixels, rendered in a square layout
+
+    Returns:
+         Figure: Plotly figure for graph and optionally highlighted subgraph
+    """
+    try:
+        import plotly.graph_objects as go
+    except ImportError:
+        raise ImportError("Plotly required for using plot()")
+    except RuntimeError:
```
trbromley: When do we think this will happen, is it e.g. running on a remote server? Should we test (i.e., just by trying, not writing tests) how this looks like?

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+        raise
+
+    s = graph.subgraph(subgraph)
+    l = nx.kamada_kawai_layout(graph)
+
+    g_nodes = go.Scatter(
+        **_node_coords(graph, l),
+        mode="markers",
+        hoverinfo="text",
+        marker=dict(color=graph_node_colour, size=graph_node_size, line_width=2),
```
trbromley: Maybe not worth it for this iteration, but in future it would be good to allow the user to specify styling too. This could be, e.g., `marker = kwargs["graph_markers"]` like this, where the user can feed in a `graph_markers` argument if they want to, that becomes the default above if not specified.

trbromley: also for the layout below, which could be fed directly from the user if they want to

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+
+        s_nodes = go.Scatter(
+            **_node_coords(s, l),
+            mode="markers",
+            hoverinfo="text",
+            marker=dict(color=subgraph_node_colour, size=subgraph_node_size, line_width=2),
+        )
+
+        s_nodes.text = [str(i) for i in s.nodes()]
```
trbromley: One thing I noticed is that when the user hovers over a subgraph node, the colour can change between red and green which looks like a bug. I think we can fix this by having `g_nodes` only be the nodes in the complement graph if a subgraph is specified.

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+
+        f = go.Figure(data=[g_edges, s_edges, g_nodes, s_nodes], layout=layout)
+
+    else:
+        f = go.Figure(data=[g_edges, g_nodes], layout=layout)
+
+    return f
+
+
+def plot_subgraph(subgraph: nx.Graph, size: float = 500) -> None:
```
trbromley:
```suggestion
def plot_subgraph(subgraph: nx.Graph, size: float = 500) -> go.Figure:
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+
+    .. image:: ../../_static/complete_graph.png
+       :width: 40%
+       :align: center
+       :target: javascript:void(0);
+
+    Args:
+        graph (nx.Graph): input graph
+        subgraph (list): optional list of nodes comprising the subgraph to highlight
+        size (float): size of the plot in pixels, rendered in a square layout
```
trbromley:
```suggestion
        size (int): size of the plot in pixels, rendered in a square layout
```

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+    >>> fig.show()
+
+    .. image:: ../../_static/complete_subgraph.png
+       :width: 40%
+       :align: center
+       :target: javascript:void(0);
+
+    Args:
+        subgraph (nx.Graph): input subgraph
+        size (dict): size of the plot
```
trbromley:
```suggestion
        size (int): size of the plot in pixels, rendered in a square layout
```

strawberryfields/gbs/plot.py
trbromley: Ok, agreed in person to drop this

tests/gbs/test_plot.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+
+from strawberryfields.gbs import plot
+
+
+class TestNodeCoords:
+    """Test for the internal function strawberryfields.gbs.plot._node_coords"""
+
+    @pytest.mark.parametrize("dim", [4, 5, 6])
+    def test_node_coords(self, dim):
```
trbromley: I enjoyed this test 😁

This is nitpicking, but if we only need one test for this function, then we could probably do away with the `TestNodeCoords` class and just have the inner function.

^ This is not important though

tests/gbs/test_plot.py
```diff
@@ -0,0 +1,59 @@
...
+
+        assert np.allclose(radii, np.ones(dim), atol=1e-5)
+
+
+class TestEdgeCoords:
+    """Test for the internal function strawberryfields.gbs.plot._edge_coords"""
+
+    @pytest.mark.parametrize("dim", [4, 5, 6])
+    def test_edge_coords(self, dim):
+        """Tests that the length of edges in a circular cycle graph are all of equal length."""
```
trbromley: Tests are great. For both cases, I would also add a test that doesn't depend on `kamada_kawai_layout` and a very simple hard coded layout (e.g. 4 nodes).

strawberryfields/gbs/plot.py
josh146: tbh I like `plot` as a function name, instead I would rename the module. The function is what the user will be typing more often than not, so `plot` is perfect, whereas the module can afford to be a bit more verbose.

josh146: An alternative would be `plot_graph` for the function name, I suppose?

strawberryfields/gbs/plot.py
josh146: The `pytest.mark.skip` decorator is only used inside the test suite, not the actual codebase.

If you want to skip testing it, all you have to do is not write a test for it 🙂 

If you don't want the coverage to show up either, you can simply add the comment `# pragma: no cover` to the right of the function signature (after the colon).

strawberryfields/gbs/plot.py
josh146: Actually, maybe the latter (`plot_graph`) makes sense, in case there are eventually a lot more different types of plotting functions?

strawberryfields/gbs/plot.py
josh146: This is something we should test as well 😉

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,254 @@
...
+        subgraph (list): optional list of nodes comprising the subgraph to highlight
+        size (float): size of the plot in pixels, rendered in a square layout
+
+    Returns:
+         Figure: Plotly figure for graph and optionally highlighted subgraph
+    """
+    try:
+        import plotly.graph_objects as go
+    except ImportError:
+        raise ImportError("Plotly required for using plot()")
```
josh146: It might be worth including more information in this exception. For example, the command required to pip install plotly, or maybe a URL to a useful resource (e.g., something in the SF documentation, or the plot.ly installation guide).

strawberryfields/gbs/plot.py
josh146: I agree. While passing axes/figure is a very common matplotlib pattern, I'm not sure the same is true with plot.ly. We can re-evaluate if we have user feedback down the line asking for it.

strawberryfields/gbs/plot.py
ixfoduap: Yes, for example there will be a function for plotting point processes, and maybe others in the future. I like using `plot_[thing_you_want_to_plot]` as the name convention for the module, so `plot_graph` in this case.

strawberryfields/gbs/plot.py
```diff
@@ -0,0 +1,229 @@
...
+        e_y.append(None)
+
+    return {"x": e_x, "y": e_y}
+
+
+GREEN = "#3e9651"
+RED = "#cc2529"
+LIGHT_GREY = "#CDCDCD"
+
+graph_node_colour = GREEN
```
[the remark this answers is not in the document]
ixfoduap: I view them as variables, since in principle you have the freedom to add more colours. For example, we could have `XANADU_TEAL='#1D9598'` and choose `graph_node_colour = XANADU_TEAL`. Also, it looks ugly to have so many caps, feels like we're shouting :scream_cat:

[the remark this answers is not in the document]
trbromley: AGREED, MAKES SENSE!

strawberryfields/gbs/plot.py
trbromley: :+1:

tests/gbs/test_plot.py
ixfoduap: Done!

tests/gbs/test_plot.py
ixfoduap: Done!

strawberryfields/gbs/plot.py
trbromley: @josh146, how would we test it? I guess this error might happen naturally on CI?

josh146: Could you mock the import so that it causes a `RuntimeError`?

PR #201, opened 2019-10-02
Add point process sampler

soranjh:
Add a function to the points module of the application layer for permanental point process sampling.


strawberryfields/gbs/points.py
```diff
@@ -66,3 +68,34 @@ def kernel(R: np.ndarray, sigma: float) -> np.ndarray:
         K (array): the kernel matrix
     """
     return np.exp(-(scipy.spatial.distance.cdist(R, R)) ** 2 / 2 / sigma ** 2)
+
+
+def perpp(R: np.ndarray, sigma: float, n_mean: int, ns: int) -> np.ndarray:
+    """This function generates permanental point process (PerPP) samples with thermal states.
```
trbromley:
```suggestion
    """Sample subsets of points using the permanental point process.

    Points are encoded through a radial basis function kernel, provided in :func:`kernel`,
    that is a function of the distance between pairs of points. Subsets of points are sampled
    with probabilities that are proportional to the permanent of the *sub*matrix of the kernel
    selected by those points.

    This permanental point process is likely to sample points that are clustered together
    :cite:`soran2019`. It can be realised using a variant of Gaussian boson sampling with thermal
    states as input.
```
Feel free to reword!
Note that, you need to add a citation to `doc/references.bib`.

strawberryfields/gbs/points.py
```diff
@@ -66,3 +68,34 @@ def kernel(R: np.ndarray, sigma: float) -> np.ndarray:
...
+
+    **Example usage:**
+    >>> R = np.array([[0, 1], [1, 0], [0, 0],[1, 1]])
```
trbromley:
```suggestion

    >>> R = np.array([[0, 1], [1, 0], [0, 0],[1, 1]])
```

strawberryfields/gbs/points.py
```diff
@@ -66,3 +68,34 @@ def kernel(R: np.ndarray, sigma: float) -> np.ndarray:
         K (array): the kernel matrix
     """
     return np.exp(-(scipy.spatial.distance.cdist(R, R)) ** 2 / 2 / sigma ** 2)
+
+
+def perpp(R: np.ndarray, sigma: float, n_mean: int, ns: int) -> np.ndarray:
```
trbromley: What is the motivation for having `R` and `sigma` as inputs as opposed to the kernel matrix directly?
It necessarily constrains us to using the RBF kernel (which is actually maybe something we want to do at this point).

trbromley: We also use lists of lists as output in the rest of the apps layer. I'm not actually sure if that's the right thing to do, but maybe to be consistent we can also do it here.

trbromley: `n_mean` should be int? I think float would be better.

tests/gbs/test_points.py
```diff
@@ -35,3 +36,14 @@ def test_kernel():
...
+
+def test_perpp():
+    r"""Tests that the generated samples have the correct dimension"""
+
+    R = np.array([[0, 1], [1, 0], [0, 0], [1, 1]])
+    samples = perpp(R, 1.0, 1, 10)
+    shape = samples.shape
+    shape_ref = (10, 4)
+
+    assert shape == shape_ref
```
trbromley: Test looks good to me!

trbromley:
Also, the reason that the CI tests are failing is because it is using the pip version of `thewalrus` which does not have the latest `csamples` module added. Possible solutions I can think of are:

- Release new version of `thewalrus` with `csamples` added. Might be good to properly integrate `csamples` before then.
- Make SF depend on the `master` branch version of `thewalrus`. Definitely not a good idea.
- Include all of `csamples` in the apps layer. Would rather not do this due to code replication, effort of doing whole process again, and my opinion that the code in `csamples` shouldn't live in the apps layer.

Should probably have thought about this before 🤦‍♂
@josh146 @nquesada

strawberryfields/gbs/points.py
soranjh: The main reason is that the user does not have to create the kernel and then sample from it. The user just calls the wrapper function. But I think it is better to separate them: the user calls both and the sampler gets the kernel matrix as input.

soranjh: Will change the function to create list of lists.

soranjh:
```suggestion
def perpp(R: np.ndarray, sigma: float, n_mean: int, n_samples: int) -> np.ndarray:
```

strawberryfields/gbs/points.py
```diff
@@ -66,3 +68,34 @@ def kernel(R: np.ndarray, sigma: float) -> np.ndarray:
...
+        sigma (float): kernel parameter
+        n_mean (int): average number of points in each sample
+        ns (int): number of samples to be generated
+
+    Returns:
+        samples (array): samples generated by PerPP
+    """
+    K = kernel(R, sigma)
+    ls, O = rescale_adjacency_matrix_thermal(K, n_mean)
+    return np.array(generate_thermal_samples(ls, O, num_samples=ns))
```
soranjh:
```suggestion
    return np.array(generate_thermal_samples(ls, O, num_samples=n_samples))
```

strawberryfields/gbs/points.py
```diff
@@ -66,3 +68,44 @@ def kernel(R: np.ndarray, sigma: float) -> np.ndarray:
...
+        sigma (float): kernel parameter
+        n_mean (int): average number of points in each sample
+        ns (int): number of samples to be generated
+
+    Returns:
+        samples (array): samples generated by PerPP
+    """
+    K = kernel(R, sigma)
+    ls, O = rescale_adjacency_matrix_thermal(K, n_mean)
+    return np.array(generate_thermal_samples(ls, O, num_samples=n_samples))
```
trbromley: Was it discussed to have this as a list?

doc/references.bib
```diff
@@ -896,3 +896,12 @@ @article{kelleher2009generating
...
 }
+
+@misc{jahangiri2019point,
+    title={Point Processes with Gaussian Boson Sampling},
+    author={Soran Jahangiri and Juan Miguel Arrazola and Nicolás Quesada and Nathan Killoran},
+    year={2019},
+    eprint={1906.11972},
+    archivePrefix={arXiv},
+    primaryClass={quant-ph}
+}
```
trbromley:
```suggestion
}

```
😅

doc/references.bib
```diff
@@ -896,3 +896,12 @@ @article{kelleher2009generating
   journal={arXiv preprint arXiv:0909.2331},
   year={2009}
 }
+
```
nquesada: I'd strongly encourage to also cite this paper:

@article{rahimi2015can,
  title={What can quantum optics say about computational complexity theory?},
  author={Rahimi-Keshari, Saleh and Lund, Austin P and Ralph, Timothy C},
  journal={Physical review letters},
  volume={114},
  number={6},
  pages={060501},
  year={2015},
  publisher={APS}
}

soranjh: We can cite this in thewalrus. We have not really used it in SF.

nquesada:
It is already in the walrus.

soranjh:
> Other than a bibliographic suggestion things look good. I wonder if other than return size you should have correctness tests.

All functions used in `samples` wrapper, have been tested separately. The shape is practically all we need here.

PR #205, opened 2019-10-07
Apps: Adds point process visualization

ixfoduap:
This pull request adds the function `plot_points` to the `plot` module of the GBS apps layer. This function takes as input: (i) the coordinate matrix of a set of points, and (ii) a sample of points expressed as a list. It outputs a plotly figure of all points, highlighting the sampled points. The figure has a white background, sampled points are highlighted in red, and all other points are shown in light grey, as seen in the images below.

normal_pp

grid_pp

Note: There are no tests for this PR as the only change is a function that outputs a figure, which we have decided not to test.


strawberryfields/gbs/plot.py
```diff
@@ -260,3 +261,72 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+        margin=dict(b=0, l=0, r=0, t=25),
+        height=size,
+        width=size,
+        plot_bgcolor="white",
+    )
+
+    points = go.Scatter(x=R[:, 0], y=R[:, 1],
+                        mode="markers",
+                        hoverinfo="text",
+                        marker=dict(color='#F2F2F2', size=30,
```
soranjh: Can we make the marker size to depend on the number of points? i.e. the points become smaller when the number of points increases such that they never overlap.

strawberryfields/gbs/plot.py
```diff
@@ -260,3 +266,79 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+        hoverinfo="text",
+        marker=dict(color="#F2F2F2", size=30, line=dict(color="black", width=1.5)),
+    )
+
+    points.text = [str(i) for i in range(len(R))]
+
+    s_x = []
+    s_y = []
+    sampled_points = [i for i in range(len(sample)) if sample[i] > 0]
+    for i in sampled_points:
```
trbromley:
```suggestion
    for s in sample:
        if s > 0:
```
to save defining `sampled_points`?

trbromley:
```suggestion
    for s in gbs.sample.modes_from_counts(sample):
```
I think this is kind of what we want from `modes_from_counts`. Also, this raises the question: why do we prefer samples to be input in the "list of counts" format as opposed to the "list of nodes" format (which equals the points selected)?

strawberryfields/gbs/plot.py
```diff
@@ -260,3 +266,79 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+        width=size,
+        plot_bgcolor="white",
+    )
+
+    points = go.Scatter(
+        x=R[:, 0],
+        y=R[:, 1],
+        mode="markers",
+        hoverinfo="text",
+        marker=dict(color="#F2F2F2", size=30, line=dict(color="black", width=1.5)),
```
trbromley: Should we define the colour `#F2F2F2` ?

strawberryfields/gbs/plot.py
trbromley: So that the diameter of a point is always equal to the minimum distance between points?

strawberryfields/gbs/plot.py
```diff
@@ -30,10 +30,12 @@
 .. autosummary::
     plot_graph
     plot_subgraph
+    plot_points
```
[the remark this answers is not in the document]
ixfoduap: I think my suggested import syntax would be

`import strawberryfields.gbs.plot as plt`
`plt.plot_graph(graph)`

which mimics matplotlib. In that case we can keep the current function names

strawberryfields/gbs/plot.py
```diff
@@ -30,10 +30,12 @@
...
 
 Code details
 ^^^^^^^^^^^^
 """
+# pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
ixfoduap: Thanks!

strawberryfields/gbs/plot.py
```diff
@@ -260,3 +266,79 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
     f = go.Figure(data=[g_edges, g_nodes], layout=layout)
 
     return f
+
+
+def plot_points(R: np.ndarray, sample: list, size: int = 500) -> None:  # pragma: no cover
```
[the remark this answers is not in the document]
ixfoduap: Hmmm, that's a good point. I think you're right, we should make the sample optional.

strawberryfields/gbs/plot.py
```diff
@@ -260,3 +266,79 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+    """Creates a Plotly plot of the input points given their coordinates, highlighting an input
+    sample of points.
```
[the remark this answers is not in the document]
ixfoduap: Yes, great point

strawberryfields/gbs/plot.py
ixfoduap: Yes!

strawberryfields/gbs/plot.py
ixfoduap: The output of the samplers is in the "list of counts" format, so this should be the default input

strawberryfields/gbs/plot.py
ixfoduap: Ok, I will test this more

soranjh: @trbromley yes, something like that.

strawberryfields/gbs/plot.py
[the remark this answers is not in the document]
soranjh: Agree with that, it is good to have both abilities: just data or data+highlights.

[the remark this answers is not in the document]
soranjh: Another useful option will be inputing the point indices that need to be highlighted.

strawberryfields/gbs/plot.py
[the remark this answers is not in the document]
josh146: > `import strawberryfields.gbs.plot as plt`

I like this, but why not just call the module `plt` then?

strawberryfields/gbs/plot.py
```diff
@@ -252,11 +259,94 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+       :width: 40%
+       :align: center
+       :target: javascript:void(0);
+
+    Args:
+        R (np.array): Coordinate matrix. Rows of this array are the coordinates of the points.
+        sample (list[int]): optional subset of sampled points to be highlighted
+
+    Returns:
+         Figure: Plotly figure of points with the sample highlighted
```
josh146: No need to repeat plotly here, right?
```suggestion
         Figure: figure of points with the sample highlighted
```

strawberryfields/gbs/plot.py
```diff
@@ -252,11 +259,94 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+    try:
+        import plotly.graph_objects as go
+    except ImportError:
+        raise ImportError(
+            "Plotly required for using plot_points(). Can be installed using pip install "
+            "plotly or visiting https://plot.ly/python/getting-started/#installation"
+        )
+    except RuntimeError:
+        print("Plotly unable to open display")
+        raise
```
josh146: This `except:` is a bit weird. If plotly already raises an error when not able to open a display, why do we need to catch it?

josh146: Also, since this `try/except` statement occurs in multiple places, might as well collate it to a single function call (such as `try_import_plotly()`).

strawberryfields/gbs/plot.py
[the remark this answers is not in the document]
trbromley: I'm happy to have an extra "o" to make "plot"

@josh146 is there any difference between

`from strawberryfields.gbs import plot`

and

`import strawberryfields.gbs.plot as plot`

?

I prefer the former.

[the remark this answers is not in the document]
trbromley: Anyway the function names should be fine for now (although I prefer `graph` to `plot_graph`)

strawberryfields/gbs/plot.py
trbromley: Possibly agree, although we should then remove the "Plotly" in the instances above for plotting graphs and subgraphs

strawberryfields/gbs/plot.py
```diff
@@ -252,11 +259,94 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+    >>> plot_points(R, sample).show()
+
+    .. image:: ../../_static/normal_pp.png
+       :width: 40%
+       :align: center
+       :target: javascript:void(0);
+
+    Args:
+        R (np.array): Coordinate matrix. Rows of this array are the coordinates of the points.
+        sample (list[int]): optional subset of sampled points to be highlighted
```
trbromley: Need to add `plot_size` and `point_size` to arguments

strawberryfields/gbs/plot.py
```diff
@@ -252,11 +259,94 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
         plot_bgcolor="#ffffff",
     )
 
     f = go.Figure(data=[g_edges, g_nodes], layout=layout)
 
     return f
+
+
+def plot_points(R: np.ndarray, sample: Optional[list] = None,
+                plot_size: int = 500, point_size: int = 30) -> None:  # pragma: no cover
```
trbromley:
```suggestion
                plot_size: int = 500, point_size: int = 30) -> go.Figure:  # pragma: no cover
```
I noticed this is also the case for the other 2 plotting functions

strawberryfields/gbs/plot.py
```diff
@@ -108,13 +112,14 @@ def _edge_coords(graph: nx.Graph, l: dict) -> dict:
 subgraph_node_size = 16
 
 
-def plot_graph(graph: nx.Graph, subgraph: Optional[list] = None,
-               size: int = 500) -> None:  # pragma: no cover
-    """Creates a plotly plot of the input graph.
+def plot_graph(
+    graph: nx.Graph, subgraph: Optional[list] = None, plot_size: int = 500
```
trbromley: Am happy with the `plot_size` change, but it should be updated in the docstring "Args" section

strawberryfields/gbs/plot.py
trbromley: Yeah I'd be tempted to go with just the `ImportError` since I'm not so confident about the `RuntimeError` and if that is the thing that's triggered when Plotly can't open a display

ixfoduap: Good catch @josh146, I'm happy with keeping only the `ImportError` and grouping it into a single function

strawberryfields/gbs/plot.py
ixfoduap: Since we're not importing plotly directly in the module but through a `try`, this raises an error. Any suggestion on what to do?

strawberryfields/gbs/plot.py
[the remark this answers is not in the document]
josh146: @trbromley no difference. I like prefer the former as well.

strawberryfields/gbs/plot.py
```diff
@@ -252,11 +252,96 @@ def plot_subgraph(subgraph: nx.Graph, size: int = 500) -> None:  # pragma: no co
...
+    .. image:: ../../_static/normal_pp.png
+       :width: 40%
+       :align: center
+       :target: javascript:void(0);
+
+    Args:
+        R (np.array): Coordinate matrix. Rows of this array are the coordinates of the points.
+        sample (list[int]): optional subset of sampled points to be highlighted
+        plot_size (int): size of the plot in pixels, rendered in a square layout
+        point_size (int): size of the points, specified by its area
```
trbromley: It's definitely area and not radius or diameter?

PR #206, opened 2019-10-07
GBS Tutorial: Graph Similarity

trbromley:
This PR introduces a sphinx-gallery based tutorial for the `similarity` module of the `gbs` applications layer.


examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,203 @@
...
+# generated uniformly at random and their resultant probabilities are calculated. If :math:`N`
+# samples :math:`\{S_{1}, S_{2}, \ldots , S_{N}\}` are generated, then the event probability can
+# be approximated as
+#
+# .. math::
+#     p(E_{k, n_{\max}}) \approx \frac{1}{N}\sum_{i=1}^N p(S_i) |E_{k, n_{\max}}|,
+#
+# with :math:`|E_{k, n_{\max}}|` denoting the cardinality of the event.
+#
+# This method can be accessed using the :func:`prob_event_mc` function. The 4-photon event is
```
soranjh: Is there a link to this function?

trbromley: https://github.com/XanaduAI/strawberryfields/blob/aacbadd8461f2aaf72e27a5d67beb1336cfff427/strawberryfields/gbs/similarity.py#L333

I decided not to show it explicitly in the tutorial because it take a while to evaluate. We can look back at this though.

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,203 @@
...
+# into **events**, which correspond to a combination of orbits with the same photon number such
+# that the number of photons counted in each mode does not exceed a fixed value
+# ``max_count_per_mode``. To understand this, let's look at all of the orbits with a photon
+# number of 5:
+
+print(list(similarity.orbits(5)))
+
+##############################################################################
+# All 5-photon samples fall into one of the orbits above. A 5-photon event with
+# ``max_count_per_mode = 3`` means that we combine the orbits: ``[[1, 1, 1, 1, 1], [2, 1, 1, 1],
```
soranjh: what "combine" means here?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,203 @@
...
+print(similarity.feature_vector_sampling(m0, event_photon_numbers=[2, 4, 6], max_count_per_mode=2))
+
+##############################################################################
+# For the second method, suppose we want to calculate the event probabilities exactly rather than
+# through sampling. To do this, we consider the event probability :math:`p_{k, n_{\max}}` as the
+# sum over all sample probabilities in the event. As discussed in the GBS `tutorial
+# <gaussian_boson_tutorial>`, each sample probability is determined by the hafnian of a relevant
+# sub-adjacency matrix. While this is tough to calculate, what makes calculating :math:`p_{k,
+# n_{\max}}` really tough is the number of samples the corresponding event contains! For example,
+# the 17-photon event :math:`E_{k=6, n_{\max}=2}` contains the following number of samples:
```
soranjh: The term "17-photon event" might be confused with ":math:`k` photon event" appeared above. Is there a better way to refer to the 17? (maybe modes?)

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports,invalid-name
+"""
+Graph Similarity Tutorial
+=========================
+
+This tutorial looks at how to use GBS to construct a similarity measure between graphs,
+what is known as a graph kernel :cite:`schuld2019quantum`. Kernels can be applied to graph-based
```
ixfoduap:
```suggestion
known as a graph kernel :cite:`schuld2019quantum`. Kernels can be applied to graph-based
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports,invalid-name
+"""
+Graph Similarity Tutorial
+=========================
+
+This tutorial looks at how to use GBS to construct a similarity measure between graphs,
+what is known as a graph kernel :cite:`schuld2019quantum`. Kernels can be applied to graph-based
+data for machine learning tasks such as classification in machine learning using a support vector
```
ixfoduap:
```suggestion
data for machine learning tasks such as classification using a support vector machines
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports,invalid-name
+"""
+Graph Similarity Tutorial
+=========================
+
+This tutorial looks at how to use GBS to construct a similarity measure between graphs,
+what is known as a graph kernel :cite:`schuld2019quantum`. Kernels can be applied to graph-based
+data for machine learning tasks such as classification in machine learning using a support vector
+machine.
```
ixfoduap:
```suggestion

```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+what is known as a graph kernel :cite:`schuld2019quantum`. Kernels can be applied to graph-based
+data for machine learning tasks such as classification in machine learning using a support vector
+machine.
+
+Graph data
+----------
+
+Let's use the MUTAG dataset of graphs for this tutorial
+:cite:`debnath1991structure,kriege2012subgraph`. This is a dataset of 188 different graphs that
+each correspond to the structure of a chemical compound. Our objective is to use GBS samples from
```
ixfoduap:
```suggestion
each correspond to the structure of a chemical compound. Our goal is to use GBS samples from
```

ixfoduap: Purely stylistic suggestion, feel free to ignore if you don't prefer the change

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+m0_a = m0.adj
+m1_a = m1.adj
+m2_a = m2.adj
+m3_a = m3.adj
+
+##############################################################################
+# We can now plot the four graphs using the :mod:`~gbs.plot` module. To use this module,
+# we need to convert the adjacency matrices into NetworkX Graphs:
+
+import networkx as nx
```
ixfoduap: Reminder that we should discuss and agree on an import structure for the tutorials

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+import plotly
+
+plot_mutag_0 = plot.plot_graph(nx.Graph(m0_a))
+plot_mutag_1 = plot.plot_graph(nx.Graph(m1_a))
+plot_mutag_2 = plot.plot_graph(nx.Graph(m2_a))
+plot_mutag_3 = plot.plot_graph(nx.Graph(m3_a))
+
+plotly.offline.plot(plot_mutag_0, filename="MUTAG_0.html")
```
ixfoduap: This is true for all other tutorials, but it's worthwhile to include a note here stating that this command is used for rendering the docs, but hte user can write `plot_mutag_0.show()` instead

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+#     :file: ../../examples_gbs/MUTAG_2.html
+
+plotly.offline.plot(plot_mutag_3, filename="MUTAG_3.html")
+
+##############################################################################
+# .. raw:: html
+#     :file: ../../examples_gbs/MUTAG_3.html
+#
+# By visual inspection, we see that the graphs of ``m1_a`` and ``m2_a`` look very similar. In fact,
+# it turns out that they are *isomorphic* to each other, which means that the graphs can be made
```
ixfoduap: Nicely put!

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# identical by permuting their node labels.
+#
+# Samples from these graphs can be accessed by indexing:
```
ixfoduap: Maybe this goes better below lines 25-28, where the data module is introduced? It seems a bit out of context here

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+
+print(m0[0])
+
+##############################################################################
+# Creating a feature vector
+# -------------------------
+#
+# Following :cite:`schuld2019quantum`, we can create a *feature vector* to describe each graph.
+# These feature vectors contain information about the graphs and can be viewed as a mapping to a
+# high dimensional feature space, a technique often used in machine learning that allows us to
```
ixfoduap:
```suggestion
# high-dimensional feature space, a technique often used in machine learning that allows us to
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# find ways of separating points in the space for classification.
```
ixfoduap: This sentence reads a bit awkwardly. Maybe "that allows us to employ properties of the feature space to separate and classify the vectors" ?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+#
+# The feature vector of a graph can be composed in a variety of ways. One approach is to built it
+# using statistics from a GBS device configured to sample from the graph, as we now discuss.
+#
+# We begin by defining the concept of an *orbit*, which contains all the GBS samples that are
+# equivalent under permutation of the modes. A sample can be converted to its corresponding orbit
+# using the :func:`~.sample_to_orbit` function. For example, the first sample of ``m0`` is ``[0,
+# 0, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0]`` and has orbit:
+
+print(similarity.sample_to_orbit(m0[0]))
```
ixfoduap: Maybe explain here what `[1,1]` means? Namely, two photons detected, each in a separate mode.

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+
+##############################################################################
+# Other samples can be randomly generated from the ``[1, 1]`` orbit using:
+
+print(similarity.orbit_to_sample([1, 1], modes=m0.modes))
+
+##############################################################################
+# Orbits provide a useful way to coarse grain the samples from GBS into outcomes that are
```
ixfoduap:
```suggestion
# Orbits provide a useful way to coarse-grain the samples from GBS into outcomes that are
```

ixfoduap: Not 100% sure, but I think this use of coarse-grain as a verb should be hyphenated

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# statistically more likely to be observed. However, we are interested in coarse graining further
```
ixfoduap:
```suggestion
# statistically more likely to be observed. However, we are interested in coarse-graining further
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# into **events**, which correspond to a combination of orbits with the same photon number such
```
ixfoduap: This is a long sentence and a bit of a mouthful. Maybe say "... coarse-graining further into **events**. An event is a combination of orbits with the same photon number ..."

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+print(similarity.sample_to_event([0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 0, 3, 0, 0, 0], 3))
+
+##############################################################################
+# Samples with more than ``max_count_per_mode`` in any mode are not counted as part of the event:
+
+print(similarity.sample_to_event([0, 4, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0], 3))
+
+##############################################################################
+# Now that we have mastered orbits and events, how can we make a feature vector? It was shown in
+# :cite:`schuld2019quantum` that one way of making a feature vector of a graph is to measure the
```
ixfoduap: I'm not sure "measure" is the right word here. Maybe "obtain"? "calculate"? "estimate"?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# where :math:`\mathbf{k} := (k_{1}, k_{2}, \ldots , k_{K})` is a list of different total photon
+# numbers.
+#
+# Calculating a feature vector
+# ----------------------------
+#
+# We provide two methods for calculating a feature vector of GBS event probabilities in
+# Strawberry Fields:
+#
+# 1. Through sampling
```
ixfoduap:
```suggestion
# 1. Through sampling.
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# 2. Using a Monte Carlo estimate of the probability
```
ixfoduap:
```suggestion
# 2. Using a Monte Carlo estimate of the probability.
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+#
+# For the first method, all one needs to do is generate some GBS samples from the graph of
```
ixfoduap: Probably good to say say first that we'll focus on three-dimensional feature vectors, which can be calculated as follows

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# interest and feed through:
+
+print(similarity.feature_vector_sampling(m0, event_photon_numbers=[2, 4, 6], max_count_per_mode=2))
+
+##############################################################################
+# For the second method, suppose we want to calculate the event probabilities exactly rather than
+# through sampling. To do this, we consider the event probability :math:`p_{k, n_{\max}}` as the
+# sum over all sample probabilities in the event. In GBS, each sample probability is determined by
+# the hafnian of a relevant sub-adjacency matrix. While this is tough to calculate, what makes
+# calculating :math:`p_{k, n_{\max}}` really tough is the number of samples the corresponding
```
ixfoduap:
```suggestion
# calculating :math:`p_{k, n_{\max}}` really challenging is the number of samples the corresponding
```

ixfoduap: ^To avoid repetition of tough

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# event contains! For example, the 6-photon event over 17 modes :math:`E_{k=6, n_{\max}=2}`
+# contains the following number of samples :
+
+print(similarity.event_cardinality(6, 2, 17))
+
+##############################################################################
+# To avoid calculating a large number of sample probabilities, an alternative for calculating the
```
ixfoduap: replace with "an alternative is to perform a Monte Carlo approximation".

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# event probability is to perform a Monte Carlo approximation. Here, samples within an event are
+# generated uniformly at random and their resultant probabilities are calculated. If :math:`N`
```
ixfoduap:
```suggestion
# selected uniformly at random and their resultant probabilities are calculated. If :math:`N`
```
^To avoid repeated use of "generated"

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# This method can be accessed using the :func:`~.prob_event_mc` function. The 4-photon event is
+# approximated as:
+
+print(similarity.prob_event_mc(nx.Graph(m0_a), 4, max_count_per_mode=2, n_mean=6))
+
+##############################################################################
+# The feature vector can then be calculated through Monte Carlo sampling using
+# :func:`~.feature_vector_mc`.
+#
+# .. note::
```
ixfoduap: Nice note!

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+#     The results of :func:`~.prob_event_mc` and :func:`~.feature_vector_mc` are probabilistic and
+#     may vary between runs. Increasing the optional ``samples`` parameter will increase accuracy
+#     but slow down calculation.
+#
+# The second method of MC approximation is intended for use in scenarios where it is
```
ixfoduap:
```suggestion
# The second method of Monte Carlo approximation is intended for use in scenarios where it is
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# computationally intensive to pre-calculate a statistically significant dataset of samples from
+# GBS.
+#
+# Machine learning with GBS graph kernels
+# ---------------------------------------
+#
+# We have seen how GBS can be used to provide a mapping of graphs into a feature space. This
+# mapping can be used for machine learning tasks such as classification: by viewing each graph as
+# a point in the high dimensional space, we can use standard approaches from machine learning
```
ixfoduap:
```suggestion
# a point in the high-dimensional space, we can use standard approaches from machine learning
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+# - MUTAG2: Class 0
+# - MUTAG3: Class 1
+
+classes = [1, 0, 0, 1]
+
+##############################################################################
+# Can we use GBS feature vectors to classify these graphs? We start by defining two-dimensional
+# feature vectors:
+
+events = [8, 10]
```
ixfoduap: Why 8 and 10? This choice may need a bit more motivation/explanation

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+print(R)
+
+##############################################################################
+# Given our points in the feature space and corresponding classifications, we can use scikit learn
+# to train an SVM:
+
+from sklearn.svm import LinearSVC
+from sklearn.preprocessing import StandardScaler
+
+R_scaled = StandardScaler().fit_transform(R)  # Transform data to zero mean and unit variance
```
ixfoduap: Why do we transform data this way?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+
+classifier = LinearSVC()
+classifier.fit(R_scaled, classes)
```
ixfoduap: What does this function do exactly?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+
+##############################################################################
+# Here, the term "linear" refers to the *kernel* function used to calculate inner products
+# between vectors in the space. We can use a linear SVM because we have already embedded the
+# graphs in a feature space based upon GBS. We can then visualize the trained SVM by plotting the
```
ixfoduap:
```suggestion
# graphs in a feature space based on GBS. We can then visualize the trained SVM by plotting the
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,278 @@
...
+##############################################################################
+# .. raw:: html
+#     :file: ../../examples_gbs/SVM.html
+#
+# This plot shows the two classes (grey points for class 0 and red points for class 1)
+# successfully separated by the linear hyperplane using the GBS feature space. Moreover,
+# recall that the two MUTAG1 and MUTAG2 graphs of class 0 are actually isomorphic. Reassuringly,
+# their corresponding feature vectors are very similar. In fact, the feature vectors of
+# isomorphic graphs should always be identical :cite:`bradler2018graph` - the small discrepancy
+# in this plot is actually due to the statistical approximation from sampling.
```
ixfoduap:
```suggestion
# in this plot is due to the statistical approximation from sampling.
```

examples_gbs/run_tutorial_similarity.py
trbromley: Great, have made this change

examples_gbs/run_tutorial_similarity.py
trbromley: Fixed

examples_gbs/run_tutorial_similarity.py
trbromley: Done

examples_gbs/run_tutorial_similarity.py
trbromley: Have changed this by adding:

> The choice of ``events`` composing the feature vectors is arbitrary and we encourage the reader
to explore different combinations. Note, however, that odd photon-numbered events have zero
probability because ideal GBS only generates and outputs pairs of photons.

But appreciate suggestions!

examples_gbs/run_tutorial_similarity.py
trbromley: Honest reason:
The fit looked weird without doing it, and seemed to work better with scaling.

Nicer sounding reason:
I believe it's actually a fairly standard thing to do:
https://scikit-learn.org/stable/modules/preprocessing.html

> Standardization of datasets is a common requirement for many machine learning estimators implemented in scikit-learn; they might behave badly if the individual features do not more or less look like standard normally distributed data: Gaussian with zero mean and unit variance.

Do you think this should be explained in the text?

examples_gbs/run_tutorial_similarity.py
trbromley: It is to train the SVM from scikit learn, I've updated the sentence in line 238 to give a link:

> Given our points in the feature space and corresponding classifications, we can use scikit-learn's [LinearSVC](https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html) as our model to train:

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+
+##############################################################################
+# Samples with more than ``max_count_per_mode`` in any mode are not counted as part of the event:
+
+print(similarity.sample_to_event([0, 4, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0], 3))
+
+##############################################################################
+# Now that we have mastered orbits and events, how can we make a feature vector? It was shown in
+# :cite:`schuld2019quantum` that one way of making a feature vector of a graph is through the
+# probabilities of events. Specifically, for a :math:`k` photon event :math:`E_{k, n_{\max}}`
```
mariaschuld: Better: frequencies? One may not normalise!

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+# Creating a feature vector
+# -------------------------
+#
+# Following :cite:`schuld2019quantum`, we can create a *feature vector* to describe each graph.
+# These feature vectors contain information about the graphs and can be viewed as a mapping to a
+# high-dimensional feature space, a technique often used in machine learning that allows us to
+# employ properties of the feature space to separate and classify the vectors.
+#
+# The feature vector of a graph can be composed in a variety of ways. One approach is to build it
+# using statistics from a GBS device configured to sample from the graph, as we now discuss.
```
mariaschuld: I would give a 1-sentence explanation here, so the reader has a motivation and guidance when reading on. For example: "One example is to associate features with the relative frequencies of certain types of measurements being recorded from a GBS device configured ..."

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+
+This tutorial looks at how to use GBS to construct a similarity measure between graphs,
+known as a graph kernel :cite:`schuld2019quantum`. Kernels can be applied to graph-based
+data for machine learning tasks such as classification using a support vector machine.
+
+The main functionality for this tutorial is provided in :mod:`~.gbs.similarity`.
+
+Graph data
+----------
+
```
mariaschuld: This section is nice, but it is a bit unclear why it is right at the top here...better to move it down to the sampling section?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+# n_{\max}}:=p_{E_{k, n_{\max}}}(G)` with respect to a graph :math:`G`, a feature vector can be
+# written as
+#
+# .. math::
+#     f_{\mathbf{k}, n_{\max}} = (p_{k_{1}, n_{\max}}, p_{k_{2}, n_{\max}}, \ldots , p_{k_{K},
+#         n_{\max}}),
+#
+# where :math:`\mathbf{k} := (k_{1}, k_{2}, \ldots , k_{K})` is a list of different total photon
+# numbers.
+#
```
mariaschuld: Maybe add an example? Or a picture showing how samples get "funneled" into orbits, and orbits get "funneled" into events?

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+
+print(similarity.feature_vector_sampling(m0, event_photon_numbers=[2, 4, 6], max_count_per_mode=2))
+
+##############################################################################
+# For the second method, suppose we want to calculate the event probabilities exactly rather than
+# through sampling. To do this, we consider the event probability :math:`p_{k, n_{\max}}` as the
+# sum over all sample probabilities in the event. In GBS, each sample probability is determined by
+# the hafnian of a relevant sub-adjacency matrix. While this is tough to calculate, what makes
+# calculating :math:`p_{k, n_{\max}}` really challenging is the number of samples the corresponding
+# event contains! For example, the 6-photon event over 17 modes :math:`E_{k=6, n_{\max}=2}`
```
mariaschuld: Nice!

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+# Machine learning with GBS graph kernels
+# ---------------------------------------
+#
+# We have seen how GBS can be used to provide a mapping of graphs into a feature space. This
+# mapping can be used for machine learning tasks such as classification: by viewing each graph as
+# a point in the high-dimensional space, we can use standard approaches from machine learning
+# such as `support vector machines <https://en.wikipedia.org/wiki/Support-vector_machine>`__ (SVMs).
+#
+# Let's build this up a bit more. The MUTAG dataset we are considering contains not only graphs
+# corresponding to the structure of chemical compounds, but also a *classification* of each
```
mariaschuld:
```suggestion
# corresponding to the structure of chemical compounds, but also a *label* of each
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+# compound based upon its mutagenic effect. The four graphs we consider here have classifications:
```
mariaschuld:
```suggestion
# compound based upon its mutagenic effect. The four graphs we consider here have labels:
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+#
+# - MUTAG0: Class 1
+# - MUTAG1: Class 0
+# - MUTAG2: Class 0
+# - MUTAG3: Class 1
+
+classes = [1, 0, 0, 1]
+
+##############################################################################
+# Can we use GBS feature vectors to classify these graphs? We start by defining two-dimensional
```
mariaschuld:
```suggestion
# Can we use GBS feature vectors to determine the labels of these graphs? We start by defining two-dimensional
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+R = np.array([f1, f2, f3, f4])
+
+print(R)
+
+##############################################################################
+# The choice of ``events`` composing the feature vectors is arbitrary and we encourage the reader
+# to explore different combinations. Note, however, that odd photon-numbered events have zero
+# probability because ideal GBS only generates and outputs pairs of photons.
+#
+# Given our points in the feature space and corresponding classifications, we can use
```
mariaschuld:
```suggestion
# Given our points in the feature space and their target labels, we can use
```

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+# scikit-learn's `LinearSVC <https://scikit-learn.org/stable/modules/generated/sklearn.svm
```
mariaschuld:
```suggestion
# scikit-learn's Support Vector Machine `LinearSVC <https://scikit-learn.org/stable/modules/generated/sklearn.svm
```

examples_gbs/run_tutorial_similarity.py
mariaschuld: This is not clear, what is missing is that kernel methods use similarities between feature vectors. Maybe use something like this:

The power of feature vectors that embed graphs in a vector space of real numbers is that we can now measure similarities between graphs. This is very useful in machine learning, where similar labels are assigned to graphs that are close to each other. GBS feature vectors therefore give rise to a similarity measure between graphs!

A way to do this is to use the GBS feature vectors in a Support Vector Machine that finds a separating hyperplane between classes in the feature space...

examples_gbs/run_tutorial_similarity.py
```diff
@@ -0,0 +1,286 @@
...
+##############################################################################
+# .. raw:: html
+#     :file: ../../examples_gbs/SVM.html
+#
+# This plot shows the two classes (grey points for class 0 and red points for class 1)
+# successfully separated by the linear hyperplane using the GBS feature space. Moreover,
+# recall that the two MUTAG1 and MUTAG2 graphs of class 0 are actually isomorphic. Reassuringly,
+# their corresponding feature vectors are very similar. In fact, the feature vectors of
+# isomorphic graphs should always be identical :cite:`bradler2018graph` - the small discrepancy
+# in this plot is due to the statistical approximation from sampling.
```
mariaschuld: It would be quite interesting to see this for different numbers of samples/ precisions!

examples_gbs/run_tutorial_similarity.py
trbromley: @mariaschuld that's odd 🤔 Would you be able to share your log from the build?

examples_gbs/run_tutorial_similarity.py
trbromley: Thanks, have added this!

examples_gbs/run_tutorial_similarity.py
trbromley: Ah that's a good point! Have changed here, but kept "probability(ies)" later on because that's how we coded it up. I do wonder if we should normalise the vectors (resulting in frequencies rather than probabilities), but currently the vectors are directly the probabilities.

examples_gbs/run_tutorial_similarity.py
ixfoduap: Maybe 'arbitrary' is not the right word? How about "there is freedom in the choice of events composing the feature vectors. We encourage the reader to explore different combinations

examples_gbs/run_tutorial_similarity.py
```diff
@@ -129,8 +135,8 @@
 
 ##############################################################################
 # Now that we have mastered orbits and events, how can we make a feature vector? It was shown in
-# :cite:`schuld2019quantum` that one way of making a feature vector of a graph is to measure the
-# probabilities of events. Specifically, for a :math:`k` photon event :math:`E_{k, n_{\max}}`
+# :cite:`schuld2019quantum` that one way of making a feature vector of a graph is through the
+# frequencies of events. Specifically, for a :math:`k` photon event :math:`E_{k, n_{\max}}`
```
ixfoduap: Why this change? The way I see it, the feature vector is defined by the probabilities, which we can estimate through their sample frequency

trbromley: I think Maria was suggesting this because the vector could be of relative frequencies

examples_gbs/run_tutorial_similarity.py
trbromley: Fixed!

examples_gbs/run_tutorial_similarity.py
trbromley: Thanks that reads very nicely! Have merged it in to the tutorial

examples_gbs/run_tutorial_similarity.py
ixfoduap: Yes, I think it's best to explain. It can be a single short sentence, maybe linking to the URL you posted here

trbromley: Great, explained! Thanks

examples_gbs/run_tutorial_similarity.py
trbromley: To follow up - not sure what stops the plots being seen, but I did try in a clean virtual environment and things did work

examples_gbs/run_tutorial_similarity.py
trbromley: Thanks Maria, have added a quick example

examples_gbs/run_tutorial_similarity.py
trbromley: Thanks Maria! I think it would be a nice thing to visualise and also isn't too hard given that we already have the samples to play with.

On the other hand, I think it might not be so useful to non-experts and can take away from the main message

Issue #135, opened 2019-07-15
Add equality comparison for Programs


co9olguy:
This is a really cumbersome way of checking the ~equality~ equivalence of two programs --- furthermore, it doesn't take into account programs with the same DAG but different temporal operations list.

A useful feature would be to overload the `Program.__eq__(self, other)` special method, in order to perform the `nx.is_isomorphic` check (similarly to that done in `CircuitSpecs.compile`, but also taking into account parameter values).

That would allow us to simply do `assert prog1 == prog2` in the tests.

It would also help provide useful tracebacks, by indicating which gates and/or parameters differ

_Originally posted by @josh146 in https://github.com/XanaduAI/strawberryfields/pull/127_

josh146:
This has now been coded up in `tests/frontend/test_circuitspecs_chip0.py` (see below). The next steps are to:

1. Move this function over to `program_utils` and write tests

2. Write an additional function, `program_equality`, which is stricter, and returns `True` if and only if both programs contain the identical commands in the identical order (by looping through `cmd1, cmd2 in zip(prog1.circuit, prog2.circuit)`).

```python

def program_equivalence(prog1, prog2, compare_params=True, atol=1e-6, rtol=0):
    r"""Checks if two programs are equivalent.
    This function converts the program lists into directed acyclic graphs,
    and runs the NetworkX `is_isomorphic` graph function in order
    to determine if the two programs are equivalent.
    Note: when checking for parameter equality between two parameters
    :math:`a` and :math:`b`, we use the following formula:
    .. math:: |a - b| \leq (\texttt{atol} + \texttt{rtol}\times|b|)

    Args:
        prog1 (strawberryfields.program.Program): quantum program
        prog2 (strawberryfields.program.Program): quantum program
        compare_params (bool): Set to ``False`` to turn of comparing
            program parameters; equivalency will only take into
            account the operation order.
        atol (float): the absolute tolerance parameter for checking
            quantum operation parameter equality
        rtol (float): the relative tolerance parameter for checking
            quantum operation parameter equality

    Returns:
        bool: returns ``True`` if two quantum programs are equivalent
    """
    DAG1 = list_to_DAG(prog1.circuit)
    DAG2 = list_to_DAG(prog2.circuit)

    circuit = []
    for G in [DAG1, DAG2]:
        # relabel the DAG nodes to integers
        circuit.append(nx.convert_node_labels_to_integers(G))

        # add node attributes to store the operation name and parameters
[...]

PR #207, opened 2019-10-08
Refactor sample module

trbromley:
This PR refactors the `sample` module based upon our discussions.
Some highlights:

- Remove options dictionaries
- Add support for threshold measurement

Note: This PR only carries out minor changes to the `sample` module docstrings. While this should also have an overhaul, that is planned for a later PR.


tests/gbs/test_sample.py
```diff
@@ -58,309 +41,97 @@ def test_invalid_adjacency(self, dim):
...
+            sample.sample(A=adj, n_mean=1, threshold=False)
+            p_func = mock_eng_run.call_args[0][0]
 
-        samples_target = np.array([sample_pnr_postselect[0] for _ in range(sample_number)])
-        samples_target[samples_target >= 1] = 1
+        b = blackbird.dumps(sf.io.to_blackbird(p))
+        b_func = blackbird.dumps(sf.io.to_blackbird(p_func))
 
-        assert np.allclose(samples, samples_target)
+        assert b == b_func
```
zzzeid: This is a little out of scope, but it would be nice if there was a method that compares two programs for "equal syntax" that did not depend on `blackbird`. Just a minor comment.

Another thing you could do here to be more explicit is to just check the last command to make sure it is the right one, in this example, something like:

```python
assert isinstance(p_func.circuit[-1].op, sf.ops.MeasureThreshold)
```

In this case, you do not need to even have `p`.

josh146: Agree :+1:

Note that there is a related issue open for this: https://github.com/XanaduAI/strawberryfields/issues/135

strawberryfields/gbs/sample.py
```diff
@@ -335,7 +134,7 @@ def to_subgraphs(graph: nx.Graph, samples: list) -> list:
     graph_nodes = list(graph.nodes)
     node_number = len(graph_nodes)
 
-    subgraph_samples = [modes_from_counts(s) for s in samples]
+    subgraph_samples = [list(set(modes_from_counts(s))) for s in samples]
```
ixfoduap: Why is using list(set( ...)) necessary here?

strawberryfields/gbs/sample.py
```diff
@@ -371,3 +170,16 @@ def modes_from_counts(s: list) -> list:
     for i, c in enumerate(s):
         modes += [i] * c
     return sorted(modes)
+
+
+def postselect(samples: list, min_count: int) -> list:
```
ixfoduap: This should have a `min_count` and also a `max_count` to be as general as possible. For example, this would allow postselecting on a specific photon number

ixfoduap:
```suggestion
def postselect(samples: list, min_count: int, max_count: int) -> list:
```

strawberryfields/gbs/sample.py
```diff
@@ -371,3 +170,16 @@ def modes_from_counts(s: list) -> list:
...
+    """Postselect samples by imposing a minimum number of photons or clicks.
+
+    Args:
+        samples (list[list[int]]): a list of samples
+        min_count (int): minimum number of photons or clicks for a sample to be included
+
+    Returns:
+        list[list[int]]: the postselected samples
+    """
+    return [s for s in samples if sum(s) >= min_count]
```
ixfoduap:
```suggestion
    return [s for s in samples if min_count <= sum(s) <= max_count]
```

strawberryfields/gbs/sample.py
```diff
@@ -33,31 +32,22 @@
 samples are informative and can be used as a form of randomness to improve solvers, as outlined
 in Refs. :cite:`arrazola2018using` and :cite:`arrazola2018quantum`.
 
-On the other hand, the :func:`uniform` function allows users to generate samples where a
-subset of modes are selected using the uniform distribution. A utility function
-:func:`modes_from_counts` is also provided to convert between two ways to represent samples from
-the device.
+A utility function :func:`modes_from_counts` is also provided to convert between two ways to
```
soranjh: "An" utility?

tests/gbs/test_sample.py
```diff
@@ -463,3 +206,16 @@ def test_modes_from_counts():
     modes = [[], [0, 3, 3], [0, 1, 2], [0, 1, 1, 2], [1, 3, 3, 4, 4, 4, 4]]
 
     assert [sample.modes_from_counts(s) for s in counts] == modes
+
+
+def test_postselect():
```
ixfoduap: Test will need to be updated to reflect postselection on a min and max photon number

tests/gbs/test_sample.py
zzzeid: @josh146 good to know. For the purposes of this test though, it would be a little cleaner to check the last op, since it is the only thing that we are testing here, and not sure we need to compare the programs fully.

strawberryfields/gbs/sample.py
soranjh: Is the set necessary here?

strawberryfields/gbs/subgraph.py
```diff
@@ -192,21 +187,16 @@ def random_search(
...
+OPTIONS_DEFAULTS = {"heuristic": {"method": random_search}, "resize": RESIZE_DEFAULTS}
 """dict[str, dict[str, Any]]: Options for dense subgraph identification heuristics. Composed of a
 dictionary of dictionaries with the first level specifying the option type, selected from keys
-``"heuristic"``, ``"backend"``, ``"resize"``, and ``"sample"``, with the corresponding value a
+``"heuristic"`` and ``"resize"``, with the corresponding value a
 dictionary of options for that type.
 """
 
 
 def resize(
```
soranjh: Is the return of this function list[list[int]]?
I see this: return method(subgraphs=subgraphs, graph=graph, target=target, resize_options=resize_options)

strawberryfields/gbs/sample.py
trbromley: So `modes_from_counts` doesn't guarantee that that output doesn't have repetitions, e.g.:
`modes_from_counts([1, 0, 2, 1]) = [0, 2, 2, 3]`
However, `list(set(...))` will remove repetitions to make a subgraph:
`list(set([0, 2, 2, 3])) = [0, 2, 3]`

The idea here is that, even if the user has generated samples with PNRs, we apply an implicit thresholding here and they can generate subgraphs in the usual way.

trbromley: Yep see the comment above, `set` appears to have been controversial! 😆 
Generally I'm feeling we should have used `set` instead of `list` for our method of communicating subgraphs, but too late now!

strawberryfields/gbs/sample.py
trbromley: Nice! Have added these in!

strawberryfields/gbs/subgraph.py
trbromley: Yes, it is returning the output of `method` which is (or should be) `list[list[int]]`

tests/gbs/test_sample.py
trbromley: Thanks @zzzeid this is awesome! 😁

PR #209, opened 2019-10-09
Refactor resizing part of subgraph module

trbromley:
This PR refactors the resizing part of the `subgraph` module based upon our discussions.
Some highlights:

- `resize` is no longer a wrapper for other methods
- the underlying `greedy_dense` and `greedy_degree` methods have been removed
- `resize` now accepts a range of input sizes to resize to
- `resize` now returns a dictionary. keys are sizes, values are resultant resized subgraphs

Note: This PR only carries out minor changes to the sample module docstrings. While this should also have an overhaul, that is planned for a later PR.


strawberryfields/gbs/subgraph.py
```diff
@@ -205,181 +195,90 @@ def random_search(
...
-    subgraphs: Iterable, graph: nx.Graph, target: int, resize_options: Optional[dict] = None
-) -> list:
-    """Method to greedily resize subgraphs based upon density.
-
-    This function uses a greedy approach to iteratively add/remove nodes to an input subgraph.
-    Suppose the input subgraph has :math:`M` nodes. When shrinking, the algorithm considers
-    all :math:`M-1` node subgraphs of the input subgraph, and picks the one with the greatest
-    density. It then considers all :math:`M-2` node subgraphs of the chosen :math:`M-1` node
-    subgraph, and continues until the desired subgraph size has been reached.
+    When growth is required, the algorithm picks individual nodes from the remainder of the graph
```
soranjh: Will be good to mention how many nodes are examined to be added to the sub-graph in each addition iteration.

strawberryfields/gbs/subgraph.py
```diff
@@ -205,181 +195,90 @@ def random_search(
...
 
     Args:
-        subgraphs (iterable[list[int]]): an iterable over subgraphs, where each subgraph is given
-            by a list of nodes
+        subgraph (list[int]): a subgraph specified by a list of nodes
         graph (nx.Graph): the input graph
-        target (int): the target number of nodes in the subgraph (must be larger than 1)
-        resize_options (dict[str, Any]): dictionary specifying options used during resizing;
-            defaults to :const:`RESIZE_DEFAULTS`
+        sizes (list[int]): a list of desired sized for the subgraph to be resized to; defaults to
```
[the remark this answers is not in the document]
soranjh: Also, why not one size to be resized to? Is it common that the user wants to find subgraphs of multiple sizes?

strawberryfields/gbs/subgraph.py
```diff
@@ -183,193 +178,98 @@ def random_search(
...
 ``"heuristic"`` and ``"resize"``, with the corresponding value a
 dictionary of options for that type.
 """
 
 
-def resize(
-        subgraphs: Iterable, graph: nx.Graph, target: int, resize_options: Optional[dict] = None
-) -> list:
-    """Resize subgraphs to a given size.
+def resize(subgraph: list, graph: nx.Graph, sizes: Optional[list] = None) -> list:
```
ixfoduap: It's not clear at this stage what are valid forms of `sizes`. Is it an arbitrary list of integers from 0 to node size? For example, for a 30-node graph, can we have something like `sizes = [9, 6, 2, 27, 11]`?

strawberryfields/gbs/subgraph.py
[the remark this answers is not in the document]
ixfoduap: I don't think sizes should be optional. I agree with Soran that the default of all nonzero sizes can be very large and not particularly informative. I'd prefer demanding that the user specify the range of sizes

strawberryfields/gbs/subgraph.py
ixfoduap: If so, then for large graphs this may become cumbersome to specify if the goal is to find subgraphs within a range. Isn't it better to specify a `min_size` and a `max_size`?

strawberryfields/gbs/subgraph.py
```diff
@@ -183,193 +178,98 @@ def random_search(
...
-            - ``"greedy-degree"``: resize subgraphs by iteratively adding/removing nodes that
-              have the highest/lowest degree.
-            - *callable*: accepting ``(subgraphs: list, graph: nx.Graph,
-              target: int, resize_options: dict)`` as arguments and returning a ``list``
-              containing the resized subgraphs as lists of nodes, see :func:`greedy_density` for
-              an example signature
+    When shrinking is required, the algorithm picks individual nodes from within the subgraph to
+    remove, choosing the nodes with lowest degree relative to the subgraph. In both growth and
+    shrink phases, ties for addition/removal with nodes of equal degree are settled by uniform
+    random choice.
```
ixfoduap: Should we have Example usage? Also in other parts of the refactored subgraph and sample modules?

strawberryfields/gbs/subgraph.py
```diff
@@ -183,193 +178,98 @@ def random_search(
...
-    graph, and picks the one with the greatest density. It then considers all :math:`M+2` node
-    subgraphs given by combining the chosen :math:`M+1` node subgraph with nodes remaining from
-    the overall graph, and continues until the desired subgraph size has been reached.
+            degrees = [
+                (c, graph.subgraph(list(grow_nodes) + [c]).degree()[c]) for c in complement_nodes
+            ]
+            np.random.shuffle(degrees)
 
-    ``resize_options`` is not currently in use in :func:`greedy_density`.
+            to_add = max(degrees, key=lambda x: x[1])
```
ixfoduap: Is there no random breaking of ties as we do in `clique_shrink`?  Or is this done implicitly by `random.shuffle`?

strawberryfields/gbs/subgraph.py
trbromley: Great, thanks! Have reworded to hopefully be more clear.

strawberryfields/gbs/subgraph.py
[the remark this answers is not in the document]
trbromley:
```suggestion
        sizes (list[int]): a list of desired sizes for the subgraph to be resized to; defaults to
```
Thanks @soranjh, have made the above change.

The idea for multiple sizes is for us to switch from the more specific "densest k-subgraph" problem to a more general "cluster identification" problem where we are looking for dense subgraphs of various sizes.

For the question of all sizes being too large - good question. I think it's ok, as we are not expecting to be looking at very large graphs any time soon due to the constraints on hardware/simulation of GBS. We also have it as a controllable parameter, so if the user finds it too slow they can adjust.

tests/gbs/test_subgraph.py
```diff
@@ -108,195 +110,88 @@ def test_returns_densest(self, graph, monkeypatch):
...
+        connecting to the stick) is removed and a starting subgraph of nodes ``[2, 3, 4, 5,
+        6]`` is selected. The objective is to add-on 1 and 2 nodes and remove 1 node."""
+        dim = 6
+        g = nx.lollipop_graph(dim, 2)
+        g.remove_edge(0, dim - 1)
+
+        s = [2, 3, 4, 5, 6]
+        sizes = [4, 6, 7]
+
+        ideal = {4: [2, 3, 4, 5], 6: [1, 2, 3, 4, 5, 6], 7: [0, 1, 2, 3, 4, 5, 6]}
```
ixfoduap: For the lollipop graph, all nodes in the "candy" have the same degree, so when choosing nodes to add/remove I would expect there to be a random breaking of ties. Then why does this test always return the "ideal" subgraph? For example, I would expect that for size 6, either [1, 2, 3, 4, 5, 6] or [0, 2, 3, 4, 5, 6] could show up. Am I missing something?

tests/gbs/test_subgraph.py
```diff
@@ -108,195 +110,88 @@ def test_returns_densest(self, graph, monkeypatch):
...
+        equal to the offset."""
+        dim = 6
+        g = nx.complete_graph(dim)
+        s = [0, 1, 2, 3]
+
+        def permute(l, offset):
+            d = len(l)
+            ll = l.copy()
+            for i in range(d):
+                l[i] = ll[(i + offset) % d]
```
ixfoduap: Nice!

tests/gbs/test_subgraph.py
trbromley: The key is the removed edge between nodes 0 and 5. This makes node 0 have lower degree with respect to the subgraph, so we always pick node 1 to add in first.

When shrinking, we always should remove node 6 as it is a stick node.

Here's a plot:
image

ixfoduap: Missed the removed node! Smart :thinking: :lollipop:

strawberryfields/gbs/subgraph.py
trbromley: This has been updated in the following commit: https://github.com/XanaduAI/strawberryfields/pull/209/commits/4f0f0f9b9efcf5987661ba16836e629cb1efb481

strawberryfields/gbs/subgraph.py
```diff
@@ -183,193 +178,93 @@ def random_search(
...
+    nodes = graph.nodes()
+    subgraph = set(subgraph)
 
-    method = resize_options["method"]
+    if not subgraph.issubset(nodes):
+        raise ValueError("Input is not a valid subgraph")
+    if min_size < 1:
+        raise ValueError("min_size must be at least 1")
+    if max_size >= len(nodes):
+        raise ValueError("max_size must be less than number of nodes in graph")
```
ixfoduap:
```suggestion
        raise ValueError("max_size cannot be larger than the number of nodes in graph")
```

ixfoduap: Cause it can be equal, right?

trbromley: Currently picking a `max_size` to be equal to the total number of nodes in the graph _is_ prevented in the code.

The motivation is that I didn't see much point in providing that last resize step because it's always going to be the same, i.e. the whole graph itself.

But we can change this - what do you think?

PR #212, opened 2019-10-10
GBS Docs: Improve similarity docstrings

trbromley:
The `similarity` module docstrings was only one sentence. This PR expands this.
It also takes some of the content out of the function docstrings and puts it into module level docstring - for example we often tried to define an orbit or event in each function, and it makes sense to take this into a single place. The function docstrings are now relatively lightweight.


strawberryfields/gbs/similarity.py
```diff
@@ -52,9 +101,6 @@
 def sample_to_orbit(sample: list) -> list:
     """Provides the orbit corresponding to a given sample.
 
-    Orbits are simply a sorting of integer photon number samples in non-increasing order with the
-    zeros at the end removed.
-
```
trbromley: I found these lined unnecessary, especially now that we have defined properly the orbits and events in the module level docstrings.

strawberryfields/gbs/similarity.py
```diff
@@ -95,62 +138,17 @@ def sample_to_event(sample: list, max_count_per_mode: int) -> Union[int, None]:
...
-            a[k] = x
-            a[l] = y
-            yield sorted(a[: k + 2], reverse=True)
-            x += 1
-            y -= 1
-        a[k] = x + y
-        y = x + y - 1
-        yield sorted(a[: k + 1], reverse=True)
-
-
```
trbromley: This was just moved down below, as I felt the order reported in the doc's "code summary" was better like this.

strawberryfields/gbs/similarity.py
```diff
@@ -275,8 +303,8 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
 
 
 def prob_orbit_mc(graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int = 1000) -> float:
```
trbromley: Note to self - have checked up to here.

strawberryfields/gbs/similarity.py
```diff
@@ -19,10 +19,33 @@
...
+This module provides tools to measure how similar graphs are using GBS. Measuring similarity is
+achieved by mapping GBS samples from each graph to a feature vector and calculating the overlap
+between these vectors.
 
-Summary
--------
+The functionality here is based upon the research papers:
+:cite:`bradler2018graph,schuld2019quantum,bradler2019duality`.
+
+An accompanying tutorial can be found `here <??>`.
```
trbromley: To be updated!

strawberryfields/gbs/similarity.py
```diff
@@ -19,10 +19,33 @@
...
+
+Coarse-graining GBS samples
+---------------------------
+
+GBS feature vectors can be composed of the probabilities of certain events from sampling. These
```
ixfoduap: How about "GBS feature vectors can be composed of a coarse-grained combination of elementary samples." ?

strawberryfields/gbs/similarity.py
```diff
@@ -19,10 +19,33 @@
...
+events are typically a coarse-grained combination of elementary samples. We consider two coarse
+grainings:
+
+- **Orbits:** Combine all samples that can be made identical under permutation. Orbits are
+  written simply as a sorting of integer photon number samples in non-increasing order with the
+  zeros at the end removed. For example, ``[1, 0, 2, 0]`` and ``[2, 0, 0, 1]`` both belong to the
```
ixfoduap:
```suggestion
  zeros at the end removed. For example, ``[1, 1, 2, 0]`` and ``[2, 1, 0, 1]`` both belong to the
```

strawberryfields/gbs/similarity.py
```diff
@@ -19,10 +19,33 @@
...
+  ``[2, 1]`` orbit.
```
ixfoduap:
```suggestion
  ``[2, 1, 1]`` orbit.
```
^ I think that including a repetition is the clearest way of explaining orbits

strawberryfields/gbs/similarity.py
```diff
@@ -32,9 +55,53 @@
     orbits
     orbit_cardinality
     event_cardinality
+
+Creating a feature vector
+-------------------------
+
+A feature vector of a graph can be created by choosing a collection of orbits or events and
+evaluating their probabilities with respect to GBS sampling with the embedded graph. These
```
ixfoduap:
```suggestion
evaluating their probabilities with respect to GBS with the embedded graph. These
```

strawberryfields/gbs/similarity.py
```diff
@@ -32,9 +55,53 @@
...
+
+Creating a feature vector
+-------------------------
+
+A feature vector of a graph can be created by choosing a collection of orbits or events and
+evaluating their probabilities with respect to GBS sampling with the embedded graph. These
+probabilities are then selected to be elements of the feature vector. Evaluating the
+probabilities of orbits or events can be achieved through two approaches:
+
+- **Direct sampling:** infer the probability of orbits or events from a set of sample data.
```
ixfoduap: Do we want to explain mathematically precisely how this is done? Otherwise I feel it contrasts too much with the explanation for the Monte Carlo approach below

strawberryfields/gbs/similarity.py
```diff
@@ -32,9 +55,53 @@
...
+uniformly at random using :func:`orbit_to_sample` or :func:`event_to_sample`. Suppose :math:`N`
+samples :math:`\{S_{1}, S_{2}, \ldots , S_{N}\}` are generated. For each sample, this function
+calculates the probability :math:`p(S_i)` of observing that sample from a GBS device programmed
+according to the input graph and mean photon number. The sum of the probabilities is then
+rescaled according to the cardinality :math:`|E|` and the total number of samples:
+
+.. math::
+    p(E) \approx \frac{1}{N}\sum_{i=1}^N p(S_i) |E|,
+
+The sample mean of this sum is an estimate of the rescaled probability :math:`p(E)`.
```
ixfoduap: Nicely worded!

strawberryfields/gbs/similarity.py
```diff
@@ -32,9 +55,53 @@
...
     prob_orbit_mc
     prob_event_mc
+
+One canonical method of constructing a feature vector is to pick event probabilities
+:math:`p_{E_{k, n_{\max}}}` with all events :math:`E_{k, n_{\max}}` having a maximum photon count
+:math:`n_{\max}` in each mode. If :math:`\mathbf{k}` is the vector of selected events,
+the resultant feature vector is
+
+.. math::
+    f_{\mathbf{k}} = (p_{k_{1}}, p_{k_{2}}, \ldots)
```
ixfoduap: I think `p_{k_{i}}` is not yet defined? We should have something like `p_{k} := p_{E_{k, n_{\max}}}`, right?

strawberryfields/gbs/similarity.py
```diff
@@ -217,23 +228,61 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
 
     **Example usage:**
 
     >>> orbit_cardinality([2, 1, 1], 4)
     12
 
     Args:
-        orbit (list[int]): orbit to count number of samples
-        modes (int): number of modes in the sample
+        orbit (list[int]): orbit to count number of samples contained within
```
ixfoduap:
```suggestion
        orbit (list[int]): orbit; we count how many samples are contained in it
```

strawberryfields/gbs/similarity.py
```diff
@@ -217,23 +228,61 @@ def event_to_sample(photon_number: int, max_count_per_mode: int, modes: int) ->
...
+        modes (int): number of modes in counted samples
```
ixfoduap:
```suggestion
        modes (int): number of modes in the samples
```

strawberryfields/gbs/similarity.py
```diff
@@ -244,23 +293,20 @@ def orbit_cardinality(orbit: list, modes: int) -> int:
...
-    """Gives the number of samples belonging to the input event.
+    r"""Gives the number of samples belonging to the input event.
 
-    An event has a number of constituting samples, which are given by all orbits in the event,
-    then all permutations over the orbit. This function counts how many samples of the given number
-    of modes there are in a given event. For example, for three modes, there are six samples in an
-    event with two photons and a maximum of photons two per mode: [1,1,0], [1,0,1], [0,1,1],[2,0,
-    0],[0,2,0], and [0,0,2].
+    For example, for three modes, there are six samples in an :math:`E_{k=2, n_{\max}=2}` event:
+    [1, 1, 0], [1, 0, 1], [0, 1, 1], [2, 0, 0], [0, 2, 0], and [0, 0, 2].
```
ixfoduap: Nice!

strawberryfields/gbs/similarity.py
```diff
@@ -275,18 +321,11 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
 
 
 def prob_orbit_mc(graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int = 1000) -> float:
-    """Gives a Monte Carlo estimate of the probability of a given orbit for a GBS device encoded
-    according to the input graph.
+    r"""Gives an MC estimate of the GBS probability of a given orbit according to the input graph.
```
ixfoduap: I prefer to spell out Monte Carlo, as it's dangerous to assume that a user will always remember what MC stands for. Personally, it makes me think of MC Hammer

strawberryfields/gbs/similarity.py
```diff
@@ -392,24 +427,17 @@ def feature_vector_sampling(
...
     **Example usage:**
 
-    >>> sample.random_seed(1967)
-    >>> adj = np.ones((4, 4))
-    >>> samples = sample.quantum_sampler(adj, 6, 10, backend_options={"threshold": False})
+    >>> from strawberryfields.gbs import data
+    >>> samples = data.Mutag0()
     >>> feature_vector_sampling(samples, [2, 4, 6])
-    [0.1, 0.2, 0.0]
+    [0.19035, 0.2047, 0.1539]
```
ixfoduap: Nice!

strawberryfields/gbs/similarity.py
```diff
@@ -441,16 +469,11 @@ def feature_vector_mc(
...
     samples: int = 1000,
 ) -> list:
-    r"""Calculates feature vector using a Monte Carlo estimation of event probabilities.
-
-    The feature vector is composed of event probabilities :math:`p_{k}` with all events
-    :math:`E_{k}` having a maximum photon count in each mode of ``max_count_per_mode``. Events
-    are specified by their total photon number :math:`k` and those chosen as part of the feature
-    vector can be specified through :math:`\mathbf{k}` (using the ``event_photon_numbers``
-    argument). The resultant feature vector is
+    r"""Calculates feature vector using MC estimation of event probabilities according to the
```
ixfoduap:
```suggestion
    r"""Calculates feature vector using Monte Carlo estimation of event probabilities according to the
```

strawberryfields/gbs/similarity.py
trbromley: Good spot! Have fixed this

strawberryfields/gbs/similarity.py
trbromley: Nice! This is what I added:

```rst
In the direct sampling approach, :math:`N` samples are taken from GBS with the embedded graph and
the number that fall within a given orbit or event :math:`E` are counted, resulting in the count
:math:`c_{E}`. The probability :math:`p(E)` of an orbit or event is then approximated as:

.. math::
    p(E) \approx \frac{c_{E}}{N}
```

ixfoduap: Looks good! Maybe replace "the embedded graph and the number that..." with "the embedded graph. The number that..." ?

trbromley: Thanks! Done

PR #215, opened 2019-10-11
GBS Tutorial: Sampling

trbromley:
This PR introduces a sphinx-gallery based tutorial for the `sample` module of the `gbs` applications layer.


examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# meaning that 2 photons were detected in mode 0, 1 photons were detected in modes 1 and 2,
+# and 0 photons were detected in modes 3 and 4. If threshold detectors were used instead,
+# the sample would be: ``[1, 1, 1, 0, 0]``.
+#
+# Sampling subgraphs
+# ------------------
+#
+# So when would threshold detection or PNR detection be preferred in GBS? Since threshold samples
+# can be post-processed from PNR samples, we might expect that PNR detection is always the
+# preferred choice. However, *simulating* PNR-based GBS is significantly slower, and it turns out
```
nquesada: Maybe I'd say, "However, in practice *simulating*..."

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# .. raw:: html
+#     :file: ../../examples_gbs/subgraph.html
+#
+# These sampled subgraphs act as the starting point for some of the applications made available
+# in Strawberry Fields, including the maximum clique and dense subgraph identification problems.
+# Go and check out the other tutorials in the applications layer to see what you can do with GBS!
+#
+# .. note::
+#       Simulating GBS can be computationally intensive when using both threshold and PNR
+#       detectors. After all, it is a quantum algorithm! To help users get to grips with the
```
nquesada: Maybe: "After all, we are using a classical algorithm to simulate a quantum process! ..."

doc/gbs_tutorials.rst
```diff
@@ -2,6 +2,6 @@ Applications layer tutorials
 ============================
 
 .. customgalleryitem::
-    :tooltip: To be filled
-    :description: :doc:`tutorials_gbs/run_tutorial_example`
-    :figure: ./_static/planted.png
\ No newline at end of file
+    :tooltip: Sampling
+    :description: :doc:`tutorials_gbs/run_tutorial_sample`
+    :figure: ./_static/thumbs/code.png
```
ixfoduap: The code.png is a placeholder while we decide on a better image, right?

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
+# pylint: disable=invalid-name,no-member,wrong-import-position,wrong-import-order,ungrouped-imports
+"""
+Sampling Tutorial
+=================
+
+This tutorial provides a quick guide to generating samples from GBS. This functionality is
```
ixfoduap: Should we mention explicitly that this employs simulators? e.g. "... generating samples from GBS simulators?"

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+provided in the :mod:`~.gbs.sample` module of the applications layer.
+
+Sampling from GBS
+-----------------
+
+A GBS device can be programmed to sample from any symmetric matrix :math:`A`. To sample,
+we must specify both the mean number of photons being generated in the device and the form of
+detection used at the output: threshold detection or photon-number resolving (PNR) detection.
+Threshold detectors are restricted to measuring whether or not one or more photons have arrived
```
ixfoduap:
```suggestion
Threshold detectors are restricted to measuring whether one or more photons have arrived
```

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+at the detector, whereas PNR detectors are able to count the number of photons.
+
+Let's take a look at both types of sample. We can generate samples from a random 5-dimensional
```
ixfoduap:
```suggestion
Let's take a look at both types of sampling methods. We can generate samples from a random 5-dimensional
```

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+matrix:
```
ixfoduap:
```suggestion
symmetric matrix:
```

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+
+from strawberryfields.gbs import sample
+import numpy as np
+
+modes = 5
+n_mean = 4
+samples = 5
+
+A = np.random.random((modes, modes))
```
ixfoduap:
```suggestion
A = np.random.normal(0, 1, (modes, modes))
```
I think this is a good opportunity to include matrices with negative entries and negative eigenvalues

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+A = A + A.T
+
+s_thresh = sample.sample(A, n_mean, samples, threshold=True)
+s_pnr = sample.sample(A, n_mean, samples, threshold=False)
+
+print(s_thresh)
+print(s_pnr)
```
ixfoduap: This is a bit dangerous in that it's likely no collisions are seen in the PNR samples, therefore not showing the difference in the types of samples obtained. At the same time, not sure what we can do about it

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+
+##############################################################################
+# In each case, a sample is a sequence of integers of ``len(modes) = 5``. Threshold samples are
```
ixfoduap: I'm not sure how to read "of `len(modes) = 5" in English. Maybe " In each case, a sample is a sequence of integers of length five, i.e., ``len(modes) = 5`` "?

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# ``0``'s and ``1``'s, corresponding to whether or not photons were detected in a mode. A ``1``
+# here is conventionally called a "click". PNR samples are non-negative integers counting the number
+# of photons detected in each mode. For example, suppose a PNR sample is ``[2, 1, 1, 0, 0]``,
+# meaning that 2 photons were detected in mode 0, 1 photons were detected in modes 1 and 2,
+# and 0 photons were detected in modes 3 and 4. If threshold detectors were used instead,
+# the sample would be: ``[1, 1, 1, 0, 0]``.
```
ixfoduap: Really good explanation!

examples_gbs/run_tutorial_sample.py
ixfoduap: ^agree with Nico. It's not obvious at this point in the tutorial why threshold sampling is faster, so we can just state that this is the case in practice

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# Sampling subgraphs
+# ------------------
+#
+# So when would threshold detection or PNR detection be preferred in GBS? Since threshold samples
+# can be post-processed from PNR samples, we might expect that PNR detection is always the
+# preferred choice. However, *simulating* PNR-based GBS is significantly slower, and it turns out
+# that threshold samples can provide enough useful information for a range of applications.
+#
+# The applications layer of Strawberry Fields focuses primarily on solving graph-based problems.
+# In this setting, we typically want to use GBS to sample subgraphs of a graph, which are likely
```
ixfoduap:
```suggestion
# In this setting, we typically want to use GBS to sample subgraphs, which are likely
```
Subgraphs can only be subgraphs of a graph :wink:

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# to be dense due to the probability distribution of GBS :cite:`arrazola2018using`. In this case,
+# threshold sampling is enough, since it lets us select nodes of the subgraph. Let's take more of
```
ixfoduap:
```suggestion
# threshold sampling is enough, since it lets us select nodes of the subgraph. Let's take
```

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# a look at this by using a small fixed graph as an example:
+
+from strawberryfields.gbs import plot
+import networkx as nx
+import plotly
+
+adj = np.array(
```
ixfoduap: Where does this graph come from? How is it generated?

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+
+graph = nx.Graph(adj)
+
+plotly.offline.plot(plot.plot_graph(graph), filename="random_graph.html")
+
+##############################################################################
+# .. raw:: html
+#     :file: ../../examples_gbs/random_graph.html
+#
+# We can sample this graph from GBS using:
```
ixfoduap: This sentence reads a bit strange to me. Also, do we want to motivate the mean photon number and number of samples?

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+samples = 20
+
+s = sample.sample(adj, n_mean, samples)
+
+print(s[:5])
+
+##############################################################################
+# Each sample in ``s`` is a list of modes with ``1``'s for nodes that have clicked and ``0``'s
+# for nodes that haven't. We want to convert a sample to another representation where the result
+# is a list of modes that have clicked. This list of modes can be used to select a subgraph of
```
ixfoduap:
```suggestion
# is a list of modes that have clicked. This list of modes can be used to select a subgraph.
```

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# the graph. For example, if ``[0, 1, 0, 1, 1, 0]`` is a sample from GBS then ``[1, 3, 4]`` are
```
ixfoduap:
```suggestion
# For example, if ``[0, 1, 0, 1, 1, 0]`` is a sample from GBS then ``[1, 3, 4]`` are
```

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+# the selected nodes of the corresponding subgraph.
+#
+# However, the number of clicks in GBS is a random variable and we are not always guaranteed to
+# have enough clicks in a sample for the resultant subgraph to be of interest. We can filter out
+# the uninteresting samples using:
```
ixfoduap: This applies to line 84 above as well --> Do we want to say "using:" or "using the `postselect` function"

examples_gbs/run_tutorial_sample.py
```diff
@@ -0,0 +1,140 @@
...
+min_clicks = 3
+max_clicks = 4
+
+s = sample.postselect(s, min_clicks, max_clicks)
+
+print(len(s))
+s.append([0, 1, 0, 1, 1, 0])
+
+##############################################################################
+# As expected, we can see that we have lost some samples. The number of samples that survive this
```
ixfoduap:
```suggestion
# As expected, we have fewer samples than before. The number of samples that survive this
```

doc/gbs_tutorials.rst
ixfoduap: Hmmm, I'd rather have a colourful, eye-catching thumbnail, like in the PL tutorials. How about this picture of one of the chips? 
chip

examples_gbs/run_tutorial_sample.py
ixfoduap: Yes, maybe increasing it to 6?

examples_gbs/run_tutorial_sample.py
ixfoduap: I think it just merits some sort of description of the structure of the graph

doc/gbs_tutorials.rst
ixfoduap: I got this from Zach, but there are official pictures from the Xanadu style guide. We could potentially crop and compress if the file is too large

trbromley: The chip is also good though, e.g.,
Xanadu-18

trbromley: Decided:
image

nquesada: That image is highly misleading. It is a single photon giving the impression that you are doing standard Boson Sampling.

trbromley: 🤔 Ah the idea was more of counting photons at the output.
Any suggestions on a good image?

nquesada: You could put a Gaussian instead, but it is not nearly as "pretty". When you say counting photons you mean threshold detectors, right?

PR #216, opened 2019-10-14
GBS Refactor: New search function for dense subgraphs

trbromley:
This updates the `search` function in the `subgraph` module. Now:

- The function accepts samples as an input rather than options dictionaries to specify how sampling should happen
- The user can input a _range_ of subgraph sizes to search over and the function outputs dense subgraphs found in this range
- The output is a dictionary of different subgraph sizes, with values being a sorted list of those subgraphs.

This PR wraps up our plan to refactor the dense subgraph module, following from: https://github.com/XanaduAI/strawberryfields/pull/207 and https://github.com/XanaduAI/strawberryfields/pull/209.

### DIFF
--- a/strawberryfields/gbs/subgraph.py
+++ b/strawberryfields/gbs/subgraph.py
@@ -150,7 +150,7 @@ def _update_subgraphs_list(l: list, t: tuple, max_count: int) -> None:
     """
     t = (t[0], sorted(set(t[1])))
 
-    for d, s in l:
+    for _d, s in l:
         if t[1] == s:
             return None
 
@@ -170,6 +170,8 @@ def _update_subgraphs_list(l: list, t: tuple, max_count: int) -> None:
             del l[-1]
             l.append(t)
             l.sort(reverse=True)
+
+    return None
 
 
 def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dict:
--- a/strawberryfields/gbs/subgraph.py
+++ b/strawberryfields/gbs/subgraph.py
@@ -73,27 +73,47 @@ def search(
     for s in subgraphs:
         s = set(s)
         if not s.issubset(nodes):
-            print("List of subgraphs contains an invalid subgraph, continuing")
             continue
 
         r = resize(s, graph, min_size, max_size)
 
-        for size in size_range:
-            current = dense.get(size, default=[])
-
-            candidate = r.get(size)
-            sub = graph.subgraph(candidate)
-            sub_dens = nx.density(sub)
-
-            if len(current) < top_count:
-                current.append((sub_dens, sub))
-                current.sort()
-            elif sub_dens > min(current):
-                current.append((sub_dens, sub))
-                current.sort()
-                del current[-1]
+        _combine_dicts(dense, r, graph, top_count)
 
     return dense
+
+
+def _combine_dicts(main: dict, new: dict, graph: nx.Graph, top_count: int) -> None:
+
+    for size, candidate in new.items():
+        current = main.get(size)
+
+        sub = graph.subgraph(candidate)
+        sub_dens = nx.density(sub)
+
+        if current is None:
+            current = [(sub_dens, candidate)]
+            main[size] = current
+        else:
+            _add_to_list(current, candidate, sub_dens)
+
+
+def _add_candidate_subgraph(l: list, t: tuple, max_count: int) ->
+
+
+def _add_to_list(l: list, subgraph: list, density: float, top_count: int) -> None:
+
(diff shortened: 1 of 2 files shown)


tests/gbs/test_subgraph.py
```diff
@@ -14,113 +14,239 @@
 r"""
 Unit tests for strawberryfields.gbs.subgraph
 """
-# pylint: disable=no-self-use,unused-argument
+# pylint: disable=no-self-use,unused-argument,protected-access,redefined-outer-name
```
trbromley: - `protected-access` is needed to test those functions (ones with a leading underscore)
- `redefined-outer-name` is needed because we use the `process_planted` fixture

josh146: Note that it might be worth adding a comment, i.e.
```python
# needed to test private functions pylint: disable=protected-access
# needed because we use the process_planted fixture pylint: disable=redefined-outer-name
```
This way it's in the source code in case a developer in future is wondering why they are required
(I _think_ that's the syntax for including comments, but it could be some permutation on that).

trbromley: @josh146 agreed it makes sense to document in code the justification for pylint disables!

Although, it seems like there are some "well-established" disable settings specifically for testing modules, e.g. `protected-access`, `no-self-use` etc. Do you think we should comment for them also?

josh146: > Do you think we should comment for them also?

Nope, just the ones you mentioned above. It should be fine from context why `no-self-use` etc. is used

tests/gbs/test_subgraph.py
```diff
@@ -14,113 +14,239 @@
...
-        [0, 1, 2, 3],
-        [0, 1, 4, 5],
-    ]
-)
+
+@pytest.fixture()
+def process_planted(min_size, max_size, max_count, n_samples):
+    """Fixture for loading samples from the Planted dataset"""
+    samples = p_planted[:n_samples]
+    samples.insert(10, [31])  # include invalid subgraph to test that function simply continues
```
ixfoduap: I don't understand the role of this insert. Is it a test within a fixture? What happens if you remove this line?

tests/gbs/test_subgraph.py
```diff
@@ -14,113 +14,239 @@
...
+                    )
+                )
+                for l in process_planted.values()
+                for t in l
+            ]
+        )
 
-        result = subgraph.search(
-            graph=adj, nodes=3, iterations=10, options={"heuristic": {"method": custom_method}}
+    def test_density(self, process_planted):
```
ixfoduap: Great test!

strawberryfields/gbs/subgraph.py
```diff
@@ -65,125 +41,137 @@
...
+    over a range using :func:`resize`. We want to add these candidates to the list of subgraph
+    tuples in ``d`` to build up our collection of dense subgraphs.
 
     Args:
-        graph (nx.Graph): the input graph
-        nodes (int): the size of desired dense subgraph
-        iterations (int): number of iterations to use in algorithm
-        options (dict[str, dict[str, Any]]): dict-of-dicts specifying options in different parts
-            of heuristic search algorithm; defaults to :const:`OPTIONS_DEFAULTS`
+        d (dict[int, list[tuple[float, list[int]]]]): dictionary of subgraph sizes and
```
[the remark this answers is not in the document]
trbromley: Ah, I _think_ I prefer it the other way where we have a `[` in line 66 - have made that change.

strawberryfields/gbs/subgraph.py
```diff
@@ -65,125 +41,137 @@
...
+            corresponding list of subgraph tuples
+        d_new (dict[int, tuple[float, list[int]]]): dictionary of subgraph sizes and corresponding
```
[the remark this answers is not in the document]
trbromley: Same as the comment above, I think I prefer `[` rather than `(`
Also not sure if there is a rule or way to decide.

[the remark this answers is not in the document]
trbromley: image

E.g., when type hinting in python we use the `[`

tests/gbs/test_subgraph.py
trbromley: Done 😄

tests/gbs/test_subgraph.py
trbromley: Good catch! The function was set to continue if an invalid subgraph was presented (i.e., one with nodes not in the actual subgraph, such as the one above). In the end I decided this was an unnecessary complication and took this out.

PR #220, opened 2019-10-16
Apps: Dense subgraph tutorial

ixfoduap:
This PR adds a new tutorial focused on dense subgraph identification. It focuses on the planted subgraph dataset and explores how to use our software to find dense subgraphs of different sizes.


doc/gbs_tutorials.rst
```diff
@@ -4,4 +4,9 @@ Applications layer tutorials
...
     :tooltip: Max Clique
     :description: :doc:`tutorials_gbs/run_tutorial_max_clique`
-    :figure: ./_static/max_clique.png
\ No newline at end of file
+    :figure: ./_static/max_clique.png
+
+.. customgalleryitem::
+    :tooltip: Dense subgraphs
+    :description: :doc:`tutorials_gbs/run_tutorial_dense`
+    :figure: ./_static/dense.png
```
trbromley: Don't forget to add the image file itself

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports
+"""
+Dense Subgraph Tutorial
+=========================
```
trbromley:
```suggestion
=======================

```

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports
+"""
+Dense Subgraph Tutorial
+=========================
+Graphs can be used to model a wide variety of objects: social networks, financial markets,
```
trbromley:
```suggestion
Graphs can be used to model a wide variety of concepts: social networks, financial markets,
```
Wasn't sure if a social network was an object 🤔

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
+# pylint: disable=wrong-import-position,wrong-import-order,ungrouped-imports
+"""
+Dense Subgraph Tutorial
+=========================
+Graphs can be used to model a wide variety of objects: social networks, financial markets,
+biological networks, and many others. A common problem of interest is to find subgraphs that
+contain a large number of connections between its nodes. These may correspond to communities in
```
trbromley:
```suggestion
contain a large number of connections between their nodes. These subgraphs may correspond to communities in
```

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+Dense Subgraph Tutorial
+=========================
+Graphs can be used to model a wide variety of objects: social networks, financial markets,
+biological networks, and many others. A common problem of interest is to find subgraphs that
+contain a large number of connections between its nodes. These may correspond to communities in
+social networks, correlated assets in a market, or mutually influential proteins in a biological
+network. This tutorial studies how GBS can be used to identify dense subgraphs.
+
+Mathematically, this task is known as the `dense subgraph problem
+<https://en.wikipedia.org/wiki/Dense_subgraph>`_. The density of a graph is equal to the number of
```
trbromley: @josh146, I still never got my head around whether we should have a single or double underscore here?

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+edges in the subgraph divided by the maximum possible number of edges. Identifying the densest
```
trbromley: How about "The density of a :math:`k`-node subgraph is equal to the number of its edges divided by the maximum possible number of edges :math:`n(n-1)/2`.

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+into a GBS device, it samples dense subgraphs with high probability. This property can be
+used to find dense subgraphs by sampling from a GBS device  and postprocessing the outputs.
+Let's take a look!
+
+Finding dense subgraphs
+-----------------------
+As usual, the first step is to import all required modules. We'll need the :mod:`~.gbs.data`
+module to load pre-generated samples, the :mod:`~.gbs.sample` module to postselect samples, the 
+:mod:`~.gbs.subgraph` module to search for dense subgraphs, and the :mod:`~.gbs.plot` module to 
+visualize the graphs. We'll also use Plotly for plotting and NetworkX for graph
```
trbromley: We built a whole ecosystem 😁

trbromley: > We'll also use Plotly for plotting

Does it raise the question of why import the `plot` module?

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+from strawberryfields.gbs import data, sample, subgraph, plot
+import plotly
+import networkx as nx
+
+##############################################################################
+# In this tutorial, we'll study a 30-node graph with a planted 10-node graph, as considered in
+# :cite:`arrazola2018using`. The graph is generated by joining two Erdős–Rényi random graphs. The
+# first graph of 20 nodes is created with edge probability of 0.5. The second planted
+# graph is generated with edge probability of 0.875. The planted nodes are the last ten nodes in the
+# graph. The two graphs are joined by selecting 8 vertices at random from both graphs and adding an
```
trbromley:
```suggestion
# graph. The two graphs are joined by selecting 8 nodes at random from both graphs and adding an
```
This was never officially decided, but should we stick with "nodes" over "vertices"?

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+# graph.
+#
+# The :mod:`~.gbs.data` module has pre-generated GBS samples from this graph. Let's load them,
+# postselect on samples with a large number of clicks, and convert them to subgraphs:
+
+planted = data.Planted()
+postselected = sample.postselect(planted, 16, 30)
+pl_graph = nx.to_networkx_graph(planted.adj)
+samples = sample.to_subgraphs(postselected, pl_graph)
+print(len(samples))
```
trbromley: Should we print
```python
len(planted)
```
as a comparison on how many we threw away?

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+#     :file: ../../examples_gbs/planted.html
+#
+# .. note::
+#     The command ``plotly.offline.plot()`` is used to display plots in the documentation. In
+#     practice, you can simply use ``plot_graph.show()`` to view the graph.
+
+##############################################################################
+# A more interesting challenge is to find dense subgraphs of different sizes; it is often
+# useful to identify many high-density subgraphs, not just the densest ones. This is the purpose of
+# the :func:`~.search` function in the :mod:`~gbs.subgraph` module: to identify collections of dense
```
trbromley:
```suggestion
# the :func:`~.subgraph.search` function in the :mod:`~.gbs.subgraph` module: to identify collections of dense
```

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+# subgraphs for a range of sizes. The output of this function is a dictionary whose keys correspond
+# to subgraph sizes within the specified range. The items in the dictionary are the top subgraphs of
```
trbromley:
```suggestion
# to subgraph sizes within the specified range. The values in the dictionary are the top subgraphs of
```
I think items could refer to dictionary-value pairs: https://www.geeksforgeeks.org/python-dictionary-items-method/

examples_gbs/run_tutorial_dense.py
```diff
@@ -0,0 +1,120 @@
...
+for k in range(8, 17):
+    print(dense[k][0])  # print only the densest subgraph of each size
+
+##############################################################################
+# From the results of the search we learn that, depending on their size, the densest subgraphs
+# belong to different regions of the graph: dense subgraphs of less than ten nodes are contained
+# within the planted subgraph, whereas larger dense subgraphs appear outside of the planted
+# subgraph. Smaller dense subgraphs can be cliques, characterized by having
+# maximum density of 1, while larger subgraphs are less dense. Let's see what the smallest and
+# largest subgraphs look like:
```
trbromley: Very cool!!!

examples_gbs/run_tutorial_dense.py
ixfoduap: Not sure if a social network is a concept :thinking: 
^joking btw

examples_gbs/run_tutorial_dense.py
ixfoduap: Me too! What's the mnemonic so we never forget?

examples_gbs/run_tutorial_dense.py
ixfoduap: Yes, let's stick with nodes

examples_gbs/run_tutorial_dense.py
ixfoduap: Nah, that information is a bit distracting and those that are curious can find it in the docs

examples_gbs/run_tutorial_dense.py
ixfoduap: :grin: I think this fixes the "issue" of the plot being able to identify the planted subgraph

examples_gbs/run_tutorial_dense.py
ixfoduap: Changed to "We'll also use Plotly which is required for the :mod:`~.gbs.plot` module"

examples_gbs/run_tutorial_dense.py
trbromley: Here are some answers:
https://stackoverflow.com/questions/27420317/restructured-text-rst-http-links-underscore-vs-use
http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#anonymous-hyperlinks

So a double underscore is for a "one off" link, whereas a single allows for that link to be used again, e.g. from the second reference above:

```rst
See the `Python home page <http://www.python.org>`_ for info.
This `link <Python home page_>`_ is an alias to the link above.
```

So I guess it's kind of like whether we add a `\label` in latex.

In our case we would do
```rst
Mathematically, this task is known as the `dense subgraph problem <https://en.wikipedia.org/wiki/Dense_subgraph>`_.
Another link to the `dense subgraph problem <dense subgraph problem_>`
```

Which is actually ok, although I'm not sure if we'd ever remember when working in another document that we've already provided the link. So for now have removed.

examples_gbs/run_tutorial_dense.py
trbromley: Actually maybe we should not mention the import of `plotly` at all, since the user shouldn't need it in general _unless_ they want to use `plotly.offline.plot()`, which we say shouldn't typically be the case.

PR #221, opened 2019-10-17
GBS Docs: Improving max clique docs

trbromley:
Polishing and expanding docstrings for the `clique` module.


strawberryfields/gbs/clique.py
```diff
@@ -22,13 +22,17 @@
...
 where all nodes are connected to each other. The maximum clique problem is to identify the
 largest clique in a graph. The problem is NP-Hard in a worst-case setting, which is why it is
-valuable to develop better algorithms to tackle it.
+valuable to develop better algorithms to tackle it. It has been shown that samples from GBS can
+be used to select dense subgraphs as a starting seed for heuristic algorithms
+:cite:`banchi2019molecular`.
+
+An accompanying tutorial can be found :ref:`here <gbs-clique-tutorial>`.
 
 Heuristics
```
trbromley:
```suggestion
Heuristic
```

ixfoduap:
```suggestion
Algorithm
```
I think it's not so useful to emphasize that it's a "heuristic" algorithm, it is arguably one of the best existing algorithms for max clique!

strawberryfields/gbs/clique.py
```diff
@@ -41,31 +45,39 @@
...
-    shrink
+    c_1
 
-Clique utility functions
-------------------------
+Using GBS to find a starting clique
+-----------------------------------
 
-The following are utility functions for calculating properties of cliques.
+Samples from GBS correspond to subgraphs that are likely to be dense. Nevertheless,
```
ixfoduap:
```suggestion
Samples from GBS correspond to subgraphs that are likely to be dense.
```

strawberryfields/gbs/clique.py
```diff
@@ -41,31 +45,39 @@
...
+these subgraphs may not be cliques, which is the required input to the :func:`search` algorithm.
```
ixfoduap:
```suggestion
These subgraphs may not be cliques, which is the required input to the :func:`search` algorithm.
```

PR #223, opened 2019-10-21
GBS: Add ability to simulate losses in GBS

trbromley:
Since realistic GBS devices will have losses, it makes sense to provide this in the applications layer. Since loss is already part of SF, we only need to change the program fed to the SF Engine, as well as updating tests and docs.

### DIFF
--- a/strawberryfields/gbs/sample.py
+++ b/strawberryfields/gbs/sample.py
@@ -29,8 +29,8 @@ Generating samples
 An :math:`M` mode GBS device can be programmed by specifying an :math:`(M \times M)`-dimensional
 symmetric matrix :math:`A` :cite:`bradler2018gaussian`. Running this device results in samples
 that carry relevant information about the encoded matrix :math:`A`. When sampling, one must also
-specify the mean number of photons in the device and the form of detection used at the output:
-threshold detection or photon-number-resolving (PNR) detection.
+specify the mean number of photons in the device, the form of detection used at the output:
+threshold detection or photon-number-resolving (PNR) detection, as well as the amount of loss.
 
 The :func:`sample` function provides a simulation of sampling from GBS:
 
@@ -41,7 +41,10 @@ Here, each output sample is an :math:`M`-dimensional list. If threshold detectio
 (``threshold = True``), each element of a sample is either a zero (denoting no photons detected)
 or a one (denoting one or more photons detected), conventionally called a "click".
 If photon-number resolving (PNR) detection is used (``threshold = False``) then elements of a
-sample are non-negative integers counting the number of photons detected in each mode.
+sample are non-negative integers counting the number of photons detected in each mode. The ``loss``
+parameter allows for simulation of photon loss in the device, which is the most common form of
+noise in quantum photonics. Here, ``loss = 0`` describes ideal loss-free GBS, while generally
+``0 <= loss <= 1`` describes the proportion of photons lost while passing through the device.
 
 Samples can be postselected based upon their total number of photons or clicks and the ``numpy``
 random seed used to generate samples can be fixed:
--- a/strawberryfields/gbs/sample.py
+++ b/strawberryfields/gbs/sample.py
@@ -108,7 +108,7 @@ import strawberryfields as sf
 
 
 def sample(
-    A: np.ndarray, n_mean: float, n_samples: int = 1, threshold: bool = True, loss: float = 0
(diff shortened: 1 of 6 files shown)


strawberryfields/gbs/sample.py
```diff
@@ -142,6 +151,10 @@ def sample(A: np.ndarray, n_mean: float, n_samples: int = 1, threshold: bool = T
     with p.context as q:
         sf.ops.GraphEmbed(A, mean_photon_per_mode=mean_photon_per_mode) | q
 
+        if loss:
```
nquesada: seems like loss is a Boolean here but a float elsewhere.

nquesada: By the way, applying loss in the Gaussian backend is very cheap so you loose very little by always applying it.

ixfoduap: Thanks Nico! Actually, `if` evaluates to `True` for any non-zero float, and is `False` if the float is zero

strawberryfields/gbs/similarity.py
```diff
@@ -345,6 +351,7 @@ def prob_orbit_mc(graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int
         orbit (list[int]): orbit for which to estimate the probability
         n_mean (float): total mean photon number of the GBS device
         samples (int): number of samples used in the Monte Carlo estimation
+        loss (float): fraction of photons lost in GBS
```
ixfoduap: This function probably needs to raise an error if the loss is not between 0 and 1

strawberryfields/gbs/similarity.py
```diff
@@ -400,6 +413,7 @@ def prob_event_mc(
         max_count_per_mode (int): maximum number of photons per mode in the event
         n_mean (float): total mean photon number of the GBS device
         samples (int): number of samples used in the Monte Carlo estimation
+        loss (float): fraction of photons lost in GBS
```
ixfoduap: Same here, function probably needs to raise an error if the loss is not between 0 and 1

strawberryfields/gbs/similarity.py
```diff
@@ -509,6 +529,6 @@ def feature_vector_mc(
         raise ValueError("Maximum number of photons per mode must be at least one")
```
ixfoduap: Same here, raise an error if the loss is not between 0 and 1

tests/gbs/test_sample.py
```diff
@@ -74,6 +80,46 @@ def test_pnr(self, monkeypatch, adj):
...
+        mock_eng_run = mock.MagicMock()
+
+        with monkeypatch.context() as m:
+            m.setattr(sf.Engine, "run", mock_eng_run)
+            sample.sample(A=adj, n_mean=1, threshold=False)
+            p_func = mock_eng_run.call_args[0][0]
+
+        assert not all([isinstance(op, sf.ops.LossChannel) for op in p_func.circuit])
+
+    def test_all_loss(self, monkeypatch, adj, dim):
```
ixfoduap: Great test!

strawberryfields/gbs/similarity.py
trbromley: This has been added in: https://github.com/XanaduAI/strawberryfields/pull/223/commits/0035650994a5feb5f643201f54ce9eee4c913885, along with some other tests that were missing. I feel we could probably get away with having fewer tests though.

trbromley:
Thanks @ixfoduap @nquesada.
Should be ready for another look!

strawberryfields/gbs/sample.py
nquesada: I always though Python got Booleans precisely to not do that, like in C. Anyway, if that is how things work, then it's fine.

josh146:
Note that we plan on adding hardware simulators in the future that have 'built in' loss, so when this is live, the `LossChannel` gates will need to be ignored if these simulators are used.

josh146:
Thinking about this a bit more: the best solution will be to return the use of `backend_options={}` as an argument at some point in the future. This will allow the user to dispatch to a backend of their choosing (remote/local/gaussian/hardware/chip0sim), while also packaging up any options that are exclusive to that backend (loss, cutoff, etc).

trbromley:
Thanks @josh146, agreed! For now we can use this but can definitely update as the lower layers of SF evolve.

PR #224, opened 2019-10-22
GBS Bug: Fix single shot sampling

trbromley:
Currently, the `gbs.sample.sample()` function raises:
`AttributeError: 'list' object has no attribute 'tolist'`
when called with `n_samples = 1` (which, ironically, is actually the default).

This is because `eng.run`, which is called within `sample()`, returns:
- A list if `n_samples = 1`
- A numpy array if `n_samples > 1`

To fix this, we have changed the `sample()` function to internally check when `n_samples = 1` and when that is the case: for a sample `s`, return `[s]`. E.g., if the sample is `[0, 1, 3, 0]` the return is `[[0, 1, 3, 0]]`. Although this seems like an unnecessary embedding, it means that the output form of `sample()` doesn't change with `n_samples`, i.e., always `len(sample()) = n_samples` and the return is a list of samples.

Alternatively, we could change `sample()` to return:
- A list if `n_samples = 1`
- A list of lists OR numpy array if `n_samples > 1`
However, this is inconsistent and can confuse users (as it did with us!).

Tests have also been updated to check for the case when `n_samples = 1`.

### DIFF
--- a/strawberryfields/gbs/sample.py
+++ b/strawberryfields/gbs/sample.py
@@ -147,7 +147,12 @@ def sample(A: np.ndarray, n_mean: float, n_samples: int = 1, threshold: bool = T
         else:
             sf.ops.MeasureFock() | q
 
-    return eng.run(p, run_options={"shots": n_samples}).samples.tolist()
+    s = eng.run(p, run_options={"shots": n_samples}).samples
+
+    if n_samples == 1:
+        return [s]
+
+    return s.tolist()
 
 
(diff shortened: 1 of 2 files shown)


heltluke:
Certainly looks as though this will work, but is it perhaps not better for `eng.run` to always return the same type of object (e.g. a numpy array)?

co9olguy:
@heltluke Yes, this is in our backlog to address this in SF. 
For now, there is really only a bug in GBS layer, so an easy fix.

PR #225, opened 2019-10-23
GBS Docs: Use sentence case for titles

trbromley:
The rest of SF uses [sentence case](https://en.wikipedia.org/wiki/Letter_case#Stylistic_or_specialised_usage) for titles. This PR makes sure we follow the same in the apps layer.

We also upgrade the title level of "Code details" and add some minor modifications to the titles/subtitles.


strawberryfields/gbs/data.py
```diff
@@ -103,7 +103,7 @@
 For example, we see that the ``data[3]`` sample has 11 clicks.
 
 Code details
-^^^^^^^^^^^^
```
trbromley: @josh146, what do you think of this change? Is there any problem to do this?

josh146: Best to keep consistent; `===` for page titles, `---` for section headings, `^^^` for subheadings, `~~~`, etc.

Note that we normally have 'Code details' as a subheading, to avoid it showing in the sidebar.

trbromley: Thanks Josh! I actually don't mind "Code details" showing in the sidebar, and think it looks better in the rendered page to have it at the same depth as the other section headings.

trbromley:
Note - need to do checks for new `vibronic` module

PR #227, opened 2019-10-25
GBS: Rename plotting functions

trbromley:
Beforehand:
`plot.plot_graph`
`plot.plot_subgraph`
`plot.plot_points`

Now:
`plot.graph`
`plot.subgraph`
`plot.points`

**Something to be aware of**: We now have a module names `points` and a function names `points` (but in the `plot` module). This could cause a problem for the user if they do, e.g.,
```python
from strawberryfields.gbs import points
from strawberryfields.gbs.plot import points
```
However, this isn't the standard workflow that we suggest, so if a user does this they probably know what they're doing and will be able to work out how to avoid the problem.


strawberryfields/gbs/plot.py
```diff
@@ -43,11 +43,11 @@
 import numpy as np
 
 
-def _node_coords(graph: nx.Graph, l: dict) -> Tuple:
+def _node_coords(g: nx.Graph, l: dict) -> Tuple:
```
ixfoduap: Nice catch! We don't want to name variables the same as functions, right?

PR #228, opened 2019-10-28
Stop TF FutureWarnings

trbromley:
If TensorFlow 1.3 is installed, an import of Strawberry Fields results in:
Annotation 2019-10-28 165831
This happens even if the user is not using the TF engine.

Although we expect users not to need TF to use SF, many might have it installed if they have looked at using ML functionality in SF. They will then receive the above error whenever importing SF.

This PR prevents the FutureWarning errors from showing when importing SF with TF==1.3 installed.

### DIFF
--- a/strawberryfields/parameters.py
+++ b/strawberryfields/parameters.py
@@ -91,6 +91,7 @@ Code details
 
 """
 import numbers
+import warnings
 
 from unittest import mock
 import numpy as np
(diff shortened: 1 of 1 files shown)


josh146:
I'm a bit hesitant to do this, as it's generally bad practice to filter warnings from an external package. The solution to this issue should instead be:

* Upgrade the backend to TF 2.0 (the future warnings indicate it will likely stop working at some point anyway)

* Upgrade from TF 1.3 locally (TF 1.3 is not a hard requirement of Strawberry Fields any more)

jswinarton:
Just wanted to add my two cents here, Tom and I discussed this over Slack but I lean toward agreement with @josh146 here. I would only suggest doing this if TF1.3 was a hard requirement, but if users can use other TF 1.X versions than just leave the warnings and let it complain, users should be keeping their software reasonably up to date anyway.

trbromley:
Given the feedback, I am closing this PR - hopefully this warning will not occur often for users.

PR #236, opened 2019-10-31
GBS: Add spectrum plotting functionality

trbromley:
This is WIP


strawberryfields/gbs/plot.py
```diff
@@ -343,3 +343,51 @@ def points(
...
+
+    Returns:
+         Figure: spectrum in the form of a histogram of energies
+    """
+    try:
+        import plotly.graph_objects as go
+    except ImportError:
+        raise ImportError(plotly_error)
+
+    h = go.Histogram(x=energies, nbinsx=bins, marker={"color": BAR_COLOUR})
```
soranjh: I think it might be good to make the limits of the x optional with some default values, such this:
x_start = 0.0
x_end = 8000.0
xbins = dict(start=x_start, end=x_end, size=0.5)

trbromley: Nice idea, similar things are done for `np.histogram` or `matplotlib.pyplot.hist`:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html
https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.hist.html

trbromley:
Example plot for formic acid:

image

strawberryfields/gbs/plot.py
```diff
@@ -343,3 +344,76 @@ def points(
...
+    Returns:
+         Figure: spectrum in the form of a histogram of energies with a Lorentzian-function curve
+    """
+    try:
+        import plotly.graph_objects as go
+    except ImportError:
+        raise ImportError(plotly_error)
+
+    if xmin is None:
+        xmin = min(energies) - 1000.0
```
ixfoduap: It's probably safest to change the x limits as a function of the range of energies. Something like:
`emin = min(energies)`
`emax = max(energies)`
`xmin = emin - 0.05*(emax-emin)`
`xmax = emax + 0.05*(emax-emin)`
The actual number 0.05 (5% of the width) should be chosen carefully. Maybe 10% or 1% is better

strawberryfields/gbs/plot.py
```diff
@@ -343,3 +344,76 @@ def points(
...
+    if xmax is None:
+        xmax = max(energies) + 1000.0
+
+    bins = int(max(energies) - min(energies)) // 5
+    bar_width = (max(energies) - min(energies)) * 0.005
+    line_width = 3.0
+
+    h = np.histogram(energies, bins)
+
```
ixfoduap: A lot of these spaces are not needed

strawberryfields/gbs/plot.py
```diff
@@ -343,3 +344,76 @@ def points(
...
+    X = np.linspace(xmin, xmax, int(xmax-xmin))
+
+    L = 0
+    for e in energies:
+        L += (gamma / 2)**2 / ((X - e)**2 + (gamma / 2)**2)
+
+    text_font = dict(color="black", family="Times New Roman")
```
ixfoduap: What other fonts are available? It would be great if we could use Computer Modern, the latex font. If not, we should probably pick something that resembles it

strawberryfields/gbs/plot.py
```diff
@@ -343,3 +344,76 @@ def points(
...
+        x=h[1].tolist(),
+        y=h[0].tolist(),
+        width=bar_width,
+        marker=dict(color=GREY))
+
+    line = go.Scatter(
+        x=X,
+        y=L,
+        mode='lines',
+        name='f(x)',
```
ixfoduap: Just a minor detail, but is this name `f(x)` ever used anywhere? Does an error occur if we just remove that line?

strawberryfields/gbs/plot.py
soranjh: It seems that we do not have many choices.

strawberryfields/gbs/plot.py
soranjh: removed.

strawberryfields/gbs/plot.py
```diff
@@ -343,3 +344,74 @@ def points(
...
+    bar_width = (emax - emin) * 0.005
+    line_width = 3.0
+
+    h = np.histogram(energies, bins)
+    X = np.linspace(xmin, xmax, int(xmax-xmin))
+    L = 0
+    for e in energies:
+        L += (gamma / 2)**2 / ((X - e)**2 + (gamma / 2)**2)
+
+    text_font = dict(color="black", family="Times New Roman")
```
ixfoduap:
```suggestion
    text_font = dict(color="black", family="Computer Modern")
```

strawberryfields/gbs/plot.py
```diff
@@ -345,49 +346,72 @@ def points(
...
 
-LIGHT_BLUE = "#94cecf"
-
-BAR_COLOUR = LIGHT_BLUE
-
-
-def spectrum(energies: list, bins: Optional[int] = 0):  # pragma: no cover
-    """Plots a spectrum based on input sampled energies.
+def spectrum(
+    energies: list, gamma: float = 100.0, xmin: float = None, xmax: float = None
```
trbromley: I think we could use the
`Optional[float] = None`
style here

strawberryfields/gbs/plot.py
```diff
@@ -345,49 +346,72 @@ def points(
...
+):  # pragma: no cover
+    """Plots a vibronic spectrum based on input sampled energies.
 
     Args:
         energies (list[float]): a list of sampled energies
-        bins (int): maximum number of bins. If zero, the number of bins is decided automatically
-            by Plotly.
+        gamma (float): parameter specifying the width of the Lorentzian function
```
trbromley: It might be worth explaining the Lorentzian function better in the docstring.

strawberryfields/gbs/plot.py
```diff
@@ -345,49 +346,72 @@ def points(
...
-    text_font = dict(color="black", family="Times New Roman")
     axis_style = dict(
         titlefont_size=30,
         tickfont=text_font,
         tickfont_size=20,
         showline=True,
         linecolor="black",
-        mirror=True,
-    )
+        mirror=True)
```
trbromley: 🤔 Is this the result of running `black`?

strawberryfields/gbs/plot.py
trbromley: No worries if this was discussed/decided, but I would have the addition of the Lorentzian curve being optional. Still `gamma` can default to a width of `100`, but if the user inputs something like `None`, they would be able to get just the histogram without the curve.

strawberryfields/gbs/plot.py
```diff
@@ -345,49 +346,72 @@ def points(
...
 
     layout = go.Layout(
         yaxis=dict(
-            title={"text": "Counts", "font": text_font}, **axis_style, gridcolor=VERY_LIGHT_GREY
-        ),
+            title={"text": "Counts", "font": text_font}, **axis_style, rangemode='tozero'),
```
trbromley: what does `rangemode='tozero'` do?

strawberryfields/gbs/plot.py
```diff
@@ -345,49 +346,72 @@ def points(
...
+    bar_width = (emax - emin) * 0.005
+    line_width = 3.0
+
+    h = np.histogram(energies, bins)
+    X = np.linspace(xmin, xmax, int(xmax-xmin))
+    L = 0
+    for e in energies:
+        L += (gamma / 2)**2 / ((X - e)**2 + (gamma / 2)**2)
+
+    text_font = dict(color="black", family="Computer Modern")
```
trbromley: I'm curious what motivated the decision to move away from Plotly's `Histogram`?

strawberryfields/gbs/plot.py
```diff
@@ -345,49 +346,72 @@ def points(
...
         raise ImportError(plotly_error)
 
-    h = go.Histogram(x=energies, nbinsx=bins, marker={"color": BAR_COLOUR})
+    emin = min(energies)
+    emax = max(energies)
+    if xmin is None:
+        xmin = emin - 0.1 * (emax - emin)
+    if xmax is None:
+        xmax = emax + 0.1 * (emax - emin)
+    bins = int(emax - emin) // 5
```
trbromley: So now we hardcode-in the number of bins such that each bin corresponds to a width of 5 in the x-axis?
Not sure whether now or later, but would be good to allow the user to specify the number of bins

trbromley: Actually this might be a case where adding a comment is necessary as I find this line and the one below not so clear

strawberryfields/gbs/plot.py
```diff
@@ -360,14 +360,13 @@ def spectrum(
     Returns:
          Figure: spectrum in the form of a histogram of energies with a Lorentzian-function curve
     """
+    if len(energies) < 2:
+        raise ValueError("Number of sampled energies must be at least two")
```
trbromley: I would prefer to keep the try-except import as the very first thing, just like how in a script the imports are the first thing

strawberryfields/gbs/plot.py
soranjh: We can do it in the polishing PRs.

strawberryfields/gbs/plot.py
soranjh: It might be. I found that ``pylint`` and ``black`` do not like each other all the time. Not sure which one we should use.

strawberryfields/gbs/plot.py
soranjh: This is a reasonable suggestion although I should say that plotting just the bar chart (histogram) form energies provided by ``vibronic.energies`` requires 2 line of code. If the user just wants the bar charts, it can be done with minimum work instead of using our ``plot.spectrum`` function.

strawberryfields/gbs/plot.py
soranjh: It forces the Y axis to start from zero. Otherwise, you might get a spectrum with empty space below it.

strawberryfields/gbs/plot.py
soranjh: We have to have exactly same bins and counts to make sure that the bars and Lorentzian curves  are generated based on the same data. Note that we need the numeric data to build the Lorentzian smooth curve, which I could not get the numeric bin and count values from plotly.

strawberryfields/gbs/plot.py
soranjh: The main problem with making the bin size optional is potential inconsistency in the position of bars and Lorentzian peaks. Note that the Lorentzian peaks appear at the exact position of the energies but the bars are bin dependent. For example, if you have one count at an energy value of 1000, the Lorentzian peak appears at 1000 while using large bins shifts the location of the bars.

strawberryfields/gbs/plot.py
soranjh: I had that initially but got an error (I believe it was CodeFactor) complaining about the order. So brought this condition to the beginning.

soranjh:
> I'm a bit late to the party with this one, but have taken a look and left some comments. Curious to see how the plots look now! Thanks!

newplot

strawberryfields/gbs/plot.py
ixfoduap: @trbromley But it looks so much better with the curve!
More seriously, we've indeed decided to keep plot options minimal, in agreement with the philosophy we've followed so far with the `plot` module.

strawberryfields/gbs/plot.py
trbromley: Ahh that makes sense! Then I'm happy with this choice! Thanks

trbromley:
Thanks @soranjh, it does look awesome!!!

PR #237, opened 2019-10-31
Simulon support

zzzeid:
**Context:**
To improve StarshipEngine to allow for multiple backends when sending remote jobs.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


co9olguy:
@zzzeid is this one needing a review?

zzzeid:
@co9olguy yes that is correct.

strawberryfields/engine.py
```diff
@@ -447,21 +448,16 @@ class StarshipEngine(BaseEngine):
 
     Args:
         backend (str, BaseBackend): name of the backend, or a pre-constructed backend instance
+        polling_delay_seconds (int): the number of seconds to wait between queries when polling for
+            job results
     """
 
-    HARDWARE_BACKENDS = ("chip0",)
+    # This engine will execute jobs remotely.
+    REMOTE = True
```
josh146: Todo: the accepted 'engine slight refactor' ADR proposes that `StarshipEngine` becomes a `HardwareEngine`/`BlackbirdEngine`, that executes both local and remote hardware simulations. That is, simulations that accept blackbird scripts, and return only samples.

`LocalEngine` will become `StateEngine`, that supports local statevector simulations.

strawberryfields/engine.py
```diff
@@ -447,21 +448,16 @@ class StarshipEngine(BaseEngine):
...
 
-    def __init__(self, polling_delay_seconds=1, **kwargs):
-        # Only chip0 backend supported initially.
-        backend = "chip0"
+    def __init__(self, backend, polling_delay_seconds=1, **kwargs):
         super().__init__(backend)
 
-        # todo: move this into backend class
```
josh146: Nice :+1:

josh146:
This PR should be merged into `demo` ASAP, to decouple `StarshipEngine` from `chip0`.

starship
```diff
@@ -67,10 +69,12 @@ if __name__ == "__main__":
             input(PROMPTS["authentication_token"].format(config.api["authentication_token"]))
             or config.api["authentication_token"]
         )
-        save = input(PROMPTS["save"]).upper() == "Y"
+        port = input(PROMPTS["port"].format(config.api["port"])) or config.api["port"]
+        use_ssl = input(PROMPTS["use_ssl"].format("y" if config.api["use_ssl"] else "n")).upper() == "Y"
+        save = input(PROMPTS["save"].format("y" if config.api["use_ssl"] else "n")).upper() == "Y"
```
co9olguy: @josh146 I guess this one's up to you and me now.
Any idea why we would consider `"use_sll"` when determining default prompt for  `save`?

strawberryfields/api_client.py
```diff
@@ -166,6 +166,8 @@ def __init__(self, **kwargs):
 
         self.HOSTNAME = self._config["hostname"]
         self.BASE_URL = "{}://{}".format("https" if self.USE_SSL else "http", self.HOSTNAME)
+        if self._config["port"] != 443:
```
co9olguy: Probably better to check against `DEFAULT_CONFIG["api"]["port"]` explicitly here

starship
josh146: `use_ssl` is stored in the config file, and the preference remembered for next time.

On a tangent though, is this argument needed? Is there ever a use case where SF will not be connecting using SSL?

co9olguy: 🤷‍♂

lneuhaus: actually, I had issues using SSL, so I disabled it for my tests. But I believe forcing its use would just make the user debug the problem and eventually use it.

jswinarton: The platform will force the use of SSL for security reasons, so in practice the only way you will be able to connect to platform is with SSL, unless you are developing against services locally. I agree that this option should be available for debugging purposes, but in production anyone not connecting over SSL will be refused.

PR #241, opened 2019-11-04
GBS: Add vibronic dataset

trbromley:
WIP: Dataset for formic acid molecule


tests/gbs/test_data.py
```diff
@@ -150,3 +160,52 @@ def test_adj_dim(self, dataset):
...
+        """Test if delta has correct shape"""
+        _, _, _, delta = dataset.w, dataset.wp, dataset.Ud, dataset.delta
+        assert delta.shape == (dataset.modes // 2,)
+
+    def test_uniform_dims(self, dataset):
+        """Test if overall dimension is correct"""
+        w, wp, Ud, delta = dataset.w, dataset.wp, dataset.Ud, dataset.delta
+        dim = w.shape[0]
+
+        assert wp.shape[0] == dim and Ud.shape[0] == dim and delta.shape[0] == dim
```
ixfoduap:
```suggestion
        assert wp.shape[0] == dim and Ud.shape == (dim, dim) and delta.shape[0] == dim
```
Otherwise we only check that the number of columns is correct

strawberryfields/gbs/data.py
```diff
@@ -408,13 +422,67 @@ class Mutag3(Dataset):
...
+
+    .. |formic| image:: ../../_static/formic.png
+        :align: middle
+        :width: 250px
+        :target: javascript:void(0);
+
+    |formic|
+
+    Attributes:
+        n_mean = 1.56
```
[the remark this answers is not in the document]
soranjh: s = eng.run(prog)
n_mean = sum([s.state.mean_photon(i)[0] for i in range(len(w))])

strawberryfields/gbs/data.py
```diff
@@ -54,16 +58,26 @@
 | **MUTAG_2** | **MUTAG_3** |
 +-------------+-------------+
 
+Molecules
+---------
+
+Using the :mod:`~.gbs.vibronic` module, GBS data has been generated by encoding a variety of
```
soranjh:
```suggestion
Using the :mod:`~.gbs.vibronic` module, GBS data has been generated for formic acid at zero
```

strawberryfields/gbs/data.py
```diff
@@ -54,16 +58,26 @@
 | **MUTAG_2** | **MUTAG_3** |
 +-------------+-------------+
 
+Molecules
+---------
+
+Using the :mod:`~.gbs.vibronic` module, GBS data has been generated by encoding a variety of
+different molecules. The GBS samples can be used to recover the vibronic spectrum of the molecule.
```
soranjh:
```suggestion
temperature. The GBS samples can be used to recover the vibronic spectrum of the molecule.
```

strawberryfields/gbs/data.py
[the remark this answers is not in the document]
soranjh: Done.

strawberryfields/gbs/data.py
```diff
@@ -461,13 +461,22 @@ def T(self) -> bool:
 class Formic(MoleculeDataset):
     """Zero temperature formic acid.
 
-    TODO Cite sources for w, wp, Ud, d + include visualisation
+    The molecular parameters are from :cite:`Huh_2015`.
+
+    **Molecule:**
+
+    .. |formic| image:: ../../_static/formic.png
```
trbromley: Image looks awesome

strawberryfields/gbs/data.py
```diff
@@ -461,13 +461,22 @@ def T(self) -> bool:
...
+        :align: middle
+        :width: 250px
+        :target: javascript:void(0);
+
+    |formic|
 
     Attributes:
         n_mean = 1.56
         threshold = False
-        n_samples = 100000
```
trbromley: Did you generate new data? Could we just go through the original data and pad with zeros?

strawberryfields/gbs/data.py
```diff
@@ -440,7 +440,7 @@ class MoleculeDataset(Dataset, ABC):
         w (array): normal mode frequencies of the electronic ground state (:math:`\mbox{cm}^{-1}`)
         wp (array): normal mode frequencies of the electronic excited state (:math:`\mbox{cm}^{-1}`)
         Ud (array): Duschinsky matrix
-        d (array): Duschinsky displacement vector corrected with wp
+        delta (array): Duschinsky displacement vector
```
trbromley: Why do we make this change? I see `d` in `gbs_params` (https://github.com/XanaduAI/strawberryfields/blob/master/strawberryfields/gbs/vibronic.py)

strawberryfields/gbs/data.py
```diff
@@ -61,8 +61,8 @@
 Molecules
 ---------
 
-Using the :mod:`~.gbs.vibronic` module, GBS data has been generated by encoding a variety of
-different molecules. The GBS samples can be used to recover the vibronic spectrum of the molecule.
+Using the :mod:`~.gbs.vibronic` module, GBS data has been generated for formic acid at zero
+temperature. The GBS samples can be used to recover the vibronic spectrum of the molecule.
```
trbromley: We decided to do just one molecule? Or add more in a follow up PR?

tests/gbs/test_data.py
```diff
@@ -196,7 +196,7 @@ def test_uniform_dims(self, dataset):
         w, wp, Ud, delta = dataset.w, dataset.wp, dataset.Ud, dataset.delta
         dim = w.shape[0]
 
-        assert wp.shape[0] == dim and Ud.shape[0] == dim and delta.shape[0] == dim
+        assert wp.shape[0] == dim and Ud.shape == (dim, dim) and delta.shape[0] == dim
```
trbromley: :+1:

strawberryfields/gbs/data.py
trbromley: Ahh I guess because of https://github.com/XanaduAI/strawberryfields/pull/247 ?

strawberryfields/gbs/data.py
```diff
@@ -54,16 +58,26 @@
 | **MUTAG_2** | **MUTAG_3** |
 +-------------+-------------+
 
+Molecules
+---------
+
+Using the :mod:`~.gbs.vibronic` module, GBS data has been generated for formic acid at zero
```
trbromley: How about also mentioning the `sample.vibronic` function?

strawberryfields/gbs/data.py
soranjh: I generated new data mainly to check everything works fine after upgrading the functions for finite T. Yes we can just pad the old data with zeros. In any case, we do not really need more than 20000 samples to get convergence in the shape of the spectrum.

strawberryfields/gbs/data.py
soranjh: There are little inconsistencies in the definitions of Duschinsky displacement parameter. One way to deal with that is to use delta instead of d, or we have to explain the difference between delta and d all the time.

strawberryfields/gbs/data.py
soranjh: Juan Miguel suggested having only one molecule.

strawberryfields/gbs/data.py
soranjh: Will add this.

strawberryfields/gbs/data.py
ixfoduap: Just one molecule will fit all our needs and we're reaching the point were we need to wrap things up, so I indeed recommended having just formic acid. As usual, if we think it is important to add more, we can always do so in the future

strawberryfields/gbs/data.py
```diff
@@ -408,13 +422,67 @@ class Mutag3(Dataset):
...
+
+# pylint: disable=abstract-method
+class MoleculeDataset(Dataset, ABC):
+    r"""Class for loading datasets of pre-generated samples from molecules.
+
+    Attributes:
+        w (array): normal mode frequencies of the electronic ground state (:math:`\mbox{cm}^{-1}`)
+        wp (array): normal mode frequencies of the electronic excited state (:math:`\mbox{cm}^{-1}`)
+        Ud (array): Duschinsky matrix
+        delta (array): Duschinsky displacement vector
```
ixfoduap:
```suggestion
        delta (array): displacement vector
```
As far as I understand, the Duschinsky displacement is the vector d that defines the transformation between mode coordinates q'= Uq+d. The delta we use here isn't quite that, but a rescaled version

soranjh: True, delta is a unitless version of d.

PR #245, opened 2019-11-08
Add vibronic sampler for finite temperature

soranjh:

### DIFF
--- a/strawberryfields/gbs/sample.py
+++ b/strawberryfields/gbs/sample.py
@@ -293,37 +293,25 @@ def vibronic(
     n_samples: int,
     loss: float = 0.0,
 ) -> list:
-    r"""Generate samples for computing vibronic spectra.
+    r"""Generate samples for computing vibronic spectra. The following gates are applied to
+    :math:`2N` vacuum states:
 
-    This function applies two procedures for generating samples depending on the temperature at
-    which the vibronic spectra is computed. In the finite temperature procedure, 2N vacuum modes
-    are prepared with N being the number of normal modes in the molecule. The first and
-    second N modes correspond to the normal modes of the final and initial electronic states,
-    respectively. These 2N modes are treated with two-mode squeezing gates:
+    1. Two-mode squeezing on all :math:`2N` modes with parameters ``t``
+    2. Interferometer ``U1`` on the first :math:`N` modes
+    3. Squeezing on the first :math:`N` modes with parameters ``r``
+    4. Interferometer ``U2`` on the first :math:`N` modes
+    5. Displacement on the first :math:`N` modes with parameters ``alpha``
 
-    #. Two-mode squeezing on all 2N modes with parameters ``t``
-
-    and then the following gates are applied to the first N modes:
-
-    #. Interferometer ``U1``
-    #. Squeezing on all N modes with parameters ``r``
-    #. Interferometer ``U2``
-    #. Displacement on all N modes with parameters ``alpha``
-
-    In the last step, the number of photons in each of the 2N modes is measured.
-
-    In the zero temperature case, only N modes are considered. The four gates mentioned above are
-    applied to N vacuum states and then the number of photons in each mode is measured. This
-    makes the zero temperature sampling significantly faster than the finite temperature one.
-    However, to keep the sizes of the samples consistent and temperature independent, we add N
-    zeros to the end of each zero temperature sample at the very end of this function. A sampling
-    example is provided in the following for formic acid where t, U1, S, U2, and alpha
-    parameters have been obtained by using the ``vibronic.gbs_params`` function.
+    Finally, the number of photons in each of the :math:`2N` modes is measured. In special
+    cases that all of the two-mode squeezing parameters ``t`` are zero, only :math:`N` modes are
+    considered. In this case, the gates 2-5 mentioned above are applied to :math:`N` vacuum states
+    and then the number of photons in each mode is measured. A sampling example is provided in the
+    following for formic acid where ``t``, ``U1``, ``S``, ``U2``, and ``alpha`` parameters have
+    been obtained by using the ``vibronic.gbs_params`` function.
 
     **Example usage:**
 
 >>> t = np.array([0., 0., 0., 0., 0., 0., 0.])
-
 >>> U1 = np.array(
 >>>     [[-0.07985219, 0.66041032, -0.19389188, 0.01340832, 0.70312675, -0.1208423, -0.10352726],
 >>>     [0.19216669, -0.12470466, -0.81320519, 0.52045174, -0.1066017, -0.06300751, -0.00376173],
@@ -331,12 +319,9 @@ def vibronic(
 >>>     [0.63690134, -0.03047939, 0.46585565, 0.50545897, 0.21194805, -0.20422433, 0.18516987],
 >>>     [0.34556293, 0.22562207, -0.1999159, -0.50280235, -0.25510781, -0.55793978, 0.40065893],
 >>>     [-0.03377431, -0.66280536, -0.14740447, -0.25725325, 0.6145946, -0.07128058, 0.29804963],
->>>     [-0.24570365, 0.22402764, 0.003273, 0.19204683, -0.05125235, 0.3881131, 0.83623564],
->>>     ])
-
+>>>     [-0.24570365, 0.22402764, 0.003273, 0.19204683, -0.05125235, 0.3881131, 0.83623564]])
 >>> r = np.array(
 >>>     [0.09721339, 0.07017918, 0.02083469, -0.05974357, -0.07487845, -0.1119975, -0.1866708])
-
 >>> U2 = np.array(
 >>>     [[-0.07012006, 0.14489772, 0.17593463, 0.02431155, -0.63151781, 0.61230046, 0.41087368],
 >>>     [0.5618538, -0.09931968, 0.04562272, 0.02158822, 0.35700706, 0.6614837, -0.326946],
@@ -344,12 +329,9 @@ def vibronic(
 >>>     [0.01788782, 0.60430409, -0.19831443, -0.73270964, -0.06393682, 0.03376894, -0.23038293],
 >>>     [0.78640978, -0.11133936, 0.03160537, -0.09188782, -0.43483738, -0.4018141, 0.09582698],
 >>>     [-0.13664887, -0.11196486, 0.86353995, -0.19608061, -0.12313513, -0.08639263, -0.40251231],
->>>     [-0.12060103, -0.01169781, -0.33937036, 0.34662981, -0.49895371, 0.03257453, -0.70709135],
->>>     ])
-
+>>>     [-0.12060103, -0.01169781, -0.33937036, 0.34662981, -0.49895371, 0.03257453, -0.70709135]])
 >>> alpha = np.array(
 >>>     [0.15938187, 0.10387399, 1.10301587, -0.26756921, 0.32194572, -0.24317402, 0.0436992])
-
 >>> vibronic(t, U1, S, U2, alpha, 2, 0.0)
 [[0, 0, 2, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0],
  [0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]
@@ -374,68 +356,45 @@ def vibronic(
 
     n_modes = len(t)
 
-    # initialize engine
-
     eng = sf.LocalEngine(backend="gaussian")
 
-    # initialize program objects
-
-    if np.all(t != 0):
+    if np.any(t != 0):
         gbs = sf.Program(n_modes * 2)
     else:
         gbs = sf.Program(n_modes)
 
-    # start circuit
     # pylint: disable=expression-not-assigned,pointless-statement
     with gbs.context as q:
 
-        # two-mode squeezing gates
-
-        if np.all(t != 0):
+        if np.any(t != 0):
             for i in range(n_modes):
                 sf.ops.S2gate(t[i]) | (q[i], q[i + n_modes])
 
-        # first interferometer
-
         sf.ops.Interferometer(U1) | q[:n_modes]
-
-        # squeezing gates
 
         for i in range(n_modes):
             sf.ops.Sgate(r[i]) | q[i]
 
-        # second interferometer
-
         sf.ops.Interferometer(U2) | q[:n_modes]
-
-        # displacement gates
 
         for i in range(n_modes):
             sf.ops.Dgate(alpha[i]) | q[i]
 
-        # apply loss
-
         if loss:
             for _q in q:
                 sf.ops.LossChannel(1 - loss) | _q
-
-        # measurement
 
         sf.ops.MeasureFock() | q
 
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=UserWarning, message="Cannot simulate non-")
 
-        # run the engine and generate samples
-
         s = eng.run(gbs, run_options={"shots": n_samples}).samples
-
-    # end circuit
 
     s = np.array(s).tolist() # convert all generated samples to list
 
     if n_samples == 1:
         s = [s]
-    if np.all(t == 0):
+    if np.any(t == 0):
         s = np.pad(s, ((0, 0), (0, n_modes))).tolist()
     return s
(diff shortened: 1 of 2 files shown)


strawberryfields/gbs/sample.py
```diff
@@ -55,21 +55,11 @@
...
-#. interferometer :math:`U_{1}` (using :class:`~.ops.Interferometer`)
-#. squeezing :math:`S_{\mathbf{r}}` on all modes with parameters :math:`\mathbf{r}` (using
-   :class:`~.ops.Sgate`)
-#. another interferometer :math:`U_{2}` (using :class:`~.ops.Interferometer`)
-#. displacement :math:`D_{\boldsymbol{\alpha}}` on all modes with parameters
-   :math:`\boldsymbol{\alpha}` (using :class:`~.ops.Dgate`)
-
-This can be followed by PNR detection. The :func:`gaussian` function
-allows users to sample from arbitrary pure Gaussian states using the above decomposition.
+The :func:`vibronic` function allows users to sample from Gaussian states for computing molecular
```
ixfoduap:
```suggestion
The :func:`vibronic` function allows users to sample Gaussian states for computing molecular
```

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,76 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
     n_samples: int,
     loss: float = 0.0,
 ) -> list:
-    r"""Generate simulated samples from pure Gaussian states.
+    r"""Generate samples for computing vibronic spectra.
 
-    The pure Gaussian states are prepared by applying the following gates to the vacuum:
+    This function applies two procedures for generating samples depending on the temperature at
+    which the vibronic spectra is computed. In the finite temperature procedure, 2N vacuum modes
+    are prepared with N being the number of normal modes in the molecule and the first and
```
ixfoduap:
```suggestion
    are prepared with N being the number of normal modes in the molecule. The first and
```

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,76 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
+    #. Displacement on all N modes with parameters ``alpha``
+
+    In the last step, the number of photons in each of the 2N modes is measured.
+
+    In the zero temperature case, only N modes are considered. The four gates mentioned above are
+    applied to N vacuum states and then the number of photons in each mode is measured. This
+    makes the zero temperature sampling significantly faster than the finite temperature one.
+    However, to keep the sizes of the samples consistent and temperature independent, we add N
+    zeros to the end of each zero temperature sample at the very end of this function.
+
```
ixfoduap: It would be good to add a sentence here briefly explaining where the parameters in the example usage come from

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,76 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
+>>>     [0.78640978, -0.11133936, 0.03160537, -0.09188782, -0.43483738, -0.4018141, 0.09582698],
+>>>     [-0.13664887, -0.11196486, 0.86353995, -0.19608061, -0.12313513, -0.08639263, -0.40251231],
+>>>     [-0.12060103, -0.01169781, -0.33937036, 0.34662981, -0.49895371, 0.03257453, -0.70709135],
+>>>     ]
+>>> )
+
+>>> alpha = np.array(
+>>>     [0.15938187, 0.10387399, 1.10301587, -0.26756921, 0.32194572, -0.24317402, 0.0436992]
+>>> )
+[[0, 0, 2, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
ixfoduap: this is missing the actual call to the function, right? `vibronic(t, U1, r, U2, alpha)` ?

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,76 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
+>>>     [0.01788782, 0.60430409, -0.19831443, -0.73270964, -0.06393682, 0.03376894, -0.23038293],
+>>>     [0.78640978, -0.11133936, 0.03160537, -0.09188782, -0.43483738, -0.4018141, 0.09582698],
+>>>     [-0.13664887, -0.11196486, 0.86353995, -0.19608061, -0.12313513, -0.08639263, -0.40251231],
+>>>     [-0.12060103, -0.01169781, -0.33937036, 0.34662981, -0.49895371, 0.03257453, -0.70709135],
+>>>     ]
+>>> )
+
+>>> alpha = np.array(
+>>>     [0.15938187, 0.10387399, 1.10301587, -0.26756921, 0.32194572, -0.24317402, 0.0436992]
+>>> )
```
ixfoduap: Anything we can do to avoid this lonely parenthesis occupying a whole line?

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +370,71 @@ def gaussian(
...
+
+        # run the engine and generate samples
+
         s = eng.run(gbs, run_options={"shots": n_samples}).samples
 
+    # end circuit
+
     if n_samples == 1:
+        if np.all(t == 0):
+            s = np.pad(s, (0, n_modes), constant_values=(0)).tolist()
```
ixfoduap: This use of `pad` is probably ok, but when I try to run the same line on my computer, I get an error `pad() missing 1 required positional argument: 'mode'`, so probably worth double-checking

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +370,71 @@ def gaussian(
...
+    # start circuit
     # pylint: disable=expression-not-assigned,pointless-statement
     with gbs.context as q:
 
-        # this interferometer has no action, but is kept to follow the decomposition given in the
-        # docstring
-        sf.ops.Interferometer(U1) | q
+        # two-mode squeezing gates
+
+        if np.all(t != 0):
```
ixfoduap: The code below this `if` statement can probably go under the `if` of line 381.  You then probably need to also add `with gbs.context as q:` under both the `if` on 381 and the `else` on 383

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +370,71 @@ def gaussian(
...
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=UserWarning, message="Cannot simulate non-")
+
+        # run the engine and generate samples
+
         s = eng.run(gbs, run_options={"shots": n_samples}).samples
 
+    # end circuit
+
     if n_samples == 1:
```
ixfoduap: If I understand correctly, the difference between a single sample or multiple samples is that a single sample is usually a list, whereas with many samples we have a list of lists. Would it be enough to just go
`if n_samples==1:`
    `s = [s]` 
and then use
 `if np.all(t == 0):`
        `s = np.pad(s, ((0, 0), (0, n_modes)))`?

soranjh: the pad syntax for a single sample and multiple samples is different. And the function should be able to pad both single and multiple samples with zeros.

strawberryfields/gbs/sample.py
agran2018: Maybe is good to mention what is ``t`` before starting the example, in the same place where you commented on ``U1``, ``U2``, etc.

strawberryfields/gbs/sample.py
soranjh: yes, I will add it.

strawberryfields/gbs/sample.py
soranjh: I will bring them to previous line. I think black does some of such corrections.

strawberryfields/gbs/sample.py
soranjh: I will add it.

strawberryfields/gbs/sample.py
soranjh: Thanks, will add.

soranjh: Mentioned, thanks.

strawberryfields/gbs/sample.py
soranjh: It works for me. Could you please check again?

nquesada: a simple of dong what that is line is doing is by writing `s = s+([0]*n_modes)`

nquesada: using numpy in that context is a little bit like using a sledgehammer to crack a nut. Still work though

soranjh: @nquesada Then have to loop over all samples, as the samples are list of list:
samples = [[1,2],[3,4]]

nquesada: Good point!

tests/gbs/test_sample.py
```diff
@@ -58,7 +60,7 @@
     [0.15938187, 0.10387399, 1.10301587, -0.26756921, 0.32194572, -0.24317402, 0.0436992]
 )
 
-p = [U1, r, U2, alpha]
+p = [t, U1, r, U2, alpha]
```
trbromley: We should also test for the nonzero squeezing case.

How about we have a p0 and a p1 which have a zero and nonzero two mode squeezing.

we can then do something like:
`@pytest.mark.parametrize("p", [p0, p1])`
above the tests.

Note the `test_all_loss` test might need an if statement added

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +372,70 @@ def gaussian(
...
+        # displacement gates
 
         for i in range(n_modes):
             sf.ops.Dgate(alpha[i]) | q[i]
 
+        # apply loss
+
         if loss:
             for _q in q:
                 sf.ops.LossChannel(1 - loss) | _q
```
trbromley: Does it still make sense to have loss now that we have made this function quite specific to vibronic spectra? I.e., would we realise a device in this way - two mode squeezing and operations on the first N modes.

@nquesada (I think?) has a way to map to 2N modes consisting of squeezing on all modes, a 2N mode interferometer, + displacement on all modes. Would this be the device we'd actually look to make - where in that case it might make more sense to talk about sampling but for now maybe not

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +372,70 @@ def gaussian(
...
+    n_modes = len(t)
+
+    # initialize engine
 
     eng = sf.LocalEngine(backend="gaussian")
-    gbs = sf.Program(n_modes)
 
+    # initialize program objects
+
+    if np.all(t != 0):
```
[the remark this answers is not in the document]
nquesada: You are right that the logic is incorrect and this should be corrected. However, it also happens that either all the ts are nonzero or they are all zero so the program will still run correctly with this incorrect logic.

strawberryfields/gbs/sample.py
nquesada: You got it all correct @trbromley ! Once it is mapped to the actual device it makes complete sense to put loss

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,78 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
     U2: np.ndarray,
     alpha: np.ndarray,
     n_samples: int,
     loss: float = 0.0,
 ) -> list:
-    r"""Generate simulated samples from pure Gaussian states.
+    r"""Generate samples for computing vibronic spectra.
+
+    This function applies two procedures for generating samples depending on the temperature at
+    which the vibronic spectra is computed. In the finite temperature procedure, 2N vacuum modes
```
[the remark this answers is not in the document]
soranjh: Done.

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,78 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
+    are prepared with N being the number of normal modes in the molecule. The first and
+    second N modes correspond to the normal modes of the final and initial electronic states,
+    respectively. These 2N modes are treated with two-mode squeezing gates:
+
+    #. Two-mode squeezing on all 2N modes with parameters ``t``
```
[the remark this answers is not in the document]
soranjh: It has been mentioned clearly now.

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,78 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
 
-    The pure Gaussian states are prepared by applying the following gates to the vacuum:
+    and then the following gates are applied to the first N modes:
 
     #. Interferometer ``U1``
-    #. Squeezing on all modes with parameters ``r``
+    #. Squeezing on all N modes with parameters ``r``
     #. Interferometer ``U2``
-    #. Displacement on all modes with parameters ``alpha``
+    #. Displacement on all N modes with parameters ``alpha``
```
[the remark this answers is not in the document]
soranjh: Done.

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,78 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
 
-    Note that, since the gates are applied to the vacuum, the first interferometer has no effect.
-    It is nevertheless included so that the inputs to this function follow a standard decomposition
-    of Gaussian unitaries shown :ref:`above <decomposition>`.
+    In the last step, the number of photons in each of the 2N modes is measured.
+
+    In the zero temperature case, only N modes are considered. The four gates mentioned above are
```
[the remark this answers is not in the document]
soranjh: Added the item numbers.

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,78 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
+    applied to N vacuum states and then the number of photons in each mode is measured. This
+    makes the zero temperature sampling significantly faster than the finite temperature one.
+    However, to keep the sizes of the samples consistent and temperature independent, we add N
+    zeros to the end of each zero temperature sample at the very end of this function. A sampling
+    example is provided in the following for formic acid where t, U1, S, U2, and alpha
+    parameters have been obtained by using the ``vibronic.gbs_params`` function.
```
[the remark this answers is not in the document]
soranjh: If we do not talk about temperature, then we should not talk about having two procedures here. In that case, the docstring will be much shorter. But we probably have to talk about having two different circuits somewhere else.

strawberryfields/gbs/sample.py
```diff
@@ -294,28 +284,78 @@ def to_subgraphs(samples: list, graph: nx.Graph) -> list:
...
+
+    **Example usage:**
+
+>>> t = np.array([0., 0., 0., 0., 0., 0., 0.])
```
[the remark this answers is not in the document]
soranjh: Yes, we will replace these parameters by a line that reads them.

strawberryfields/gbs/sample.py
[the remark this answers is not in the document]
soranjh: Initially I had it the other way: if np.all(t = 0) then do T=0 operations. I corrected it now.

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +372,70 @@ def gaussian(
...
+        # run the engine and generate samples
+
         s = eng.run(gbs, run_options={"shots": n_samples}).samples
 
-    if n_samples == 1:
-        return [s]
+    # end circuit
 
-    return s.tolist()
+    s = np.array(s).tolist() # convert all generated samples to list
```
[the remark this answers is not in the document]
soranjh: We should have it because the sampler is inconsistent in the format of the returned samples:
For one single sample it give you a list while for more than one sample it gives you an array. If we remove this line, then we have to add an extra if condition to deal with that, otherwise we get list to list conversion error.

tests/gbs/test_sample.py
soranjh: The difference in T=0 and T!=0 samples is the absence of photons in the last N modes of the zero temperature case. Since majority of the modes might already have zero photons in them, any test will be probabilistic and probably not very useful. Am I missing something here?

strawberryfields/gbs/sample.py
```diff
@@ -332,37 +372,70 @@ def gaussian(
...
+    # start circuit
     # pylint: disable=expression-not-assigned,pointless-statement
     with gbs.context as q:
 
-        # this interferometer has no action, but is kept to follow the decomposition given in the
-        # docstring
-        sf.ops.Interferometer(U1) | q
+        # two-mode squeezing gates
+
+        if np.all(t != 0):
```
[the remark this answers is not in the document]
soranjh: corrected.

PR #249, opened 2019-11-14
Creates a Chip2 circuit spec

josh146:


lneuhaus:
Thanks for this. I tested it successfully by merging this branch into the demo branch (to have the Starship engine). In order to make this work, I had to replace all occurrences of `chip0` in the starship engine in engine.py by `chip2` though. Do we want to add these changes here, too, or should I keep them only in my local version?

josh146:
> Do we want to add these changes here, too, or should I keep them only in my local version?

I think the changes you refer to have already been made in #237, and are just awaiting review :slightly_smiling_face:

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+from strawberryfields.utils import random_interferometer
+from strawberryfields.circuitspecs.chip2 import Chip2Specs, CircuitSpecs
+
+
+pytestmark = pytest.mark.frontend
+
+np.random.seed(42)
+
+
+def program_equivalence(prog1, prog2, compare_params=True, atol=1e-6, rtol=0):
```
co9olguy: did this never get put in the main library??

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+        parameter_mapping = {i: par_evaluate(n.op.p) for i, n in enumerate(G.nodes())}
+
+        # CXgate and BSgate are not symmetric wrt to permuting the order of the two
+        # modes it acts on; i.e., the order of the wires matter
+        wire_mapping = {}
+        for i, n in enumerate(G.nodes()):
+            if n.op.__class__.__name__ == "CXgate":
+                if np.allclose(n.op.p[0], 0):
+                    # if the CXgate parameter is 0, wire order doesn't matter
+                    wire_mapping[i] = 0
```
co9olguy: is `0` a placeholder/indicator? the two branches of the conditional seem to lead to different types in `wire_mapping[i]`

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+
+    modes = 8
+    remote = False
+    local = True
+    interactive = True
+    primitives = {"S2gate", "MeasureFock", "Rgate", "BSgate"}
+    decompositions = {"Interferometer": {}, "MZgate": {}}
+
+
+class TestChip0Compilation:
```
co9olguy:
```suggestion
class TestChip2Compilation:
```

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+    """Tests for compilation using the Chip0 circuit specification"""
```
co9olguy:
```suggestion
    """Tests for compilation using the Chip2 circuit specification"""
```

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def test_exact_template(self, tol):
```
co9olguy: can we parametrize a few more test cases?

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+    def test_incorrect_s2gates(self):
+        """Test exceptions raised if S2gates do not appear on correct modes"""
+        prog = sf.Program(8)
+        U = random_interferometer(4)
+
+        with prog.context as q:
+            ops.S2gate(0.5) | (q[0], q[1])
+            ops.S2gate(0.5) | (q[2], q[3])
+            ops.S2gate(0.5) | (q[4], q[5])
+            ops.S2gate(0.5) | (q[4], q[6])
```
co9olguy: Might be better to test two things:
i) s2gates do not appear multiple times for same mode (this test)
ii) s2gates do not cross over between signal and idler modes (perhaps what this test's docstring would suggest)

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+        with expected.context as q:
+            ops.S2gate(0.5, 0) | (q[0], q[4])
+            ops.S2gate(0.5, 0) | (q[1], q[5])
+            ops.S2gate(0.5, 0) | (q[2], q[6])
+            ops.S2gate(0.5, 0) | (q[3], q[7])
+            ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=False) | (q[0], q[1], q[2], q[3])
+            ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=False) | (q[4], q[5], q[6], q[7])
+            ops.MeasureFock() | q
+
+        expected = expected.compile(DummyCircuit())
```
co9olguy: not clear where the MZ gates mentioned in the docstring come in

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+        prog = sf.Program(8)
+        U = random_interferometer(8)
+
+        with prog.context as q:
+            ops.S2gate(0.5, 0) | (q[0], q[4])
+            ops.S2gate(0.5, 0) | (q[1], q[5])
+            ops.S2gate(0.5, 0) | (q[2], q[6])
+            ops.S2gate(0.5, 0) | (q[3], q[7])
+            ops.Interferometer(U) | q
+            ops.BSgate() | (q[2], q[3])
```
co9olguy: what is the role of the `BSgate` here?? 🤔

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,260 @@
...
+        MeasureFock() | [0, 1, 2, 3, 4, 5, 6, 7]
+        """
+    )
+
+    def compile(self, seq, registers):
+        """Try to arrange a quantum circuit into a form suitable for Chip0.
+
+        Args:
+            seq (Sequence[Command]): quantum circuit to modify
+            registers (Sequence[RegRefs]): quantum registers
```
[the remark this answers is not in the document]
josh146: Yep. The parameters PR removes the `RegRefTransform` class.

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,260 @@
...
+            return seq
+
+        # first do general GBS compilation to make sure
+        # Fock measurements are correct
+        # ---------------------------------------------
+        seq = GBSSpecs().compile(seq, registers)
+        A, B, C = group_operations(seq, lambda x: isinstance(x, ops.MeasureFock))
+
+        if len(B[0].reg) != self.modes:
+            raise CircuitError("All modes must be measured.")
```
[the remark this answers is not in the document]
josh146: I can't quite remember, but I think the reason we enforce all programs ending with `MeasureFock` on all modes is so that hardware simulation output matches software simulation output. The hardware will always return samples from all modes, whereas simulators currently return `None` if the mode is not explicitly measured.

This is to avoid the same compiled program returning different values on different backends.

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,260 @@
...
+
+        # Check circuit begins with two mode squeezers
+        # --------------------------------------------
+        A, B, C = group_operations(seq, lambda x: isinstance(x, ops.S2gate))
+
+        print([o.op for o in A])
```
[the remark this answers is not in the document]
josh146: oops!
```suggestion
```

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if A:
+            raise CircuitError("Circuits must start with four S2gates.")
+
+        # get circuit registers
+        regrefs = {q for cmd in B for q in cmd.reg}
+
+        if len(regrefs) != self.modes:
+            raise CircuitError("S2gates do not appear on the correct modes.")
```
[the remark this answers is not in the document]
josh146: Yeah this one puzzles me. I might be going crazy, but I can't quite tell what this should be (copied from chip0)

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,260 @@
...
+                    raise CircuitError(
+                        "Unitary must be applied separately to modes [0, 1, 2, 3] and modes [4, 5, 6, 7]."
+                    )
+
+        # multiply all unitaries together, to get the final
+        # unitary representation on modes [0, 1] and [2, 3].
+        U0 = multi_dot(U_list0)
+        U4 = multi_dot(U_list4)
+
+        # check unitaries are equal
```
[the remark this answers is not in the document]
josh146: The reason a separate U0 and U4 is created is so that we can call two separate `ops.Interferometer` calls. Currently, the MZ decomposition on a single `U` matrix will result in (identity) beamsplitters and rotation gates occurring across the modes that shouldn't be connected. Kind of annoying.

tests/frontend/test_circuitspecs_chip2.py
josh146: Nope :( I want to add it to `program_utils`, or make it a program method.

tests/frontend/test_circuitspecs_chip2.py
josh146: Let me think, it's been a while since I wrote this. I suppose the thinking is that `CXgate(0) | [2, 1]` should be equivalent to `CXgate(0) | [1, 2]`, but `CXgate(0.5) | [1, 2]` and `CXgate(0.5) | [2, 1]` must be treated as distinct gates.

tests/frontend/test_circuitspecs_chip2.py
josh146: `mesh="rectangular_symmetric"` causes the interferometer to decompose using `MZgate`'s. Will update the docstring.

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -0,0 +1,373 @@
...
+            ops.Rgate(0) | (q[5])
+            ops.Rgate(np.pi) | (q[6])
+            ops.Rgate(-np.pi) | (q[7])
+
+            ops.MeasureFock() | q
+
+        assert program_equivalence(res, expected, atol=tol)
+
+    def test_interferometers(self, tol):
+        """Test interferometers correctly decompose to MZ gates"""
```
josh146:
```suggestion
        """Test that the compilation correctly decomposes the interferometer using
        the rectangular_symmetric mesh (which decomposes to MZ gates)"""
```

tests/frontend/test_circuitspecs_chip2.py
josh146: I don't know :thinking: 
```suggestion
```

tests/frontend/test_circuitspecs_chip2.py
josh146: I'm not sure it's worth it - the parameter values don't do anything and are currently ignored by the compilation (since the program to be compiled already matches the template, no unitary re/de-compilation occurs).

The only reason they are in the test is that I am forced to provide them (they are used in the `.format()` call of the circuit template string).

tests/frontend/test_circuitspecs_chip2.py
josh146: Yep I agree. This will need some thinking, I'm not sure the circuit spec compile method currently does this correctly.

The logic works for `chip0` I think by chance due to the small number of modes (4 modes, 2 S2gates), but I think the logic is incorrect, and doesn't scale for anything larger than 4 modes.

Worth fixing here, or making an issue? Since it likely affects chip0 as well.

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if not B:
+            # no interferometer was applied
+            A, B, C = group_operations(seq, lambda x: isinstance(x, ops.S2gate))
+            A = B  # move the S2gates to A
+        else:
+            for cmd in B:
+                # calculate the unitary matrix representing each
+                # rotation gate and each beamsplitter
+                # Note: this is done separately on modes [0, 1, 2, 3]
```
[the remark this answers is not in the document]
josh146:
```suggestion
```

strawberryfields/circuitspecs/chip2.py
```diff
@@ -0,0 +1,259 @@
...
+                # and modes [4, 5, 6, 7]
```
josh146:
```suggestion
```

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
co9olguy: But we can easily put logic in HW jobs pipeline to not return results for unwanted measured modes, can't we? Can save on the size of results that has to be communicated. 

Anyway, we shouldn't tie too closely to what the simulators return, since we have a long-open ticket to fix up and unify that

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
co9olguy: We can fix here  and in chip0 spec to just give a more relevant msg, i.e., "Wrong number of modes" or something

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
co9olguy: Not sure I follow what the issue is here, but in any case, can't we just create one matrix `U` and then call `ops.Interferometer` twice with the same argument? (`U0` and `U4` are effectively the same argument already, but taking up twice as much memory)

tests/frontend/test_circuitspecs_chip2.py
co9olguy: I understood that part. I was just asking about why use `0` if the gate clearly has two wires?

tests/frontend/test_circuitspecs_chip2.py
co9olguy: Let's start with the second suggested test, then fix things in the circuitspec if the test is failing

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
josh146: Oh I see what you mean. I thought you meant one `Interferometer` call.

tests/frontend/test_circuitspecs_chip2.py
josh146: I don't remember :laughing: Programmatic shortcut? Probably a placeholder like you mentioned.

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
josh146: Good point. I suppose the logic for template matching would need to be modified to mark lines as optional? :thinking:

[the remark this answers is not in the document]
co9olguy: the measurement operator will need to be non-optional, and some (logically consistent) set of modes must be provided. But we should probably allow for the modes to be a legal a subset (not important for current time, but something to plan going forward)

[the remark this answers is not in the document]
josh146: I think this is looking complicated enough to be its own issue/PR? It also applies to other circuitspecs not touched in this PR.

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
josh146: I've made the logic here a lot smarter. Now, it explicitly checks the allowed mode tuples for the S2gates, and inserts `S2gate(0, 0)` where no S2gate is specified.

```python
# Check circuit begins with two mode squeezers
# --------------------------------------------
A, B, C = group_operations(seq, lambda x: isinstance(x, ops.S2gate))

regrefs = set()

if B:
    # get set of circuit registers as a tuple for each S2gate
    regrefs = {(cmd.reg[0].ind, cmd.reg[1].ind) for cmd in B}

# the set of allowed mode-tuples the S2gates must have
allowed_modes = set(zip(range(0, 4), range(4, 8)))

if not regrefs.issubset(allowed_modes):
    raise CircuitError("S2gates do not appear on the correct modes.")

# determine which modes do not have input S2gates specified
missing = allowed_modes - regrefs

for i, j in missing:
    # insert S2gates with 0 squeezing
    seq.insert(0, Command(ops.S2gate(0, 0), [registers[i], registers[j]]))
```

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
josh146: I had a think about this, and for now (because of the way SF is written), it is possible for a user to apply separate unitaries to modes [0,1,2,3] and [4,5,6,7]. For example:

```python
with prog.context as q:
    ops.S2gate(0.2, 0) | (q[0], q[4])
    ops.S2gate(0.3, 0) | (q[1], q[5])
    ops.S2gate(0.1, 0) | (q[2], q[6])
    ops.S2gate(0.4, 0) | (q[3], q[7])

    # unitary on modes [0,1,2,3]
    ops.Interferometer(U) | (q[0], q[1], q[2], q[3])

    # unitary on modes [4,5,6,7]
    ops.Interferometer(U) | (q[4], q[5], q[6], q[7])
    ops.BSgate() | (q[2], q[3])

    # measurements
    ops.MeasureFock() | q
```
Since the (tunable) squeezers couple these two sets of modes, we need to create this as a 8 mode program (not a 4 mode program), so I think this is unavoidable for now.

tests/frontend/test_circuitspecs_chip2.py
josh146: This is still an open issue, so resolving this comment for now.

tests/frontend/test_circuitspecs_chip2.py
josh146: Split this into two tests.

josh146:
@co9olguy, this should now be ready for re-review, with significantly improved S2gate logic. For the time being, the S2gates are restricted to have `r=1` or `r=0`. We can update this to the proper values once known.

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
co9olguy: But this program doesn't fit the circuitspec anyway, so why are we considering it?###

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -305,15 +376,16 @@ def test_no_unitary(self, tol):
         assert program_equivalence(res, expected, atol=tol)
 
     def test_interferometers(self, tol):
-        """Test interferometers correctly decompose to MZ gates"""
+        """Test that the compilation correctly decomposes the interferometer using
+        the rectangular_symmetric mesh (which decomposes to MZ gates)"""
```
co9olguy: just remove the mention of MZ gates, since it doesn't seem to be relevant for the actual test

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -135,17 +135,17 @@ class DummyCircuit(CircuitSpecs):
...
 
     def test_exact_template(self, tol):
         """Test compilation works for the exact circuit"""
         bb = blackbird.loads(Chip2Specs.circuit)
         bb = bb(
-            squeezing_amplitude_0=0.43,
-            squeezing_amplitude_1=0.65,
-            squeezing_amplitude_2=0.543,
-            squeezing_amplitude_3=0.665,
+            squeezing_amplitude_0=1,
```
co9olguy: to save effort later, can we just make the non-zero squeezing value a magic number at the top of the file?

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -200,45 +200,116 @@ def test_no_s2gates(self):
...
+            ops.Interferometer(U) | (q[0], q[1], q[2], q[3])
+            ops.Interferometer(U) | (q[4], q[5], q[6], q[7])
+            ops.MeasureFock() | q
+
+        res = prog.compile("chip2")
+        expected = expected.compile("chip2")
+        assert program_equivalence(res, expected, atol=tol)
+
+    def test_missing_s2gates(self, tol):
+        """Test identity S2gates are inserted"""
```
co9olguy: same docstring as above

strawberryfields/circuitspecs/chip2.py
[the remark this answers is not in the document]
co9olguy: nevermind, I think I see it now

tests/frontend/test_circuitspecs_chip2.py
josh146: :+1:

tests/frontend/test_circuitspecs_chip2.py
josh146: Done

PR #255, opened 2019-11-26
Polish docs

ixfoduap:
This PR makes a final polish to the apps layer docs


strawberryfields/gbs/sample.py
```diff
@@ -308,28 +308,15 @@ def vibronic(
...
->>> vibronic(t, U1, S, U2, alpha, 2, 0.0)
+>>> from strawberryfields.gbs import data, vibronic
+>>> formic = data.Formic()
+>>> w = formic.w
+>>> wp = formic.wp
+>>> Ud = formic.Ud
+>>> delta = formic.delta
+>>> T = 0
+>>> t, U1, r, U2, alpha = vibronic.gbs_params(w, wp, Ud, delta, T)
+>>> vibronic(t, U1, r, U2, alpha, 2, 0.0)
```
soranjh: 4 spaces (or tab) should be added to the beginning of these lines.

strawberryfields/gbs/points.py
```diff
@@ -26,7 +26,7 @@
 ---------------
 
 A point process is a mechanism that randomly generates points among a set of possible outcomes.
```
soranjh: I think this first sentence can be slightly modified. Something such as "A point process is a mechanism that generates random point patterns among a set of possible outcomes."

strawberryfields/gbs/sample.py
```diff
@@ -308,28 +308,15 @@ def vibronic(
 
     **Example usage:**
```
trbromley:
```suggestion
    **Example usage:**
```

strawberryfields/gbs/sample.py
```diff
@@ -308,28 +308,15 @@ def vibronic(
...
->>>     [0.5618538, -0.09931968, 0.04562272, 0.02158822, 0.35700706, 0.6614837, -0.326946],
->>>     [-0.16560687, -0.7608465, -0.25644606, -0.54317241, -0.12822903, 0.12809274, -0.00597384],
->>>     [0.01788782, 0.60430409, -0.19831443, -0.73270964, -0.06393682, 0.03376894, -0.23038293],
->>>     [0.78640978, -0.11133936, 0.03160537, -0.09188782, -0.43483738, -0.4018141, 0.09582698],
->>>     [-0.13664887, -0.11196486, 0.86353995, -0.19608061, -0.12313513, -0.08639263, -0.40251231],
->>>     [-0.12060103, -0.01169781, -0.33937036, 0.34662981, -0.49895371, 0.03257453, -0.70709135]])
->>> alpha = np.array(
->>>     [0.15938187, 0.10387399, 1.10301587, -0.26756921, 0.32194572, -0.24317402, 0.0436992])
->>> vibronic(t, U1, S, U2, alpha, 2, 0.0)
+>>> from strawberryfields.gbs import data, vibronic
```
trbromley:
```suggestion
```

We don't typically have import statements in these codeblocks

strawberryfields/gbs/sample.py
trbromley: I realised this codeblock might not work, as it's unclear whether `vibronic` is the module or the function. We implicitly assume in codeblocks that the user has done `from strawberryfields.gbs.currentmodule import *`, but I think this would be problematic here. How about we just do:
```suggestion
>>> sample.vibronic(t, U1, r, U2, alpha, 2, 0.0)
```

trbromley:
> Thanks for the corrections Juan Miguel. The two sentences of lines 296 and 297 of the sample.py module should separated in two lines: (otherwise, the first line appears in bold in the html file).
> 
> 296 """Generate samples for computing vibronic spectra.
> 
> 297 The following gates are applied to input vacuum states:

@soranjh, not sure, but I think the problem as instead the lack of indentation in the codeblock, causing the formatting to go off generally

strawberryfields/gbs/vibronic.py
```diff
@@ -94,8 +94,8 @@ def gbs_params(
         wp (array): normal mode frequencies of the final electronic state in units of
             :math:`\mbox{cm}^{-1}`
         Ud (array): Duschinsky matrix
-        delta (array): Displacement vector, with entries :math:`\delta_i=\sqrt{\omega_i/\hbar}d_i`,
-            and :math:`d_i` is the Duschinsky displacement
+        delta (array): Displacement vector, with entries :math:`\delta_i=\sqrt{\omega'_i/\hbar}d_i`,
+            and :math:`d` is the Duschinsky displacement
```
soranjh: :math:`d_i` is correct in the original file. I do not know why :math:`d` appears here.

PR #257, opened 2019-12-02
Add virtual backends

josh146:
**Context:** Currently, all backends must have a corresponding `BaseBackend` class to pass to the engine. However, this doesn't make sense for backends that might only be accessible remotely.

* Makes `chip0` and `chip2` virtual backends. If loaded (via `sf.Engine("chip2")`), then a `BaseBackend` class is created dynamically in the backend loader, and links it to the corresponding circuit spec.

**Benefits:** Avoids superfluous empty classes being maintained.

**Possible Drawbacks:** ?


strawberryfields/backends/__init__.py
```diff
@@ -46,8 +48,17 @@ def load_backend(name):
...
 
-    if name in supported_backends:
-        backend = supported_backends[name]()
+    if name in virtual_backends:
+        # Backend is a remote device/simulator, that has a
+        # defined circuit spec, but no local backend class.
+        # By convention, the short name and corresponding
+        # circuit spec are the same.
+        backend_attrs = {"short_name": name, "circuit_spec": name}
+        backend_class = type(name, (BaseBackend,), backend_attrs)
```
jswinarton: Nice

PR #259, opened 2019-12-05
Apps plotting: allow user to specify both x and y dimension of plots

trbromley:
Currently, users could only access square-layout plots. This remains the default, but users are now able to specify a rectangular layout.

### DIFF
--- a/strawberryfields/apps/plot.py
+++ b/strawberryfields/apps/plot.py
@@ -116,7 +116,7 @@ graph_node_size = 14
 subgraph_node_size = 16
 
 
-def graph(g: nx.Graph, s: Optional[list] = None, plot_size: int = 500):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


strawberryfields/apps/plot.py
```diff
@@ -125,7 +125,7 @@ def graph(g: nx.Graph, s: Optional[list] = None, plot_size: int = 500):  # pragm
     **Example usage:**
 
     >>> graph = nx.complete_graph(10)
-    >>> fig = graph(graph, [0, 1, 2, 3])
```
trbromley: @ixfoduap @soranjh One thing to note is that the codeblocks looked a bit odd - we had for example
```python
>>> fig = graph(graph, [0, 1, 2, 3])
```
for plotting. It looks clearer when we specify the module:
```python
>>> fig = plot.graph(graph, [0, 1, 2, 3])
```
Although we don't normally do this, I think it's fair to make an exception here.

PR #260, opened 2019-12-05
Apps: add module to SF __init__

trbromley:
We normally suggest users can import the applications layer through commands like:
```python
from strawberryfields.apps import clique
```

However, users may end up on occasion using:
```python
import strawberryfields as sf

sf.apps.clique.grow(...)
```

This PR allows for this.


trbromley:
> I approve, assuming you've already checked that
> `import strawberryfields as sf`
> `sf.apps.[module].[function]`
> works!

@ixfoduap, yes it works (and didn't before):
image

PR #261, opened 2019-12-05
Apps: Rename kernel to rbf_kernel

trbromley:
Based on user feedback, we have decided to rename the `kernel` function in the `points` module to `rbf_kernel`. This helps specify the type of kernel used, which is the radial basis function kernel.

### DIFF
--- a/strawberryfields/apps/points.py
+++ b/strawberryfields/apps/points.py
@@ -41,11 +41,22 @@ encodes information about the similarity between points. When this kernel matrix
 semidefinite, it is possible to use GBS to implement a *permanental* point process and employ
 fast classical algorithms to sample from the resulting distribution.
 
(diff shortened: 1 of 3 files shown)


strawberryfields/apps/points.py
```diff
@@ -41,11 +41,22 @@
 semidefinite, it is possible to use GBS to implement a *permanental* point process and employ
 fast classical algorithms to sample from the resulting distribution.
 
+One choice of kernel matrix is the radial basis function (RBF) kernel whose elements are computed
```
trbromley: @soranjh this involved a bit of a change to the docstrings, just wanted to check if you think this is ok?

soranjh: All good.

PR #263, opened 2019-12-10
Tf2 dev

ziofil:
I haven't produced the new docs, but I have updated the docstrings

Changes:
1) All TF tensor objects are now eager tensors, so there's no need for Sessions and feed_dicts. All computations should be inside a `tf.GradientTape` context if we want TF to be able to return gradients. There's no need for tapes otherwise.

Example: 
```python
def loss(result):
    return tf.abs(tf.linalg.norm(np.array([0,1,0,0,0]) - result))

vac = tf.constant([1,0,0,0,0], dtype=tf.complex64)
a = tf.Variable(0.1+0.1j, dtype=tf.complex64)

prog = sf.Program(1)
eng = sf.Engine('tf', backend_options={'cutoff_dim':5})

with prog.context as mode:
        sf.ops.Ket(vac) | mode[0]
        sf.ops.Dgate(a) | mode[0]
        
with tf.GradientTape(persistent=True) as tape:
    out = eng.run(prog)
    L = loss(out.state.ket())

tape.gradient(_loss, a)
>>> <tf.Tensor: shape=(), dtype=complex64, numpy=(-0.7301567+0.0073753153j)>
```

2) There is no distinction between Tensor and evaluated tensor, so there's no need for the `eval` kwarg anymore (tensors are always evaluated).

3) There are no graphs anymore, however a user can use the `@tf.function` decorator to compile their code if they wish to.
Example:
```python
@tf.function
def loss(result):
    return tf.abs(tf.linalg.norm(np.array([0,1,0,0,0]) - result))
```

4) Some tests have been modified to accommodate for new functionality and/or syntax. All tests pass on my machine (MacOS 10.15.1, python 3.7.3). I tried both with TF 2.0.0 and TF 2.1.0-dev20191203.


josh146:
(Closing this PR for now, as we work on a better way of merging these code changes in)

PR #270, opened 2019-12-17
Added Quadrature Values

Lucaman99:
**Context:**

- Adds the `p_quad_values` functions, which integrates across the Wigner distribution to find the quadrature probability density landscape for the p-quadrature.
- Adds the `x_quad_values`, which also integrates across the Wigner distribution to find the quadrature probability density landscape for the x-quadrature.

**Benefits:**

- This will allow users to construct probability distributions for the different quadratures, rather than only being able to obtain values sampled from Wigner function (which ranges over the entire x-p phase space).

**Possible Drawbacks:**

- I don't see any particular drawbacks to this feature.

**Related GitHub Issues:**


strawberryfields/backends/states.py
```diff
@@ -621,6 +622,23 @@ def wigner(self, mode, xvec, pvec):
 
         return W / (self._hbar)
 
+    def p_quad_values(self, mode, xvec, pvec):
+
+        W = BaseFockState.wigner(mode, xvec, pvec)
```
josh146:
```suggestion
        W = self.wigner(mode, xvec, pvec)
```

strawberryfields/backends/states.py
```diff
@@ -621,6 +622,23 @@ def wigner(self, mode, xvec, pvec):
...
+        y = []
+        for i in range(0, len(xvec)):
+            res = simps([W[i][k] for k in range(0, len(xvec))], xvec)
+            y.append(res)
+        return np.array(y)
+
+    def x_quad_values(self, mode, xvec, pvec):
+        W = BaseFockState.wigner(mode, xvec, pvec)
```
josh146:
```suggestion
        W = self.wigner(mode, xvec, pvec)
```

strawberryfields/backends/states.py
```diff
@@ -621,6 +622,23 @@ def wigner(self, mode, xvec, pvec):
 
         return W / (self._hbar)
 
+    def p_quad_values(self, mode, xvec, pvec):
+
+        W = BaseFockState.wigner(mode, xvec, pvec)
+        y = []
+        for i in range(0, len(xvec)):
+            res = simps([W[i][k] for k in range(0, len(xvec))], xvec)
```
josh146: Can this be changed to the following?
```suggestion
            res = simps(W[i, :len(xvec)], xvec)
```

This _should_ work since `W` is a NumPy array so supports fancier indexing (and should be faster than a list comprehension)

strawberryfields/backends/states.py
```diff
@@ -621,6 +622,23 @@ def wigner(self, mode, xvec, pvec):
...
+            y.append(res)
+        return np.array(y)
+
+    def x_quad_values(self, mode, xvec, pvec):
+        W = BaseFockState.wigner(mode, xvec, pvec)
+        y = []
+        for i in range(0, len(pvec)):
+            res = simps([W[k][i] for k in range(0, len(pvec))], pvec)
```
josh146: Same comment about the list comprehension as the comment above :)

strawberryfields/backends/states.py
```diff
@@ -621,6 +622,23 @@ def wigner(self, mode, xvec, pvec):
 
         return W / (self._hbar)
 
+    def p_quad_values(self, mode, xvec, pvec):
+
```
josh146: If you add a docstring here (see the other methods for the style used), it will automatically show up in the docs!

strawberryfields/backends/states.py
```diff
@@ -621,6 +622,23 @@ def wigner(self, mode, xvec, pvec):
...
+        W = BaseFockState.wigner(mode, xvec, pvec)
+        y = []
+        for i in range(0, len(xvec)):
+            res = simps([W[i][k] for k in range(0, len(xvec))], xvec)
+            y.append(res)
+        return np.array(y)
+
+    def x_quad_values(self, mode, xvec, pvec):
```
josh146: If you add a docstring here (see the other methods for the style used), it will automatically show up in the docs!

Lucaman99:
Hi @josh146! Thanks for the review, I'll make the changes and re-submit! You're right, it would make more sense to have this in `BaseState`, I'll move it over!

Lucaman99:
Hi @josh146! I just made the changes you requested! 😄

strawberryfields/backends/states.py
```diff
@@ -308,6 +309,48 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
+
+        Returns:
+            array: 1D array of size len(xvec), containing reduced x-quadrature
+            probability values for a specified range of x and p.
+        """
+
+        W = self.wigner(mode, xvec, pvec)
+        y = []
+        for i in range(0, len(xvec)):
+            res = simps([W[ :len(pvec), i], pvec)
```
josh146: Looks like there is a stray parenthesis here!
```suggestion
            res = simps(W[:len(pvec), i], pvec)
```

Lucaman99:
@josh146 Awesome! Would testing the vacuum state be sufficient? Maybe a sample coherent state (alpha = 1)?

Lucaman99:
@josh146 Would you like me to make a new file for the tests, or add them into one of the pre-existing test files?

josh146:
Both sound great - perhaps even the |1> state if you want to get fancy? Although it might involve hermite polynomials! I think a great place for these tests would be in `tests/backend/test_states_wigner.py`

Lucaman99:
Excellent, I'll add that stuff right away!

Lucaman99:
@josh146 Alright, I just added the tests for the |1> state, the vacuum state, and the alpha = 1 coherent state!

tests/backend/test_states_wigner.py
```diff
@@ -108,3 +108,68 @@ def test_two_mode_squeezed(setup_backend, hbar, tol):
...
+
+    #Exact probability distribution
+    def exact(a):
+        return np.sqrt(1/(np.pi * hbar)) * np.exp(-1*(a**2)/hbar)
+
+    exact_x = np.array([exact(x) for x in XVEC])
+    exact_p = np.array([exact(p) for p in XVEC])
+
+    assert np.allclose(x_vals, exact_x, atol=0.01, rtol=0)
+    assert np.allclose(p_vals, exact_p, atol=0.01, rtol=0)
```
josh146: Is there a reason you are not using the `tol` variable here?

tests/backend/test_states_wigner.py
```diff
@@ -108,3 +108,68 @@ def test_two_mode_squeezed(setup_backend, hbar, tol):
...
+    def x_exact(a):
+        return np.sqrt(1/(np.pi * hbar)) * np.exp(-1*((a - np.sqrt(2 * hbar))**2)/hbar)
+    def p_exact(a):
+        return np.sqrt(1 / (np.pi * hbar)) * np.exp(-1 * (a ** 2) / hbar)
+
+    exact_x = np.array([x_exact(x) for x in XVEC])
+    exact_p = np.array([p_exact(p) for p in XVEC])
+
+    assert np.allclose(x_vals, exact_x, atol=0.01, rtol=0)
+    assert np.allclose(p_vals, exact_p, atol=0.01, rtol=0)
```
josh146: Same question here with respect to `atol=tol`. If this is because the cutoff dimension of the Fock backend is causing numerical inaccuracy, you can always try reducing the value of alpha. Other than that, fantastic tests!

tests/backend/test_states_wigner.py
Lucaman99: Yeah, I was going to use `tol`, but I saw in previous tests of the Wigner function, the tolerance was just set to `0.01`, so I wasn't totally sure what to do. I'll change it to `tol`!

tests/backend/test_states_wigner.py
Lucaman99: Just to be safe, I'll reduce the value of alpha to `0.5` 😄

Lucaman99:
Alright, CHANGELOG.md has been updated!

PR #280, opened 2019-12-27
sympy.functions.atan2 has been fixed in sympy 1.5, workaround removed.

smite:
**Context:**

In sympy 1.4, the `atan2` function has a bug in its implementation.  In PR #163 `sf.ops.CXgate` is implemented using a decomposition that needs `atan2`, so a workaround was used. The workaround could not handle the case `s=0`, however. The bug has been fixed in sympy 1.5.

* Adds `sympy >= 1.5` as an installation requirement (this was missing before)
* Removes the workaround code
* Modifies the tests so they test the `s=0` case as well.

**Benefits:**

The code is simplified, and can also handle the `s=0` case.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,11 @@
-networkx>=2.0
 numpy>=1.17.4
 scipy>=1.0.0
-tensorflow==1.3
-tensorflow-tensorboard>=0.1.8
+sympy>=1.5
+networkx>=2.0
 quantum-blackbird>=0.2.0
 thewalrus>=0.10
 toml
 appdirs
+tensorflow==1.3
+tensorflow-tensorboard>=0.1.8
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,10 +1,11 @@
...
-tensorflow-tensorboard>=0.1.8
+sympy>=1.5
+networkx>=2.0
 quantum-blackbird>=0.2.0
 thewalrus>=0.10
 toml
 appdirs
+tensorflow==1.3
+tensorflow-tensorboard>=0.1.8
 plotly==4.4.1
```
smite: Now the requirements are in the same order as in `setup.py`

smite:
Don't merge this yet, I'll have to check something first.

PR #281, opened 2019-12-30
Moved CV conventions into ops.py docstrings

josh146:
**Context:**

We currently have information about the CV operations split among multiple pages:

* Conventions/Operators (information about creation/annhilation, quad operator conventions)
* Conventions/Gates (CV gate set definitions)
* Conventions/States (CV initial state definitions)
* Conventions/Decompositions (common CV decompositions)
* Conventions/Measurements (CV measurements)
* Conventions/Other (loss channels)
* `ops.py` docstrings (operator arguments)

This is confusing to navigate, and can cause users to end up on the wrong page. Furthermore, the `ops.py` docstrings currently have very little information in them.

Merges the conventions rst files with the ops.py docstrings, under a *Details and Conventions* section (collapsed by default). For example,

Screenshot_20191230_133629

In particular:

* Conventions/Operators is left as-is, as it is more technical/doesn't have any direct code correspondence.

* Everything in Conventions/Gates is moved to the corresponding `Gate` docstrings in `ops.py`

* Everything in Conventions/States, except the introduction on Gaussian states, is moved to the corresponding `Preparation` docstrings in `ops.py`

* Everything in Conventions/Decompositions, except the intro on Gaussian unitaries, is moved to the corresponding `Decomposition` docstrings in `ops.py`

* Everything in Conventions/Measurements is moved to the corresponding `Measurement` docstrings in `ops.py`

* Everything in Conventions/Other, except a very brief section on commutation relations, is moved to the corresponding `Channels` docstrings in `ops.py`.

**Benefits:**

* Now that each Python object has it's own page, it makes sense to collate all information regarding that object (including conventions) in the same place.

* Easier to find and navigate

**Still to do:**

What should be done with the remaining conventions pages? All that remains is:

* Annhilation/creation operators
* Quadrature operators
* Commutation relations
* Gaussian unitaries
* Gaussian states

These likely make up two new pages, one on operator conventions, and one on the Gaussian representation.

**Possible Drawbacks:** n/a


nquesada:
Maybe one could include the discussion on creation and annihilation operators, quadrature operators, commutation relations and Gaussian states and operations as a preface of `ops.py`. This will be helpful for people reading this files read things like "the `Dgate` is Gaussian"? Is that reasonable?

doc/xanadu_theme/static/xanadu.css_t
```diff
@@ -2206,10 +2206,10 @@ dl.class > dd > table.docutils {
     /*border-bottom: 1px solid lightgray;*/
 }
 
-dd > .collapse-header:first-of-type > h2 {
+/*dd > .collapse-header:first-of-type > h2 {
   margin-top: 70px;
 }
-
+*/
```
trbromley: What do these changes do to the theme?

strawberryfields/ops.py
```diff
@@ -966,16 +1352,77 @@ def _decompose(self, reg, **kwargs):
...
+
+        The squeeze gate affects the position and momentum operators as
+
+        .. math::
+            S^\dagger(z) \x_{\phi} S(z) = e^{-r}\x_{\phi}, ~~~ S^\dagger(z) \p_{\phi} S(z) = e^{r}\p_{\phi}
+
+        The Fock basis decomposition of displacement and squeezing operations was analysed
+        by Krall :cite:`kral1990`, and the following quantity was calculated,
+
+        .. math::
```
trbromley: Do we have any way to avoid crossing the background border?

image

trbromley:
What's the plan to fix the CodeFactor errors?

strawberryfields/ops.py
josh146: Only options I can think of are:

1. Remove the page shadow (similar to the walrus documentation: https://the-walrus.readthedocs.io)

2. Somehow add a line break in that last line of the latex?

josh146:
> What's the plan to fix the CodeFactor errors?

Two of the errors can be fixed by removing unneeded imports. The others were existing pylint warnings that have been picked up again, so I think it's safe to continue to ignore for now (out of scope to fix in this PR)

doc/xanadu_theme/static/xanadu.css_t
josh146: The first one adds a background colour to the `.. admonition: defn` box (previously was white, so it didn't display differently from the rest of the text). Second one removes a weird linebreak that occured between the 'Details and conventions' header and the rest of the docstring

trbromley: Thanks!

strawberryfields/ops.py
trbromley: Yeah I think we probably just need to be careful about equation width... but not worth sorting out now

josh146:
@trbromley, I've fixed the transition error occurring with module-level variables.

strawberryfields/ops.py
```diff
@@ -1245,16 +1913,46 @@ def _decompose(self, reg, **kwargs):
...
+
+            .. math::
+                \text{CZ}(s) =  \iint dx dy \: e^{i s x_1 x_2/\hbar }
+                \xket{x_1,x_2}\xbra{x_1,x_2} = \exp\left({i s \: \hat{x_1}
+                \otimes \hat{x_2} /\hbar}\right).
+
+        It is related to the addition gate by a :doc:`phase space rotation <strawberryfields.ops.Rgate>`
+        in the second mode:
+
+        ..math::
```
trbromley, line 1940:
```suggestion
        .. math::
```

Issue #271, opened 2019-12-17
issue when running the example quantum_neural_network.py


MichelNowak1:
#### Issue description

When the example of strawberryfields quantum_neural_network.py is run, it breaks when calling the Dgate in ops.py. After a git bissect, one finds that the problem comes from the commit:
 94036642f3008bfa7202774a51b907ba51f35138 Symbolic Operation parameters using SymPy.
It seems that the wrapper around the exp function is not working correctly.

* *Expected behavior:* quantum_neural_network.py should return (result thrown by SF when checking out to the first commit before the commit that broke the example):
```
Beginning optimization
Fidelity before optimization: 4.7938176e-08
Fidelity after optimization: 0.9851418

Target state: [0.+0.j 1.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j]
Output state: [ 0.   +0.001j  0.993-0.001j  0.   +0.001j -0.   -0.001j  0.   +0.j
 -0.   +0.j   ]
```

* *Actual behavior:* (What actually happens)
Nothing happens: see the traceback below

* *Reproduces how often:* Always

* *System information:*
Strawberry Fields: a Python library for continuous-variable quantum circuits.
Copyright 2018-2019 Xanadu Quantum Technologies Inc.
Python version:            3.6.8
Platform info:             Linux-5.0.0-37-generic-x86_64-with-Ubuntu-18.04-bionic
Installation path:         /home/michel/codes/strawberryfields/strawberryfields
Strawberry Fields version: 0.12.0
Numpy version:             1.17.4
Scipy version:             1.3.1
The Walrus version:           0.9.0
Blackbird version:         0.2.1
TensorFlow version:        1.3.0

#### Source code and tracebacks

```
Traceback (most recent call last):
  File "quantum_neural_network.py", line 99, in <module>
    results = eng.run(qnn, run_options={"eval": False})
  File "/home/michel/codes/strawberryfields/strawberryfields/engine.py", line 481, in run
    result = super()._run(program, args=args, compile_options=compile_options, **eng_run_options)
  File "/home/michel/codes/strawberryfields/strawberryfields/engine.py", line 349, in _run
    self._run_program(p, **kwargs)
  File "/home/michel/codes/strawberryfields/strawberryfields/engine.py", line 423, in _run_program
    cmd.op.apply(cmd.reg, self.backend, **kwargs)  # NOTE we could also handle storing measured vals here
[...]

PR #282, opened 2020-01-02
Symbolic parameter / TF compatibility, documentation

### LINKED ISSUE #161: `io.save` doesn't work for TensorFlow programs
Strawberry Fields allows creating programs using Python literals, NumPy arrays, and TensorFlow tensors as gate parameters.

However, if using TensorFlow tensors, this breaks the ability to save the program to a Blackbird script. For example, using `io.save`, results in the following output:

```python
Rgate(Tensor("strided_slice:0", shape=(), dtype=float32)) | 0
```

There are two ways to fix this:

1. If using TensorFlow 1.3, the conversion of the program to a Blackbird script must be done within an active TensorFlow session, in order to convert the tensors to arrays/numeric values.

2. If using a newer version of TensorFlow that supports eager execution, the tensor can be converted to a NumPy array via the `tensor.numpy()` method.

As the tensorflow backend currently only support TF v1.3, we should implement solution (1), and change it to solution (2) once the backend is updated to support TF 2.0.

smite:
**Context:**

The symbolic parameter system should be able to handle TensorFlow objects being passed through it,  without actually depending on TF. This can be done using the `tensorflow` printer in SymPy. SymPy requires TF 2.0 to work.

* Symbolic parameter syntax and use is described in the docs.
* TF objects can now appear as `Operation` parameters only by binding free parameters to them when the `Program` is run.
* TF-related symbolic parameter tests and examples are fixed and temporarily disabled, until SF has been migrated to TF 2.0.
* Symbolic parameters are now consistently evaluated right before the backend API call, not before.

**Benefits:**

Once the TF 2.0 migration is done, SF can again be used in symbolic TF computations.

**Related GitHub Issues:**

* Will eventually solve #271
* Closes #161 (because now TF objects can only enter a program as free symbolic parameters)

### DIFF
--- a/doc/introduction/circuits.rst
+++ b/doc/introduction/circuits.rst
@@ -25,7 +25,7 @@ Circuits
 
     circuits/glossary
 
-In Strawberry Fields, photonic quantum circuits are represented as :class:`~.Program`
+In Strawberry Fields, photonic quantum circuits are represented as :class:`~strawberryfields.Program`
 objects. By creating a program, quantum operations can be applied, measurements performed,
 and the program can then be simulated using the various Strawberry Fields backends.
 
@@ -37,7 +37,7 @@ and the program can then be simulated using the various Strawberry Fields backen
 Creating a quantum program
 --------------------------
 
-To construct a photonic quantum circuit in Strawberry Fields, a :class:`~.Program` object
+To construct a photonic quantum circuit in Strawberry Fields, a :class:`~strawberryfields.Program` object
 must be created, and operations applied.
 
 .. code-block:: python3
@@ -77,8 +77,8 @@ the same structure as the above example; in particular,
 
 .. note::
 
-    The contents of a program can be viewed by calling :meth:`.Program.print` method,
-    or output as a qcircuit :math:`\LaTeX{}` document by using the :meth:`.Program.draw_circuit`
+    The contents of a program can be viewed by calling the :meth:`~strawberryfields.Program.print` method,
+    or output as a qcircuit :math:`\LaTeX{}` document by using the :meth:`~strawberryfields.Program.draw_circuit`
     method.
 
 .. seealso::
@@ -178,26 +178,27 @@ for accessing the results of your program execution:
 Symbolic parameters
 -------------------
 
-The quantum operations can take both numerical and symbolic parameters. The latter fall into two types:
+The quantum operations can take both numerical and symbolic parameters.
+The latter fall into two types:
 
-* **Measured parameters**: Certain quantum programs require that
+* **Measured parameters**: Certain quantum programs (e.g. quantum teleportation) require that
   operations can be conditioned on measurement results obtained during the execution of the
   program. In this case the parameter value is not known until the measurement is made
-  (or simulated). The latest measurement result of qumode ``i`` is referred to as ``q[i].par``.
+  (or simulated). The latest measurement result of qumode ``i`` is available via ``q[i].par``.
 
 * **Free parameters**: A *parametrized circuit template* is a program that
-  depends on a number of free parameters. These parameters are bound to new fixed
+  depends on a number of free parameters. These parameters can be bound to new fixed
   values each time the program is executed.
-  The free parameters are created accessed using the :meth:`Program.params` method.
+  The free parameters are created and accessed using the
+  :meth:`~strawberryfields.Program.params` method.
 
-The symbolic parameters can be combined and transformed using basic algebraic operations and
-common mathematical functions in the ``strawberryfields.parameters.parfuncs`` namespace.
+The symbolic parameters can be combined and transformed using basic algebraic operations, and
+the mathematical functions in the :data:`strawberryfields.math` namespace.
 
 .. code-block:: python3
 
     import strawberryfields as sf
     from strawberryfields import ops
-    from strawberryfields.parameters import parfuncs as pf
 
     # create a 2-mode quantum program
     prog = sf.Program(2)
@@ -209,7 +210,7 @@ common mathematical functions in the ``strawberryfields.parameters.parfuncs`` na
     with prog.context as q:
         ops.Dgate(a ** 2)    | q[0]  # free parameter
         ops.MeasureX         | q[0]  # measure qumode 0, the result is used in the next operation
-        ops.Sgate(1 - pf.sin(q[0].par)) | q[1]  # measured parameter
+        ops.Sgate(1 - sf.math.sin(q[0].par)) | q[1]  # measured parameter
         ops.MeasureFock()    | q[1]
 
     # intialize the Fock backend
@@ -222,7 +223,7 @@ common mathematical functions in the ``strawberryfields.parameters.parfuncs`` na
 Compilation
 -----------
 
-The :class:`~.Program` object also provides *compilation* methods, that
+The :class:`~strawberryfields.Program` object also provides *compilation* methods, that
 automatically transform your circuit into an *equivalent* circuit with
 a particular layout or topology. For example, the ``gbs`` compile target will
 compile a circuit consisting of Gaussian operations and Fock measurements
--- a/doc/introduction/circuits.rst
+++ b/doc/introduction/circuits.rst
@@ -45,7 +45,7 @@ must be created, and operations applied.
     import strawberryfields as sf
     from strawberryfields import ops
 
-    # create a 3 mode quantum program
+    # create a 3-mode quantum program
     prog = sf.Program(3)
 
     with prog.context as q:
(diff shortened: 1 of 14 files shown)


tests/integration/test_tf_symbolic.py
```diff
@@ -90,20 +91,29 @@ def test_eng_run_eval_false_measurements_are_tensors(self, setup_eng):
...
         eng, prog = setup_eng(1)
-        x = prog.params('a')  # free parameter
+        x, y = prog.params('a', 'phi')  # free parameters
         with prog.context as q:
-            Dgate(x) | q
+            Dgate(x, y) | q
 
-        state = eng.run(prog, args={'a': a}, run_options=tf_params).state
+        # use TF to evaluate symbolic parameter expressions
+        strawberryfields.parameters.par_evaluate.lambdify_printer = ['tensorflow']
```
smite: Here we change the symbolic parameter lambdify printer to one that can handle TF objects, and after running the program we change it back. I'm not 100% sure if it's a good idea to leave this to the user though.

Another possibility would be to add a check in `Engine.run` to see if the `'session'` run_option is present, which I think is required to use TF objects as parameters, and switch to the tensorflow printer just for that run if this is the case.

doc/introduction/circuits.rst
```diff
@@ -171,6 +175,50 @@ for accessing the results of your program execution:
...
 
+Symbolic parameters
+-------------------
+
+The quantum operations can take both numerical and symbolic parameters. The latter fall into two types:
+
+* **Measured parameters**: Certain quantum programs require that
+  operations can be conditioned on measurement results obtained during the execution of the
+  program. In this case the parameter value is not known until the measurement is made
+  (or simulated). The latest measurement result of qumode ``i`` is referred to as ``q[i].par``.
```
josh146: Is referred the right word here? 'denoted', or even 'available via ``q[i].par``'?

doc/introduction/circuits.rst
```diff
@@ -171,6 +175,50 @@ for accessing the results of your program execution:
...
+
+* **Free parameters**: A *parametrized circuit template* is a program that
+  depends on a number of free parameters. These parameters are bound to new fixed
+  values each time the program is executed.
+  The free parameters are created accessed using the :meth:`Program.params` method.
+
+The symbolic parameters can be combined and transformed using basic algebraic operations and
+common mathematical functions in the ``strawberryfields.parameters.parfuncs`` namespace.
```
josh146: I'm not sure I really like the name `parfuncs`. At the same time, the `strawberryfields.parameters` module is not  user-facing apart from this one import.

I think we should import this to top level, and give it a better name, how about the following?

```python
ops.Sgate(1 - sf.math.sin(q[0].par))
```

This matches nicely with TF 2.0, where a lot of the previous top level functions are now in the `math` module (i.e., `tf.math.log`).

strawberryfields/ops.py
```diff
@@ -23,7 +23,7 @@
 import numpy as np
 
 from scipy.linalg import block_diag
-from scipy.special import factorial as fac
+import scipy.special as ssp
```
josh146: How come this was changed?

strawberryfields/ops.py
```diff
@@ -625,28 +625,29 @@ def __init__(self, alpha=0, p=0):
...
-        phi = np.pi * p[1]
+        alpha = self.p[0]
+        phi = np.pi * self.p[1]
         D = backend.get_cutoff_dim()
         l = np.arange(D)[:, np.newaxis]
 
         # normalization constant
-        temp = np.exp(-0.5 * np.abs(alpha)**2)
-        N = temp / np.sqrt(2*(1 + np.cos(phi) * temp**4))
+        temp = pf.exp(-0.5 * pf.Abs(alpha)**2)
```
josh146: Is it really `sympy.Abs`? Weird :thinking:

strawberryfields/parameters.py
```diff
@@ -179,6 +179,9 @@ def do_evaluate(p):
         return ret[0]
     return ret
 
+# default printer for lambdify to use (we also use 'tensorflow' for TF objects)
+par_evaluate.lambdify_printer = ['numpy']
```
josh146: How would this be modified to allow it to work with both TF and numpy? Currently it looks like something that must be set manually.

strawberryfields/program.py
```diff
@@ -202,6 +111,9 @@ class Program:
             ops.BSgate(0.43, 0.1) | (q[1], q[2])
             ops.MeasureFock() | q
 
+    The program objects keep track of the state of the quantum register they act on, using a dictionary of :class:`RegRef` objects.
```
josh146: Since this is very user facing, I would suggest removing this line (quite advanced, and might confuse if they click on `RegRef`).

doc/introduction/circuits.rst
smite: Would `sf.math` be a module, or would `math` be a `SimpleNamespace` object inside `sf`? The latter is how `parfuncs` works now.

josh146: I was imagining just a `SimpleNamespace` object. However, it would be nice if documentation and autocomplete could continue to work, maybe it needs `__dir__` defined?

strawberryfields/ops.py
smite: I think most style guides recommend importing entire modules over importing stuff from modules, as it makes the code more readable. `factorial` is only used in one method so it does not even add much extra characters.

https://docs.python-guide.org/writing/structure/#modules

strawberryfields/ops.py
smite: It's to make sure it's not mixed up with the Python built-in function `abs`. It's the same thing with `min` and `max`.

strawberryfields/parameters.py
smite: Right now it is. I was planning to change it so that `par_evaluate` switches to the `tensorflow` printer if any of the `vals` are TF objects. I'm kinda leaning towards simply checking if `type(x).__module__` begins with `'tensorflow'`.

strawberryfields/program.py
smite: OK, but wasn't the new convention that the basic user-facing docs are rst files in the `doc` directory (e.g. `doc/introduction/*.rst`), and the docstrings in the Python files are for the API docs for devs and more advanced users? What's the correct place for the slightly lower-level explanations on how the code works?

doc/introduction/circuits.rst
smite: Available might be the best choice

doc/introduction/circuits.rst
smite: Looks like a `SimpleNamespace` object works nicely, it supports autocompletion and the docs are inherited from the wrapped Sympy funcs.

strawberryfields/parameters.py
josh146: :+1: Would be good to have this commented out in the code with a todo

strawberryfields/program.py
josh146: I think the best place are comments in the source code itself

doc/introduction/circuits.rst
josh146:
```suggestion
  (or simulated). The latest measurement result of qumode ``i`` is available via ``q[i].par``.
```

strawberryfields/parameters.py
smite: I can also change it right away, would that be better?

josh146: yep, that would be perfect

doc/introduction/circuits.rst
```diff
@@ -25,7 +25,7 @@ Circuits
 
     circuits/glossary
 
-In Strawberry Fields, photonic quantum circuits are represented as :class:`~.Program`
+In Strawberry Fields, photonic quantum circuits are represented as :class:`~strawberryfields.Program`
```
smite: This is to get rid of an error message when building the docs (when module A does `from B import X`, Sphinx finds the class X in both A and B and gets confused :/ )

strawberryfields/parameters.py
```diff
@@ -125,7 +125,7 @@ def wrapper(*args):
     return wrapper
 
 #: SimpleNamespace: Namespace of mathematical functions for manipulating Parameters. Consists of all :mod:`sympy.functions` public members, which we wrap with :func:`wrap_mathfunc`.
-parfuncs = types.SimpleNamespace(**{name: wrap_mathfunc(getattr(sf, name)) for name in dir(sf) if name[0] != '_'})
+par_funcs = types.SimpleNamespace(**{name: wrap_mathfunc(getattr(sf, name)) for name in dir(sf) if name[0] != '_'})
```
smite: For consistency.

doc/introduction/circuits.rst
smite: Okay, it's done.

strawberryfields/parameters.py
smite: Done.

PR #283, opened 2020-01-05
TF2 compatibility 

ziofil:
**Context:**
Minimal TFBackend update to TF2 (no new features, just the bare necessary to pass the tests with TF2)

1) Updated from deprecated TF1 notation to new TF2 notation (e.g. `tf.log` -> `tf.math.log`) and removed graph, session, eval, feed_dict (also from docstrings), but left **kwargs in tfbackend/states.py (even if now unused) to avoid breaking user code.

2) Minimal updates to tests to reflect new notation. When we will hook up the old TF1 backend, we will add the old tests and mark them to be run only under TF1.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/backends/tfbackend/states.py
```diff
@@ -213,158 +156,116 @@ def fidelity(self, other_state, mode, **kwargs):
...
-
-                * If this contains the key
-                  ``eval``, then the corresponding argument will be used to determine the return behaviour of this function.
-                  When ``eval=True``, the return value is numerical; when ``eval=False``, it is symbolic.
-                * If eval is not present in kwargs, then state falls back to the an internal evaluation behaviour,
-                  which is specified at initialization.
-                * A Tensorflow Session or feed_dict may also be passed via the keys ``session`` or ``feed_dict``, respectively.
-                  If a Session is supplied, then ``eval`` is overriden and the numerical evaluation takes place in the provided Session.
-                  If session and/or feed_dict are not given, then a temporary session and/or empty feed_dict will be used.
+            **kwargs: Optional keyword arguments. (Unused)
```
josh146: same here

strawberryfields/backends/tfbackend/states.py
```diff
@@ -373,40 +274,29 @@ def reduced_dm(self, modes, **kwargs):
...
```
josh146: same here

strawberryfields/backends/tfbackend/states.py
```diff
@@ -415,169 +305,125 @@ def quad_expectation(self, mode, phi=0., **kwargs):
...
```
josh146: can remove

strawberryfields/backends/tfbackend/states.py
```diff
@@ -415,169 +305,125 @@ def quad_expectation(self, mode, phi=0., **kwargs):
...
```
josh146: Can remove from the docstring

strawberryfields/backends/tfbackend/backend.py
```diff
@@ -337,18 +307,8 @@ def measure_homodyne(self, phi, mode, shots=1, select=None, **kwargs):
...
-                    v = vac_elem
-                else:
-                    sess = tf.Session()
-                    v = sess.run(vac_elem)
-                    sess.close()
-
-            result = np.abs(v-1) <= tol
-        return result
+        vac_elem = self.circuit.vacuum_element()
+        return np.abs(vac_elem-1) <= tol
```
smite: Why is `tf.name_scope` dropped here?

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -684,8 +647,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
                 phi (float): phase angle of quadrature to measure
                 mode (int): which mode to measure.
                 select (float): user-specified measurement value (used instead of random sampling)
-                **kwargs: can be used to pass a session or a feed_dict. Otherwise a temporary session
-                and no feed_dict will be used.
+                **kwargs:
 
         Returns:
             The measured value (or a list of measured values when running in batch mode).
```
smite: It'd be nice if the incomplete/wrong docstrings would be fixed when each method is updated, e.g. the Returns: statement should be of the format `type: explanation`

strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -154,42 +149,42 @@
...
     #pylint: disable=unused-argument
     print(value)
 
 
-if not (tf_available and tf_version[:3] == "1.3"):
+if not (tf_available and tf_version[:2] == "2."):
     sys.excepthook = excepthook
 
-    if sys.version_info[1] > 6:
+    if sys.version_info[1] not in {5, 6, 7}:
```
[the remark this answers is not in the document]
ziofil: No, and 3.7 is quite recent actually...

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -540,139 +511,131 @@ def measure_fock(self, modes, select=None, **kwargs):
...
+                removed_ctr = 0
+                red_state_is_pure = self._state_is_pure
+                for m in range(self._num_modes):
+                    if m not in modes:
+                        new_mode_idx = m - removed_ctr
+                        reduced_state = ops.partial_trace(reduced_state, new_mode_idx, red_state_is_pure, self._batched)
+                        red_state_is_pure = False
+                        removed_ctr += 1
+                # go from bra_A,ket_A,bra_B,ket_B,... -> bra_A,bra_B,ket_A,ket_B,... since this is what diag_part expects
+                # workaround for getting multi-index diagonal since tensorflow doesn't support getting diag of more than one subsystem at once
```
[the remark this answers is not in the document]
ziofil: I'm not sure what you're asking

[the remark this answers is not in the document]
josh146: The part where it says "since tensorflow doesn't support getting diag of more than one subsystem at once". I think this comment was written when TF 1.3 was being used, but maybe TF2.0 now supports this behaviour?

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -709,140 +671,132 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
...
+                    eqn_rhs += conditional_indices[mode_size * cond_ctr : mode_size * (cond_ctr + 1)]
+                    cond_ctr += 1
+            eqn = eqn_lhs + "->" + batch_index + eqn_rhs
+            new_state = tf.einsum(eqn, meas_mode_vac, normalized_conditional_state)
 
-                self._update_state(new_state)
+            self._update_state(new_state)
 
-            return meas_result
+        return meas_result.numpy()
```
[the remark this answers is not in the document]
ziofil: In the previous version it was a float (if I remember correctly the tests require that).

strawberryfields/backends/tfbackend/ops.py
```diff
@@ -316,11 +316,12 @@ def beamsplitter_matrix(t, r, D, batched=False, save=False, directory=None):
...
-        # if cached prefactors live on another graph, we'll have to reload them into this graph.
-        # In future versions, if 'copy_variable_to_graph' comes out of contrib, consider using that
-        get_prefac_tensor.cache_clear()
-        prefac = get_prefac_tensor(D, directory, save)
+    # NOTE: I don't think this is necessary anymore in TF2, but to avoid issues 
+    # if prefac.graph != phase.graph:
+    #     # if cached prefactors live on another graph, we'll have to reload them into this graph.
+    #     # In future versions, if 'copy_variable_to_graph' comes out of contrib, consider using that
+    #     get_prefac_tensor.cache_clear()
+    #     prefac = get_prefac_tensor(D, directory, save)
```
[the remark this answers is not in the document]
ziofil: I wasn't sure, hence the note. If it's no longer needed let's remove it.

strawberryfields/backends/tfbackend/states.py
```diff
@@ -654,41 +500,19 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs): # pragma: no c
...
-                self._data = session.run(self._data, feed_dict=feed_dict)
-                if close_session:
-                    session.close()
-                return super().poly_quad_expectation(A, d, k, phi, **kwargs)
-            else:
-                raise NotImplementedError("Calculation of multi-mode quadratic expectation values is currently "
-                                          "only supported when batched=False and when evaluating numerically.")
+
+        if not self.batched:
+            return super().poly_quad_expectation(A, d, k, phi, **kwargs)
```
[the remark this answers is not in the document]
ziofil: I don't know. I just made it TF2-compatible, which may come with hidden gifts such as stuff now being trainable.

strawberryfields/backends/tfbackend/states.py
```diff
@@ -606,11 +452,11 @@ def wigner(self, mode, xvec, pvec):
             array: 2D array of size [len(xvec), len(pvec)], containing reduced Wigner function
             values for specified x and p values.
         """
-        if self._eval and not self.batched:
+        if not self.batched:
             return super().wigner(mode, xvec, pvec)
 
         raise NotImplementedError("Calculation of the Wigner function is currently "
-                                  "only supported when eval=True and batched=False")
+                                  "only supported when batched=False")
```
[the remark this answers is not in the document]
ziofil: It should be trainable, yes. Whether this is efficient/convenient it's a different matter!

strawberryfields/engine.py
```diff
@@ -384,7 +378,7 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
         temp_run_options.setdefault('modes', None)
 
         # avoid unexpected keys being sent to Operations
-        eng_run_keys = ["eval", "session", "feed_dict", "shots"]
+        eng_run_keys = ["shots"] #NOTE: used to contain "session", "feed_dict" and "eval" (from TF 1.x)
```
co9olguy: I think we can just drop the note here

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -709,140 +669,132 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
...
 
-                # normalize
-                if self._state_is_pure:
-                    norm = tf.norm(tf.reshape(conditional_state, [batch_size, -1]), axis=1)
+            # normalize
+            if self._state_is_pure:
+                norm = tf.norm(tf.reshape(conditional_state, [batch_size, -1]), axis=1)
+            else:
+                # calculate norm of conditional_state
+                # cheap hack since tensorflow doesn't allow einsum equation for trace:
```
co9olguy: Is this supported in TF2 now?

strawberryfields/backends/tfbackend/circuit.py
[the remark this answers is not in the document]
co9olguy: I think this type was previously kept flexible, i.e., whether it was a tensor or an array was determined by the `eval` argument. See `meas_result = homodyne_sample.eval(feed_dict, session)` on line 782 of previous version..

Most tests use `eval=True`, so they expect a numpy array. The tests in `test_tf_symbolic` would have checked the other cases, where a `tf.Tensor` was returned.

Obviously the old tf symbolic stuff has gone by the wayside. But what are the consequences (in particular I'm thinking for differentiability) of casting to numpy explicitly here? Maybe there are none, since the measurement result being stochastic precludes the worry about it needing to be differentiable?

strawberryfields/backends/tfbackend/ops.py
```diff
@@ -37,7 +37,7 @@
 
 from strawberryfields.backends.shared_ops import generate_bs_factors, load_bs_factors, save_bs_factors, squeeze_parity
 
-def_type = tf.complex64
+def_type = tf.complex64 #NOTE: what if a user wants higher accuracy?
```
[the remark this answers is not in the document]
co9olguy: Sounds good, a user might want higher or lower precision. I do seem to recall that some tests might not pass on lower precisions though :thinking:

strawberryfields/backends/tfbackend/ops.py
```diff
@@ -113,7 +113,7 @@ def get_prefac_tensor(D, directory, save):
 def squeezed_vacuum_vector(r, theta, D, batched=False, eps=1e-32):
     """returns the ket representing a single mode squeezed vacuum state"""
     if batched:
-        batch_size = r.shape[0].value
+        batch_size = r.shape[0]
```
co9olguy: Nice that it works this way now :+1:

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -360,121 +343,110 @@ def prepare_multimode(self, state, modes, input_state_is_pure=False):
...
+            vac_component = tf.reshape(self._state, [-1])[0]
 
-            if self._state_is_pure:
-                return tf.abs(vac_component) ** 2
-            return vac_component
+        if self._state_is_pure:
+            return tf.abs(vac_component) ** 2
+        return vac_component
 
     def measure_fock(self, modes, select=None, **kwargs):
```
co9olguy: Just a hunch, but I have a feeling the measurement ops will give us the biggest headaches when we push for tf.autograph compatibility

strawberryfields/backends/tfbackend/states.py
```diff
@@ -170,41 +121,30 @@ def all_fock_probs(self, **kwargs):
...
-                probs = tf.squeeze(probs, 0)  # drop fake batch dimension
+        if not self.batched:
+            probs = tf.squeeze(probs, 0)  # drop fake batch dimension
 
-            probs = tf.identity(probs, name="all_fock_probs")
-            probs = self._run(probs, **kwargs)
+        probs = tf.identity(probs, name="all_fock_probs")
 
-            return probs
+        return probs.numpy()
```
co9olguy: Unlike the measurements, where casting back to numpy is probably fine, here I feel like this might break differentiability?

strawberryfields/backends/tfbackend/states.py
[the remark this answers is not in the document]
co9olguy: How is it trainable? Doesn't the call to `super().wigner()` fall back to a numpy implementation?

tests/backend/test_tf_import.py
```diff
@@ -35,23 +35,23 @@ class TestBackendImport:
 
     def test_incorrect_tf_version(self, monkeypatch):
         """Test that an exception is raised if the version
-        of TensorFlow installed is not version 1.3"""
+        of TensorFlow installed is not version 1.x"""
```
co9olguy:
```suggestion
        of TensorFlow installed is not version 2.x"""
```

tests/backend/test_tf_import.py
```diff
@@ -71,24 +71,24 @@ class TestFrontendImport:
 
     def test_incorrect_tf_version(self, monkeypatch):
         """Test that an exception is raised if the version
-        of TensorFlow installed is not version 1.3"""
+        of TensorFlow installed is version 2.x"""
```
co9olguy: docstring is not correct

tests/frontend/test_ops_gate.py
```diff
@@ -172,7 +172,7 @@ def test_merge_measured_pars():
         ops.MeasureX | q[0]
         mpar = q[0].par  # measured parameter
         D = ops.Dgate(mpar)
-        F = ops.Dgate(1.0)
+        F = ops.Dgate(1)
```
co9olguy: What was the issue that required a change here?

tests/integration/test_utils_integration.py
```diff
@@ -451,15 +448,19 @@ def test_extract_arbitrary_unitary_two_modes_vectorized(
...
-        else:
-            final_state = U @ initial_state.reshape([-1])
+            U = U.numpy()
 
-        expected_state = eng_ref.run([p0, prog]).state.ket().reshape([-1])
-        assert np.allclose(final_state, expected_state, atol=tol, rtol=0)
+        final_state = U @ initial_state.reshape([-1])
+        expected_state = eng_ref.run([p0, prog]).state.ket()
+        
+        if isinstance(expected_state, tf.Tensor):
```
co9olguy: Since this is a test, shouldn't we be asserting that `expected_state` has a particular type? :thinking:

tests/integration/test_utils_integration.py
```diff
@@ -451,15 +448,19 @@ def test_extract_arbitrary_unitary_two_modes_vectorized(
...
+            expected_state = expected_state.numpy()
+
+        if bsize >= 2:
+            for exp_state in expected_state:
+                assert np.allclose(final_state, exp_state.reshape([-1]), atol=tol, rtol=0)
```
co9olguy: I seem to remember that, since the tests just broadcast arrays over the batch axis if `bsize>1`, then a call to `np.allclose` still worked. How come this specific test needed an adjustment, but the same pattern elsewhere in the codebase works without change?

strawberryfields/engine.py
ziofil: done

strawberryfields/backends/tfbackend/circuit.py
ziofil: Apparently not yet, I tried on a small example and I get:
`ValueError: Subscript not supported: an axis appears more than once: ijkkll`

strawberryfields/backends/tfbackend/ops.py
ziofil: The call to .numpy() made the tests pass, so I assumed it was the required behaviour.
If we want a different behaviour then we should revise the tests first.
But I agree with the consideration that measurement results are by definition not trainable.

strawberryfields/backends/tfbackend/circuit.py
ziofil: How so?
In any case, the best way to be safe I think is to use stock TF functions wherever possible.

strawberryfields/backends/tfbackend/states.py
ziofil: If I remember correctly also in this case the tests would not pass without the call to .numpy().
I can try again and perhaps modify the tests (i.e. call .numpy() there for example)?

tests/backend/test_tf_import.py
ziofil: docstring fixed.
Regarding the wigner function, if it does fall back to a numpy implementation then it's not trainable indeed.

tests/frontend/test_ops_gate.py
ziofil: a few lines below there's the assert: `assert F.merge(D).p[0] == mpar + 1`, which fails if the argument of Dgate is a float (+1 <-- int). Either we change one or the other.

tests/integration/test_utils_integration.py
ziofil: Isn't `expected_state` a tf.Tensor only when we use the tfbackend?

tests/integration/test_utils_integration.py
ziofil: All the other tests don't have a `batch_size`.
The problem is that reshape([-1]) destroys the batch dimension, so if it's there we need the for loop.

tests/integration/test_utils_integration.py
co9olguy: You're right, my mistake

strawberryfields/backends/tfbackend/circuit.py
josh146: Just a future thought, but if we go down the python `@tf.custom_gradient` route with the gates, autograph won't be supported. I think you need to implement the TF API directly in C++ to get autograph support.

tests/integration/test_utils_integration.py
co9olguy: Was this test working previously without the for loop? :thinking:

tests/frontend/test_ops_gate.py
co9olguy: - Was this test working before? What changed so that line now causes failing tests?
- Since the natural argument to `Dgate` is a float, wouldn't we want to test that case? (i.e., changing to int here might mask a bug which a user would experience during natural use)

tests/integration/test_utils_integration.py
ziofil: For some reason now it works. Mystery.

tests/frontend/test_ops_gate.py
ziofil: I have changed the argument of the Dgate to a float, and the others accordingly. I don't know if they were passing before. They pass now.

strawberryfields/backends/tfbackend/ops.py
ziofil: I have pushed the numpy call to the tests (only when the tensors are tf.Tensor objects). It works well, so I think problem solved 👍

strawberryfields/backends/tfbackend/circuit.py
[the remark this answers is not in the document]
co9olguy: I'm a bit worried about adding `.numpy()` within the backend itself. All stuff related to types (numpy arrays or tf tensors) should be in the user's hands. Previously we allowed the measurement results to be returned as symbolic tensors (when the user asked for it). I'd like to keep that functionality for tf2 (except the tensors would no longer be symbolic by default). 

If a test is failing, it is likely because it was written with different assumptions in mind. I suspect it might even be a bug or implicit assumption in the `Parameters` class (which was not written with tf2 in mind). In that case, it's better to let the test fail as an indication that we need to fix/revisit it, rather than forcing the backend itself to have a more restricted behaviour.

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -794,7 +794,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
 
             self._update_state(new_state)
 
-        return meas_result
+        return meas_result.numpy()
```
co9olguy:
```suggestion
        return meas_result
```

josh146:
@ziofil, just double checking the status of this PR. It has approval to merge, but it looks like the tests stopped passing with the latest commits

co9olguy:
Tests are failing intentionally here. @ziofil had submitted a version where all tests are passing. I believe the tests are actually checking the wrong thing now (they assume measurement results have to be numpy arrays, but they should be allowed to be not tf.tensors---i.e., the native type), so I reverted @ziofil's fix to remind us to update the tests

josh146:
> Tests are failing intentionally here.

We could mark them as `@pytest.mark.xfail`, and merge this into the `tf2` branch. This allows them to be fixed in a new PR (which might be easier)

ziofil:
Sounds good to me. I should check the tests again (now I'm on battery power), but if I remember well there might be many that fail (all of those that are using the Fock probabilities).

co9olguy:
Ready to merge in everyone? (can correct `xfail`ing tests in future PR)

josh146:
Yep!

ziofil:
👍

PR #289, opened 2020-01-30
S2gate to fock

thisac:
**Context:**
There was no primitive two-mode squeezing gate (S2 gate) in Strawberry Fields, only being able to be executed by decomposition. A direct implementation is available in The Walrus and thus the S2 gate can be implemented as a primitive instead.

Added the S2 gate to the fock backend, calling the S2gate method in The Walrus, modified the fock circuit spec to include the gate and added an apply method to the S2 gate in the backend api. Also moved the S2 gate from decompostions to primitives in the fockspecs.

**Benefits:**
The S2 gate is now available as a primitive, and runs fast due to calling the The Walrus method.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -421,9 +421,9 @@ def two_mode_squeezing(r, theta, trunc):
     r"""The two-mode squeezing operator :math:`S_2(re^{i\theta})`.
 
     Args:
-            r (float): two-mode squeezing magnitude
-            theta (float): two-mode squeezing phase
-            trunc (int): Fock ladder cutoff
+        r (float): two-mode squeezing magnitude
+        theta (float): two-mode squeezing phase
+        trunc (int): Fock ladder cutoff
     """
     ret, _, _ = S2gate(r, theta, cutoff=trunc)
 
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -28,6 +28,7 @@ from scipy.special import factorial as fac
 from scipy.linalg import expm as matrixExp
 
 from strawberryfields.backends import shared_ops as so
+from thewalrus.fock_gradients import S2gate
 
 
 def_type = np.complex128
@@ -411,6 +412,22 @@ def squeezing(r, theta, trunc, save=False, directory=None):
         scale = mask * np.power(sinh(r)/2, mask*(N+n-2*k)/2) / (cosh(r)**((N+n+1)/2))
         ph = exp(1j*theta*(N-n)/2)
         ret = np.sum(scale*ph*prefac, axis=-1)
+
+    return ret
+
+
+@functools.lru_cache()
+def two_mode_sqeezing(r, theta, trunc):
+    r"""The two-mode squeezing operator :math:`S_2(re^{i\theta})`.
+
+    Args:
+            r (float): two-mode squeezing magnitude
+            theta (float): two-mode squeezing phase
+            trunc (int): Fock ladder cutoff
+    """
+    ret, _, _ = S2gate(r, theta, cutoff=trunc)
+
+    ret = np.transpose(ret, [0, 2, 1, 3])
 
     return ret
 
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -417,7 +417,7 @@ def squeezing(r, theta, trunc, save=False, directory=None):
 
 
 @functools.lru_cache()
-def two_mode_sqeezing(r, theta, trunc):
+def two_mode_squeezing(r, theta, trunc):
     r"""The two-mode squeezing operator :math:`S_2(re^{i\theta})`.
 
     Args:
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1143,7 +1143,7 @@ class S2gate(Gate):
     def _apply(self, reg, backend, **kwargs):
         p = self.p[0] * pf.exp(1j * self.p[1])
         z = par_evaluate(p)
-        backend.two_mode_squeezing(z, *reg)
+        backend.two_mode_squeeze(z, *reg)
 
     def _decompose(self, reg, **kwargs):
         # two opposite squeezers sandwiched between 50% beamsplitters
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -18,8 +18,6 @@ The syntax is modeled after ProjectQ :cite:`projectq2016`.
 """
 from collections.abc import Sequence
 import copy
-import types
-import sys
 import warnings
 
 import numpy as np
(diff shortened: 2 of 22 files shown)


josh146:
:muscle:

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -415,6 +416,22 @@ def squeezing(r, theta, trunc, save=False, directory=None):
...
+    r"""The two-mode squeezing operator :math:`S_2(re^{i\theta})`.
+
+    Args:
+            r (float): two-mode squeezing magnitude
+            theta (float): two-mode squeezing phase
+            trunc (int): Fock ladder cutoff
+    """
+    ret, _, _ = S2gate(r, theta, cutoff=trunc)
+
+    ret = np.transpose(ret, [0, 2, 1, 3])
```
josh146: I suppose this requires a new PR to the walrus, so that the transposition is no longer needed? @nquesada

strawberryfields/ops.py
```diff
@@ -1140,6 +1140,11 @@ class S2gate(Gate):
     def __init__(self, r, phi=0.):
         super().__init__([r, phi])
 
+    def _apply(self, reg, backend, **kwargs):
+        p = self.p[0] * pf.exp(1j * self.p[1])
+        z = par_evaluate(p)
```
josh146: We really should fix up the Strawberry Fields conventions, we are doing too many unnecessary conversions :laughing:

tests/backend/test_twomode_squeezing_operation.py
```diff
@@ -0,0 +1,70 @@
...
+PHASE = np.linspace(0, 2 * np.pi, 3, endpoint=False)
+MAG = np.linspace(0.0, 0.2, 5, endpoint=False)
+
+
+def get_amplitude(k, r, p):
+    """ Get amplitude for two mode squeezing operation"""
+    if k[0] == k[1]:
+        tmsv = (np.exp(1j * p) * np.tanh(r)) ** k[0] / np.cosh(r)
+        return tmsv
+    return 0.0
```
josh146: Nice!

tests/backend/test_twomode_squeezing_operation.py
```diff
@@ -0,0 +1,70 @@
...
+
+        state = backend.state()
+
+        if state.is_pure:
+            for k in it.product(range(cutoff), repeat=2):
+                if k[0] == k[1]:
+                    tmsv = get_amplitude(k, r, p)
+                    assert np.allclose(state.data[k], tmsv, atol=tol, rtol=0)
+                else:
+                    assert np.allclose(state.data[k], 0, atol=tol, rtol=0)
```
josh146: It looks like the get amplitudes function already returns zero when k0 = k1, so is this if/else statement needed here?

tests/backend/test_twomode_squeezing_operation.py
```diff
@@ -0,0 +1,70 @@
...
+        else:
+            for k in it.product(range(cutoff), repeat=2):
+                for l in it.product(range(cutoff), repeat=2):
+                    t = (k[0], l[0], k[1], l[1])
+                    tmsv2 = get_amplitude(k, r, p) * np.conj(get_amplitude(l, r, p))
+
+                    assert np.allclose(state.data[t], tmsv2, atol=tol, rtol=0)
```
josh146: This is a great test @thisac :+1:

tests/backend/test_twomode_squeezing_operation.py
```diff
@@ -0,0 +1,70 @@
...
+        :math: `\delta_{kl} \frac{e^{in\phi} \tanh^n{r}}{\cosh{r}}`
+        """
+        z = r * np.exp(1j * p)
+
+        backend = setup_backend(2)
+        backend.two_mode_squeeze(z, 0, 1)
+
+        state = backend.state()
+
+        if state.is_pure:
```
josh146: It might be better to include the `pure` fixture in this test function, and then do `if pure:` instead. This way if the test fails, the pytest traceback message will indicate whether it failed for pure=True or pure=False.

tests/backend/test_twomode_squeezing_operation.py
```diff
@@ -0,0 +1,70 @@
...
+@pytest.mark.backends("fock")
+class TestTwomodeSqueezing:
+    """Tests two-mode squeezing operation"""
+
+    @pytest.mark.parametrize("r", MAG)
+    @pytest.mark.parametrize("p", PHASE)
+    def test_two_mode_squeezing(self, setup_backend, r, p, cutoff, tol):
+        r""" Test two-mode squeezing on vacuum-state for both pure states and
+        mixed states with the amplitude given by
+        :math: `\delta_{kl} \frac{e^{in\phi} \tanh^n{r}}{\cosh{r}}`
```
josh146:
```suggestion
	        :math:`\delta_{kl} \frac{e^{in\phi} \tanh^n{r}}{\cosh{r}}`
```
(I don't think whitespace is allowed when defining Sphinx domains)

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -415,6 +416,22 @@ def squeezing(r, theta, trunc, save=False, directory=None):
...
 
 
+@functools.lru_cache()
+def two_mode_sqeezing(r, theta, trunc):
+    r"""The two-mode squeezing operator :math:`S_2(re^{i\theta})`.
+
+    Args:
+            r (float): two-mode squeezing magnitude
+            theta (float): two-mode squeezing phase
+            trunc (int): Fock ladder cutoff
```
josh146: Seems to be too much indentation here, should only be 4 space indentation after the `Args:`.

If you want to test how the Strawberry Fields documentation looks locally, you can run
```bash
pip install -e doc/requirements.txt
make docs
```

That being said, I don't think this file is currently being displayed in the documentation!

tests/backend/test_twomode_squeezing_operation.py
josh146: It might be worth adding to the docstring a description of the arguments and return type

thisac: Added docstrings with args and return

tests/backend/test_twomode_squeezing_operation.py
thisac: You're right! I've removed the unnecessary statement. Thanks!

tests/backend/test_twomode_squeezing_operation.py
thisac: Makes sense. Done!

strawberryfields/backends/fockbackend/ops.py
thisac: Thanks for spotting this. I'm not sure why I indented it so much. It's fixed now! Thanks for the make docs. I'll check it out.

thisac:
The same issue as before was found (unnecessary "else" after "return"). Should I fix it?

strawberryfields/backends/fockbackend/ops.py
nquesada: Yes, it would require a new PR to the Walrus. And I definitely agree, it would be great to do some timing comparisons to see the gain.

strawberryfields/ops.py
nquesada: Yes, this would be great, write every gate explicitly in terms of two real parameters.

josh146:
> The same issue as before was found (unnecessary "else" after "return"). Should I fix it?

It might be good to fix it, it looks like when I turn it off, it comes back with new commits 🤔

PR #292, opened 2020-02-11
Custom contraction for beamsplitter, two-mode squeezing and diagonal operators

### LINKED ISSUE #291: Custom contraction for beamsplitter and two-mode squeezing
The beamsplitter  and two-mode squeezing operation have special symmetries, namely SU(2) and SU(1,1). This implies that their matrix elements B_{nm}^{kl}, T_{ij}^{kl} satisfy the selection rules

image

This implies that one can contract pure states (without loss of generality we assume only two modes) using one less loop as follows
image
and 
image
where c_{ij} is the tensor representing the two mode state and n is the cutoff in Fock space.
Similarly for a mixed state one should be able to do the update of the state using two less summations.

These observations should result in significant savings when acting beamsplitters and two-mode squeezing in the Fock backend.

thisac:
**Context:**
The beamsplitter  and two-mode squeezing operation have special symmetries, namely SU(2) and SU(1,1). This implies that their matrix elements B_{nm}^{kl}, T_{ij}^{kl} satisfy the selection rules

image

This implies that one can contract pure states (without loss of generality we assume only two modes) using one less loop as follows
image
and 
image
where c_{ij} is the tensor representing the two mode state and n is the cutoff in Fock space.
Similarly for a mixed state one should be able to do the update of the state using two less summations.

These observations should result in significant savings when acting beamsplitters and two-mode squeezing in the Fock backend.

Added a new apply_twomode_gate function to the fockbackend that can be used instead of apply_gate_BLAS for faster tensor contraction calculations for the beamsplitter gate and the two-mode squeeze gate. It is powered by the numba compiler making it much faster than before.

The beamsplitter and the two-mode squeeze functions in the circuit class now calls the apply_twomode_gate function in the fockbackend.

Add functionality to apply_gate_BLAS that checks if the operator is diagonal (fast) and, if so, performs faster applications of the gates.

**Note:** the new tests for the mode subsets (e.g. running the BSgate on modes [1, 3] in a 4-mode circuit) do not work with the tensorflow backend, which I believe is a problem with the tf-backend and not the tests. That's why the beamsplitter test is in a separate class although it would fit into the already existing TestFockRepresentation class.

**Benefits:**
Much faster calculations of the tensor contractions for the beamsplitter, the two-mode squeeze gate and all diagonal (single-mode and multimode) operations.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
closes #291

### DIFF
--- a/strawberryfields/backends/fockbackend/circuit.py
+++ b/strawberryfields/backends/fockbackend/circuit.py
@@ -23,6 +23,7 @@ from itertools import product
 import numpy as np
 from numpy import sqrt, pi
 from scipy.special import factorial as bang
+from numba.typed import List
 
 from . import ops
 
@@ -314,7 +315,14 @@ class Circuit():
         """
         Applies a beamsplitter.
         """
-        self._apply_gate(ops.beamsplitter(t, r, phi, self._trunc), [mode1, mode2])
+        mat = ops.beamsplitter(t, r, phi, self._trunc)
+
+        modes = List()
+        modes.append(mode1)
+        modes.append(mode2)
+
+        args = [mat, self._state, self._pure, modes, self._num_modes, self._trunc]
+        self._state = ops.apply_twomode_gate(*args, gate="BSgate")
 
     def squeeze(self, r, theta, mode):
         """
@@ -326,7 +334,15 @@ class Circuit():
         """
         Applies a two-mode squeezing gate.
         """
-        self._apply_gate(ops.two_mode_squeezing(r, theta, self._trunc), [mode1, mode2])
+        # self._apply_gate(ops.two_mode_squeezing(r, theta, self._trunc), [mode1, mode2])
+        mat = ops.two_mode_squeezing(r, theta, self._trunc)
+
+        modes = List()
+        modes.append(mode1)
+        modes.append(mode2)
+
+        args = [mat, self._state, self._pure, modes, self._num_modes, self._trunc]
+        self._state = ops.apply_twomode_gate(*args, gate="S2gate")
 
     def kerr_interaction(self, kappa, mode):
         """
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -348,7 +348,7 @@ def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
         mat (array[complex]): The numeric operator to be applied to the state, of shape `[trunc]*(2*n)`
         state (array[complex]): The state that the operator is applied to
         pure (bool): whether the state is pure or mixed
-        modes (list[int]): A list of modes to which the BS is applied
+        modes (list[int]): the list of modes to which the operator is applied on
         n (int): The total number of modes
         trunc (int): The Hilbert space truncation/cutoff
         gate (str): the gate which should be called (BSgate, S2gate)
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -358,6 +358,9 @@ def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
     Returns:
         array[complex]: state after application of the two-mode operation
     """
+    # the transpositions below are necessary for using the jitted apply-gate
+    # functions by moving and grouping the modes on which the gate is applied
+    # to the front indices, since the other modes can then simply be ignored
     if pure:
         t1 = modes[0]
         t2 = modes[1]
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -351,7 +351,8 @@ def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
         modes (list[int]): the list of modes to which the operator is applied on
         n (int): The total number of modes
         trunc (int): The Hilbert space truncation/cutoff
-        gate (str): the gate which should be called (BSgate, S2gate)
+        gate (str): The gate that is being applied. This argument determines the selection rules that
+            are used. Options are ``"BSgate"`` and ``"S2gate"``.
 
     Returns:
         ndarray: State where the two-mode operation has been applied
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -355,7 +355,7 @@ def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
             are used. Options are ``"BSgate"`` and ``"S2gate"``.
 
     Returns:
-        ndarray: State where the two-mode operation has been applied
+        array[complex]: state after application of the two-mode operation
     """
     if pure:
         t1 = modes[0]
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -253,7 +253,8 @@ def apply_gate_BLAS(mat, state, pure, modes, n, trunc):
         ret = np.zeros([trunc for i in range(n)], dtype=def_type)
         for i in product(*([range(trunc) for j in range(n - size)])):
             if diag:
-                ret[i] = np.multiply(matview.diagonal(), view[i].ravel()).reshape(stshape)
+                mat_diag = matview.diagonal()
+                ret[i] = np.multiply(mat_diag, view[i].ravel()).reshape(stshape)
             else:
                 ret[i] = np.dot(matview, view[i].ravel()).reshape(stshape)
 
@@ -281,7 +282,7 @@ def apply_gate_BLAS(mat, state, pure, modes, n, trunc):
     ret = np.zeros([trunc for i in range(n*2)], dtype=def_type)
     for i in product(*([range(trunc) for j in range((n - size)*2)])):
         if diag:
-            mat_diag = mat.diagonal().reshape(-1, 1)
+            mat_diag = matview.diagonal().reshape(-1, 1)
             ret[i] = np.multiply(mat_diag, np.multiply(view[i].reshape((dim, dim)), dagger(mat_diag))).reshape(stshape + stshape)
         else:
             ret[i] = np.dot(matview, np.dot(view[i].reshape((dim, dim)), dagger(matview))).reshape(stshape + stshape)
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -224,12 +224,6 @@ def apply_gate_BLAS(mat, state, pure, modes, n, trunc):
     if reshaping is efficient this should be faster.
     """
 
-    # checks if mat is diagonal
-    if mat.ndim == 2:
-        diag = np.all(mat == np.diag(np.diagonal(mat)))
-    else:
-        diag = False
-
     size = len(modes)
     dim = trunc**size
     stshape = [trunc for i in range(size)]
@@ -238,6 +232,9 @@ def apply_gate_BLAS(mat, state, pure, modes, n, trunc):
     # |m1><m1| |m2><m2| ... |mn><mn| -> |m1>|m2>...|mn><m1|<m2|...<mn|
     transpose_list = [2*i for i in range(size)] + [2*i + 1 for i in range(size)]
     matview = np.transpose(mat, transpose_list).reshape((dim, dim))
+
+    # checks if matview is diagonal and, if so, use faster contractions
+    diag = np.all(matview == np.diag(np.diagonal(matview)))
 
     if pure:
         if n == 1:
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -342,7 +342,7 @@ def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
     """Applies a two-mode gate to a state.
 
     Applies the specified two-mode gate to the state using custom tensor contractions and
-    the numba compiler for faster application.
+    the Numba compiler for faster application.
 
     Args:
         mat (ndarray): The BS operator to be applied to the state
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -341,7 +341,7 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
 def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
     """Applies a two-mode gate to a state.
 
-    Applies the two-mode gate to the state using custom tensor contractions and
+    Applies the specified two-mode gate to the state using custom tensor contractions and
     the numba compiler for faster application.
 
     Args:
--- a/strawberryfields/backends/fockbackend/ops.py
+++ b/strawberryfields/backends/fockbackend/ops.py
@@ -339,7 +339,7 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
 
 
 def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
-    """Applies a two-mode gate to a state
+    """Applies a two-mode gate to a state.
 
     Applies the two-mode gate to the state using custom tensor contractions and
     the numba compiler for faster application.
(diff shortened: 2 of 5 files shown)


strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -314,7 +315,14 @@ def beamsplitter(self, t, r, phi, mode1, mode2):
         """
         Applies a beamsplitter.
         """
-        self._apply_gate(ops.beamsplitter(t, r, phi, self._trunc), [mode1, mode2])
+        mat = ops.beamsplitter(t, r, phi, self._trunc)
+
+        modes = List()
+        modes.append(mode1)
+        modes.append(mode2)
```
josh146: How come `List` needs to be imported from `numba.typed`, and the modes constructed in this way? Might be worth adding a couple of comments explaining, for other developers.

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -314,7 +315,14 @@ def beamsplitter(self, t, r, phi, mode1, mode2):
...
+
+        args = [mat, self._state, self._pure, modes, self._num_modes, self._trunc]
+        self._state = ops.apply_twomode_gate(*args, gate="BSgate")
```
josh146: Based on the arguments being passed to `apply_twomode_gate`, it sounds like it should be a method in the `Circuit` class 🙂 This way, you won't need to pass `self._state`, `self._pure`, `self._num_modes`, `self._trunc`

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -231,8 +234,14 @@ def apply_gate_BLAS(mat, state, pure, modes, n, trunc):
...
     matview = np.transpose(mat, transpose_list).reshape((dim, dim))
 
+    # checks if matview is diagonal and, if so, use faster contractions
+    diag = np.all(matview == np.diag(np.diagonal(matview)))
+
     if pure:
         if n == 1:
+            if diag:
+                mat_diag = mat.diagonal()
+                return np.multiply(mat_diag, state)
```
josh146: Nice!

josh146: Does this cause specific speedup in certain use-cases?

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
     return np.einsum(einstring, mat, state, mat.conj())
 
 
+def apply_twomode_gate(mat, state, pure, modes, n, trunc, gate="BSgate"):
+    """Applies a two-mode gate to a state
```
josh146:
```suggestion
    """Applies a two-mode gate to a state.
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+
+    Applies the two-mode gate to the state using custom tensor contractions and
```
josh146:
```suggestion
    Applies the specified two-mode gate to the state using custom tensor contractions and
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+    the numba compiler for faster application.
```
josh146:
```suggestion
    the Numba compiler for faster application.
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+
```
josh146: Would be worth adding a note mentioning that only the `BSgate` and `S2gate` are currently supported.

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+    Args:
+        mat (ndarray): The BS operator to be applied to the state
```
josh146:
```suggestion
        mat (array[complex]): The numeric operator to be applied to the state, of shape `[trunc]*(2*n)`
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+        state (ndarray): The state that the BS is applied to
```
josh146:
```suggestion
        state (array[complex]): The state that the operator is applied to
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+        pure (bool): If the state is pure or mixed
```
josh146:
```suggestion
        pure (bool): whether the state is pure or mixed
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+        modes (list[int]): A list of modes to which the BS is applied
```
josh146:
```suggestion
        modes (list[int]): the list of modes to which the operator is applied on
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+        n (int): The total number of modes
+        trunc (int): The Hilbert space truncation/cutoff
+        gate (str): the gate which should be called (BSgate, S2gate)
```
josh146:
```suggestion
        gate (str): The gate that is being applied. This argument determines the selection rules that
            are used. Options are ``"BSgate"`` and ``"S2gate"``.
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+
+    Returns:
+        ndarray: State where the two-mode operation has been applied
```
josh146:
```suggestion
        array[complex]: state after application of the two-mode operation
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+    """
+    if pure:
+        t1 = modes[0]
+        t2 = modes[1]
+
+        # put the ket-values in front to be operated on in the apply function
```
josh146: It might be good to expand on this comment --- what was the previous dimension ordering, and what is the new one? Just to avoid issues with indexing as we had before due to bad documentation!

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+
+        state = state.transpose(switch_list_1)
+        state = state.transpose(switch_list_2)
+
+        if gate == "BSgate":
+            state = _apply_BS(mat, state, trunc)
+        elif gate == "S2gate":
+            state = _apply_S2(mat, state, trunc)
+        else:
+            raise NotImplementedError
```
josh146: Good practice to specify why the error is raised:
```suggestion
            raise NotImplementedError("Currently, selection rules are only implemented for the BSgate "
                                      "and the S2gate. The {} gate is not supported".format(gate)")
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+        switch_list_1[[0, 1, t1, t1+1]] = switch_list_1[[t1, t1+1, 0, 1]]
+        switch_list_2[[0, 1, t2, t2+1]] = switch_list_2[[t2, t2+1, 0, 1]]
+
+
+        # put bra-values to the left, and ket-values to the right (ignoring values not operated on)
+        transpose_list = np.arange(2 * n)
+        transpose_list[[t1+1, t2]] = transpose_list[[t2, t1+1]]
+
+        state = state.transpose(transpose_list)
+        state = state.transpose(switch_list_1)
```
josh146: Same here re: more commenting 🙂 Highly vectorized code can be a bit obfuscated, even for ourselves when we re-visit the same code we wrote later!

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+            state = state.transpose(switch_list_1)
+
+            state = state.transpose(switch_list_2)
+            state = _apply_S2(mat.conj(), state, trunc)
+
+        else:
+            raise NotImplementedError
+
+        state = state.transpose(switch_list_2)
+        ret = state.transpose(transpose_list)
```
josh146: What transposition is happening here?

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+    """
+    if pure:
+        t1 = modes[0]
+        t2 = modes[1]
+
+        # put the ket-values in front to be operated on in the apply function
+        switch_list_1 = np.arange(n)
+        switch_list_2 = np.arange(n)
+        switch_list_1[[0, t1]] = switch_list_1[[t1, 0]]
+        switch_list_2[[1, t2]] = switch_list_2[[t2, 1]]
```
josh146: Are these transpositions temporary and required due to the differing conventions in the walrus? If so, they can be removed later, so we can add a TODO.

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+
+        state = state.transpose(switch_list_2)
+        ret = state.transpose(transpose_list)
+
+    return ret
+
+
+@jit(nopython=True)
+def _apply_BS(mat, state, trunc):
+    """Applies the BS gate to the first bra in state"""
```
josh146: May as well expand this docstring with the details you added in the PR description! Nice to have it in the code itself for later reference.

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,109 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+    for i in range(trunc):
+        for j in range(trunc):
+            for k in range(max(1+i+j-trunc, 0), min(i+j, trunc-1) + 1):
+                ret[i, j] += mat[i, k, j, i+j-k] * state[k, i+j-k]
+    return ret
+
+
+@jit(nopython=True)
+def _apply_S2(mat, state, trunc):
+    """Applies the S2 gate to the first bra in state"""
```
josh146: Same comment here. The math can even be included using the :math:<backtick>\sum<backtick> notation.

tests/backend/test_beamsplitter_operation.py
```diff
@@ -151,3 +153,55 @@ def test_coherent_vacuum_interfered_fock_elements(
...
+class TestModeSubsets:
+    """Tests on mode subsets using the Fock basis representation."""
+
+    @pytest.mark.parametrize("t", T_VALUES)
+    @pytest.mark.parametrize("mag_alpha", MAG_ALPHAS[1:])
+    @pytest.mark.parametrize("r_phi", PHASE_R)
+    def test_beamsplitter_on_mode_subset(
+            self, setup_backend, mag_alpha, t, r_phi, cutoff, pure, tol
+    ):
+        """Tests that apply the BS on different mode subsets."""
```
josh146:
```suggestion
        """Tests applying the beamsplitter on different mode subsets."""
```

tests/backend/test_beamsplitter_operation.py
```diff
@@ -151,3 +153,55 @@ def test_coherent_vacuum_interfered_fock_elements(
...
+
+        phase_alpha = np.pi / 5
+        alpha = mag_alpha * np.exp(1j * phase_alpha)
+        r = np.exp(1j * r_phi) * np.sqrt(1.0 - np.abs(t) ** 2)
+
+        n_modes = 4
+        for modes in it.combinations(range(n_modes), 2):
```
josh146: This for loop can be moved into a `@pytest.mark.parametrize()` decorator. The advantage of doing this is that pytest will treat each parametrized loop iteration as a separate test, and failed tracebacks will include the mode list it failed on.

tests/backend/test_twomode_squeezing_operation.py
```diff
@@ -74,3 +74,28 @@ def test_two_mode_squeezing(self, setup_backend, r, p, cutoff, pure, tol):
...
+        for modes in it.combinations(range(n_modes), 2):
+            backend = setup_backend(n_modes)
+            backend.two_mode_squeeze(z, *modes)
+
+            state = backend.state()
+
+            state_data = state.reduced_dm(list(modes))
+
+            for k in it.product(range(cutoff), repeat=2):
+                for l in it.product(range(cutoff), repeat=2):
```
josh146: Same comment here re: parametrize 🙂 Note that 'stacking' decorators is equivalent to nesting for loops.

josh146: One downside here though: this would require performing the `two_mode_squeeze` method many more times

strawberryfields/backends/fockbackend/ops.py
nquesada: yup, for the Kerr gate and the rotation gate since they are diagonal.

strawberryfields/backends/fockbackend/ops.py
nquesada: I don't think these ones are related to The Walrus

strawberryfields/backends/fockbackend/ops.py
josh146: Awesome 🎆

strawberryfields/backends/fockbackend/circuit.py
thisac: It's needed for Numba to know the type of the parameter passed to the jitted function. I'll add some comments about it.

strawberryfields/backends/fockbackend/circuit.py
thisac: I agree about that, although I put the `apply_twomode_gate` in ops.py since all other tensor contraction functions are there (`apply_gate_blas` and `apply_gate_einsum`). All three have the same arguments. I'm not sure if separating them is a good idea.

thisac: Another option would be to put all tensor contractions in the `Circuit` class, since they seem to only be called from within there. I don't know if that would brake something else though. :slightly_smiling_face:

strawberryfields/backends/fockbackend/ops.py
thisac: They're not related to The Walrus. `switch_list` is there to move the modes on which the gates should be applied to the front indices of `state`, so that `apply_BS_gate` and `apply_S2_gate` can be used (they'll apply the gates to the first two modes in `state`). Then the modes are moved back to their original indices.

strawberryfields/backends/fockbackend/ops.py
thisac: For mixed states the ket-indices are put next to one-another (and the same for the bra-indices), so that the same `apply_BS_gate` and `apply_S2_gate` can be used for mixed states as for pure states. The `transpose_list` puts the modes on which the gate should be applied together.

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,157 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+        else:
+            raise NotImplementedError
+
+        state = state.transpose(switch_list_2)
+        ret = state.transpose(transpose_list)
+
+    return ret
+
+
+# ignored in covtest (doesn't work well with the jit decorator)
```
josh146: :+1:

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -318,6 +338,157 @@ def apply_gate_einsum(mat, state, pure, modes, n, trunc):
...
+
+        elif gate == "S2gate":
+            state = _apply_S2(mat, state, trunc)
+            state = state.transpose(switch_list_1)
+
+            state = state.transpose(switch_list_2)
+            state = _apply_S2(mat.conj(), state, trunc)
+
+        else:
+            raise NotImplementedError
```
josh146: I think one of my previous suggested changes was missed here (see https://github.com/XanaduAI/strawberryfields/pull/292#discussion_r378606418)

PR #293, opened 2020-02-14
Fockbackend apply-gate cleanup

thisac:
**Context:**
Calling the apply-gate functions in `ops.py` meant passing along many `Circuit` class attributes, meaning that it would be more efficient to put these functions in the Circuit class directly. They weren't called from anywhere else than inside the `Circuit` class.

Futhermore, `apply_gate_einsum` was no longer in use, in turn causing the `Circuit._apply_gate` to become superfluous.

Moved `apply_gate_BLAS`, `apply_twomode_gate`, `_apply_BS_gate` and `_apply_S2_gate` to the `Circuit` class, and removed `apply_gate_einsum` and `Circuit._apply_gate`.

**Benefits:**
Cleaner code

**Possible Drawbacks:**
If `apply_gate_BLAS` is used as a function somewhere outside of SF it would no longer work (since it is now a `Circuit` class method).

**Related GitHub Issues:**
None


strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -135,6 +113,249 @@ def reset(self, pure=None, cutoff_dim=None, num_subsystems=None):
...
+            n (int): The total number of modes
+
+        Returns:
+            array[complex]: The state after application of the two-mode operation
+        """
+        # pylint: disable=too-many-branches
+
+        state = kwargs.get("state", self._state)
+        pure = kwargs.get("pure", self._pure)
+        n = kwargs.get("n", self._num_modes)
```
antalszava: Currently, I'm still grasping how the internals of SF work, what would be the use cases when these would be passed as a keyword argument (and not default to the values of `self._state`, `self._pure` and `self._num_modes`)?

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -96,16 +80,10 @@ def _apply_channel(self, kraus_ops, modes):
...
             self._state = np.zeros([self._trunc for i in range(self._num_modes*2)], dtype=ops.def_type)
-        elif self._mode == 'blas':
-            states = [ops.apply_gate_einsum(k, np.copy(self._state), False, modes, self._num_modes, self._trunc)\
-                                    for k in kraus_ops]
-            self._state = sum(states)
-        elif self._mode == 'einsum':
-            states = [ops.apply_gate_einsum(k, self._state, False, modes, self._num_modes, self._trunc)\
-                                    for k in kraus_ops]
+        else:
+            states = [self.apply_gate_BLAS(k, modes, pure=False) for k in kraus_ops]
```
josh146: much cleaner!

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -135,6 +113,249 @@ def reset(self, pure=None, cutoff_dim=None, num_subsystems=None):
...
+
+        Gate application based on custom indexing and matrix multiplication.
+        Assumes the input matrix has shape (out1, in1, ...).
+
+        This implementation uses indexing and BLAS. As per stack overflow,
+        einsum doesn't actually use BLAS but rather a c implementation. In theory
+        if reshaping is efficient this should be faster.
+
+        Args:
+            mat (array[complex]): The numeric operator to be applied to the state, of shape `[trunc]*(2*n)`
```
josh146:
```suggestion
            mat (array[complex]): The numeric operator to be applied to the state, of shape ``[trunc]*(2*n)``
```

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -135,6 +113,249 @@ def reset(self, pure=None, cutoff_dim=None, num_subsystems=None):
...
+        size = len(modes)
+        dim = self._trunc**size
+        stshape = [self._trunc for i in range(size)]
+
+        # Apply the following matrix transposition:
+        # |m1><m1| |m2><m2| ... |mn><mn| -> |m1>|m2>...|mn><m1|<m2|...<mn|
+        transpose_list = [2*i for i in range(size)] + [2*i + 1 for i in range(size)]
+        matview = np.transpose(mat, transpose_list).reshape((dim, dim))
+
+        # checks if matview is diagonal and, if so, use faster contractions
```
josh146:
```suggestion
        # checks if matview is diagonal (for example, Kerr/rotation gates) and, if so, use faster contractions
```

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -135,6 +113,249 @@ def reset(self, pure=None, cutoff_dim=None, num_subsystems=None):
...
+            for i in product(*([range(self._trunc) for j in range(n - size)])):
+                if diag:
+                    mat_diag = matview.diagonal()
+                    ret[i] = np.multiply(mat_diag, view[i].ravel()).reshape(stshape)
+                else:
+                    ret[i] = np.dot(matview, view[i].ravel()).reshape(stshape)
+
+            # "untranspose" the return matrix ret
+            untranspose_list = [0] * len(transpose_list)
+            for i in range(len(transpose_list)): # pylint: disable=consider-using-enumerate
```
josh146: Just curious, was this pylint statement in the original version? I'm wondering why we don't instead just change it to
```python
for i, t in enumerate(transpose_list):
    untranspose_list[t] = i
```

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -492,11 +716,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
             reduced = ops.partial_trace(state, self._num_modes, unmeasured)
 
             # Rotate to measurement basis
-            args = [ops.phase(-phi, self._trunc), reduced, False, [0], 1, self._trunc]
-            if self._mode == 'blas':
-                reduced = ops.apply_gate_BLAS(*args)
-            elif self._mode == 'einsum':
-                reduced = ops.apply_gate_einsum(*args)
+            reduced = self.apply_gate_BLAS(ops.phase(-phi, self._trunc), [0], state=reduced, pure=False, n=1)
```
josh146: Makes this much cleaner and more maintainable :slightly_smiling_face:

strawberryfields/backends/fockbackend/circuit.py
thisac: That was there before (I didn't add it). I'm not sure why enumerate isn't used instead; that would be much cleaner. Should I try to change it as part of this PR?

strawberryfields/backends/fockbackend/circuit.py
thisac: There are some calls to `apply_gate_BLAS` that specify other values for `state`, `pure` and `n` than the the corresponding class attributes in `measure_homodyne`. I haven't looked into it further though.

PR #294, opened 2020-02-19
Refactor remote job engine to Engine/Connection API

pauljxtan:
**Context:**

Strawberry Fields provides engines for computing quantum programs against local and remote (currently on the `demo` branch) backends. The purpose of this PR is to refactor and simplify the existing functionality on `demo` to an API structured around two core objects, an `Engine` and a `Connection`, to facilitate envisioned job execution workflows.

- Refactor the existing `StarshipEngine` and `APIClient` to define interfaces at two distinct levels of abstraction: a higher-level `StarshipEngine` that provides the minimal methods needed to run a job (more intended for users), and a lower-level `Connection` that provides more flexibility around job management (more intended for developers or automated testing environments).
- The `StarshipEngine` allows remote jobs to be run synchronously (via `engine.run()`) or asynchronously (via `engine.run_async()`). In the former mode, the engine blocks until the job completes, then returns the final result. In the latter mode, the engine immediately returns a job object to be manually queried and refreshed by the user.
- The `Connection` exposes additional remote platform operations, such as cancelling a job in progress or directly querying its status or result. The `StarshipEngine` serves as a high-level wrapper around this object, but the `Connection` can also be instantiated and used directly, independent of the engine.
- The `starship` CLI is refactored to use the new classes, and also simplified to a more minimal interface.

N.B. The changed files are auto-formatted with `black` (default settings), which produces a number of purely cosmetic diffs.

**Benefits:**

- Provide simplified abstractions to ease various job execution workflows from both a user and developer perspective.
- Job results are retrieved from the remote platform in `numpy` format (with `dtype=np.int8`) instead of JSON, reducing bandwidth usage and conversion overhead.

**Possible Drawbacks:**

- This will likely break downstream applications currently submitting jobs via the demo branch, so some coordination will be needed there.

**Related GitHub Issues:**

N/A


tests/frontend/test_engine.py
```diff
@@ -166,209 +174,297 @@ def inspect():
...
+        result = connection.get_job_result("123")
 
-    def test__run(self, starship_engine, monkeypatch):
-        """
-        Tests StarshipEngine._run, with the assumption that the backend is a hardware backend
-        that supports running only a single program. This test ensures that a program is compiled
-        for the hardware backend, is locked, is added to self.run_progs, that it is run and that
-        a Result object is returned populated with the result samples.
-        """
+        assert np.array_equal(result.samples.T, result_samples)
```
pauljxtan: This probably deserves some explanation. As the `Result` class is currently implemented, `Result.samples` has dimensions `(modes, shots)`; on the other hand, job results received from the remote backend have the shape `(shots, modes)`, so we transpose the samples here. Modifying `Result.samples` to return `(shots, modes)` appears to break various tests elsewhere in SF, so I opted to leave it unchanged. I also believe that the issue of sample dimensions is being (or has been) addressed on `master`, so it may be better to revisit this in another PR down the line.

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+    def __init__(self, target: str, connection: Connection = None):
+        if target not in self.VALID_TARGETS:
+            raise ValueError(
+                "Invalid engine target: {} (valid targets: {})".format(
+                    target, self.VALID_TARGETS
+                )
+            )
+        if connection is None:
+            # TODO use `load_config` when implemented
+            config = DEFAULT_CONFIG["api"]
```
pauljxtan: This line should be updated to use `load_config()` after https://github.com/XanaduAI/strawberryfields/pull/298 is merged.

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        # response = self._get("/jobs?size={}".format(self.MAX_JOBS_REQUESTED))
+        # if response.status_code == 200:
+            # return [
+                # Job(id_=info["id"], status=info["status"], connection=self)
+                # for info in response.json()["data"]
+                # if datetime.strptime(info["created_at"], self.JOB_TIMESTAMP_FORMAT)
+                # > after
+            # ]
+        # raise RequestFailedError(self._format_error_message(response))
+        raise NotImplementedError("This feature is not yet implemented")
```
pauljxtan: I've left the above commented for context, but I think we probably want to remove it before merging.

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+class RequestMethod(enum.Enum):
+    """Defines the valid request methods for messages sent to the remote platform."""
+
+    GET = "get"
+    POST = "post"
+    PATCH = "patch"
+
+
+class Connection:
```
antalszava: Hmm, this file seems to be getting quite big, maybe it would be worth having the `Connection` class, `Job` and related `Error` classes in a separate one called `connection.py`? :thinking: Or to split them is some other meaningful way :blush:

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+        # Get a specific job by ID
+        job = connection.get_job("59a1c0b1-c6a7-4f9b-ae37-0ac5eec9c413")
+        job  # <Job: id=59a1c0b1-c6a7-4f9b-ae37-0ac5eec9c413, ...>
+
+    Args:
+        token (str): the API authentication token
+        host (str): the hostname of the remote platform
+        port (int): the port to connect to on the remote host
+        use_ssl (bool): whether to use SSL for the connection
```
antalszava: In the configuration, a `debug` boolean is also specified, I'm missing more details on where it should be incorporated. Maybe it would be intended here as well? :thinking:  @josh146 could you help out with that one?

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+    def _patch(
+        self, path: str, headers: Dict[str, str] = None, **kwargs
+    ) -> requests.Response:
+        return self._request(RequestMethod.PATCH, path, headers, **kwargs)
+
+    def _request(
+        self, method: RequestMethod, path: str, headers: Dict[str, str] = None, **kwargs
+    ) -> requests.Response:
+        headers = {} if headers is None else headers
+        return getattr(requests, method.value)(
```
antalszava: It might be worth having a small auxiliary naming here for readabilitiy such as
```python
request_method = getattr(requests, method.value)
return request_method(urljoin(self.base_url, path),
            headers={"Authorization": self.token, **headers},
            **kwargs
        )
```

pauljxtan: Thanks for pointing that out! Updated.

strawberryfields/engine.py
pauljxtan: Great idea! Especially as the `Connection`/`Job` interface is intended to be usable independent of the `StarshipEngine`, and the module boundary would help to enforce and make explicit that separation.

One problem I see though is that creates a circular dependency due to the `engine.Result` class, which would to be imported in `connection.py`---and `engine.py` would need to import `Connection` from `connection.py`, etc.

A solution to that might be to move `Result` into a separate module. What do you think?

strawberryfields/engine.py
antalszava: Hmm, after going through `_get`, `_post`, `_patch` and `_request`, I'm wondering, is there any advantage in having `_request` and having a `RequestMethod` enum class?

Maybe the explicit values of `get`, `post`, and `patch` could simply be used in their corresponding methods to making the return statement for example for `_get`:
```python
return requests.get(urljoin(self.base_url, path), headers={"Authorization": self.token, **headers}, **kwargs)
```

But let me know if I missed something or you'd rather keep the implementation as is, for future extendability, this is more of a suggestion after parsing each of them separately. :)

strawberryfields/engine.py
antalszava: Sounds good to me :)

strawberryfields/engine.py
pauljxtan: Ah, yes - the former `APIClient` implemented a `debug` mode that simply collected responses and errors in list attributes on the client object. I didn't see a particular use case for that, so left it out for simplicity; but it should not be a very big change to add that in if deemed necessary.

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        response = self._post("/jobs", data=json.dumps({"circuit": circuit}))
+        if response.status_code == 201:
+            return Job(
+                id_=response.json()["id"],
+                status=JobStatus(response.json()["status"]),
+                connection=self,
+            )
+        raise RequestFailedError(
+            "Job creation failed: {}".format(self._format_error_message(response))
+        )
```
antalszava: Cool! :)

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
-    Executes :class:`.Program` instances on the chosen remote backend, and makes
-    the results available via :class:`.Result`.
+    OPEN = "open"
+    QUEUED = "queued"
+    CANCELLED = "cancelled"
+    COMPLETE = "complete"
+    FAILED = "failed"
+
+    @property
+    def is_terminal(self) -> bool:
```
antalszava: Just a minor suggestion here: would it make sense to go with the docstring and name this `is_final`? :slightly_smiling_face:

strawberryfields/engine.py
pauljxtan: Yes, that's a perfectly valid point :+1:

I was on the fence myself about whether that introduced additional indirection for no substantial benefit. My thinking was to avoid repeating shared elements, like the `"Authorization"` header, between requests; but now that you mention it, it seems that is the _only_ thing that is actually duplicated, so it's probably better to just be explicit and direct about it.

I will look at simplifying the code as you described. Thanks!

strawberryfields/engine.py
pauljxtan: Good catch, updated :+1:

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        Args:
+            job_id (str): the job ID
         """
-        if self.jobs:
-            raise OneJobAtATimeError(
-                "A job is already queued. Please reset the engine and try again."
-            )
+        response = self._patch(
+            "/jobs/{}".format(job_id), data={"status", JobStatus.CANCELLED.value}
+        )
```
antalszava: Just for clarification: you are using `PATCH` here such that you can modify *only* the status, right?

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        """
+        return self._status
+
+    @property
+    def result(self) -> Result:
+        """The job result.
 
-    def __init__(self, backend, polling_delay_seconds=1, **kwargs):
-        super().__init__(backend)
+        This is only defined for complete jobs, and raises an exception for any other
```
antalszava:
```suggestion
        This is only defined for completed jobs and raises an exception for any other
```

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+        Only an open or queued job can be refreshed; an exception is raised otherwise.
+        """
+        if self.status.is_final:
+            raise InvalidJobOperationError(
+                "A {} job cannot be refreshed".format(self.status.value)
+            )
+        self._status = self._connection.get_job_status(self.id)
+        if self._status == JobStatus.COMPLETE:
+            self._result = self._connection.get_job_result(self.id)
```
antalszava: Cool! :)

strawberryfields/engine.py
antalszava: Oh, this suggestion regards other parts as well, such as `JobStatus.COMPLETE`. I thought it was more descriptive to have `completed`, but this is pretty subjective and let me know if you'd rather keep it as is :)

strawberryfields/engine.py
```diff
@@ -126,6 +128,8 @@ def state(self):
         Returns:
             BaseState: quantum state returned from program execution
         """
+        if not self._is_stateful:
+            raise AttributeError("The state is undefined for a stateless computation.")
```
antalszava: Great! :)

strawberryfields/engine.py
antalszava: I see! Let's see if there would be a use case. Currently, it is included in #298, but that could also be adjusted accordingly. :)

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+    def _post(
+        self, path: str, headers: Dict[str, str] = None, **kwargs
+    ) -> requests.Response:
+        return self._request(RequestMethod.POST, path, headers, **kwargs)
+
+    def _patch(
+        self, path: str, headers: Dict[str, str] = None, **kwargs
+    ) -> requests.Response:
+        return self._request(RequestMethod.PATCH, path, headers, **kwargs)
```
antalszava: Maybe these could be made static, what do you think?

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+                ``None`` otherwise
+        """
+        job = self.run_async(program, shots)
+        try:
+            while True:
+                job.refresh()
+                if job.status == JobStatus.COMPLETE:
+                    return job.result
+                if job.status == JobStatus.FAILED:
+                    raise JobFailedError("The computation failed; please try again.")
```
antalszava: I wonder if we'll be able to provide more information here to the user on the cause of failure (e.g. timeout, etc.), but right now I cannot really think of anything else.

strawberryfields/engine.py
```diff
@@ -531,161 +544,542 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+        Args:
+            program (strawberryfields.Program): the quantum circuit
+            shots (int): the number of shots for which to run the job
 
         Returns:
-            Result: results of the computation
+            strawberryfields.engine.Job: the created remote job
         """
+        return self._connection.create_job(self.target, program, shots)
```
antalszava: This looks really neat! :)

tests/frontend/test_engine.py
```diff
@@ -166,209 +174,299 @@ def inspect():
...
+
+    def stateless_result_raises_on_state_access(self):
+        """Tests that `result.state` raises an error for a stateless result.
         """
-        methods = MagicMock()
-        inputs = MagicMock()
+        result = Result([[1, 2], [3, 4]], is_stateful=False)
+
+        with pytest.raises(AttributeError):
+            _ = result.state
```
antalszava: Nice!

strawberryfields/engine.py
```diff
@@ -531,161 +544,507 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        newly completed.
+
+        Only an open or queued job can be refreshed; an exception is raised otherwise.
         """
-        # Do nothing for now...
-        pass
+        if self.status.is_final:
+            raise InvalidJobOperationError(
+                "A {} job cannot be refreshed".format(self.status.value)
+            )
```
antalszava: I'm wondering about raising an error in this case. If the user mistakenly calls on `refresh`, then this would throw an `InvalidJobOperationError` in their script and terminate it. Instead of that, maybe information could be provided, why no refreshing can take place (e.g. the result is already contained, or the job failed, etc.) and perhaps a warning could be raised? :thinking: 

What do you think?

strawberryfields/engine.py
pauljxtan: That's right!

strawberryfields/engine.py
antalszava: (For cancellation, it would seem in order to raise an error, as the user expects to take action, whereas `refresh` seems to be something more along the lines of trying to extract more information about the `Job`. But this might just be my point of view.)

strawberryfields/engine.py
pauljxtan: No, I totally agree that "completed" makes more sense in this context (as a synonym for "finished"), while "complete" sounds to me along the lines of "full" or "whole".

The reason I had it as "complete" to begin with is because that is the identifier used by the remote platform :smiley: So the value of the `JobStatus` enum would have to be `"complete"` to match up with the responses, but the enum name could certainly be `COMPLETED` (along with comments, docstrings, etc.). I will make the change, thanks!

tests/frontend/test_engine.py
```diff
@@ -166,209 +175,299 @@ def inspect():
...
-        """
-        Tests that StarshipEngine.jobs is correctly cleared when callling StarshipEngine.reset.
-        """
-        starship_engine.jobs.append(MagicMock())
-        assert len(starship_engine.jobs) == 1
-        starship_engine.reset()
-        assert len(starship_engine.jobs) == 0
 
-    def test__get_blackbird(self, starship_engine, monkeypatch):
+def mock_response(status_code, json_body=None, binary_body=None):
```
antalszava: If you would like, this could also become a `MockResponse` class and then there would be no need to use `MagicMock` (if I'm not missing something). The `__init__` could then contain the branching logic. The upside of that would be having `pytest` as the only testing framework in this file.

tests/frontend/test_engine.py
antalszava: Thanks for the explanation! One thing to note is that there is an upcoming PR to be merged on this: #300

tests/frontend/test_engine.py
```diff
@@ -166,209 +175,299 @@ def inspect():
...
+    def test_get_job_result_error(self, connection, monkeypatch):
+        """Tests a failed job result request."""
+        monkeypatch.setattr(requests, "get", mock_return(mock_response(404, {})))
+
+        with pytest.raises(RequestFailedError):
+            connection.get_job_result("123")
+
+    def test_cancel_job(self, connection, monkeypatch):
+        """Tests a successful job cancellation request."""
+        monkeypatch.setattr(requests, "patch", mock_return(mock_response(204, {})))
```
antalszava: One thing to note here, that if I understand it correctly any PATCH request is mocked out this way, correct?

One thing that could be checked here, is that `requests.patch` receives the correct arguments. This would include creating a function similar to `mock_return`, but one, that also tracks the arguments passed, such as:
```python

arguments = []
def mock_return(return_value):
    """A helper function for defining a mock function that returns the given value for
    any arguments.
    """
    def function(*args, **kwargs):
        arguments.append(args)
        return return_value
    return function
```
or something comparable.

```python
In [2]: arguments = [] 
   ...: def mock_return(return_value): 
   ...:     """A helper function for defining a mock function that returns the g
   ...: iven value for 
   ...:     any arguments. 
   ...:     """ 
   ...:     def function(*args, **kwargs): 
   ...:         arguments.append(args) 
   ...:         return return_value 
   ...:     return function 
   ...:                                                                         

In [3]: mock_return(3)                                                          
Out[3]: <function __main__.mock_return.<locals>.function(*args, **kwargs)>

In [4]: a = mock_return(3)                                                      

In [5]: a(4, 5, 6)                                                              
Out[5]: 3

In [6]: arguments                                                               
Out[6]: [(4, 5, 6)]
```

tests/frontend/test_engine.py
pauljxtan: Great idea. I've replaced `mock_response` with a `MockResponse` class!

tests/frontend/test_engine.py
```diff
@@ -166,209 +175,299 @@ def inspect():
...
-        mock_job.is_failed = True
+        with pytest.raises(RequestFailedError):
+            connection.get_job_status("123")
 
-        monkeypatch.setattr("strawberryfields.engine.to_blackbird", mock_to_blackbird)
-        monkeypatch.setattr(starship_engine, "_get_blackbird", mock__get_blackbird)
-        monkeypatch.setattr(starship_engine, "_queue_job", lambda job_content: mock_job)
+    def test_get_job_result(self, connection, monkeypatch):
+        """Tests a successful job result request."""
+        result_samples = np.array([[1, 2], [3, 4]], dtype=np.int8)
```
antalszava: The `dtype` argument could be parametrized as well, just to see that it works fine for types like `np.float64` and the likes as well :)

tests/frontend/test_engine.py
```diff
@@ -166,209 +175,299 @@ def inspect():
...
 
-        some_params = {"param": MagicMock()}
+        with io.BytesIO() as buf:
+            np.save(buf, result_samples)
+            buf.seek(0)
+            monkeypatch.setattr(
+                requests,
+                "get",
+                mock_return(mock_response(200, binary_body=buf.getvalue())),
+            )
```
antalszava: Cool! :)

tests/frontend/test_engine.py
```diff
@@ -166,209 +175,299 @@ def inspect():
...
+
+    def test_cancel_job(self, connection, monkeypatch):
+        """Tests a successful job cancellation request."""
+        monkeypatch.setattr(requests, "patch", mock_return(mock_response(204, {})))
+
+        # A successful cancellation does not raise an exception
+        connection.cancel_job("123")
+
+    def test_cancel_job_error(self, connection, monkeypatch):
+        """Tests a successful job cancellation request."""
```
antalszava:
```suggestion
        """Tests a failed job cancellation request."""
```

tests/frontend/test_engine.py
```diff
@@ -166,209 +174,304 @@ def inspect():
...
-        fetching the result.)
-        """
+        engine = StarshipEngine("chip2", connection=connection)
+        job = engine.run_async(prog)
+        assert job.status == JobStatus.OPEN
+
+        for _ in range(server.REQUESTS_BEFORE_COMPLETED):
+            job.refresh()
+
+        assert job.status == JobStatus.COMPLETED
```
antalszava: Nice!

strawberryfields/engine.py
antalszava: Oh, I see! Feel free to leave it as is, if you'd rather not have any further confusions coming from this difference :)

tests/frontend/test_engine.py
thisac: Just mentioning that #300 is already merged (earlier today). It unified the sample returns for all simulator engines so that they return NumPy arrays with dimension `(shots, modes)`.

strawberryfields/engine.py
josh146: Agree

strawberryfields/engine.py
josh146: :+1: I commented the exact same thing last night, but couldn't submit my review because github buttons stopped working 😆 But yes, I agree. Options include:

* Package: `api/connection.py`, `api/job,py`, etc.

* A module `connection.py` that contains all classes

I tend to prefer option (1), an API package, as it collects all objects as the same level of abstraction.

strawberryfields/engine.py
josh146: Perhaps, rather than debug, a better options would be `verbose=True`. In verbose mode, SF simply prints out details of the job submission to the terminal as it occurs.

@antalszava, might make sense to remove `debug` from #298 for now

strawberryfields/engine.py
```diff
@@ -531,161 +544,507 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        ^C # KeyboardInterrupt cancels the job
+
+        # Run a job asynchronously
+        job = engine.run_async(program, shots=1)
+        job.status  # <JobStatus.QUEUED: 'queued'>
+        job.result  # InvalidJobOperationError
+        # (After some time...)
+        job.refresh()
+        job.status  # <JobStatus.COMPLETED: 'complete'>
+        job.result  # [[0 1 0 2 1 0 0 0]]
```
josh146: This might be easier to read using Python prompt syntax. Eg:

```rest
Run a job synchronously:

>>> engine = StarshipEngine("chip2")
>>> result = engine.run(program, shots=1) # blocking call
>>> result
[[0 1 0 2 1 0 0 0]]

Run a job synchronously, but cancel it before it is completed
using a keyboard interrupt (`ctrl+c`):

>>> result = engine.run(program, shots=1)
^C---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-4-f1a1495c6d9c> in <module>()
----> 1 time.sleep(10)

Run a job asynchronously:

>>> job = engine.run_async(program, shots=1)
>>> job.status 
<JobStatus.QUEUED: 'queued'>
>>> job.result
InvalidJobOperationError
>>> job.refresh()
>>> job.status 
<JobStatus.COMPLETED: 'complete'>
>>> job.result
[[0 1 0 2 1 0 0 0]]
```

The nice thing about using prompts is that Sphinx renders it automatically without needing to specify a code-block or indent

strawberryfields/engine.py
josh146: This is something that should be probably done via logging (i.e., `log.info(message)`, `log.warning(message)` etc).

strawberryfields/engine.py
```diff
@@ -531,161 +544,507 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+                if job.status == JobStatus.FAILED:
+                    raise JobFailedError("The computation failed; please try again.")
+                time.sleep(self.POLLING_INTERVAL_SECONDS)
+        except KeyboardInterrupt:
+            self._connection.cancel_job(job.id)
+
+    def run_async(self, program: Program, shots: int = 1) -> Job:
+        """Runs a remote job asynchronously.
+
+        In the asynchronous mode, a `Job` is returned immediately, and the user can
```
josh146:
```suggestion
        In the asynchronous mode, a ``Job`` is returned immediately, and the user can
```

tests/frontend/test_engine.py
```diff
@@ -166,209 +174,304 @@ def inspect():
         assert inspect() == expected2
 
 
-class TestStarshipEngine:
-    """
-    Tests various methods on the remote engine StarshipEngine.
+@pytest.fixture
+def connection():
```
josh146: All API specific tests I would move into their own folder (i.e., `tests/api/test_connection.py`, `tests/api/test_job.py`)

josh146: It might even make sense to move the API-specific starship tests here as well.

tests/frontend/test_engine.py
```diff
@@ -166,209 +174,304 @@ def inspect():
...
+    """Tests for the ``Result`` class."""
+
+    def stateless_result_raises_on_state_access(self):
+        """Tests that `result.state` raises an error for a stateless result.
         """
-        methods = MagicMock()
-        inputs = MagicMock()
+        result = Result([[1, 2], [3, 4]], is_stateful=False)
+
+        with pytest.raises(AttributeError):
```
josh146: It would be good to also include the `match=` argument here, just to make sure that the _right_ attribute error is raised!

tests/frontend/test_engine.py
```diff
@@ -166,209 +174,304 @@ def inspect():
...
+
+
+class TestJob:
+    """Tests for the ``Job`` class."""
+
+    def incomplete_job_raises_on_result_access(self):
+        """Tests that `job.result` raises an error for an incomplete job."""
+        job = Job("abc", status=JobStatus.QUEUED, connection=Connection)
+
+        with pytest.raises(AttributeError):
```
josh146: Same applies throughout (better to match the error message in these assertions)

strawberryfields/engine.py
antalszava: It got merged just now :)

strawberryfields/engine.py
pauljxtan: I've updated all occurrences of "complete" to "completed" except for the single value in the enum member. Don't think it should lead to much confusion, but I'll add a comment to clarify,

strawberryfields/engine.py
pauljxtan: Went one better and removed them completely as we now use the `requests` functions directly :slightly_smiling_face:

strawberryfields/engine.py
pauljxtan: For that particular case (job status becomes "failed"), I believe the most detail we can provide to the user is something along the lines of "The computation failed due to an internal system error." (There may be more specific causes, e.g. hardware error, that is only available in our internal logs.)

@josh146 Do you mean logging a failure message and not raising an exception at all?

strawberryfields/engine.py
antalszava: Great, thanks! :)

strawberryfields/engine.py
pauljxtan: Yeah, that makes sense. I'll replace that raised exception with a warning message.

tests/frontend/test_engine.py
pauljxtan: Yes, any PATCH request with any combination of arguments will return the same response.

That's a good point about validating the arguments. I'll look at adding that into the relevant tests :+1:

tests/frontend/test_engine.py
pauljxtan: Good idea, will do so!

starship
```diff
@@ -14,114 +14,51 @@
...
-                str(response.status_code), response.content))
-            sys.exit(1)
 
     program = load(args.input)
-    eng = StarshipEngine(program.target)
-    result = eng.run(program)
+
+    eng = StarshipEngine("chip2", connection)
+    sys.stdout.write("Computing...\n")
+    result = eng.run(program, shots=1)
```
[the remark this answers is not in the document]
pauljxtan: Ah yes, I think you're right. I'll remove that, thanks!

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+class JobFailedError(Exception):
+    """Raised when a remote job enters a 'failed' status."""
+
+
+class JobStatus(enum.Enum):
+    """Represents the status of a remote job.
+
+    This class maps a set of job statuses to the string representations returned by the
+    remote platform.
```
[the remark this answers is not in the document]
pauljxtan: I think the first option sounds good, just to separate things more cleanly and have a dedicated namespace. Which do you prefer?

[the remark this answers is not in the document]
pauljxtan: Oh, never mind, I see you said it in another comment :smiley:

strawberryfields/engine.py
pauljxtan: Oh nice! I'll update all the relevant docstrings to use the prompt syntax. (I assume these won't be run as doctests?)

tests/frontend/test_engine.py
pauljxtan: Sounds good, I will do so!

tests/frontend/test_engine.py
pauljxtan: Cool, thanks for the heads-up guys. I'm assuming we will have another substantial PR at some point to line up `demo` with `master`, so that may be a good time to get all the dimensions agreeing?

strawberryfields/engine.py
pauljxtan: I've moved the relevant code into the following modules: `api/connection.py`, `api/job.py`, and `api/result.py`.

strawberryfields/engine.py
pauljxtan: Great, I've updated the relevant logic to use the new configuration functions!

strawberryfields/engine.py
pauljxtan: On further thought, I wonder if it's worth having this method stub here at all, or if we should just remove it and put it back in a future PR. Thoughts @antalszava @josh146?

strawberryfields/engine.py
pauljxtan: The `verbose` option has been added!

strawberryfields/engine.py
pauljxtan: I've removed the wrapper methods for sending requests, and the `create_job`, `get_job` (etc.) methods now call the relevant `requests` method directly.

strawberryfields/engine.py
pauljxtan: That exception has been replaced with a logged warning message.

tests/frontend/test_engine.py
pauljxtan: This has been added!

strawberryfields/engine.py
```diff
@@ -531,161 +544,541 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+
+    # pylint: disable=bad-continuation
+    # See: https://github.com/PyCQA/pylint/issues/289
+    def __init__(
+        self,
+        token: str,
+        host: str = "platform.strawberryfields.ai",
+        port: int = 443,
+        use_ssl: bool = True,
+    ):
```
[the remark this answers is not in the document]
pauljxtan: Done!

tests/frontend/test_engine.py
pauljxtan: All of the new tests have been moved under a new `api` directory.

tests/frontend/test_engine.py
pauljxtan: Thanks, I've added `match` checks to all the relevant tests.

tests/frontend/test_engine.py
pauljxtan: Done :+1:

strawberryfields/engine.py
pauljxtan: The docstrings for `StarshipEngine` and `Connection` have been updated to use the prompt syntax.

strawberryfields/api/connection.py
```diff
@@ -0,0 +1,284 @@
...
+    given API authentication token, tests the connection, submits a new job, and
+    retrieves an existing job.
+
+    >>> connection = Connection(token="abc")
+    >>> success = connection.ping() # `True` if successful, `False` if the connection fails
+    >>> job = connection.create_job("chip2", program, shots=123)
+    >>> job
+    <Job: id=d177cbf5-1816-4779-802f-cef2c117dc1a, ...>
+    >>> job.status
+    <JobStatus.QUEUED: 'queued'>
```
co9olguy: Users won't likely ever use a `Connection` directly, but they will certainly interact with a `Job`. The string printed here could be improved to be simpler and less redundant, e.g., `<JobStatus: queued>`

strawberryfields/api/connection.py
```diff
@@ -0,0 +1,284 @@
...
+    >>> job.result
+    AttributeError
```
co9olguy: Is this the exact string that will be printed? Honestly not sure what the best way is to "print" an exception in a docstring :thinking:

strawberryfields/api/connection.py
```diff
@@ -0,0 +1,284 @@
...
+    >>> job = connection.get_job(known_job_id)
+    >>> job
+    <Job: id=59a1c0b1-c6a7-4f9b-ae37-0ac5eec9c413, ...>
+    >>> job.status
+    <JobStatus.COMPLETED: 'complete'>
```
co9olguy: see above comment

strawberryfields/api/connection.py
```diff
@@ -0,0 +1,284 @@
...
+            "s" if self.use_ssl else "", self.host, self.port
+        )
+        self._headers = {"Authorization": self.token}
+
+    @property
+    def token(self) -> str:
+        """The API authentication token.
+
+        Returns:
+            str: the authentication token
```
co9olguy: Docstring seems redundant. See my general question (may appear elsewhere in PR) about whether it is even necessary to report the `Return` of a `@property`

strawberryfields/api/connection.py
```diff
@@ -0,0 +1,284 @@
...
+        Args:
+            job_id (str): the job ID
+
+        Returns:
+            strawberryfields.engine.Job: the job
+        """
+        path = "/jobs/{}".format(job_id)
+        response = requests.get(self._url(path), headers=self._headers)
+        if response.status_code == 200:
+            return Job(
```
co9olguy: Might be worthwhile to make this `Job`(...)` block a separate function, since it is used elsewhere in the same way

strawberryfields/api/connection.py
```diff
@@ -0,0 +1,284 @@
...
+            with io.BytesIO() as buf:
+                buf.write(response.content)
+                buf.seek(0)
+                samples = np.load(buf)
+            return Result(samples, is_stateful=False)
+        raise RequestFailedError(
+            "Failed to get job result: {}".format(self._format_error_message(response))
+        )
+
+    def cancel_job(self, job_id: str):
```
co9olguy: Better to use `id_` as used elsewhere?

tests/frontend/test_engine.py
pauljxtan: I've added an argument check to the test - I'm especially grateful that you suggested this because it actually caught a bug!

tests/api/test_starship_engine.py
```diff
@@ -0,0 +1,97 @@
...
+        return (
+            JobStatus.COMPLETED
+            if self.request_count >= self.REQUESTS_BEFORE_COMPLETED
+            else JobStatus.QUEUED
+        )
+
+
+@pytest.fixture
+def job_to_complete(connection, monkeypatch):
+    """Mocks a remote job that is completed after a certain number of requests."""
```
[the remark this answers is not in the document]
pauljxtan: It's not very clear and I'm not enamoured with how the code looks, but basically the `MockServer` (instantiated as `server` here) holds the number of requests as a state attribute. It returns a job status of "queued" if the number of requests is below a certain number, then returns "completed" once it exceeds the threshold.

It's a little hacky but I think it does the job in lieu of a full integration test (which we plan to tackle in a future PR).

tests/api/test_starship_engine.py
```diff
@@ -0,0 +1,97 @@
...
+    monkeypatch.setattr(
+        Connection,
+        "create_job",
+        mock_return(Job(id_="123", status=JobStatus.OPEN, connection=connection)),
+    )
+    server = MockServer()
+    monkeypatch.setattr(Connection, "get_job_status", server.get_job_status)
```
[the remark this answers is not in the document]
pauljxtan: I agree that it looks a little jarring, but the files in this PR are all auto-formatted with `black` (which has a default line length limit of 88 characters), so I'm hesitant to manually impose ad hoc formatting :slightly_smiling_face:

tests/api/test_starship_engine.py
```diff
@@ -0,0 +1,97 @@
...
+        mock_return(Result([[1, 2], [3, 4]], is_stateful=False)),
+    )
+
+
+class TestStarshipEngine:
+    """Tests for the ``StarshipEngine`` class."""
+
+    def test_run_complete(self, connection, prog, job_to_complete):
+        """Tests a successful synchronous job execution."""
+        engine = StarshipEngine("chip2", connection=connection)
```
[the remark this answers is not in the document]
pauljxtan: Ah, thanks for letting me know! I'll make a note to change this once the new names are finalized.

tests/api/test_starship_engine.py
```diff
@@ -0,0 +1,97 @@
...
+        result = engine.run(prog)
+
+        assert np.array_equal(result.samples.T, np.array([[1, 2], [3, 4]]))
```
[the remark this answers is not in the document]
pauljxtan: Please see: https://github.com/XanaduAI/strawberryfields/pull/294#discussion_r383527915

Long story short, @thisac has already resolved this on `master` but as this branch is based on `demo`, I felt that it seemed less of a headache to address in a future commit (probably when these changes get merged into `master`) as various tests throughout the codebase would have to updated.

strawberryfields/api/connection.py
pauljxtan: The exact output has a bunch of traceback info preceding it. It seemed a little distracting/superfluous so I omitted it, but I can add that in if we consider it necessary.

strawberryfields/api/connection.py
pauljxtan: This one is the actual string as implemented by `__str__`/`__repr__`.

strawberryfields/engine.py
```diff
@@ -19,120 +19,21 @@
...
-            BaseState: quantum state returned from program execution
-        """
-        return self._state
-
-    def __str__(self):
-        """String representation."""
-        return "Result: {} subsystems, state: {}\n samples: {}".format(
-            len(self.samples), self.state, self.samples
-        )
+log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
pauljxtan: That was used to log a warning message in `StarshipEngine.run()`.

tests/api/test_connection.py
```diff
@@ -0,0 +1,215 @@
...
+        assert job.status == status
+
+    def test_create_job_error(self, prog, connection, monkeypatch):
+        """Tests a failed job creation flow."""
+        monkeypatch.setattr(requests, "post", mock_return(MockResponse(400, {})))
+
+        with pytest.raises(RequestFailedError, match="Failed to create job"):
+            connection.create_job("chip2", prog, 1)
+
+    @pytest.mark.skip(reason="method not yet implemented")
```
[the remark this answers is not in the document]
pauljxtan: Thanks, updated.

tests/api/test_connection.py
```diff
@@ -0,0 +1,215 @@
...
+        ]
+        monkeypatch.setattr(
+            requests, "get", mock_return(MockResponse(200, {"data": jobs})),
+        )
+
+        jobs = connection.get_all_jobs(after=datetime(2020, 1, 5))
+
+        assert [job.id for job in jobs] == [str(i) for i in range(5, 10)]
+
+    @pytest.mark.skip(reason="method not yet implemented")
```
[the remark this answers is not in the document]
pauljxtan: And updated.

tests/api/test_starship_engine.py
```diff
@@ -0,0 +1,97 @@
...
+# limitations under the License.
+"""TODO"""
+import numpy as np
+import pytest
+
+from strawberryfields.api import Connection, Job, JobStatus, Result
+from strawberryfields.engine import StarshipEngine
+
+
+def mock_return(return_value):
```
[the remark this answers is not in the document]
pauljxtan: Good catch, updated.

tests/api/test_job.py
```diff
@@ -0,0 +1,41 @@
...
+
+        with pytest.raises(
+            AttributeError,
+            match="The result is undefined for jobs that are not completed",
+        ):
+            _ = job.result
+
+    def test_final_job_raises_on_cancel(self, connection):
+        """Tests that `job.cancel()` raises an error for a complete, failed, or
+        aleady cancelled job."""
```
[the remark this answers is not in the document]
pauljxtan: Thanks for pointing that out, updated.

strawberryfields/engine.py
```diff
@@ -531,161 +441,151 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+    running remote jobs in a synchronous or asynchronous manner.
 
-    # This engine will execute jobs remotely.
-    REMOTE = True
+    **Example:**
 
-    def __init__(self, backend, polling_delay_seconds=1, **kwargs):
-        super().__init__(backend)
+    The following examples instantiate an engine with the default configuration, and
+    runs jobs both synchronously and asynchronously.
```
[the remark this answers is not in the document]
pauljxtan: Not sure :thinking: I like the symmetry of "synchronous" as the opposite of "asynchronous", but open to changing it if we don't find that clear enough!

strawberryfields/engine.py
```diff
@@ -531,161 +441,151 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+    >>> result
+    [[0 1 0 2 1 0 0 0]]
 
-    def _init_backend(self, *args):
-        """
-        TODO: This does not do anything right now.
-        """
-        # Do nothing for now...
-        pass
+    Run a job synchronously, but cancel it before it is completed using a keyboard
```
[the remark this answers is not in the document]
pauljxtan: Good point. I've removed the example keyboard interrupt and added a line to the docstring for `run()` describing the feature.

tests/api/test_connection.py
```diff
@@ -0,0 +1,215 @@
...
+        """Tests a successful ping to the remote host."""
+        monkeypatch.setattr(requests, "get", mock_return(MockResponse(200, {})))
+
+        assert connection.ping()
+
+    def test_ping_failure(self, connection, monkeypatch):
+        """Tests a failed ping to the remote host."""
+        monkeypatch.setattr(requests, "get", mock_return(MockResponse(500, {})))
+
+        assert not connection.ping()
```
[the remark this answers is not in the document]
pauljxtan: :stuck_out_tongue:

strawberryfields/engine.py
```diff
@@ -531,161 +441,151 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        self._target = target
+        self._connection = connection
+
+    @property
+    def target(self) -> str:
+        """The target device used by the engine.
 
         Returns:
-            (strawberryfields.api_client.Job): a Job instance referencing the queued job
+            str: the target device used by the engine
```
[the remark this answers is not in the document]
pauljxtan: Good point - I've removed the "Returns" part from the relevant property docstrings.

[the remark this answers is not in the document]
pauljxtan: On second thought, we probably want at least the types in there. I've put them back in, but without the repetitive text.

strawberryfields/engine.py
```diff
@@ -531,161 +441,151 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
-        return job
+        return self._target
 
-    def _run_program(self, program, **kwargs):
+    @property
+    def connection(self) -> Connection:
+        """The connection object used by the engine.
+
+        Returns:
+            strawberryfields.engine.Connection: the connection object used by the engine
```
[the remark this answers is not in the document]
pauljxtan: Please see above.

strawberryfields/api/job.py
```diff
@@ -0,0 +1,139 @@
...
+        return self._result
+
+    def refresh(self):
+        """Refreshes the status of the job, along with the job result if the job is
+        newly completed.
+
+        Only an open or queued job can be refreshed; an exception is raised otherwise.
+        """
+        if self.status.is_final:
+            log.warning("A %s job cannot be refreshed", self.status.value)
```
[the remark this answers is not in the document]
pauljxtan: I believe it's the recommended way of formatting `logging` messages.

https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation

strawberryfields/api/connection.py
josh146: Another thought - if the user wants to programmatically check job status, they will need to import the `JobStatus` enum from within the API. Will they also be able to compare via a simple string? i.e., `job.status == "queued"`?

strawberryfields/engine.py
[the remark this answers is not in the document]
josh146: Perhaps 'blocking' and 'non-blocking' might make more sense to people coming from a scientific computing background?

tests/api/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+def connection():
+    """A mock connection object."""
+    return Connection(token="token", host="host", port=123, use_ssl=True)
+
+
+def mock_return(return_value):
+    """A helper function for defining a mock function that returns the given value for
+    any arguments.
+    """
+    return lambda *args, **kwargs: return_value
```
josh146: Having separate conftests makes it much clearer which fixtures are used in which part of the test suite!

tests/api/test_starship_engine.py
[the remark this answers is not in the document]
josh146: @pauljxtan I've been periodically merging master into demo every week, to make sure they keep in sync and lesson merge conflicts. The changes made by @thisac should already be in the demo branch now, if you want to merge them in here.

strawberryfields/engine.py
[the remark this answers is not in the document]
pauljxtan: Sounds good! I've updated the relevant parts to use those terms.

tests/api/test_starship_engine.py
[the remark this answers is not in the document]
pauljxtan: Thanks for letting me know! I've merged the latest `demo` and made a small change to `Result.__init__` removing the `np.stack` (which is no longer needed). I've modified the new API tests accordingly (removed all transposes), and all relevant existing tests also appear to be passing.

starship
```diff
@@ -14,114 +14,51 @@
...
-            sys.exit(1)
-        else:
-            sys.stderr.write("Could not connect to server: {}, {}\n".format(
-                str(response.status_code), response.content))
-            sys.exit(1)
 
     program = load(args.input)
-    eng = StarshipEngine(program.target)
+
+    eng = StarshipEngine("chip2", connection)
```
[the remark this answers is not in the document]
pauljxtan: Good catch - I've changed that to use `program.target` as before.

strawberryfields/api/connection.py
pauljxtan: They can compare to a string with `job.status.value == "queued"`. It's not ideal, but since the alternative is dropping the enum entirely, I feel that it's a decent tradeoff :thinking: 

I guess the options here are:
- Leave it as is
- Override the `__eq__` method to enable comparison with a string, i.e. `JobStatus.QUEUED == "queued"` -> `True`
- Add properties on `JobStatus`, e.g. `is_queued`, `is_completed`, etc.
- Get rid of the enum and replace all usages with string literals

What do you think?

strawberryfields/api/connection.py
pauljxtan: Do you mean the `Job(...)` "constructor"? Since it is already is (sort of) a function in itself, I'm not sure I see the benefit of factoring it out further, unless I'm missing something :thinking:

strawberryfields/api/connection.py
pauljxtan: I think using `id_` within the `Job` class itself makes sense, since it is quite obvious what the variable is identifying; but I think `job_id` is more explicit for methods on other classes and makes the interface a bit clearer.

pauljxtan:
Hey @antalszava @co9olguy @josh146, I think I've addressed all of your comments, but let me know if anything is missing. (GitHub seems to be having some issues at the moment causing comments to be intermittently dropped.)

strawberryfields/api/connection.py
pauljxtan: Actually, please ignore the above.

After more thought, the obvious solution (to me) would be to keep `JobStatus` internally but expose it as a string via the property. I will make that change - let me know if you think there is a better way!

doc/index.rst
```diff
@@ -174,7 +174,6 @@ Strawberry Fields is **free** and **open source**, released under the Apache Lic
    code/sf_apps
    code/sf_ops
    code/sf_utils
-   code/sf_api_client
```
antalszava: Instead, the new `api` directory will have to be included, such that when `make docs` is run, the new directory is documented.

strawberryfields/api/job.py
```diff
@@ -0,0 +1,139 @@
...
+    def is_final(self) -> bool:
+        """Checks if this status represents a final and immutable state.
+
+        This method is primarily used to determine if an operation is valid for a given
+        status.
+
+        Returns:
+            bool: ``True`` if the job status is final, and ``False`` otherwise
+        """
+        return self in (JobStatus.CANCELLED, JobStatus.COMPLETED, JobStatus.FAILED)
```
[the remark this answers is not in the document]
antalszava: (not too sure if there was a change in the meantime, but cannot really see the diff betteen the code and the suggestion)

strawberryfields/api/result.py
```diff
@@ -0,0 +1,116 @@
...
+        # the latter needs to be converted to a 2-D array with shape (shots, modes)
+        if np.ndim(samples) == 1:
+            samples = np.array([samples])
+        self._samples = samples
+
+    @property
+    def samples(self):
+        """Measurement samples.
+
+        Returned measurement samples will have shape ``(modes,)``. If multiple
```
antalszava:
```suggestion
        Returned measurement samples will have shape ``(shots, modes)``. If multiple
```

strawberryfields/api/result.py
```diff
@@ -0,0 +1,116 @@
...
+        shots are requested during execution, the returned measurement samples
+        will instead have shape ``(shots, modes)``.
```
antalszava:
```suggestion
        Returned measurement samples will have shape ``(shots, modes)``.
```

antalszava: As per #300

strawberryfields/api/result.py
```diff
@@ -0,0 +1,115 @@
...
+        Returns:
+            BaseState: quantum state returned from program execution
+        """
+        if not self._is_stateful:
+            raise AttributeError("The state is undefined for a stateless computation.")
+        return self._state
+
+    def __str__(self):
+        """String representation."""
+        return "Result: {} subsystems, state: {}\n samples: {}".format(
```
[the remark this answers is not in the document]
antalszava: Hmm, perhaps only smaller states could be outputted by this and there could be a separate auxiliary method that outputs every state.

That is, this could have a check along the lines of
```python
if len(self.state)>10 or len(self.samples)>1000:
# output lengths and advice users to use the Result.verbose_str option
# else proceed with outputting everything
```
Open to suggestions if this could work and if so, what the limit for these should be.

strawberryfields/engine.py
```diff
@@ -552,161 +467,141 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
         return result
 
 
-class StarshipEngine(BaseEngine):
-    """
-    Starship quantum program executor engine.
+class StarshipEngine:
+    """A quantum program executor engine that that provides a simple interface for
```
antalszava:
```suggestion
    """A quantum program executor engine that provides a simple interface for
```

strawberryfields/api/result.py
pauljxtan: Good catch :+1:

strawberryfields/api/result.py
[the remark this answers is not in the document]
co9olguy: Might be possible that the string-printing methods of numpy already take the size into account and truncate (haven't tested myself)

strawberryfields/api/job.py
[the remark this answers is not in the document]
co9olguy: yep, ignore this for now. I was reviewing at the same time that @pauljxtan was updating with new commits

strawberryfields/api/job.py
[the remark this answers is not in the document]
co9olguy: Didn't know, thanks

strawberryfields/api/connection.py
co9olguy: oh, for this I was referring to the comment about `<JobStatus.QUEUED: 'queued'>`

tests/api/test_job.py
```diff
@@ -32,14 +34,31 @@ def test_incomplete_job_raises_on_result_access(self, connection):
             AttributeError,
             match="The result is undefined for jobs that are not completed",
         ):
-            _ = job.result
+            job.result
```
co9olguy: @pauljxtan I see, was the linter complaining if you didn't assign the property to something? In that case, I might defer to the linter (sorry for making pointless work :laughing:)

strawberryfields/engine.py
```diff
@@ -297,7 +299,7 @@ def _broadcast_nones(val, dim):
             prev = p
 
         if self.samples is not None:
-            return Result(self.samples.copy())
+            return Result(np.array(self.samples).T)
```
co9olguy: @pauljxtan @thisac How come there is a transpose needed here? Is there still an inconsistent return shape between some parts of the pipeline?

strawberryfields/engine.py
```diff
@@ -443,45 +467,32 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
-    """A quantum program executor engine that that provides a simple interface for
-    running remote jobs in a synchronous or asynchronous manner.
+    """A quantum program executor engine that provides a simple interface for
+    running remote jobs in a blocking or non-blocking manner.
 
     **Example:**
 
     The following examples instantiate an engine with the default configuration, and
-    runs jobs both synchronously and asynchronously.
+    runs both blocking and non-blocking jobs.
```
co9olguy:
```suggestion
    run both blocking and non-blocking jobs.
```

strawberryfields/engine.py
pauljxtan: I think it's due to the way the samples are computed by the simulator backend (?) I don't understand the local backend well enough to modify the code there without a risk of breaking something, so will have to defer to someone else's expertise here :slightly_smiling_face:

tests/api/test_job.py
pauljxtan: Haha yes it was :slightly_smiling_face:

doc/index.rst
pauljxtan: I added a `sf_api.rst` as you described, and I _think_ the docs are showing up properly in the expected places - but would appreciate if someone could confirm I did it right!

strawberryfields/api/result.py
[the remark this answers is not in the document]
pauljxtan: Sorry, could've sworn I replied to this but maybe the GitHub outage ate it.

I tested this out, and it does seem that the samples get automatically truncated:

```python
In [4]: result = Result(np.random.randint((1e6, 8)))

In [6]: str(result)
Out[6]: 'Result: 1 subsystems, state: None\n samples: [[148373      4]]'
```

I should also note that this was a holdover from existing code that I simply moved to a different module, so I assume this was probably also discussed when it was first implemented?

strawberryfields/engine.py
thisac: `self.samples` has shape `(modes, shots)` and thus needs to be transposed to `(shots, modes)` somewhere. In the master branch the transposition happens in the `Results` class at the line `np.stack(samples, 1)`, so the `BaseEngine._run` function simply returns `self.samples.copy()` (with shape `(modes, shots)`). I'm not sure why this is. It was like that before I touched it; I only added the specific check in `Results` for a single-shot list so that it returned an array of shape `(shots, modes)` even in that case. But this was all done in the master branch, and not the demo branch. Now both branches seem to have different implementations of the `Results` class. I'm not too sure about what the demo branch is though, so maybe this doesn't matter that much. :slightly_smiling_face:

doc/index.rst
antalszava: @pauljxtan thanks so much! :blush: Could you move the line with the `api` change here? (such that it is in alphabetical order, this is also the order in which it appears on the website)

doc/index.rst
```diff
@@ -185,3 +184,4 @@ Strawberry Fields is **free** and **open source**, released under the Apache Lic
    code/sf_program
    code/sf_program_utils
    code/sf_parameters
+   code/sf_api
```
antalszava: So moving this line for where the deleted line is

strawberryfields/engine.py
pauljxtan: Thanks! This is making more sense to me now. What I essentially did was move the transposition slightly down the pipeline, from `Result.__init__` to the end of `BaseEngine._run`. This prevents `np.stack(samples, 1)` (now removed) from incorrectly transposing samples coming in from the remote platform (which have `(shots, modes)`). So I believe this all works out.

doc/index.rst
pauljxtan: Thanks for pointing that out, done :+1:

strawberryfields/engine.py
thisac: That makes sense. Thanks! :slightly_smiling_face:  Should this change preferably be done in the master branch as well (or will the demo branch be merged with the master branch at some point)? @josh146 @co9olguy

josh146: The plan is to merge the demo branch into master in the coming week(s), before the upcoming SF release!

josh146: However, it sounds like there will be a merge conflict now, as master does the transposition in `Result.__init__`, and demo does it in `BaseEngine._run`, so we will need to be careful to resolve this conflict when we merge.

josh146: I think we will have to look at this again carefully after the merge. In an ideal world, the API _as well as_ the Gaussian/Fock/TF backends, should be returning samples of the form `(shots, modes)`. The engine or the results class should not have to do any transposition.

tests/api/test_starship_engine.py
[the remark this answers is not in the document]
josh146: @pauljxtan, we actually use a line length of 100 characters on SF! So you might need to reblack before merging (`black -l 100 strawberryfields/`).

Before the next release, I am going to add a CI check that ensures blacking is done with the right line length, sorry about that!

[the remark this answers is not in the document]
pauljxtan: Ah I see, I will reformat the relevant files - thanks for pointing that out!

strawberryfields/api/result.py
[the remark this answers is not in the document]
pauljxtan: Oh, my bad, got confused with the syntax there :stuck_out_tongue: Thanks for checking that!

strawberryfields/api/connection.py
pauljxtan: Nice, I've updated the docstring to use something similar (with `AttributeError` and the attached error message).

starship
```diff
@@ -14,114 +14,52 @@
...
-    parser = argparse.ArgumentParser(description="run a blackbird script on StarshipEngine")
+    parser = argparse.ArgumentParser(description="run a blackbird script")
+    parser.add_argument(
+        "--token", "-t", help="the API authentication token", required=True
+    )
     group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument("--input", "-i", help="the xbb file to run")
+    group.add_argument(
+        "--ping", "-p", action="store_true", help="test the API connection"
+    )
```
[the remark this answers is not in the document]
pauljxtan: Hmm, after taking another look at this, it occurs to me that the API token should probably be loaded in the same way as the other connection parameters (host, port, etc.) via the new configuration logic implemented by Antal. This is more consistent, would reduce the required arguments down to the bare minimum (input file), and we wouldn't even need to instantiate a `Connection` object in the CLI script - allowing us to work at the highest possible level of abstraction with only an `Engine`.

I've removed the `--token` parameter from the CLI - which can now be run simply as:

```
$ ./starship -i test.xbb
Executing program on remote hardware...
[[0 0 1 1 0 0 1 0]]%
```

tests/api/test_starship_engine.py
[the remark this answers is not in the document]
pauljxtan: @josh146 - running `black -l 100` it on the full `strawberryfields/` seemed to change a lot of unrelated modules, so I've run it only on the files relevant to this PR.

doc/code/sf_api.rst
```diff
@@ -0,0 +1,14 @@
...
+
+.. warning::
+
+    Unless you are a Strawberry Fields developer, you likely do not need
+    to use these classes directly.
+
+.. automodapi:: strawberryfields.api
+    :no-heading:
+    :no-inheritance-diagram:
+    :include-all-objects:
```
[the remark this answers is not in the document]
pauljxtan: Yes, @antalszava has been a massive help on various things :+1:

pauljxtan:
Planning to merge this afternoon if/when the latest CI build passes.

strawberryfields/api/result.py
```diff
@@ -0,0 +1,114 @@
...
+    can be returned, as well as quantum state information.
+
+    >>> eng = sf.Engine("gaussian")
+    >>> results = eng.run(prog)
+    >>> print(results)
+    Result: 3 subsystems
+        state: <GaussianState: num_modes=3, pure=True, hbar=2>
+        samples: [0, 0, 0]
+    >>> results.samples
+    [0, 0, 0]
```
thisac:
```suggestion
        samples: [[0, 0, 0]]
    >>> results.samples
    np.array([[0, 0, 0]])
```

Sorry for the late suggestion. Just noticed this.

pauljxtan: Good catch, thanks!

strawberryfields/api/result.py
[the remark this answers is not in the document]
thisac: @josh146 It seems like it's the `samples = np.stack(samples, 1)` that is taking time when printing the results (I'm using the master branch). This should be gone in this PR though, since it's an unnecessary line and apparently quite slow (which isn't too surprising since it not only transposes the array but also re-stacks all the inputs; I'm not sure why that was done that way though).

thisac:
I've looked through the files relevant to the changes I made for the unified results PR earlier, and it looks good. This is a much better solution and I don't think that anything will break when merging (at least not anything obvious, related to the `Result` class as well as the transposition being relocated). :+1:

tests/api/test_starship_engine.py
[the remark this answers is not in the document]
co9olguy: good idea @pauljxtan

starship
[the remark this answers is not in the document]
antalszava: No worries @pauljxtan, if further final touches are required to the `starship` CLI, they could come down the line in upcoming separate PRs. :blush:

pauljxtan:
> I've looked through the files relevant to the changes I made for the unified results PR earlier, and it looks good. This is a much better solution and I don't think that anything will break when merging (at least not anything obvious, related to the `Result` class as well as the transposition being relocated). +1

Thanks for taking the time to look this over @thisac!

strawberryfields/engine.py
thisac: I looked through how the `LocalEngine` gets samples from the backends, and it doesn't look like there's an easy fix to get the correct shape directly from the backends. The only backend that supports multi-shot measurements seems to be the Gaussian backend, and it returns an array of values for one mode at a time (one value for each shot). In that case the combining of modes is done in `engine.py` either way, so the transposing would have to be there also. Maybe I misunderstood something though.

As for the other backends they don't support multiple shots together with any measurements, and simply return a list of length = number of modes filled with `None`, which is then converted into a nested array in the desired shape in `engine.py` (the `_broadcast_nones` function in `BaseEngine._run`).

tests/api/test_starship_engine.py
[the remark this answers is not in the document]
josh146: :+1: Just prior to the new SF release, I plan on blacking the entire library, and adding a formatting CI check. This will stop new PRs from allowing unformatted code through.

strawberryfields/api/result.py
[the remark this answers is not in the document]
josh146: Nice, good to know :+1:

PR #295, opened 2020-02-19
Apps layer: Add WAW adjacency matrix rescaling function

trbromley:
This adds a `waw_matrix` function that rescales in input adjacency matrix according to the WAW encoding, with the objective of embedding node weights into the adjacency matrix.

### DIFF
--- a/strawberryfields/apps/sample.py
+++ b/strawberryfields/apps/sample.py
@@ -99,9 +99,9 @@ graph's adjacency matrix. One canonical approach to doing this is to use the ``W
 
     A \rightarrow WAW,
 
-with :math:`W` the diagonal matrix formed by the weighted nodes. The rescaled adjacency matrix
-can be passed to :func:`sample`, resulting in a bias toward sampling subgraphs with a large
-product of node weights. :func:`waw_matrix`
+with :math:`W=w_{i}\delta_{ij}` the diagonal matrix formed by the weighted nodes :math:`w_{i}`. The
+rescaled adjacency matrix can be passed to :func:`sample`, resulting in a distribution that
+increases the probability of observing a node proportionally to its weight.
 """
 import warnings
 from typing import Optional, Union
@@ -383,15 +383,15 @@ def waw_matrix(A: np.ndarray, w: Union[np.ndarray, list]) -> np.ndarray:
     r"""Rescale adjacency matrix to account for node weights.
 
     Given a graph with adjacency matrix :math:`A` and a vector :math:`w` of weighted nodes,
-    this function rescales the adjacency matrix according to:
+    this function rescales the adjacency matrix according to :cite:`banchi2019molecular`:
 
     .. math::
 
         A \rightarrow WAW,
 
-    with :math:`W` the diagonal matrix formed by the weighted nodes. This encoding can be used to
-    add a bias in GBS toward selecting subgraphs with a large product of node weights
-    :cite:`banchi2019molecular`. The resulting matrix can be passed to :func:`sample`.
+    with :math:`W=w_{i}\delta_{ij}` the diagonal matrix formed by the weighted nodes :math:`w_{i}`.
+    The rescaled adjacency matrix can be passed to :func:`sample`, resulting in a distribution that
+    increases the probability of observing a node proportionally to its weight.
 
     **Example usage:**
 
@@ -412,31 +412,7 @@ def waw_matrix(A: np.ndarray, w: Union[np.ndarray, list]) -> np.ndarray:
     if not np.allclose(A, A.T):
         raise ValueError("Input must be a NumPy array corresponding to a symmetric matrix")
 
-    w = _process_w(len(A), w)
-
-    return (w * A).T * w
-
-
-def _process_w(dim: int, w: Union[np.ndarray, list]) -> np.ndarray:
-    """Ensure correct form of weight vector.
-
-        Args:
-            dim (int): dimension of graph
-            w (array or list): weight vector to be processed
-
-        Returns:
-            array: the weight vector as a numpy array
-        """
     if isinstance(w, list):
         w = np.array(w)
 
-    w_s = w.shape
-
-    w_check = (len(w_s) == 1 and w_s[0] == dim) or (
-        len(w_s) == 2 and sorted(np.unique(w_s)) == [1, dim]
-    )  # Check if the vector has one dimension or is a two-dimensional row or column vector
-
-    if not w_check:
-        raise ValueError("Vector of node weights must be a row or column vector")
-
-    return w.flatten()
+    return (w * A).T * w
--- a/strawberryfields/apps/sample.py
+++ b/strawberryfields/apps/sample.py
@@ -361,3 +361,47 @@ def vibronic(
     if np.any(t == 0):
         s = np.pad(s, ((0, 0), (0, n_modes))).tolist()
     return s
+
+
+def waw_matrix(A: np.ndarray, w: np.ndarray) -> np.ndarray:
+    """Rescale adjacency matrix to account for node weights.
+
+    Given a graph with adjacency matrix :math:`A` and a vector :math:`w` of weighted nodes,
+    this function rescales the adjacency matrix according to:
+
+    .. math::
+
+        A \rightarrow WAW,
+
+    with :math:`W` the diagonal matrix formed by the weighted nodes. This encoding can be used to
+    add a bias in GBS toward selecting subgraphs with a large product of node weights
+    :cite:`banchi2019molecular`. The resulting matrix can be passed to :func:`sample`.
+
+    **Example usage:**
+
+    >>> g = nx.erdos_renyi_graph(5, 0.7)
+    >>> a = nx.to_numpy_array(g)
+    >>> w = [1, 1, 3, 1, 0.5]
+    >>> a = waw_matrix(a, w)
+    >>> sample(a, 3, 4)
+
+    Args:
+        A (array): adjacency matrix to rescale
+        w (array): vector of real node weights
+
+    Returns:
+        array: matrix rescaled according to the WAW encoding
+    """
+    if not np.allclose(A, A.T):
+        raise ValueError("Input must be a NumPy array corresponding to a symmetric matrix")
+    w_s = w.shape
+    dim = len(A)
+
+    w_check = (len(w_s) == 1 and w_s[0] == dim) or (
+        len(w_s) == 2 and sorted(np.unique(w_s)) == [1, dim]
+    )  # Check if the vector has one dimension or is a two-dimensional row or column vector
+
+    if not w_check:
+        raise ValueError("Vector of node weights must be a row or column vector")
+
+    return (w * A).T * w
--- a/strawberryfields/apps/sample.py
+++ b/strawberryfields/apps/sample.py
@@ -412,16 +412,16 @@ def waw_matrix(A: np.ndarray, w: Union[np.ndarray, list]) -> np.ndarray:
     if not np.allclose(A, A.T):
         raise ValueError("Input must be a NumPy array corresponding to a symmetric matrix")
 
-    w = _process_w(A, w)
+    w = _process_w(len(A), w)
 
     return (w * A).T * w
 
 
-def _process_w(A: np.ndarray, w: Union[np.ndarray, list]) -> np.ndarray:
+def _process_w(dim: int, w: Union[np.ndarray, list]) -> np.ndarray:
     """Ensure correct form of weight vector.
 
         Args:
-            A (array): adjacency matrix, used to fix dimension
+            dim (int): dimension of graph
             w (array or list): weight vector to be processed
 
         Returns:
@@ -431,7 +431,6 @@ def _process_w(A: np.ndarray, w: Union[np.ndarray, list]) -> np.ndarray:
         w = np.array(w)
 
     w_s = w.shape
-    dim = len(A)
 
     w_check = (len(w_s) == 1 and w_s[0] == dim) or (
         len(w_s) == 2 and sorted(np.unique(w_s)) == [1, dim]
--- a/strawberryfields/apps/sample.py
+++ b/strawberryfields/apps/sample.py
@@ -397,10 +397,10 @@ def waw_matrix(A: np.ndarray, w: Union[np.ndarray, list]) -> np.ndarray:
 
     >>> g = nx.erdos_renyi_graph(5, 0.7)
     >>> a = nx.to_numpy_array(g)
-    >>> w = [1, 1, 3, 1, 0.5]
+    >>> w = [10, 1, 0, 1, 1]
     >>> a = waw_matrix(a, w)
     >>> sample(a, 3, 4)
-    [[0, 0, 0, 0, 0], [1, 0, 0, 0, 1], [0, 0, 0, 0, 0], [0, 1, 1, 1, 1]]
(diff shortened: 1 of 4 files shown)


tests/apps/test_sample.py
```diff
@@ -28,7 +28,7 @@
 
 adj_dim_range = range(2, 6)
 
-t0 = np.array([0., 0., 0., 0., 0., 0., 0.])
+t0 = np.array([0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])
```
trbromley: (result of running `black`)

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,67 @@ def vibronic(
...
+    dim = len(A)
+
+    w_check = (len(w_s) == 1 and w_s[0] == dim) or (
+        len(w_s) == 2 and sorted(np.unique(w_s)) == [1, dim]
+    )  # Check if the vector has one dimension or is a two-dimensional row or column vector
+
+    if not w_check:
+        raise ValueError("Vector of node weights must be a row or column vector")
+
+    return w.flatten()
```
trbromley: I plan to use this function for input validation of the weight vector.

We definitely want to let users input either a list or numpy array without throwing an error.
This function also checks the shape of the numpy array.

One question - where is the best place for this function to live? It will be used in other modules (e.g. `clique`). I think it's perhaps ok for it to just live here for now, it is not a user-facing function and is not shown in the documentation.

strawberryfields/apps/sample.py
```diff
@@ -86,9 +86,25 @@
 
 The subgraphs sampled from GBS are likely to be dense :cite:`arrazola2018using`, motivating their
 use within heuristics for problems such as maximum clique (see :mod:`~.apps.clique`).
+
+Including node weights
+----------------------
+
+Some graphs are composed of nodes with weights. These node weights can correspond to relevant
```
ixfoduap:
```suggestion
Some graphs are composed of nodes with weights. These weights can correspond to relevant
```

strawberryfields/apps/sample.py
```diff
@@ -86,9 +86,25 @@
...
+information in an optimization problem and it is desirable to encode them into GBS along with the
+graph's adjacency matrix. One canonical approach to doing this is to use the ``WAW`` encoding
+:cite:`banchi2019molecular`, which rescales the adjacency matrix according to:
+
+.. math::
+
+    A \rightarrow WAW,
+
+with :math:`W` the diagonal matrix formed by the weighted nodes. The rescaled adjacency matrix
+can be passed to :func:`sample`, resulting in a bias toward sampling subgraphs with a large
```
ixfoduap: I would rephrase this as: 
can be passed to sample, resulting in a distribution that increases the probability of observing a node proportionally to its weight.

strawberryfields/apps/sample.py
```diff
@@ -86,9 +86,25 @@
...
+Some graphs are composed of nodes with weights. These node weights can correspond to relevant
+information in an optimization problem and it is desirable to encode them into GBS along with the
+graph's adjacency matrix. One canonical approach to doing this is to use the ``WAW`` encoding
+:cite:`banchi2019molecular`, which rescales the adjacency matrix according to:
+
+.. math::
+
+    A \rightarrow WAW,
+
+with :math:`W` the diagonal matrix formed by the weighted nodes. The rescaled adjacency matrix
```
ixfoduap: It may be good to write W explicitly. One compact way of doing so is using `W=w_{ij}\delta_{ij}`

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,66 @@ def vibronic(
...
+    r"""Rescale adjacency matrix to account for node weights.
+
+    Given a graph with adjacency matrix :math:`A` and a vector :math:`w` of weighted nodes,
+    this function rescales the adjacency matrix according to:
+
+    .. math::
+
+        A \rightarrow WAW,
+
+    with :math:`W` the diagonal matrix formed by the weighted nodes. This encoding can be used to
```
ixfoduap: Again maybe good to write W explicitly and to rephrase the effect on the distribution

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,66 @@ def vibronic(
...
+    :cite:`banchi2019molecular`. The resulting matrix can be passed to :func:`sample`.
+
+    **Example usage:**
+
+    >>> g = nx.erdos_renyi_graph(5, 0.7)
+    >>> a = nx.to_numpy_array(g)
+    >>> w = [1, 1, 3, 1, 0.5]
+    >>> a = waw_matrix(a, w)
+    >>> sample(a, 3, 4)
+    [[0, 0, 0, 0, 0], [1, 0, 0, 0, 1], [0, 0, 0, 0, 0], [0, 1, 1, 1, 1]]
```
ixfoduap: Can we `post-select` on samples that more clearly show that nodes with large weight appear more often?

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,66 @@ def vibronic(
...
+
+    with :math:`W` the diagonal matrix formed by the weighted nodes. This encoding can be used to
+    add a bias in GBS toward selecting subgraphs with a large product of node weights
+    :cite:`banchi2019molecular`. The resulting matrix can be passed to :func:`sample`.
+
+    **Example usage:**
+
+    >>> g = nx.erdos_renyi_graph(5, 0.7)
+    >>> a = nx.to_numpy_array(g)
+    >>> w = [1, 1, 3, 1, 0.5]
```
ixfoduap: Maybe `w=[1,2,3,4,5]` so that it's more clear which nodes have larger weight?

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,66 @@ def vibronic(
...
+
+    Returns:
+        array: matrix rescaled according to the WAW encoding
+    """
+    if not np.allclose(A, A.T):
+        raise ValueError("Input must be a NumPy array corresponding to a symmetric matrix")
+
+    w = _process_w(len(A), w)
+
+    return (w * A).T * w
```
ixfoduap: Clever!

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,66 @@ def vibronic(
...
+
+
+def _process_w(dim: int, w: Union[np.ndarray, list]) -> np.ndarray:
```
ixfoduap: I don't mind this function, but isn't it a bit of overkill?

ixfoduap: Is the idea that we are allowing the weight vector to have different forms? Isn't it better to have a fixed form?

strawberryfields/apps/sample.py
trbromley: Thanks!

strawberryfields/apps/sample.py
trbromley: Nice! Done

strawberryfields/apps/sample.py
trbromley: Thanks have added

strawberryfields/apps/sample.py
trbromley: Done

strawberryfields/apps/sample.py
trbromley: The user might want to input the weight vector as:
- a list
- a numpy array of shape (dim,) - standard
- a numpy array of shape (1, dim) - a 2d row vector
- a numpy array of shape (dim, 1) - a 2d column vector

Actually the above line 417 is fine with any of them... so maybe you're right and we can skip the checking. Although one thing is we definitely should let the user input as a list or numpy array.

trbromley: Ok, simplified

strawberryfields/apps/sample.py
trbromley: I like having it be a bit different to consecutive ascending numbers, as maybe `[1,2,3,4,5]` could confuse someone that the weights have to be linked to the node numbers?

strawberryfields/apps/sample.py
trbromley:
```rst
>>> g = nx.erdos_renyi_graph(5, 0.7)
>>> a = nx.to_numpy_array(g)
>>> w = [10, 1, 0, 1, 1]
>>> a = waw_matrix(a, w)
>>> sample(a, 3, 4)
[[1, 1, 0, 1, 0], [1, 1, 0, 1, 0], [1, 1, 0, 0, 0], [1, 1, 0, 1, 0]]
```
This is what I came up with. There is a zero weight one which nicely never clicks. There is a 10 weight one which always clicks

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,66 @@ def vibronic(
...
+    with :math:`W` the diagonal matrix formed by the weighted nodes. This encoding can be used to
+    add a bias in GBS toward selecting subgraphs with a large product of node weights
+    :cite:`banchi2019molecular`. The resulting matrix can be passed to :func:`sample`.
+
+    **Example usage:**
+
+    >>> g = nx.erdos_renyi_graph(5, 0.7)
+    >>> a = nx.to_numpy_array(g)
+    >>> w = [1, 1, 3, 1, 0.5]
+    >>> a = waw_matrix(a, w)
```
soranjh: It might be nice to print the new a here.

strawberryfields/apps/sample.py
soranjh: Or maybe use large weights such that all samples reflect that without post selecting?

tests/apps/test_sample.py
```diff
@@ -368,3 +369,48 @@ def test_vibronic_integration(p, integration_sample_number):
...
+    @pytest.mark.parametrize("inst", zip(adjs, wvecs, resc_adjs))
+    def test_valid_w(self, inst):
+        """Test if function returns the correct answer on some pre-calculated instances."""
+        assert np.allclose(sample.waw_matrix(inst[0], inst[1]), inst[2])
+
+    @pytest.mark.parametrize("inst", zip(adjs, wvecs, resc_adjs))
+    def test_valid_w_list(self, inst):
+        """Test if function returns the correct answer on some pre-calculated instances,
+        when w is a list"""
+        assert np.allclose(sample.waw_matrix(inst[0], list(inst[1])), inst[2])
```
josh146: Nice!

strawberryfields/apps/sample.py
```diff
@@ -361,3 +377,42 @@ def vibronic(
...
+    Returns:
+        array: matrix rescaled according to the WAW encoding
+    """
+    if not np.allclose(A, A.T):
+        raise ValueError("Input must be a NumPy array corresponding to a symmetric matrix")
+
+    if isinstance(w, list):
+        w = np.array(w)
+
+    return (w * A).T * w
```
josh146: Maybe I'm missing something (because the tests below pass), but is this doing WAW? I did a small test in NumPy:

```python
>>> import numpy as np
>>> A = np.random.random(size=[4,4])
>>> w = np.random.random(size=[4])
>>> (w * A).T * w == np.diag(w) @ A @ np.diag(w)
array([[ True, False, False, False],
       [False,  True, False, False],
       [False, False,  True, False],
       [False, False, False,  True]])
```

josh146: It seems that the diagonal is rescaled correctly, but the off-diagonals differ. But let me know if I'm doing something silly 😆

tests/apps/test_sample.py
```diff
@@ -368,3 +369,48 @@ def test_vibronic_integration(p, integration_sample_number):
...
+            ]
+        ),
+        np.array([[0, -2, 0, -4], [-2, 0, -6, 0], [0, -6, 0, -12], [-4, 0, -12, 0]]),
+        np.array([[0, 0, 0], [0, 0, 1], [0, 1, 0]]),
+    ]
+
+    def test_invalid_adjacency(self):
+        """Test if function raises a ``ValueError`` for a matrix that is not symmetric."""
+        with pytest.raises(ValueError, match="Input must be a NumPy array"):
+            adj_asym = np.triu(np.ones((4, 4)))
```
josh146: You could place this line before the pytest.raises, since it will not raise the ValueError (just makes it slightly cleaner)

tests/apps/test_sample.py
```diff
@@ -368,3 +369,48 @@ def test_vibronic_integration(p, integration_sample_number):
...
+            sample.sample(A=adj_asym, n_mean=1.0)
+
+    @pytest.mark.parametrize("inst", zip(adjs, wvecs, resc_adjs))
+    def test_valid_w(self, inst):
+        """Test if function returns the correct answer on some pre-calculated instances."""
+        assert np.allclose(sample.waw_matrix(inst[0], inst[1]), inst[2])
```
josh146: The passing tests make me think that my comment above about the WAW calculation is incorrect :)

tests/apps/test_sample.py
josh146: How were the matrices pre-calculated? By hand?

strawberryfields/apps/sample.py
```diff
@@ -86,9 +86,25 @@
 
 The subgraphs sampled from GBS are likely to be dense :cite:`arrazola2018using`, motivating their
 use within heuristics for problems such as maximum clique (see :mod:`~.apps.clique`).
+
+Including node weights
+----------------------
+
+Some graphs are composed of nodes with weights. These weights can correspond to relevant
+information in an optimization problem and it is desirable to encode them into GBS along with the
+graph's adjacency matrix. One canonical approach to doing this is to use the ``WAW`` encoding
```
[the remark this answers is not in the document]
trbromley: Nice, fixed to math mode.

As a side note... I keep accidentally writing WAR instead of WAW :thinking:

strawberryfields/apps/sample.py
trbromley: That's a great idea, have done this. Makes the code block a bit longer but I think it's worth it.
image

strawberryfields/apps/sample.py
trbromley: Your code block looks right, but in this case A is symmetric. We have:
`w * A =  A @ np.diag(w)`
`(w * A).T = np.diag(w) @ A.T`
`(w * A).T = np.diag(w) @ A`
`(w * A).T * w = np.diag(w) @ A.T @ np.diag(w)`
So the `(w * A).T * w` approach does require symmetry of `A`, and wouldn't work if `A` was complex and hermitian. However, I think we are reasonably restricting to graphs with real edge weights for now.

Although, perhaps going with `np.diag(w) @ A @ np.diag(w)` is simpler. One minor drawback of `np.diag(w) @ A @ np.diag(w)` is that it doesn't work vectors that are 2D arrays (e.g. row or column vector), while the `(w * A).T * w` approach does - but this setting should be relatively rare.

tests/apps/test_sample.py
trbromley: Nice spot, done!

tests/apps/test_sample.py
trbromley: Yes all by :hand:
Tried to cover a variety of cases (e.g. edges with different weights in `A`, positive/negative/fractional node weights)

tests/apps/test_sample.py
trbromley: Yes, due to the symmetry of the test `A`'s

strawberryfields/apps/sample.py
trbromley: Thanks, yes we're now using some more biased weights and picking some indicative samples.

PR #297, opened 2020-02-21
Apps: Update subgraph functions to accept node weights

trbromley:
This PR adds support for node-weights in the search algorithms of the `subgraph` module.

### DIFF
--- a/strawberryfields/apps/subgraph.py
+++ b/strawberryfields/apps/subgraph.py
@@ -46,6 +46,8 @@ or least degree with respect to to the subgraph, with ties settled uniformly at 
 This function returns a dictionary over the range of sizes specified, with each value being the
 corresponding resized subgraph.
 """
+from typing import Union
+
 import networkx as nx
 import numpy as np
 
@@ -189,7 +191,13 @@ def _update_subgraphs_list(l: list, t: tuple, max_count: int) -> None:
     return
 
 
-def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dict:
+def resize(
+    subgraph: list,
+    graph: nx.Graph,
+    min_size: int,
+    max_size: int,
+    node_select: Union[str, np.ndarray, list] = "uniform",
+) -> dict:
     """Resize a subgraph to a range of input sizes.
 
     This function uses a greedy approach to iteratively add or remove nodes one at a time to an
@@ -201,9 +209,15 @@ def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dic
     the algorithm performs the procedure again.
 
     When shrinking is required, the algorithm examines all nodes from within the subgraph as
-    candidates and removes the single node with lowest degree relative to the subgraph. In both
-    growth and shrink phases, ties for addition/removal with nodes of equal degree are settled by
-    uniform random choice.
+    candidates and removes the single node with lowest degree relative to the subgraph.
+
+    In both growth and shrink phases, there may be multiple candidate nodes with equal degree to
+    add to or remove from the subgraph. The method of selecting the node is specified by the
+    ``node_select`` argument, which can be either:
+
+    - ``"uniform"`` (default): uniform randomly choose a node from the candidates;
+    - A list or array: specifying the node weights of the graph, resulting in choosing the node
+      from the candidates with the lowest weight, settling ties by uniform random choice.
 
     **Example usage:**
 
@@ -222,6 +236,9 @@ def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dic
         graph (nx.Graph): the input graph
         min_size (int): minimum size for subgraph to be resized to
         max_size (int): maximum size for subgraph to be resized to
+        node_select (str, list or array): method of settling ties when more than one node of
+            equal degree can be added/removed. Can be ``"uniform"`` (default), or a numpy array or
+            list.
 
     Returns:
         dict[int, list[int]]: a dictionary of different sizes with corresponding subgraph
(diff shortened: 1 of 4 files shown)


strawberryfields/apps/subgraph.py
```diff
@@ -39,19 +39,32 @@
...
 The key element of the :func:`search` algorithm is the resizing of each subgraph, allowing a
-range of subgraph sizes to be tracked. Resizing proceeds by greedily adding or removing nodes to
-a subgraph one-at-a-time. Node selection is carried out by picking the node with the greatest
-or least degree with respect to to the subgraph, with ties settled uniformly at random.
-
-This function returns a dictionary over the range of sizes specified, with each value being the
-corresponding resized subgraph.
+range of subgraph sizes to be tracked. Resizing proceeds in the :func:`resize` function by greedily
+adding or removing nodes to a subgraph one-at-a-time. Node selection is carried out by picking
+the node with the greatest or least degree with respect to the subgraph. This function returns a
```
soranjh: with respect to the subgraph or the whole graph?

strawberryfields/apps/subgraph.py
```diff
@@ -80,6 +102,9 @@ def search(
         min_size (int): minimum size to search for dense subgraphs
         max_size (int): maximum size to search for dense subgraphs
         max_count (int): maximum number of densest subgraphs to keep track of for each size
+        node_select (str, list or array): method of settling ties when more than one node of
```
soranjh: The term "settling ties" is a bit confusing to me. Method of settling ties can be a numpy array or a list?

strawberryfields/apps/subgraph.py
```diff
@@ -271,15 +309,51 @@ def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dic
...
 
             new_size = shrink_subgraph.order()
 
             if min_size <= new_size <= max_size:
                 resized[new_size] = sorted(shrink_subgraph.nodes())
 
     return resized
+
+
+def _validate_inputs(
```
soranjh: This function probably needs some explanation.

strawberryfields/apps/subgraph.py
ixfoduap: Maybe good to say that the numpy array or list specifies the node weights

strawberryfields/apps/subgraph.py
```diff
@@ -222,21 +260,16 @@ def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dic
...
-
-    if not subgraph.issubset(nodes):
-        raise ValueError("Input is not a valid subgraph")
-    if min_size < 1:
-        raise ValueError("min_size must be at least 1")
-    if max_size >= len(nodes):
-        raise ValueError("max_size must be less than number of nodes in graph")
-    if max_size < min_size:
-        raise ValueError("max_size must not be less than min_size")
+    node_select, w = _validate_inputs(subgraph, graph, min_size, max_size, node_select)
```
ixfoduap: _validate_inputs gets defined after it is used here. Is that good coding practice?

strawberryfields/apps/subgraph.py
```diff
@@ -253,14 +286,19 @@ def resize(subgraph: list, graph: nx.Graph, min_size: int, max_size: int) -> dic
...
+            degrees = np.array(
+                [(c, graph.subgraph(list(grow_nodes) + [c]).degree()[c]) for c in complement_nodes]
+            )
+            degrees_max = np.argwhere(degrees[:, 1] == degrees[:, 1].max()).flatten()
 
-            to_add = max(degrees, key=lambda x: x[1])
-            grow_subgraph.add_node(to_add[0])
+            if node_select == "uniform":
+                to_add_index = np.random.choice(degrees_max)
+            elif node_select == "weight":
```
ixfoduap: I'm confused. I thought node_select could be either `"uniform"` or a list/array. Where did we name the list/array option `"weight"`?

tests/apps/test_subgraph.py
```diff
@@ -323,25 +344,73 @@ def test_tie(self, monkeypatch):
...
+        the function so that instead it returns a fixed element. This element is set to 0 or 1 in
+        the test, resulting in different nodes being added and removed from the starting
+        subgraph."""
+        g = nx.complete_graph(dim)
+        s = list(range(dim - 2))
+        min_size = dim - 3
+        max_size = dim - 1
+        w = [1] * dim
+
+        def choice(x, element):
```
ixfoduap: This function was already defined @ line 354

strawberryfields/apps/subgraph.py
trbromley: Have added more details.

One reason for factoring out this function was because pylint/codefactor complained on the number of `if` statements in `resize`.

trbromley:
Also had a play around with the functions, seems to work ok

strawberryfields/apps/subgraph.py
josh146: Since Python does not use name checking at 'compile' time, only when the defined function/method is executed, this is typically quite idiomatic in Python. You see it quite a lot in class methods as well.

trbromley: Thanks @josh146

strawberryfields/apps/subgraph.py
```diff
@@ -80,6 +102,9 @@ def search(
         min_size (int): minimum size to search for dense subgraphs
         max_size (int): maximum size to search for dense subgraphs
         max_count (int): maximum number of densest subgraphs to keep track of for each size
+        node_select (str, list or array): method of settling ties when more than one node of
+            equal degree can be added/removed. Can be ``"uniform"`` (default), or a numpy array or
```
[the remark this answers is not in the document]
trbromley: Thanks, I added this as a separate commit as there were multiple instances.

PR #298, opened 2020-02-21
Config refactor

antalszava:
**Context:**
Currently, to configure a connection to a remote service the `Configuration` class is used which encompasses the creation and updating of configuration-related data, allowing environment variables and configuration files to be used. However, no keyword arguments can be passed to this class for configuration. Furthermore, storing an instance of this class that it is available from different components of SF (e.g. `RemoteEngine`, `LocalEngine`) seems to be problematic, as a *global* `Configuration` instance would be needed for such use cases.

This is the first part of replacing the `Configuration` class with functions, adding the `load_config` and auxiliary functions.

The configuration object (that is a nested dictionary) would be created based on the following (order defines the importance, going from most important to least important):

- keyword arguments passed to load_config
- data contained in environmental variables (if any)
- data contained in a configuration file (if exists)

If a user wants to simply use a configuration, the `configuration` object is available from `confiugration.py` which calls on `load_config` without any arguments. This solves the issue of having a *global* configuration if needed. A default configuration object called `DEFAULT_CONFIG` is also available containing pre-defined data for configuration.

*Further changes regarding current implementation*
* If there are several configuration files at multiple locations, the very first to be found will be regarded
* Moved the creation of a default config object into a function: now the defaults are through having the `kwargs.get("option", default)` calls
* Iteration through the dictionary will occur multiple times -> means two iterations but creates a cleaner logic

*Further comments*
* Functionality and tests have been removed for saving configuration files as well, these will be added in the subsequent PR
* Tests that were failing due to parts of SF that will be soon refactored were marked `xfail`

**Benefits:**
- The `load_config` function can be called anytime for configuration, without having to keep and update the state that was contained in a `Configuration` instance.
- There is now an opportunity to load the configuration from keyword arguments and environment variables as well.
- When parsing environment variables, a logic for parsing integers has been added.
- The current use cases were taken into consideration.
- More tests were added.

**Possible Drawbacks:**
Currently, all three options of keyword arguments, environment variables and configuration file are taken into consideration, eventually giving a union of the options while also taking into consideration the precedence. For now, this wouldn't seem to be a problem with limited details available.

**Related GitHub Issues:**
N/A


tests/frontend/test_api_client.py
```diff
@@ -132,7 +132,7 @@ def json(self):
     def raise_for_status(self):
         raise requests.exceptions.HTTPError()
 
-
```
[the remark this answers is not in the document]
antalszava: These will be updated in the corresponding API Client refactor PR https://github.com/XanaduAI/strawberryfields/pull/294 :blush:

tests/frontend/test_engine.py
```diff
@@ -317,6 +317,7 @@ def test_run(self, starship_engine, monkeypatch):
             compile_options={}
         )
 
+    @pytest.mark.xfail
```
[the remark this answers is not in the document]
antalszava: Will be coming up in #294 :)

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+        with monkeypatch.context() as m:
+            m.setattr(os, "getcwd", lambda: tmpdir)
+            configuration = conf.load_config()
+
+        assert configuration == OTHER_EXPECTED_CONFIG
+
+        tear_down_all_env_var_defs()
+
+    def test_conf_file_loads_well(self, monkeypatch, tmpdir):
+        """Test that the data in environment variables precedence over data in
```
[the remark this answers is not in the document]
antalszava: Thanks!

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+        with monkeypatch.context() as m:
+            m.setattr(os, "getcwd", lambda: tmpdir)
+            m.setattr(conf, "load_config_file", lambda filepath: filepath)
+            config_file = conf.look_for_config_in_file(filename=filename)
+
+        assert config_file == tmpdir.join(filename)
+
+    def test_loading_env_variable(self, tmpdir, monkeypatch):
+        """Test that the correct configuration file is found using the correct
+        environment variable.
```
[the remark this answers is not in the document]
antalszava: Yes, I'll include it in the docstring. Thank you!

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+            raise ex
+
+        with monkeypatch.context() as m:
+            m.setattr(os, "getcwd", lambda: "NoConfigFileHere")
+            m.setattr(os.environ, "get", lambda x, y: tmpdir if x=="SF_CONF" else "NoConfigFileHere")
+            m.setattr(conf, "load_config_file", lambda filepath: raise_wrapper(FileNotFoundError()) if "NoConfigFileHere" in filepath else filepath)
+
+            # Need to mock the module specific function
+            # m.setattr(conf, "user_config_dir", lambda *args: "NotTheFileName")
+            # os.environ["SF_CONF"] = lambda: FileNotFoundError
```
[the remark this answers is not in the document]
antalszava: Thanks for getting this! The first was related to mocking out the third directory (`user_config_dir`), which although in principle shouldn't make a difference (as the path contained in `os.environ["SF_CONF"]` takes precedence, so the line of execution should break before it reaches `user_config_dir`), it is good to include such that the test does not pass this way for sure. The second one was included in a more refined version on line 208.

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+        config = conf.create_config_object()
+        assert config["api"]["authentication_token"] == ""
 
-        with open(filename, "w") as f:
-            f.write(TEST_FILE)
+        conf.update_with_other_config(config, EXPECTED_CONFIG)
+        assert config == EXPECTED_CONFIG
 
-        host = "https://6.4.2.4"
+    ONLY_AUTH_CONFIG = {
```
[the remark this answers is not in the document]
antalszava: Could do that, one reason to leave them here would be that the upcoming test uses them (and so there is no scrolling needed if they were to be checked out). Would be happy to move them to the top though if you think it's better that way! :)

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+
+    ONLY_DEBUG_CONFIG = {
+                "api": {
+                        "debug": "PlaceHolder",
+                                                            }
+    }
 
-    def test_parse_environment_variable(self, monkeypatch):
+    ONLY_PORT_CONFIG = {
+            "api": {"port": "PlaceHolder"}
```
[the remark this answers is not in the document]
antalszava: Sure!

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+        """Tests that only one item is updated in the configuration object is updated."""
+        config = conf.create_config_object()
+        assert config["api"][specific_key] != "PlaceHolder"
+
+        conf.update_with_other_config(config, config_to_update_with)
+        assert config["api"][specific_key] == "PlaceHolder"
+        assert all(v != "PlaceHolder" for k, v in config["api"].items() if k != specific_key)
+
+
+value_mapping = [
```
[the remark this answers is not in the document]
antalszava: In contrast to `environment_variables_with_keys_and_values`, these were not used when defining fixtures, but purely inside of the test functions. I'm up for having a cleaner way, though this was the closest I could get :smile:

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,371 @@
...
+
+        conf.update_from_environment_variables(config)
+        assert config["api"][key] == parsed_values_mapping[env_var]
+
+        for v, (key, parsed_value) in zip(config["api"].values(), parsed_values_mapping.items()):
+            if key != env_var:
+                assert v != parsed_value
+
+        # Tear-down
+        del os.environ[env_var]
```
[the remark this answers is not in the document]
antalszava: Yeah, so [apparently these are process specific changes](https://www.geeksforgeeks.org/python-os-environ-object/):

> `os.environ` behaves like a python dictionary, so all the common dictionary operations like get and set can be performed. We can also modify os.environ but any changes will be effective only for the current process where it was assigned and it will not change the value permanently.

The reason why they are included here is that if the entire test suite is run, then there might be state left between tests. That is, sometimes I came across funky situations :)

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+    config_from_keyword_arguments = {"api": kwargs}
+    update_with_other_config(config, other_config=config_from_keyword_arguments)
+
+    return config
+
+def create_config_object(authentication_token="", **kwargs):
+    """Create a configuration object that stores configuration related data
+    organized into sections.
+
+    Currently API related configuration options are defined. This function
```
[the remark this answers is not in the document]
antalszava: Sounds good to me!

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+        config (dict[str, dict[str, Union[str, bool, int]]]): the
+            configuration to be updated
+        other_config (dict[str, dict[str, Union[str, bool, int]]]): the
+            configuration used for updating
+
+    Returns:
+        dict[str, dict[str, Union[str, bool, int]]]): the updated
+            configuration
+    """
+    # Here an example for sectionconfig is API
```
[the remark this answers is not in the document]
antalszava: Yes, so this is connected to the repeated use of the word "Currently" :smile: The comment just wanted to indicate that the `"api"` is an example for a `sectionconfig` (and actually the only example for now). However, since it seems to be confusing, I'll just remove it :)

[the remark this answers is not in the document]
antalszava: I just realized that the entire confusion comes from the fact that the comment itself should say `section` instead of `sectionconfig`, as each `section` is a key in the outer dictionary. Would it be a useful comment as such or would it be better to remove it as intended?

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+    for section, sectionconfig in config.items():
+        for key in sectionconfig:
```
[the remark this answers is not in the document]
antalszava: That's a fair point! That would include, however, making the implicit assumption that `other_config` is a valid configuration object. That is if `other_config={"SomeNonDefinedSection":{"SomeConfig":None}}`, and we iterate through its keys, then we might introduce non-recognized options in the configuration.

As `load_config` is intended to be developer-facing, the chance for this is rather low, however.

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
-
-    def save(self, filepath):
-        """Save a configuration file.
-
-        Args:
-            filepath (str): path to the configuration file
-        """
-        with open(filepath, "w") as f:
-            toml.dump(self._config, f)
+DEFAULT_CONFIG = create_config_object()
```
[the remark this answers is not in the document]
antalszava: This would accessible through the following way for example:
```python
In [1]: from strawberryfields import configuration as conf                      

In [2]: conf                                                                    
Out[2]: <module 'strawberryfields.configuration' from '/strawberryfields/strawberryfields/configuration.py'>

In [3]: conf.configuration                                                      
Out[3]: 
{'api': {'authentication_token': "MyAuth",
  'hostname': "HostName",
  'use_ssl': True,
  'port': 123,
  'debug': False}}

In [4]: conf.DEFAULT_CONFIG                                                     
Out[4]: 
{'api': {'authentication_token': '',
  'hostname': 'localhost',
  'use_ssl': True,
  'port': 443,
  'debug': False}}
```
`conf.configuration` stores the configuration details that were obtained when the code in `configuration.py` and so `load_config` was run.

antalszava:
Thank you for the comments @co9olguy! I've implemented and replied to them

antalszava:
Locally I got the following for `configuration.py`:
```
Coverage for strawberryfields/configuration.py : 100%
69 statements   69 run 0 missing 3 excluded
```
Whereas the changes in other files will be affected by an upcoming refactor there.

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+
+
+def load_config(filename="config.toml", **kwargs):
+    """Load configuration from keyword arguments, configuration file or
+    environment variables.
 
-BOOLEAN_KEYS = ("debug", "use_ssl")
+    .. note::
+
+        The configuration object (a nested dictionary) will be created based
```
josh146: While technically true, referring to it as an object first and dictionary later can be a bit confusing, as I first think it is an object of a class
```suggestion
        The configuration dictionary will be created based
```

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+        authentication_token (str): the token to be used for user
+            authentication
+        hostname (str): the name of the host to connect to
+        use_ssl (bool): specifies if requests should be sent using SSL
+        port (int): the port to be used when connecting to the remote service
+        debug (bool): determines if the debugging mode is requested
+
+    Returns:
+        dict[str, dict[str, Union[str, bool, int]]]: the configuration
+            object
```
josh146:
```suggestion
```

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+        3. data contained in a configuration file (if exists)
+
+    Keyword arguments:
+        filename (str): the name of the configuration file to look for
+        authentication_token (str): the token to be used for user
+            authentication
+        hostname (str): the name of the host to connect to
+        use_ssl (bool): specifies if requests should be sent using SSL
+        port (int): the port to be used when connecting to the remote service
+        debug (bool): determines if the debugging mode is requested
```
josh146: I'm thinking it might be best to move all the configuration options to the module docstring. This has several advantages:

* Less redundancy --- configuration options only need to be specified in one place, and linked from everywhere else

* We can use Sphinx glossary notation to add more details

* It is user facing, whereas these docstrings aren't

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+
+    Returns:
+        dict[str, dict[str, Union[str, bool, int]]]): the updated
+            configuration
+    """
+    # Here an example for section is API
+    for section, sectionconfig in config.items():
+        for key in sectionconfig:
+            if key in other_config[section]:
+                config[section][key] = other_config[section][key]
```
josh146: Could this be made more succinct using the `dict.update` method?

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+    for section, sectionconfig in config.items():
+        env_prefix = "SF_{}_".format(section.upper())
+        for key in sectionconfig:
+            env = env_prefix + key.upper()
+            if env in os.environ:
+                config[section][key] = parse_environment_variable(key, os.environ[env])
+
+
+BOOLEAN_KEYS = ("debug", "use_ssl")
+INTEGER_KEYS = ("port")
```
josh146: Since these are defined as module level variables, I would suggest:

1. Moving them to the top of the file

2. Including the default values as well.

These could then be programmatically accessed from `create_config`. This is nice from a readability point of view --- if the default keys/types of the keys ever need to be updated, there is a 'default_config' dict at the top of the file that contains the sections, keys, types, and default values.

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+        * The current working directory
+        * The directory specified by the environment variable SF_CONF (if specified)
+        * The user configuration directory (if specified)
+
+    Keyword arguments:
+        filename (str): the configuration file to look for
+
+    Returns:
+         dict[str, dict[str, Union[str, bool, int]]] or None: the
+             configuration object that was loaded
```
josh146: This method returns a tuple it looks like

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+    directories = [current_dir, sf_env_config_dir, sf_user_config_dir]
+    for directory in directories:
+        filepath = os.path.join(directory, filename)
+        try:
+            parsed_config = load_config_file(filepath)
+            break
+        except FileNotFoundError:
+            parsed_config, filepath = None, None
+
+    return parsed_config, filepath
```
josh146: How come the filepath needs to be returned?

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+            "hostname": hostname,
+            "use_ssl": use_ssl,
+            "port": port,
+            "debug": debug
+            }
+    }
+    return config
+
+def look_for_config_in_file(filename="config.toml"):
+    """Looks for the first configuration file to be found at certain paths.
```
josh146:
```suggestion
    """Loads the first configuration file found from the defined configuration directories.
```

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+            "authentication_token": authentication_token,
+            "hostname": hostname,
+            "use_ssl": use_ssl,
+            "port": port,
+            "debug": debug
+            }
+    }
+    return config
+
+def look_for_config_in_file(filename="config.toml"):
```
josh146: I might suggest an updated method name here, to signify that this function doesn't just look for the config file, but actually loads and returns it. Something concise :)

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+
+    Returns:
+         dict[str, dict[str, Union[str, bool, int]]]: the configuration
+            object that was loaded
+    """
+    with open(filepath, "r") as f:
+        config_from_file = toml.load(f)
+    return config_from_file
+
+def update_with_other_config(config, other_config):
```
josh146: This could be a case where a more verbose name makes it slightly harder to parse. I would suggest `update_config` is sufficient here, as it is similar to the `dict.update()` (and feels quite pythonic)

josh146: Actually, is this method needed at all? Would `config.update(other_config)` work?

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+        log.info("No Strawberry Fields configuration file found.")
 
+    update_from_environment_variables(config)
+
+    config_from_keyword_arguments = {"api": kwargs}
+    update_with_other_config(config, other_config=config_from_keyword_arguments)
+
+    return config
+
+def create_config_object(authentication_token="", **kwargs):
```
josh146: Suggest changing this to `create_config`

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+        update_with_other_config(config, other_config=parsed_config)
+    else:
+        log.info("No Strawberry Fields configuration file found.")
 
+    update_from_environment_variables(config)
+
+    config_from_keyword_arguments = {"api": kwargs}
+    update_with_other_config(config, other_config=config_from_keyword_arguments)
+
+    return config
```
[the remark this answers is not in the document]
josh146: The logic flow is really clear!

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
+    parsed_config, _ = look_for_config_in_file(filename=filename)
+
+    if parsed_config is not None:
+        update_with_other_config(config, other_config=parsed_config)
+    else:
+        log.info("No Strawberry Fields configuration file found.")
 
+    update_from_environment_variables(config)
+
+    config_from_keyword_arguments = {"api": kwargs}
```
josh146: It might make sense to define the kwargs as 'belonging' to the "api" section when the defaults/types are defined.

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
-            raise ValueError("Boolean could not be parsed")
-    else:
-        return value
-
-
-class ConfigurationError(Exception):
-    """Exception used for configuration errors"""
 
+        if value in falses:
+            return False
```
josh146: Much clearer!

strawberryfields/configuration.py
```diff
@@ -15,138 +15,229 @@
...
-        """Save a configuration file.
-
-        Args:
-            filepath (str): path to the configuration file
-        """
-        with open(filepath, "w") as f:
-            toml.dump(self._config, f)
+DEFAULT_CONFIG = create_config_object()
+configuration = load_config()
+config_file_path = look_for_config_in_file()[1]
```
josh146: Perhaps this indicates that `look_for_config_in_file()` should simply be changed to `get_config_filepath()`, and only return the filepath? Them `load_config()` can be called manually.

tests/frontend/test_configuration.py
```diff
@@ -46,107 +46,360 @@
...
+                    "SF_API_DEBUG",
+                    "SF_API_PORT"
+                    ]
+
+def tear_down_all_env_var_defs():
+    """Making sure that no environment variables are defined after."""
+    for key in environment_variables:
+        if key in os.environ:
+            del os.environ[key]
+            assert key not in os.environ
```
josh146: This is neat, but I would recommend using the pytest `monkeypatch.setenv()` and `monkeypatch.delenv()` inside the tests, rather than doing it by hand

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
josh146: Mentioned this earlier, but better to use `monkeypatch.setenv()` here to avoid side effects 🙂

strawberryfields/configuration.py
antalszava: Would that be referring to the defaults used in `create_config`?

One of the problems I came across was that if I use `create_config` and `kwargs` here, then having defaults will result in an unexpected overwrite.

That is
```python
In [1]: import strawberryfields.configuration as conf
In [3]:     config = { 
   ...:         "api": { 
   ...:             "authentication_token": 'NotDefaultAuth', 
   ...:             "hostname": 'NotDefaultHostname', 
   ...:             "use_ssl": True, 
   ...:             "port": 123456789, 
   ...:             "debug": False 
   ...:             } 
   ...:     } 
   ...: other_config = conf.create_config(authentication_token="MyToken") 
   ...: conf.update_config(config, other_config) 
   ...: config                                                                  
Out[3]: 
{'api': {'authentication_token': 'MyToken',
  'hostname': 'localhost',
  'use_ssl': True,
  'port': 443,
  'debug': False}}
```
where  `'hostname'` and `'port'` were overwritten, whereas they were not defined in `kwargs`.

strawberryfields/configuration.py
antalszava: Was relying on the previous implementation here :)

strawberryfields/configuration.py
antalszava: That is really amazing comment, thank you so much! I've updated the function names, logic and tests accordingly :)

tests/frontend/test_configuration.py
antalszava: Great, thanks for that!

strawberryfields/configuration.py
antalszava: I've moved this to the top now :) As for a default config, it is accessible from `DEFAULT_CONFIG` which calls on `create_config` with no parameters. Would you prefer to have it such that it is explicitly included as well?

strawberryfields/configuration.py
antalszava: I've changed it so, however, will have to add certain validity check most probably, such that no problems can arise from updating with invalid configurations.

strawberryfields/configuration.py
josh146: No I think that's a good way of doing it!

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
co9olguy: Should they be deleted?

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
co9olguy: No big preference to do that, just found the location of different things inconsistent

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
co9olguy: Again no strong preference, it was just hard to tell the reasoning why some things were located in/outside of classes

strawberryfields/configuration.py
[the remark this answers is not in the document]
co9olguy: new version looks fine :+1:

strawberryfields/configuration.py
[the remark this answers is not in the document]
co9olguy: Got it!

strawberryfields/configuration.py
```diff
@@ -124,26 +127,20 @@ def look_for_config_in_file(filename="config.toml"):
...
-        try:
-            parsed_config = load_config_file(filepath)
-            break
-        except FileNotFoundError:
-            parsed_config, filepath = None, None
+        if os.path.exists(filepath):
+            return filepath
 
-    return parsed_config, filepath
+    return None
```
co9olguy: redundant
```suggestion
```

strawberryfields/configuration.py
antalszava: I've changed this such that there the `dict.update` method is being used :)

One thing to note is that it cannot be replaced entirely. This is because assuming that not all options are defined in `kwargs`, calling on `update` on the outer dictionary does not update the inner dictionary, but overwrites it (updating is not recursive for the nested structure):

```python
In [1]: a = {'a': {'b': 3}}                                                     

In [2]: a                                                                       
Out[2]: {'a': {'b': 3}}

In [3]: b = {'a': {'c': 5}}                                                     

In [4]: b                                                                       
Out[4]: {'a': {'c': 5}}

In [5]: a.update(b)                                                             

In [6]: a                                                                       
Out[6]: {'a': {'c': 5}}

# Restoring a

In [7]: a = {'a': {'b': 3}}                                                     

In [8]: a                                                                       
Out[8]: {'a': {'b': 3}}

In [9]: a['a'].update(b['a'])                                                   

In [10]: a                                                                      
Out[10]: {'a': {'b': 3, 'c': 5}}
```

strawberryfields/configuration.py
antalszava: I've added them to the module docstring :) Added the list of environment variables, let me know if you think they need more adjustments :)

antalszava:
Reason for the force-push: Traversing the parents of each commit starting from the source `43d385d`, it can be seen that the master branch was merged accidentally.

strawberryfields/configuration.py
```diff
@@ -14,139 +14,231 @@
 r"""
 This module contains the :class:`Configuration` class, which is used to
 load, store, save, and modify configuration options for Strawberry Fields.
+
+Configuration options
+*********************
+
+.. note::
+    The following configuration options are taken into consideration:
```
josh146:
```suggestion
    The following configuration options are available:```

File: strawberryfields/configuration.py [hunk 32]
```diff
@@ -14,139 +14,231 @@
...
+
+        * **authentication_token (str)** (*required*): the token used for user authentication
+        * **hostname (str)** (*optional*): the name of the host to connect to
+        * **use_ssl (bool)** (*optional*): specifies if requests should be sent using SSL
+        * **port (int)** (*optional*): the port to be used when connecting to the remote service
+        * **debug (bool)** (*optional*): determines if the debugging mode is requested
```
josh146: The bullet point list needs to be de-indented to match the indentation level of `The following...` above.

strawberryfields/configuration.py
```diff
@@ -14,139 +14,231 @@
...
+
+.. note::
+
+    When loading the configuration, the following environment variables are checked:
+
+    * SF_API_AUTHENTICATION_TOKEN
+    * SF_API_HOSTNAME
+    * SF_API_USE_SSL
+    * SF_API_DEBUG
+    * SF_API_PORT
```
josh146: This is sufficient for now, but we should create a new ticket to fix up the configuration keys documentation 🙂

strawberryfields/configuration.py
```diff
@@ -14,139 +14,231 @@
...
+        important):
+
+        1. keyword arguments passed to ``load_config``
+        2. data contained in environmental variables (if any)
+        3. data contained in a configuration file (if exists)
+
+    Kwargs:
+        filename (str): the name of the configuration file to look for
+
+        Furthermore configuration options as detailed in
```
josh146:
```suggestion
        Configuration options are detailed in
```

strawberryfields/configuration.py
```diff
@@ -14,139 +14,231 @@
...
+        :mod:`strawberryfields.configuration`
```
josh146: This is great 🙂 From a maintainability point of view, there is a single place which is the source of truth for allowed configuration options!

strawberryfields/configuration.py
josh146: Ah, I see. This is because the `other_config` is replacing the default values 🤔

strawberryfields/configuration.py
```diff
@@ -14,139 +14,231 @@
...
+    """
+    current_dir = os.getcwd()
+    sf_env_config_dir = os.environ.get("SF_CONF", "")
+    sf_user_config_dir = user_config_dir("strawberryfields", "Xanadu")
+
+    directories = [current_dir, sf_env_config_dir, sf_user_config_dir]
+    for directory in directories:
+        filepath = os.path.join(directory, filename)
+        if os.path.exists(filepath):
+            return filepath
```
josh146: This is much cleaner now!

tests/frontend/test_configuration.py
```diff
@@ -46,107 +47,364 @@
...
+
+        with open(path_to_write_file, "w") as f:
+            f.write(TEST_FILE)
+
+        def raise_wrapper(ex):
+            raise ex
+
+        with monkeypatch.context() as m:
+            m.setattr(os, "getcwd", lambda: "NoConfigFileHere")
+            m.setenv("SF_CONF", tmpdir)
```
josh146: Nice :)

antalszava:
These last bit of modifications included:
- removing the `update_config` function
- adding the `keep_valid_options` function and adding tests.

The `keep_valid_options` function makes sure that only options that are recognized can be included coming from a configuration file or keyword arguments.

These modifications make up for using the `dict.update` function where applicable. As currently the only sectionconfig is `"api"`, this is used more than previously. This can be extended to more complex code once new sectionconfigs are introduced (and potentially reintroducing the logic from `update_config`)

Further changes were carried out to `configuration.rst` to update it with:
* keyword arguments
* environment variables
* port (where applicable)

strawberryfields/configuration.py
```diff
@@ -14,139 +14,214 @@
 r"""
 This module contains the :class:`Configuration` class, which is used to
 load, store, save, and modify configuration options for Strawberry Fields.
```
josh146: This sentence is now out of date!

doc/introduction/configuration.rst
```diff
@@ -33,15 +33,29 @@ and has the following format:
...
+    **port (int)** (*optional*)
+        The port to be used when connecting to the remote service.
+
+Environment variables
+*********************
+
+* SF_API_AUTHENTICATION_TOKEN
+* SF_API_HOSTNAME
+* SF_API_USE_SSL
+* SF_API_PORT
```
josh146: I would combine the two sections above into one, something like (feel free to modify/clean):

```rest
Configuration options
*********************

 **authentication_token (str)** (*required*)
    The authentication token to use when connecting to the API. Will be sent with every request in
    the header. Corresponding environment variable: ``SF_API_AUTHENTICATION_TOKEN``

**hostname (str)** (*optional*)
    The hostname of the server to connect to. Defaults to ``localhost``. Must be one of the allowed
    hosts. Corresponding environment variable: ``SF_API_HOSTNAME``

**use_ssl (bool)** (*optional*)
    Whether to use SSL or not when connecting to the API. True or False.
    Corresponding environment variable: ``SF_API_USE_SSL``

**port (int)** (*optional*)
    The port to be used when connecting to the remote service.
    Corresponding environment variable: ``SF_API_PORT``
```

strawberryfields/configuration.py
```diff
@@ -14,139 +14,214 @@
...
+
+.. warning::
+
+    The following configuration options are available:
+    :doc:`/introduction/configuration`.
```
josh146: For consistency with our other developer pages:
```rest
.. warning::

    Unless you are a Strawberry Fields developer, you likely do not need to access this module directly.
    See more details regarding Strawberry Fields configuration and available configuration options 
    on the :doc:`/introduction/configuration` page.
```

strawberryfields/configuration.py
```diff
@@ -14,139 +14,214 @@
...
+        "authentication_token": (str, ""),
+        "hostname": (str, "localhost"),
+        "use_ssl": (bool, True),
+        "port": (int, 443),
+    }
 }
 
-BOOLEAN_KEYS = ("debug", "use_ssl")
+BOOLEAN_KEYS = {"use_ssl"}
+INTEGER_KEYS = {"port"}
```
josh146: Now that the above `DEFAULT_CONFIG_SPEC` specifies the types, are these still needed?

strawberryfields/configuration.py
```diff
@@ -14,139 +14,214 @@
...
+        important):
+
+        1. keyword arguments passed to ``load_config``
+        2. data contained in environmental variables (if any)
+        3. data contained in a configuration file (if exists)
+
+    Kwargs:
+        filename (str): the name of the configuration file to look for
+
+        Configuration options are detailed in
```
josh146:
```suggestion
        Additional configuration options are detailed in
```

strawberryfields/configuration.py
```diff
@@ -14,139 +14,214 @@
...
-
-        Args:
-            filepath (str): path to the configuration file
-        """
-        with open(filepath, "w") as f:
-            toml.dump(self._config, f)
+VALID_KEYS = set(create_config()["api"].keys())
+DEFAULT_CONFIG = create_config()
+configuration = load_config()
+config_filepath = get_config_filepath()
```
josh146: This is very nice :)

strawberryfields/configuration.py
antalszava: Oh not at all, thanks for that!

strawberryfields/configuration.py
antalszava: Oh good one, thanks!

doc/introduction/configuration.rst
antalszava: Oh thanks for that!

strawberryfields/configuration.py
antalszava: Sure! Here the first part will already be included from https://github.com/XanaduAI/strawberryfields/blob/demo/doc/code/sf_configuration.rst

so I'll add the reworded second part :)

PR #299, opened 2020-02-24
Docs S2gate consistent with backends

nquesada:
The description of the S2gate given in the docs is incosistent with the one in the backends.
This PR correct the docs so that they are consistent with the backends. I have checked the decomposition given in terms of BSgate and Sgates and it is correct.

The way to confirm that indeed we implement S2gate(r,phi) = \exp( r e^(i \phi) a^\dagger b^\dagger  - \hc) is to note that

S2gate(r,phi) |0,0> = |0,0> + r e^{i \phi} |1,1>  for  r<<1.

If instead one used what was written in the docs one would get  |0,0> - r e^{i \phi} |1,1>.

### DIFF
--- a/doc/introduction/intro_photonics/conventions/gates.rst
+++ b/doc/introduction/intro_photonics/conventions/gates.rst
@@ -290,7 +290,7 @@ Two-mode squeezing
    :class: defn
 
    .. math::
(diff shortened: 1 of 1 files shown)


nquesada:
all of this is implicitly tested in ``test_two_mode_squeezing_operation.py`` that @thisac added.

PR #300, opened 2020-02-24
Unified backend results

thisac:
**Context:**
The sample results that are returned from the different backends do not have the same type nor the same dimensions. The three different return styles that exist now are:

1. When ``shots = 1``, the simulators all return a list with shape ``(modes,)``.
2. When ``shots > 1`` (which only works for the Gaussian backend with the ``MeasureFock`` measurement) they return a NumPy array with shape ``(shots, modes)``.
3. When using the hardware, it always return a NumPy array with shape ``(modes, shots)`` (even for one shot).

If ``shots == 1``, the return should be a NumPy array with shape ``(1, modes)``, consistent with ``shots > 1``. Furthermore, an error will be raised when using batching, feed-forwarding and/or post-selection with ``shots > 1``.

**Benefits:**
Consistent return types and shapes when running simulators.

**Possible Drawbacks:**
If there is code that assumes that a one-dimensional list is returned when using ``shots == 1``, it might break (since the return is now a two-dimensional NumyArray).

**Related GitHub Issues:**


strawberryfields/apps/sample.py
```diff
@@ -153,9 +153,6 @@ def sample(
         warnings.filterwarnings("ignore", category=UserWarning, message="Cannot simulate non-")
         s = eng.run(p, run_options={"shots": n_samples}).samples
 
-    if n_samples == 1:
-        return [s]
-
```
trbromley: Thanks!

thisac: You're welcome! :wink:

strawberryfields/engine.py
```diff
@@ -62,9 +62,9 @@ class Result:
     >>> print(results)
     Result: 3 subsystems
         state: <GaussianState: num_modes=3, pure=True, hbar=2>
-        samples: [0, 0, 0]
+        samples: [[0, 0, 0]]
     >>> results.samples
-    [0, 0, 0]
+    np.array([[0, 0, 0]])
```
josh146: Very thorough!

strawberryfields/engine.py
```diff
@@ -485,6 +490,22 @@ def run(self, program, *, args=None, compile_options=None, run_options=None):
...
+        for p in program_lst:
+            for c in p.circuit:
+                try:
+                    if c.op.select and eng_run_options["shots"] > 1:
+                        raise NotImplementedError("Post-selection cannot be used together with multiple shots.")
+                except AttributeError:
+                    pass
+
+                if c.op.measurement_deps and eng_run_options["shots"] > 1:
+                    raise NotImplementedError("Feed-forwarding of measurements cannot be used together with multiple shots.")
```
josh146: This is very nice and quite elegant 🤔 I wouldn't have though of using `measurement_deps`.

However, I wonder if this might add some overhead for large programs - perhaps at some point we might want to add properties to the program that return `True` for feedforward/postselection.

tests/frontend/test_engine.py
```diff
@@ -156,3 +168,40 @@ def inspect():
...
         assert inspect() == expected2
+
+
+class TestMultipleShotsErrors:
+    """Test if errors are raised correctly when using multiple shots."""
+
+    @pytest.mark.parametrize("meng", batch_engines)
+    def test_batching_error(self, meng, prog):
+        """Check that correct error is raised with batching and shots > 1."""
+        with pytest.raises(NotImplementedError) as excinfo:
```
josh146: Note that pytest has a 'match' option:
```suggestion
        with pytest.raises(NotImplementedError, match="Batching"):
```

tests/frontend/test_engine.py
```diff
@@ -156,3 +168,40 @@ def inspect():
...
+            meng.run(prog, run_options={"shots": 2})
+
+        assert "Batching" in str(excinfo.value)
```
josh146: Using match as above, this line can be deleted (the assert is implicit).

strawberryfields/engine.py
josh146: Might be better to make an issue for this, and do it in a new PR.

thisac: I was thinking the same. This was the best way I could figure out how to this early in `engine.py` though (and it seemed to work well).

tests/frontend/test_engine.py
thisac: Oh, that's nicer. I didn't know that. :smile: Thanks!

tests/integration/test_engine_integration.py
```diff
@@ -313,14 +313,19 @@ def test_measurefock_shots(self, setup_eng):
...
-            eng.run(p1, run_options={"shots": shots}).samples
+        if batch_size:
+            with pytest.raises(NotImplementedError,
+                               match="Batching cannot be used together with multiple shots."):
+                eng.run(p1, run_options={"shots": shots}).samples
+        else:
+            with pytest.raises(NotImplementedError,
+                               match=r"""(Measure|MeasureFock) has not been implemented in {} """
+                               """for the arguments {{'shots': {}}}""".format(backend_name, shots)):
+                eng.run(p1, run_options={"shots": shots}).samples
```
josh146: Should this be two separate tests, one for each assert? Seems a bit weird to have a test testing different logic.

thisac: I guess it could be. The problem was that it takes the environmental variables `BATCHED` and `BATCH_SIZE` for different runs and depending on their values different errors are raised. It's not straight forward to me how to tell a test only to run if BATCHED is True/False. I'll look at it and see if I can figure something out. Two different tests perhaps make more sense. :slightly_smiling_face:

thisac: I guess I could simply have the `if batch_size:` and `if not batch_size:` in each respective test, and do nothing if `False`, but that would mean running many instances of the tests without actually doing anything, which seems a bit unnecessary.

josh146: What I normally do is have:
```python
if batch_size:
    pytest.skip("Test only runs for non-batched backends")
```
at the top of the test. Not the best looking, but it usually does the job!

PR #301, opened 2020-02-25
Fixes a bug in the MZgate convention by reversing the internal and external phase arguments.

josh146:
**Context:** There was previously a discrepancy in the Chip2 compilation pipeline for the MZgate --- the hardware decomposition expected the internal phase followed by the external phase, whereas the MZgate was defined in the reverse.

**Description of the Change:** Updates the docstring and arguments of the MZgate to make it `MZgate(phase_in, phase_ex)`.

**Benefits:** Matches the hardware decomposition.

**Possible Drawbacks:** None.


josh146:
> And should we maybe add a note to https://github.com/XanaduAI/strawberryfields/blob/demo/.github/CHANGELOG.md, so this change does not make it into master silently?

Thanks @lneuhaus for catching this --- I keep telling everyone I'm going to add a bot that comments 'Don't forget to update the changelog!', but I'm guilty of it just as often.

strawberryfields/ops.py
```diff
@@ -1105,13 +1105,13 @@ class MZgate(Gate):
...
     """
     ns = 2
 
-    def __init__(self, phi_ex, phi_in):
-        super().__init__([phi_ex, phi_in])
+    def __init__(self, phi_in, phi_ex):
+        super().__init__([phi_in, phi_ex])
 
     def _decompose(self, reg, **kwargs):
         # into local phase shifts and two 50-50 beamsplitters
```
lneuhaus: I think we also need to adjust this method, also swapping the indices of the parameters used for the various phase shifts.

lneuhaus:
```

        # into local phase shifts and two 50-50 beamsplitters
        return [
            Command(Rgate(self.p[1]), reg[0]),
            Command(BSgate(np.pi/4, np.pi/2), reg),
            Command(Rgate(self.p[0]), reg[0]),
            Command(BSgate(np.pi/4, np.pi/2), reg)
        ]
```
right?

lneuhaus:
I was actually wondering why the chip2 circuit spec does not simply use the decomposition of the MZgate into more atomic gates for avoiding the extra logic that eventually caused this bug. Was this a deliberate choice for numerical accuracy or the like, or is that maybe something else we should aim for (if this circuit spec will be obsolete very soon, maybe its not worth the effort)?

.github/CHANGELOG.md
```diff
@@ -36,10 +36,16 @@
...
   [#280](https://github.com/XanaduAI/strawberryfields/pull/280)
-  
+
 * Removed two unnecessary else statements that pylint complained about.
   [#290](https://github.com/XanaduAI/strawberryfields/pull/290)
-  
+
+* Fixed a bug in the `MZgate`, where the internal and external phases were
+  in the wrong order in both the docstring and the argument list. The new
+  signature is `MZgate(phase_in, phase_ex)`, matching the existing hardware
```
heltluke: What do you mean `hardware decomposition`? There were two definitions directly in SF, no?

strawberryfields/ops.py
heltluke: Isn't it enough to only make the swap in one place (either the order of the arguments or where p[0] and p[1] appear)?

lneuhaus: `p[0]` is the first argument (=now internal phase), `p[1]` the second (=now external phase). The command sequence in my snippet above should first apply the external phase, then a BS, then the internal phase, then another BS. So it needs to be changed to what I proposed above, no?

heltluke: Right!

heltluke: And the math/docstring on line 1103!

heltluke:
Should we also update the tests to catch this sort of thing? I guess what would be needed would be to start from a simpler circuit and compile it to arrive at what [is already written](https://github.com/XanaduAI/strawberryfields/blob/92b5730db323cce08ce309a6024bfa6e0ef62ad2/tests/frontend/test_circuitspecs_chip2.py#L342)?

strawberryfields/ops.py
josh146: Good catch!

.github/CHANGELOG.md
josh146: Was referring to the `rectangular_symmetric` decomposition, will change it!
```suggestion
  signature is `MZgate(phase_in, phase_ex)`, matching the existing `rectangular_symmetric`
```

co9olguy:
> I was actually wondering why the chip2 circuit spec does not simply use the decomposition of the MZgate into more atomic gates for avoiding the extra logic that eventually caused this bug. Was this a deliberate choice for numerical accuracy or the like, or is that maybe something else we should aim for (if this circuit spec will be obsolete very soon, maybe its not worth the effort)?

@lneuhaus could you elaborate on what you mean here? :thinking: From what I recall, that was originally how we had it, but then it was decided to add MZgates

co9olguy:
I'll defer my review until @heltluke's two suggestions are incorporated (an updated test that would have caught this, and an additional fix to the docstring)

.github/CHANGELOG.md
```diff
@@ -36,10 +36,16 @@
 * Added `sympy>=1.5` to the list of dependencies.
   Removed the `sympy.functions.atan2` workaround now that SymPy has been fixed.
   [#280](https://github.com/XanaduAI/strawberryfields/pull/280)
-  
+
```
josh146: My text editor automatically removes trailing whitespace, hence the changes above.

strawberryfields/ops.py
```diff
@@ -1472,7 +1472,7 @@ def _decompose(self, reg, **kwargs):
 
                 if "symmetric" in mesh:
                     # Mach-Zehnder interferometers
-                    cmds.append(Command(MZgate(np.mod(phi, 2*np.pi), np.mod(theta, 2*np.pi)), (reg[n], reg[m])))
+                    cmds.append(Command(MZgate(np.mod(theta, 2*np.pi), np.mod(phi, 2*np.pi)), (reg[n], reg[m])))
```
josh146: Another one we missed 🙂

josh146:
I've added in two tests --- one that checks a specific unitary (@heltluke I'm simply using the example you posted in the Slack channel), and one that checks that having no MZgate's present correctly results in the compiled MZgates applying the identity.

In addition, I found a couple of other places where the MZgate phases had to be reversed.

> @lneuhaus could you elaborate on what you mean here? 🤔 From what I recall, that was originally how we had it, but then it was decided to add MZgates

iirc, the main reason behind the `MZgate` was to make the topology scripts more concise. Previously, we were simply defining the chips by writing 

```python
Rgate(phase_ex) | 0
BSgate(np.pi/4, np.pi/2) | [0, 1]
Rgate(phase_in) | 0
BSgate(np.pi/4, np.pi/2) | [0, 1]
```

whereas after the addition of the MZgate, this simply becomes `MZgate(phase_in, phase_ex)`.

This is essentially a failure of the test suite/test coverage for the MZgate, while BSgate and Rgate were older and more battle tested. Major thanks to @nquesada for implementing the Gaussian unitary compiler, it made writing better tests much easier!

strawberryfields/ops.py
lneuhaus: Did you search for MZgate in all files to find this one?

josh146: No, it revealed itself as I added the new tests. Goes to show the important of having proper tests with good coverage!

lneuhaus: I just ran a full text search, and did not find any additional issues withing SF.

lneuhaus: However, I realized there is now a problem in line 60+61 of circuitspecs/chip2.py (as now even and odd indices should be swapped in below docstring which is the "current, outdated" version):
```
        # even phase indices correspond to external Mach-Zehnder interferometer phases
        # odd phase indices correspond to internal Mach-Zehnder interferometer phases
```
which propagates all the way to chipdata and even our interferometer calibration routines, namely the phase array convention.

lneuhaus: So I guess the best, least error-prone fix would be to keep the phase indices in the templates as they are, just change the docstrings, and implement these changes in github.com/xanaduai/interferometerCalibration.

lneuhaus:
> I've added in two tests --- one that checks a specific unitary (@heltluke I'm using the example you posted in the Slack channel), and one that checks that having no MZgate's present correctly results in the compiled MZgates applying the identity.

They look very nice, and sorry I was too slow to provide you one.

> Note that I have also removed chip0, as it is no longer available - let me know if this should be reversed.

Good!

> @lneuhaus could you elaborate on what you mean here? thinking From what I recall, that was originally how we had it, but then it was decided to add MZgates

This was not what I meant, I like the MZgate very much for the readability reasons Josh explained above. The comment meant that there should be need to have an explicit call to mzgate (lowercase, the function that returns a unitary for an MZgate) in 
https://github.com/XanaduAI/strawberryfields/blob/8097041f844acb27836cb723d04158010dd5c2b8/strawberryfields/circuitspecs/chip2.py#L189
since we already know the gate sequence an MZgate can be decomposed into from 
https://github.com/XanaduAI/strawberryfields/blob/8097041f844acb27836cb723d04158010dd5c2b8/strawberryfields/ops.py#L1119 and could decompose it into the more elementary commands, avoiding having to use an additional function. If computation time and machine precision are no concerns, I think that, too, would (have) made the code immune to this issue, so I wondered if we wanted to implement that in retrospective.

josh146:
Oh, I see @lneuhaus. I believe the original compilation decomposed into `MZgate`, `Rgate`, and `BSgate` simply for performance --- if the user provides a circuit using `MZgate`'s, we didn't want the additional overhead of decomposing those into more `Rgate` and `BSgate`'s, expanding the circuit depth, and _then_ recompiling again.

strawberryfields/ops.py
josh146: Summarizing a conversation @lneuhaus and I had offline, the solutions here are:

* Keep the changes in this PR, but modify the templates in the rest of the stack so that internal phase comes first in MZgates. Pros: having internal phase come first is a better convention, matches the BSgate. Cons: more work.

* Revert the changes in this PR, and instead keep the MZgate using the convention where external phase comes first.

lneuhaus:
> we didn't want the additional overhead of decomposing those into more Rgate and BSgate's, expanding the circuit depth, and then recompiling again.

Sounds like a good reason.

strawberryfields/ops.py
co9olguy: @josh146 @lneuhaus which of the two solutions would you recommend?

lneuhaus: The first one, half way done already.

co9olguy:
Safe to review this one now?

josh146:
> All clear to review this one now?
Yep! Although, it looks like codecov is saying the coverage decreased, I'll look into this.

tests/frontend/test_engine.py
```diff
@@ -46,7 +46,7 @@ def starship_engine(monkeypatch):
     """
```
co9olguy: I have a feeling this file might have conflicts when we merge #294

tests/frontend/test_engine.py
```diff
@@ -348,7 +348,7 @@ def test_engine_with_mocked_api_client_sample_job(self, monkeypatch):
         monkeypatch.setattr(APIClient, "post", mock_api_client_post)
         monkeypatch.setattr(APIClient, "get", mock_get)
 
-        prog = sf.Program(4)
+        prog = sf.Program(8)
 
         sqz0 = 1.0
```
co9olguy: if not used anymore, should be deleted

tests/frontend/test_engine.py
```diff
@@ -348,7 +348,7 @@ def test_engine_with_mocked_api_client_sample_job(self, monkeypatch):
...
         sqz1 = 1.0
```
co9olguy: same here

tests/frontend/test_engine.py
```diff
@@ -357,16 +357,14 @@ def test_engine_with_mocked_api_client_sample_job(self, monkeypatch):
...
+            ops.S2gate(1, 0) | (q[2], q[6])
+            ops.S2gate(1, 0) | (q[3], q[7])
             ops.Rgate(phi0) | q[0]
             ops.BSgate(pi / 4, pi / 2) | (q[0], q[1])
-            ops.Rgate(phi1) | q[0]
-            ops.BSgate(pi / 4, pi / 2) | (q[0], q[1])
-            ops.Rgate(phi0) | q[2]
-            ops.BSgate(pi / 4, pi / 2) | (q[2], q[3])
-            ops.Rgate(phi1) | q[2]
-            ops.BSgate(pi / 4, pi / 2) | (q[2], q[3])
```
co9olguy: Any reason this was changed? From what I can see, it shouldn't actually matter what the specific program is here

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -367,6 +395,62 @@ def test_no_unitary(self, tol):
...
+        S = TMS(SQ_AMPLITUDE, 0)
+
+        expected = np.zeros([2*8, 2*8])
+        idx = np.arange(2*8).reshape(4, 4).T
+        for i in idx:
+            expected[i.reshape(-1, 1), i.reshape(1, -1)] = S
+
+        assert np.allclose(O, expected, atol=tol)
+
+    def test_mz_gate(self, tol):
```
co9olguy: @nquesada can you have a look at this test please?

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -367,6 +395,62 @@ def test_no_unitary(self, tol):
...
+
+        # extract the Gaussian symplectic matrix
+        O = res.compile("gaussian_unitary").circuit[0].op.p[0]
+
+        # construct the expected symplectic matrix corresponding
+        # to just the initial two mode squeeze gates
+        S = TMS(SQ_AMPLITUDE, 0)
+
+        expected = np.zeros([2*8, 2*8])
+        idx = np.arange(2*8).reshape(4, 4).T
```
co9olguy: What's the purpose of the transpose here? :thinking:

tests/frontend/test_circuitspecs_chip2.py
```diff
@@ -367,6 +395,62 @@ def test_no_unitary(self, tol):
...
+
+    def test_mz_gate(self, tol):
+        """Test that the Mach-Zehnder gate compiles to give the correct unitary"""
+        prog = sf.Program(8)
+
+        with prog.context as q:
+            ops.MZgate(np.pi/2, np.pi) | (q[0], q[1])
+            ops.MZgate(np.pi, 0) | (q[2], q[3])
+            ops.MZgate(np.pi/2, np.pi) | (q[4], q[5])
+            ops.MZgate(np.pi, 0) | (q[6], q[7])
```
co9olguy: To avoid sneaky edge cases, might be nice to have a test case at non-exceptional values, e.g., `np.pi/13`?

tests/frontend/test_circuitspecs_chip2.py
josh146: Without the transpose, the S2gates would be applied to modes in a 'cascade', i.e., `[0, 1]`, `[2, 3]`, `[4, 5]` etc.

With the transpose, the cascade now properly couples between the modes: `[0, 4]`, `[1, 5]`, `[2, 6]` etc.

tests/frontend/test_engine.py
josh146:
```suggestion
```

tests/frontend/test_engine.py
josh146:
```suggestion
```

tests/frontend/test_engine.py
josh146: Potentially 🤔

tests/frontend/test_engine.py
josh146: The unitary restriction on chip2 differed from chip0. On chip2, the unitary on modes 0-3 must match the unitary on modes 4-7. So this test needed to be modified to take this into account.

josh146:
> I approve, provided that we modify the docstring in the template at the beginning of the chip2 circuit definition

@lneuhaus this is now done!

tests/frontend/test_engine.py
pauljxtan: I believe so - it seems that the `StarshipEngine` tests are mocking `APIClient`, which is probably going to lead to a conflict as `APIClient` is removed in https://github.com/XanaduAI/strawberryfields/pull/294.

tests/frontend/test_circuitspecs_chip2.py
josh146: @nquesada, once you've given the ✔️ on this test, I will merge in this PR :)

josh146:
> The test seems correct. It would be good to have one where not all the phases are multiples of half pi.

Will add this now!

nquesada:
Really nice to see that the `gaussian_unitary` compiler is seeing some use :+1:

PR #306, opened 2020-02-26
Add the store account function

antalszava:
**Context:**
The features of the `Configuration` class are being replaced by functions in `configuration.py`. Saving a configuration to a configuration file is yet to be added.

Adds the `store_account` user-convenience function that uses the `save_config_to_file` auxiliary function.

Changes the `configuration.rst` documentation page to mirror the changes and provide simple examples.

**Benefits:**
Provides a simple way from with Python to configure the API related details in Strawberry Fields.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


strawberryfields/configuration.py
```diff
@@ -143,6 +143,8 @@ def get_config_filepath(filename="config.toml"):
         if os.path.exists(filepath):
             return filepath
 
+    return None
```
antalszava: This was readded to satisfy the linter :smile:

co9olguy: :thinking:

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,35 @@ Configuration options
...
-    Corresponding environment variable: ``SF_API_PORT``
\ No newline at end of file
+    Corresponding environment variable: ``SF_API_PORT``
+
+Store your account
+------------------
+
+Using the :func:`~.store_account` function, a configuration file can be created easily. It only requires specifying the authentication token. Apart from that, further configuration options can be passed as keyword arguments.
+
+Configure for the current SF project
```
co9olguy: Title could use some polish

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,35 @@ Configuration options
...
+************************************
+
+The following is an example for using ``store_account`` with defaults:
+
+.. code-block:: python
+
+    import strawberryfields as sf
+    sf.store_account("MyToken")
+
+where ``"MyToken"`` contains the user specific authentication token.
```
co9olguy:
```suggestion
where ``"MyToken"`` contains the user-specific authentication token.
```

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,35 @@ Configuration options
...
+
+It is advised to execute this code snippet **only once** per configuration. This should be done in the same directory where the SF project can be found, separately from executing any other Strawberry Fields scripts. Using the default options it will store the account in the *current working directory* by creating a ``config.toml`` file.
```
co9olguy: Will a user know what "per configuration" means? Suggest making a bit more clear

co9olguy: What happens if the user runs this code snippet twice? Does anything break?

co9olguy: What is "the SF project"? :thinking:

co9olguy: This should also be put on multiple lines

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,35 @@ Configuration options
...
+******************************
+
+The following code snippet can be used to create a configuration file for *every Strawberry Fields project*.
+
+.. code-block:: python
+
+    import strawberryfields as sf
+    sf.store_account("MyToken", location="user_config")
+
+where ``"MyToken"`` is the user specific authentication token.
```
co9olguy: I'm not clear what a "Strawberry Fields project" is :thinking:

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
 
     return value
 
+def store_account(authentication_token, filename="config.toml", location="local", **kwargs):
+    """Stores an account in a configuration file.
+
+    The configuration file can be created in the following locations:
+       - current working directory (local)
```
co9olguy:
```suggestion
       - current working directory ("local")
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
...
+       - user configuration directory (user_config)
```
co9olguy:
```suggestion
       - user configuration directory ("user_config")
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
...
+
+    Args:
+        authentication_token (str): the authentication token to use when
+            connecting to the API, it will be sent with every request in the
```
co9olguy:
```suggestion
            connecting to the API; it will be sent with every request in the
```

tests/frontend/test_configuration.py
```diff
@@ -396,3 +396,89 @@ def test_parse_environment_variable_integer(self, monkeypatch):
...
+        test_filename = "test_config.toml"
+
+        call_history = []
+        with monkeypatch.context() as m:
+            m.setattr(os, "getcwd", lambda: tmpdir)
+            m.setattr(conf, "user_config_dir", lambda *args: "NotTheCorrectDir")
+            m.setattr(conf, "save_config_to_file", lambda a, b: call_history.append((a, b)))
+            conf.store_account(authentication_token, filename=test_filename, location="local", **DEFAULT_KWARGS)
+
+        assert call_history[0][0] == EXPECTED_CONFIG
```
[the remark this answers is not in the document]
antalszava: Hmm, it doesn't necessarily have to be a list, though I'm actually not too sure how we could have a one-element solution here with `lambda` that mocks the line of execution, but mutates a variable in the test :thinking:

doc/introduction/configuration.rst
antalszava: Changing it, thanks!

doc/introduction/configuration.rst
antalszava: Sure, I'll reword it :slightly_smiling_face: Great question, it simply creates the same file each time (so upon the second call, the first file will be overwritten)

The reason why I've included it is not to encourage users to have them in their scripts and also mainly to use this once and then in most cases not really bother about configuration at all. For other cases they could delve deeper into understanding the different options.

antalszava: It was meant to refer to Strawberry Fields project, but I'll rephrase, thanks! :)

doc/introduction/configuration.rst
antalszava: I've reworded it, thanks!

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
antalszava: Yep, so after giving it another round of thought, with this flow of logic we would like to use some mutable built-ins (and not `unittest.mock MagicMock` for example). But I'm up for changing it to something more fitting :)

doc/introduction/configuration.rst
antalszava: It is worth noting that after a discussion with @josh146, I've now updated this such that the "user_config" option (that is *global*) is the default. The behaviour would be the same nonetheless, the file would be overwritten.

antalszava:
@co9olguy thank you so much for your comments, I've reworded the description where necessary and applied the changes!

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,39 @@ Configuration options
...
+The following is an example for using ``store_account`` with defaults:
+
+.. code-block:: python
+
+    import strawberryfields as sf
+    sf.store_account("MyToken")
+
+where ``"MyToken"`` contains the user-specific authentication token.
+
+It is advised to execute this code snippet **only once**, separately any other Python scripts.
```
josh146: I'm not quite sure what you mean by this statement?

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,39 @@ Configuration options
...
 **port (int)** (*optional*)
     The port to be used when connecting to the remote service.
-    Corresponding environment variable: ``SF_API_PORT``
\ No newline at end of file
+    Corresponding environment variable: ``SF_API_PORT``
+
+Store your account
+------------------
+
+Using the :func:`~.store_account` function, a configuration file can be created easily. It only requires specifying the authentication token. Apart from that, further configuration options can be passed as keyword arguments.
```
josh146:
```suggestion
Using the :func:`~.store_account` function, a configuration file containing your Xanadu Cloud credentials
can be created easily. By default, this configuration file is saved *globally*, and will be used every time
a remote job is submitted.
```

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,39 @@ Configuration options
...
+***************************
+
+The following code snippet can be run in the *same directory* of a Python script or Jupyter Notebook that uses Strawberry Fields to create a configuration file *only for this project*:
+
+.. code-block:: python
+
+    import strawberryfields as sf
+    sf.store_account("MyToken", location="local")
+
+where ``"MyToken"`` is the user specific authentication token.
```
josh146: Slight suggestion for the above: I would recommend shortening it (removing the two subsections), and just having one or two small examples. For additional details, the user can be redirected to the docstring for `store_account`, which should include all details about the function, plus more complicated examples.

Think of it as similar to how we do the 'quickstart' section in PennyLane. For example, the [measurements quickstart](https://pennylane.readthedocs.io/en/stable/introduction/measurements.html) briefly introduces the available functionality, with the user directed to go to the [expval docstring](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.expval.html) for more examples and details.

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
 
     return value
 
+def store_account(authentication_token, filename="config.toml", location="user_config", **kwargs):
+    """Stores an account in a configuration file.
+
+    The configuration file can be created in the following locations:
+       - user configuration directory ("user_config")
+       - current working directory ("local")
```
josh146:
```suggestion
    The configuration file can be created in the following locations:
     
     - A global user configuration directory ("user_config")
     - The current working directory ("local")
     
     This global user configuration directory differs depending on the operating system:
     
     * On Linux: ``~/.config/strawberryfields``
     * On Windows: ``~C:\Users\USERNAME\AppData\Local\Xanadu\strawberryfields``
     * On MacOS: ``~/Library/Application Support/strawberryfields``
     
     By default, Strawberry Fields will load the configuration and account credentials from the global
     user configuration directory, no matter the working directory. However, if there exists a configuration
     file in the *local* working directory, this takes precedence. The ``"local"`` option is therefore useful
     for maintaining per-project configuration settings.
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
...
+
+    Args:
+        authentication_token (str): the authentication token to use when
+            connecting to the API; it will be sent with every request in the
+            header
```
josh146: Since this is user facing, we can make it less technical. How about something like the following?
```suggestion
        authentication_token (str): API token for authentication to the Xanadu Cloud platform.
            This is required for submitting remote jobs using :class:`~.StarshipEngine`.
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
...
+
+    Kwargs:
+        location (str): determines where the configuration file should be saved
+
+        filename (str): the name of the configuration file to look for
+
+        Configuration options are detailed in
+        :doc:`/introduction/configuration`
```
josh146:
```suggestion
        location (str): determines where the configuration file should be saved
        filename (str): the name of the configuration file to look for

Additional configuration options are detailed in :doc:`/introduction/configuration` and can be passed
as keyword arguments.
```

strawberryfields/configuration.py
josh146: Would be good to also include examples here, using the standard `**Examples:**` heading we use in other docstrings 🙂

tests/conftest.py
```diff
@@ -47,6 +47,11 @@
     pytest.param("gaussian", marks=pytest.mark.gaussian),
 ]
 
+@pytest.fixture(scope="session")
+def test_filename():
+    """Using a filename for the tests."""
+    return "test_config.toml"
```
josh146: This seems more like test data, so it should be sufficient to have this as a test level variable, rather than as a fixture

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
josh146: Is it sufficient to simply check that the file `tmpdir/config.toml` exists? It might not be as 'pure' as the above approach, but might be cleaner(?)

tests/frontend/test_configuration.py
```diff
@@ -396,3 +396,81 @@ def test_parse_environment_variable_integer(self, monkeypatch):
...
+        with open(filepath, "w") as f:
+            f.write(TEST_FILE)
+
+        result_for_existing_file = toml.load(filepath)
+        assert result_for_existing_file == EXPECTED_CONFIG
+
+        conf.save_config_to_file(OTHER_EXPECTED_CONFIG, filepath)
+
+        result_for_new_file = toml.load(filepath)
+        assert result_for_new_file == OTHER_EXPECTED_CONFIG
```
josh146: Very thorough tests!

doc/introduction/configuration.rst
```diff
@@ -52,4 +52,39 @@ Configuration options
...
+
+Default configuration
+*********************
+
+The following is an example for using ``store_account`` with defaults:
+
+.. code-block:: python
+
+    import strawberryfields as sf
+    sf.store_account("MyToken")
```
josh146: Is `store_account` being imported into `__init__.py`?

strawberryfields/configuration.py
```diff
@@ -218,6 +220,49 @@ def parse_environment_variable(key, value):
 
     return value
 
+def store_account(authentication_token, filename="config.toml", location="user_config", **kwargs):
+    """Stores an account in a configuration file.
```
josh146: This should perhaps be modified to emphasize that this is for storing Xanadu cloud credentials primarily, since this is more user facing i.e., "Configure Strawberry Fields for access to the Xanadu cloud platform by saving your account credentials"

doc/introduction/configuration.rst
antalszava: I was trying to describe that a user should in most cases execute this code only once (and afterwards their configuration would be done) and that this code should be kept separate from other Strawberry Fields related code that they have (such that it does not get committed or shared). Just trying to help out with suggestions such that they wouldn't give away their tokens. Let me know if you would suggest rewording or including different ideas here!

trbromley: How about tying this in with a big warning not to commit your token?

doc/introduction/configuration.rst
trbromley: Also, if you do go into the details of the project-level configuration, maybe have a motivation of why one would want to do this... e.g., is it so you can have a configuration that is different to the default?

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
antalszava: I've added a solution that:
* uses a new class used for mocking
* mocks out `create_config` as well

This is because this unit test would like to merely check, that the functions created in `configuration.py` are called with the correct arguments (as per the logic of `store_account`). This way, we are not depending on the logic contained in those at all (hence there could be no side effects).

Further to this, I have included two integration tests, where the only thing that is mocked out, is either `os.getcwd` or `user_config_dir`, such that the `tmpdir` fixture can be used in these tests (and no state is left after the tests).

These tests test that the functions created in `configuration.py` indeed integrate correctly when `store_account` calls on them.

strawberryfields/configuration.py
antalszava: Sounds good! I'll update it in `configuration.rst` as well (got it from there)

strawberryfields/configuration.py
antalszava: This reads really good, thank you so much! :blush:

doc/introduction/configuration.rst
antalszava: Not yet, thanks for catching that!

doc/introduction/configuration.rst
```diff
@@ -69,18 +71,22 @@ The following is an example for using ``store_account`` with defaults:
     import strawberryfields as sf
     sf.store_account("MyToken")
 
-where ``"MyToken"`` contains the user specific authentication token.
+where ``"MyToken"`` contains the user-specific authentication token.
 
-It is advised to execute this code snippet **only once** per configuration. This should be done in the same directory where the SF project can be found, separately from executing any other Strawberry Fields scripts. Using the default options it will store the account in the *current working directory* by creating a ``config.toml`` file.
+It is advised to execute this code snippet **only once**, separately any other Python scripts.
```
co9olguy: Still seems to be a grammatical omission in this sentence. Anyway I still wasn't 100% happy with it, so I gave a go at modifying it
```suggestion
Typically, a user should only ever have to execute this code snippet once, when initially configurating their system to connect to the Xanadu cloud platform.
```

doc/introduction/configuration.rst
co9olguy: If the main message we want to convey is to treat your token carefully and not commit it, then that should be what is stated here :).

doc/introduction/configuration.rst
```diff
@@ -69,18 +71,22 @@ The following is an example for using ``store_account`` with defaults:
...
 
-Configure for every SF project
-******************************
+.. note::
 
-The following code snippet can be used to create a configuration file for *every Strawberry Fields project*.
+    Using the default options will store the account in the *Strawberry Fields configuration directory* by creating a ``config.toml`` file.
+
+Configuration for a project
```
co9olguy:
```suggestion
Project-specific configurations
```

doc/introduction/configuration.rst
```diff
@@ -69,18 +71,22 @@ The following is an example for using ``store_account`` with defaults:
...
+***************************
```
co9olguy:
```suggestion
*******************************
```

doc/introduction/configuration.rst
co9olguy: Still seems a bit confusing to discuss specific projects having separate configurations. Maybe we shouldn't anticipate the user's needs too much here and just state plainly how to set global and local configurations

tests/conftest.py
co9olguy: This was my suggestion (since the same object was being defined independently in several tests).

Upon further consideration, maybe better to just put the data directly in the asserts? e.g., `assert call_history[0][1] == tmpdir.join("test_config.toml")`

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
co9olguy: @antalszava which commit added the solution/tests you mention above? I can't seem to find them

doc/introduction/configuration.rst
antalszava: Thank you! I've added this part to the warning message

doc/introduction/configuration.rst
antalszava: I've implemented the suggestion coming from @josh146 such that here we have a smaller section only with two small examples :blush:

tests/frontend/test_configuration.py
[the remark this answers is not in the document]
antalszava: Sorry, I had this commit locally at this time, now it is indeed changed!

tests/conftest.py
antalszava: I've reverted this such that the test data are contained in the tests!

doc/introduction/configuration.rst
antalszava: @josh146 I've further added examples to the docstring of `store_account`

strawberryfields/configuration.py
```diff
@@ -67,10 +67,10 @@ def load_config(filename="config.toml", **kwargs):
     """
     config = create_config()
 
-    config_filepath = get_config_filepath(filename=filename)
+    filepath = get_config_filepath(filename=filename)
```
antalszava: This change is due to a conflict with the module attribute `config_filepath`

strawberryfields/configuration.py
```diff
@@ -218,6 +220,106 @@ def parse_environment_variable(key, value):
...
+        filename (str): the name of the configuration file to look for
+
+    Additional configuration options are detailed in :doc:`/introduction/configuration` and can be passed
+    as keyword arguments.
+    """
+    if location == "user_config":
+        directory = user_config_dir("strawberryfields", "Xanadu")
+
+        # Create target Directory if it doesn't exist
+        os.makedirs(directory, exist_ok=True)
```
antalszava: I've added this since an error would be otherwise raised if the configuration directory did not exist.

This solution follows the suggestions from here:
https://stackoverflow.com/questions/273192/how-can-i-safely-create-a-nested-directory

antalszava:
Updated the texts where applicable as best as I could, reverted the test data and included examples in the docstring. Let me know if you have further suggestions! :blush:

doc/introduction/configuration.rst
```diff
@@ -36,20 +36,61 @@ and has the following format:
...
     The port to be used when connecting to the remote service.
-    Corresponding environment variable: ``SF_API_PORT``
\ No newline at end of file
+    Corresponding environment variable: ``SF_API_PORT``
+
+Store your account
+------------------
+
+Using the :func:`~.store_account` function, a configuration file containing your Xanadu Cloud credentials
+can be created easily. By default, this configuration file is saved *globally*, and will be used every time
```
josh146:
```suggestion
will be created. By default, this configuration file is saved *globally*, and will be used every time
```

doc/introduction/configuration.rst
```diff
@@ -36,20 +36,61 @@ and has the following format:
...
+
+    By default, the account is stored in a ``config.toml`` file located in the
+    *Strawberry Fields configuration directory*.
+
+.. warning::
+    Typically, a user should only ever have to execute this code snippet once, when
+    initially configurating their system to connect to the Xanadu Cloud platform.
+    It is advised to call ``store_account`` **separately** from any other
+    Python code, such that the authentication token is not shared or committed
+    accidentally.
```
josh146: Since this is user facing, better to not call the user using third person :)

```suggestion
.. warning::

    The ``store_account`` function only needs to be executed once, when
    initially configuring your system to connect to the Xanadu cloud platform.
    
    Take care not to share or publicly commit your authentication token, as it provides
    full access to your account.
```

doc/introduction/configuration.rst
```diff
@@ -36,20 +36,61 @@ and has the following format:
...
+    Corresponding environment variable: ``SF_API_PORT``
+
+Store your account
+------------------
+
+Using the :func:`~.store_account` function, a configuration file containing your Xanadu Cloud credentials
+can be created easily. By default, this configuration file is saved *globally*, and will be used every time
+a remote job is submitted.
+
+In these examples ``"MyToken"`` contains the user-specific authentication token.
```
josh146:
```suggestion
In these examples ``"MyToken"`` should be replaced with a valid authentication token.
```

doc/introduction/configuration.rst
```diff
@@ -36,20 +36,61 @@ and has the following format:
...
+The following code snippet can be run to create a configuration file locally in
+the *same directory* of a Python script or Jupyter Notebook that uses
+Strawberry Fields:
+
+.. code-block:: python
+
+    import strawberryfields as sf
+    sf.store_account("MyToken", location="local")
+
+To check out more detailed examples visit the :func:`~.store_account`
```
josh146:
```suggestion
For more detailed examples, visit the :func:`~.store_account`
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,107 @@ def parse_environment_variable(key, value):
...
     return value
 
+def store_account(authentication_token, filename="config.toml", location="user_config", **kwargs):
+    r"""Configure Strawberry Fields for access to the Xanadu Cloud platform by
+    saving your account credentials.
+
+    The configuration file can be created in the following locations:
+     
+     - A global user configuration directory ("user_config")
+     - The current working directory ("local")
```
josh146:
```suggestion
    - A global user configuration directory (``"user_config"``)
    - The current working directory (``"local"``)
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,107 @@ def parse_environment_variable(key, value):
...
+     * On MacOS: ``~/Library/Application Support/strawberryfields``
+     
+     By default, Strawberry Fields will load the configuration and account credentials from the global
+     user configuration directory, no matter the working directory. However, if there exists a configuration
+     file in the *local* working directory, this takes precedence. The ``"local"`` option is therefore useful
+     for maintaining per-project configuration settings.
+
+    **Examples:**
+
+    In these examples ``"MyToken"`` contains the user-specific authentication
```
josh146:
```suggestion
    In these examples ``"MyToken"`` should be replaced with a valid authentication
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,107 @@ def parse_environment_variable(key, value):
...
+    token.
+
+    The access to the Xanadu Cloud can be configured simply by running the
+    following Python code:
+
+    >>> import strawberryfields as sf
```
josh146: We can assume the user knows what the `sf` module refers to here :)
```suggestion
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,107 @@ def parse_environment_variable(key, value):
...
+     file in the *local* working directory, this takes precedence. The ``"local"`` option is therefore useful
+     for maintaining per-project configuration settings.
+
+    **Examples:**
+
+    In these examples ``"MyToken"`` contains the user-specific authentication
+    token.
+
+    The access to the Xanadu Cloud can be configured simply by running the
+    following Python code:
```
josh146:
```suggestion
    Access to the Xanadu cloud can be configured as follows:
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,107 @@ def parse_environment_variable(key, value):
...
+        port = 443
+
+    You can also create the configuration file locally (in the **current
+    working directory**) the following way:
+
+    >>> import strawberryfields as sf
+    >>> sf.store_account("MyToken", location="local")
+
+    Each of the configuration options can be passed as further keyword
+    arguments as well (check out the :doc:`/introduction/configuration` page
```
josh146:
```suggestion
    arguments as well (see the :doc:`/introduction/configuration` page
```

strawberryfields/configuration.py
```diff
@@ -218,6 +220,107 @@ def parse_environment_variable(key, value):
...
+    Args:
+        authentication_token (str): API token for authentication to the Xanadu Cloud platform.
+            This is required for submitting remote jobs using :class:`~.StarshipEngine`.
+
+    Kwargs:
+        location (str): determines where the configuration file should be saved
+        filename (str): the name of the configuration file to look for
+
+    Additional configuration options are detailed in :doc:`/introduction/configuration` and can be passed
+    as keyword arguments.
```
josh146: Looks much better now! 🥇

doc/introduction/configuration.rst
josh146: Agree with @co9olguy above

PR #312, opened 2020-03-09
Refactoring the starship CLI and adding unit tests

antalszava:
**Context:**
The command line interface (CLI), `starship` has a few options which can be used. However, its functionality can be extended and tests need to be included such that most caveats for bugs are covered.

In terms of portability, the idea is to keep the command line interface platform independent.

Created a fixed CLI structure where there are two main types of options:
* General options: specified with the `--` prefix, e.g. `--ping`
* Commands: specified by the name of the command e.g. `run`

In more details this looks as follows:
_There are two general options: `--help` and `--ping`, in more details:_
```
starship: returns the help menu
starship --help: returns the help menu
starship --ping: tests the connection to the remote backend
```
_There are two commands: `configure` and `run`, in more details:_

_**Configure**_
```
starship configure [--local]: interactive wizard
starship configure [--local] --token: only the authentication token with default configurations
(in both cases the --local option creates the configuration file locally)
```

_**Run**_
```
starship run bbscript.xbb: run a blackbird script
starship run bbscript.xbb --ouput myhome/mydirectory/myresult.txt: run a blackbird script and store the result
```

In terms of code structure, the ability to create unit tests was regarded (as an example, suggestions on [how to write tests for the `argparse` method](https://stackoverflow.com/questions/18160078/how-do-you-write-tests-for-the-argparse-portion-of-a-python-module) were followed).

**Benefits:**

The idea behind separating the previously mentioned two groups:
* _General options_ are options which are auxiliary in nature and do not branch off into multiple separate options, they are less complex
* _Commands_ can have optional arguments and their functionality can be extended since their nature is more complex (e.g. `starship configure --local` where `--local` is an optional argument)

In terms of functionality, users can:
* _ping_ the remote server to see if their configuration is correct and they receive a valid response
* _configure_ their API connection by either simply setting their tokens or setting all their configuration details
* _run_ their blackbird scripts with the results either appearing on the standard output or can be written to a specified file

**Possible Drawbacks:**
To have `General options` appear when the help menu is queried, the protected member `_optionals` of the `ArgumentParser` class was set. This was the best solution to be found for this task; caution must be taken in case the implementation of the `ArgumentParser` class changes.

**Related GitHub Issues:**
N/A


strawberryfields/api/connection.py
```diff
@@ -278,3 +278,5 @@ def __repr__(self):
 
     def __str__(self):
         return self.__repr__()
+
+connection = Connection()
```
josh146: What is the reason for this change here? If this file is being imported when `strawberryfields` is imported, than it seems that a connection object will attempt to be created on import. This will cause error messages for the user if the configuration does not exist

antalszava: That's a fair point! This was added such that pinging could be used in the CLI. Previously no such object was created. Maybe it would be better to create a `Connection` object only when pinging is used with the CLI?

josh146: Sounds good!

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+        dict[str, Union[str, bool, int]]: the configuration options
+    """
+    default_config = create_config()["api"]
+
+    authentication_token = (
+        input(PROMPTS["authentication_token"].format(default_config["authentication_token"]))
+        or default_config["authentication_token"]
+    )
+
+    if authentication_token == "":
```
josh146:
```suggestion
    if not authentication_token:
```

setup.py
```diff
@@ -47,13 +47,19 @@
...
         "strawberryfields.apps",
+        "strawberryfields.cli",
         "strawberryfields.backends",
         "strawberryfields.backends.tfbackend",
         "strawberryfields.backends.fockbackend",
         "strawberryfields.backends.gaussianbackend",
     ],
     "package_data": {"strawberryfields": ["backends/data/*", "apps/data/*"]},
     "include_package_data": True,
+    "entry_points" : {
```
antalszava: An approach using `entry_points` was chosen such that the CLI can be tested with ease. This allowed creating tests comparable to those of other components.

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    configure_parser = subparsers.add_parser(
+        "configure", help="Configure each detail of the API connection."
+    )
+    configure_parser.set_defaults(func=configure)
+
+    configure_parser.add_argument(
+        "--token",
+        "-t",
+        type=str,
+        help="Configure only the token by using defaults for other options.",
```
josh146:
```suggestion
        help="Configure Strawberry Fields with your Xanadu cloud platform API token.",
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    )
+    configure_parser.add_argument(
+        "--local", "-l", action="store_true", help="Create the configure for the project."
```
josh146:
```suggestion
        "--local", "-l", action="store_true", help="Create a local configuration file in the current directory."
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+        "input", type=str, help="The input blackbird script to run.",
+    )
+    run_parser.set_defaults(func=run_blackbird_script)
+    run_parser.add_argument(
+        "--output",
+        "-o",
+        help="Path to the output file, where the results of the program will be written (stdout by default).",
+    )
+
+    return parser
```
josh146: 💪

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    "use_ssl": "Should the client attempt to connect over SSL? [{}] ",
+}
+
+
+def main():
+    """The Xanadu cloud platform command line interface.
+
+    Commands:
+        * run
+        * configure
```
josh146:
```suggestion
    
    * run
    * configure
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    """An auxiliary function for configuring the API connection to the Xanadu
+    cloud platform.
+
+    Supports configuration by:
+    * only specifying the token and using default configuration options;
+    * specifying every configuration option one by one.
+
+    Related arguments:
+    * token: the authentication token to use
+    * local: whether or not to create the configuration file locally
```
josh146:
```suggestion
    Flags:
    
    * ``--token``: The authentication token to use.
    * ``--local``: Whether or not to create the configuration file locally in the current directory.
      If not provided, the configuration file will be saved in the users configuration directory.
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    return parser
+
+
+def configure(args):
+    """An auxiliary function for configuring the API connection to the Xanadu
+    cloud platform.
+
+    Supports configuration by:
+    * only specifying the token and using default configuration options;
+    * specifying every configuration option one by one.
```
josh146: Might be best to rewrite this to make it clearer, i.e., "Can be used to simply save the authentication token with default configuration options. Alternatively, a wizard is provided for full configurability."

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Related arguments:
+    * token: the authentication token to use
+    * local: whether or not to create the configuration file locally
+
+    Args:
+        args (ArgumentParser): arguments that were specified on the command
+            line stored as attributes in an argument parser object
```
josh146: Never seen a docstring with so many instances of `[A|a]rg*` appearing 😆

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+def ping():
+    """Tests the connection to the remote backend."""
+    if Connection().ping():
+        sys.stdout.write("You have successfully authenticated to the platform!\n")
+    else:
+        sys.stdout.write("There was a problem when authenticating to the platform!\n")
+    sys.exit()
+
+
+def configure_everything():
```
josh146: Nitpick: the name `configuration_wizard` might be more descriptive; `configure_everything` somewhat hides the interactivity of this function, and the fact that the main logic in this function is the wizard rather than determining defaults.

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+        store_account(**kwargs)
+
+
+def ping():
+    """Tests the connection to the remote backend."""
+    if Connection().ping():
+        sys.stdout.write("You have successfully authenticated to the platform!\n")
+    else:
+        sys.stdout.write("There was a problem when authenticating to the platform!\n")
+    sys.exit()
```
josh146: Not 100% sure, but is the `sys.exit` necessary? Reason I ask is that if it is removed, then the user could call this from within Python:
```python
>>> sf.cli.ping()
You have successfully authenticated to the platform!
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    default_config = create_config()["api"]
+
+    authentication_token = (
+        input(PROMPTS["authentication_token"].format(default_config["authentication_token"]))
+    )
+
+    if not authentication_token:
+        sys.stdout.write("No authentication token was provided, please configure again.")
+        sys.exit()
```
josh146: Same question here with `exit()`. Does `return` allow similar behaviour?

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    ssl_default = "y" if default_config["use_ssl"] else "n"
+    ssl_input = input(PROMPTS["use_ssl"].format(ssl_default)) or ssl_default
+    use_ssl = ssl_input.upper() == "Y"
+
+    port = input(PROMPTS["port"].format(default_config["port"])) or default_config["port"]
+
+    kwargs = {
+        "authentication_token": authentication_token,
+        "hostname": hostname,
+        "use_ssl": use_ssl,
```
josh146: Should this be `"use_ssl": True`, rather than `"use_ssl": "Y"`? I thought this was a boolean property

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    Default configuration options are provided as defaults to the user.
+    Configuration options as detailed in :doc:`/introduction/configuration`.
+
+    Returns:
+        dict[str, Union[str, bool, int]]: the configuration options
+    """
+    default_config = create_config()["api"]
+
+    authentication_token = (
+        input(PROMPTS["authentication_token"].format(default_config["authentication_token"]))
```
josh146: Nitpick: does it make sense to replace `PROMPTS` here with the actual prompt string? As far as I can tell, `PROMPTS` isn't used anywhere else in this file, and would make this function a bit more readable.

Currently, the way this function is structured, any updates/additions to the module-level `PROMPTS` variable would require equivalent changes to this function as well.

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+        "port": port,
+    }
+    return kwargs
+
+def run_blackbird_script(args):
+    """Run a blackbird script.
+
+    Related arguments:
+    * input: the input blackbird script to be run
+    * output: the output file to store the results in (optional)
```
josh146: Perhaps good to have cli examples in these docstrings?

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+        sys.stdout.write("The {} blackbird script was not found.".format(args.input))
+        sys.exit()
+
+    eng = StarshipEngine(program.target)
+
+    sys.stdout.write("Executing program on remote hardware...\n")
+    result = eng.run(program)
+
+    if result and result.samples is not None:
+        write_script_results(result.samples, output_file=args.output)
```
josh146: How come we are only writing if `result.samples` is not None? Shouldn't we always attempt to write the output, even if it is non-numeric?

tests/frontend/cli/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+    """Program fixture."""
+    program = Program(2)
+    with program.context as q:
+        ops.Dgate(0.5) | q[0]
+    return program
+
+
+@pytest.fixture
+def connection():
+    """A mock connection object."""
```
josh146: This looks like a real connection object 😆

tests/frontend/cli/test_starship.py
```diff
@@ -0,0 +1,506 @@
...
+
+    def test_output(self):
+        """Test that specifying input, --output and passing the arguments to
+        the CLI sets the correct attributes."""
+        parser = cli.create_parser()
+        args = parser.parse_args(['run', 'SomeInputPath', '--output', 'SomeOutputPath'])
+
+        assert args.func is cli.run_blackbird_script
+        assert args.input == 'SomeInputPath'
+        assert args.output == 'SomeOutputPath'
```
josh146: Really nice tests!

tests/frontend/cli/test_starship.py
```diff
@@ -0,0 +1,506 @@
...
+
+class MockStoreAccount:
+    """A mock class used for capturing the arguments with which the store_account
+    function is being called."""
+
+    def __init__(self):
+        self.kwargs = None
+
+    def store_account(self, **kwargs):
+        self.kwargs = kwargs
```
josh146: How come some of the mocking functions/classes are here, and some are in conftest? Not suggesting any change here, just curious!

tests/frontend/cli/test_starship.py
```diff
@@ -0,0 +1,506 @@
...
+        self.pinging = "SuccessfulPing"
+
+class MockSysStdout:
+    """A mock class used for mocking the sys.stdout object while testing."""
+
+    def __init__(self):
+        self.write_output = []
+
+    def write(self, message):
+        self.write_output = message
```
josh146: In general, I would suggest using `capfd` or `capsys` built in fixtures rather than mocking out sys.stdout manually

tests/frontend/cli/test_starship.py
```diff
@@ -0,0 +1,506 @@
...
+        """Test that by default the configure_everything function exits with a
+        relevant message."""
+        with monkeypatch.context() as m:
+            m.setattr(builtins, "input", lambda *args: False)
+            mocked_stdout = MockSysStdout()
+            m.setattr(sys, "stdout", mocked_stdout)
+            with pytest.raises(SystemExit):
+                cli.configure_everything()
+
+            mocked_stdout.write_output == "No authentication token was provided, please configure again."
```
josh146: Should this be an assert?

tests/frontend/cli/test_starship.py
```diff
@@ -0,0 +1,506 @@
...
+        """Tests that the write_script_results function writes to the standard
+        output correctly."""
+        mocked_stdout = MockSysStdout()
+        some_samples = [1,2,3,4,5]
+
+        with monkeypatch.context() as m:
+            m.setattr(sys, "stdout", mocked_stdout)
+            cli.write_script_results(some_samples)
+
+        assert mocked_stdout.write_output == str(some_samples)
```
josh146: Really nice tests!

strawberryfields/cli/__init__.py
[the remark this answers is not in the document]
antalszava: Yes, this is also something I've been wondering about actually. That sounds good to me!

strawberryfields/cli/__init__.py
antalszava: Sounds good to me, updating it. Thank you!

strawberryfields/cli/__init__.py
antalszava: Yes, this whole argument parsing topic is really packed with nice examples of those haha

strawberryfields/cli/__init__.py
antalszava: Oh that's a great point, I'll update it! :blush:

strawberryfields/cli/__init__.py
antalszava: Hmm, that's a valid point! I've removed it completely for pinging .

strawberryfields/cli/__init__.py
antalszava: Here this would be slightly different, as simply having `return` would return to the `main` function and then a call on `store_account` for example would be placed. In the pinging case, it was lying fortunately in a separate branch.

Overall, if the user has chosen to use the configuration wizard, but has not provided an authentication token, then I find it reasonable that the configuration process (and the execution of the script) stops (hence a call on `exit` seems to be reasonable).

strawberryfields/cli/__init__.py
josh146: Never mind, I see my mistake!

antalszava: Right! It is, after all, a boolean as a result of `ssl_input.upper() == "Y"`.

This is merely used such that the user can see `[y]` in the wizard and that `True` would be stored in case they hit any of `y`, `Y` or simply enter (this results in the empty string being parsed `''` and for this reason, is the `or ssl_default` part of the statement needed in `ssl_input = input(PROMPTS["use_ssl"].format(ssl_default)) or ssl_default`). Let me know, however, if you think this could be simplified and/or made more intuitive in the code or in the interaction

strawberryfields/cli/__init__.py
antalszava: Sounds good, updated it! Thanks :)

tests/frontend/cli/conftest.py
antalszava: Oh, this is a really good catch, as it remained from a previous iteration of testing! As a matter of fact, these were not even used anymore in the test file.

For now, I've removed this file and moved the test file up a directory into `frontend`. Would you, however, suggest that indeed the `Mock` classes from the test file would go into a conftest file? (I can recreate this structure easily, if so :) )

tests/frontend/cli/test_starship.py
antalszava: See my comment on this above (namely the `conftest` was eventually not used, but can move any of these mocked functions&classes there if it would be better that way).

tests/frontend/cli/test_starship.py
antalszava: Oh that's a good one, thanks so much! Modified these ones where applicable

tests/frontend/cli/test_starship.py
antalszava: Yes! Thank you!

strawberryfields/cli/__init__.py
antalszava: I was mostly relying on previous code here. What would you mean by the output being non-numeric?

One thing to note is that currently only `samples` get written, though this can be modified with ease. Would you have any suggestions on that? :)

josh146: Ah, I'm just wondering if it's possible that the user could submit a job where `result.samples` is None, in which case they would be wondering why there is no output. But, thinking about it more, I don't think this will ever be the case!

Still, might be better to catch the case where result doesn't exist/result.samples is None so that a warning or error can be returned to the user, rather than nothing happening.

tests/frontend/cli/conftest.py
josh146: Nope, looks good how it is! My preference is always:

* Start off with fixtures/test data local/within the test file itself.

* If the fixtures/test data is too large, _or_ we identify that the fixtures can be used in other test files, we move them to conftest.

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    )
+
+    # Adding subparsers configure and input
+    subparsers = parser.add_subparsers(title="Commands")
+
+    # Adding the configure subparser
+    configure_parser = subparsers.add_parser(
+        "configure", help="Configure each detail of the API connection."
+    )
+    configure_parser.set_defaults(func=configure)
```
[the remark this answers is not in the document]
antalszava: Sure! 

_tl;dr:_
`func` is the default function to be called for a given command. This line sets that for any valid variation of `starship configure`, the `configure` function should be called with the parsed arguments (that are contained as attributes to a single `args` object).

_More details:_
With `argparse`, the basic situation is that an `ArgumentParser` object is used. This is a simple object, to which for example certain arguments can be added using the `add_argument` method. Once the parser is set-up, the last step is calling the `ArgumentParser.parse_args` method. Whatever arguments, the user passes to the CLI, they will be contained in an `args` object based on the configuration of the parser (e.g. in our case `--ping` will set the `args.ping` attribute to `True`).

In our case, the logic is slightly complex in a way that there is a main `ArgumentParser` that covers `starship` and there are two subparsers created as well. One for the `input` and one for the `run` commands. Each subparser gets "activated", whenever the corresponding command is specified (e.g. if `starship input [args]` was executed, then the options for the `configure` main parser and `input` subparser will be relevant).

`func` is the default function when being passed to the `set_defaults` method that should be called with the parsed arguments.

The logic was obtained from [a very similar example as detailed in the docs](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers) below the following:

> One particularly effective way of handling sub-commands is to combine the use of the add_subparsers() method with calls to set_defaults()

strawberryfields/cli/__init__.py
antalszava: That's a good idea! I've added examples for the configuration commands, would you have any ideas for a basic bb script that could be used here?

strawberryfields/cli/__init__.py
antalszava: Oh, that's a great idea!

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+def main():
+    """The Xanadu cloud platform command line interface.
+
+    **Example:**
+
+    The following is a simple example on getting the help message of the cloud platform command line interface. It details each of the options available.
+
+    .. code-block:: bash
+
+        $starship
```
josh146:
```suggestion
        $ starship
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+
+    Can be used to simply save the authentication token with default
+    configuration options. Alternatively, a wizard is provided for full
+    configurability.
+
+    **Example:**
+
+    The following is the easiest way to configure the connection to the cloud platform.
+
+    Here we show how the `starship configure --token MYAUTH` command can create a
```
josh146:
```suggestion
    Here we show how the ``starship configure --token MYAUTH`` command can create a
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+    return parser
+
+
+def configure(args):
+    r"""An auxiliary function for configuring the API connection to the Xanadu
+    cloud platform.
+
+    Can be used to simply save the authentication token with default
+    configuration options. Alternatively, a wizard is provided for full
+    configurability.
```
josh146: nice!

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+
+    **Example:**
+
+    The following is the easiest way to configure the connection to the cloud platform.
+
+    Here we show how the `starship configure --token MYAUTH` command can create a
+    configuration. It is worth noting that `MYAUTH` stands for the
```
josh146:
```suggestion
    configuration. It is worth noting that ``MYAUTH`` stands for the
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+    authentication token. Once done, we have a look at the newly created
+    configuration file.
+
+    Example run in a Linux based operating system:
+
+    .. code-block:: bash
+
+        $starship configure --token MYAUTH
+        $cat  ~/.config/strawberryfields/config.toml
```
josh146:
```suggestion
        $ starship configure --token MYAUTH
        $ cat  ~/.config/strawberryfields/config.toml
```

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+        hostname = "platform.strawberryfields.ai"
+        use_ssl = true
+        port = 443
+
+    The same example run on a Windows operating system:
+
+    .. code-block:: powershell
+
+        $starship configure --token MYAUTH
+        $cat C:\Users\USERNAME\AppData\Local\Xanadu\strawberryfields\config.toml
```
josh146: Awesome to also have a windows example! I didn't know `cat` worked in powershell 👀 
```suggestion
        $ starship configure --token MYAUTH
        $ cat C:\Users\USERNAME\AppData\Local\Xanadu\strawberryfields\config.toml
```

strawberryfields/cli/__init__.py
pauljxtan: Yes, it was like this before but I think it makes sense to alert the user if the result is malformed somehow :+1:

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+
+        $ starship configure --token MYAUTH
+        $ cat  ~/.config/strawberryfields/config.toml
+        [api]
+        authentication_token = "MYAUTH"
+        hostname = "platform.strawberryfields.ai"
+        use_ssl = true
+        port = 443
+
+    The same example run on a Windows operating system:
```
pauljxtan: Really great that you provided examples on multiple platforms! :+1:

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+def write_script_results(samples, output_file=None):
+    """Write the results of the script either to a file or to the standard output.
+
+    Args:
+        samples (array[float]): array of samples
+        output_file (str or None): the path to the output file, None if no output was defined
+    """
+    if output_file:
+        with open(output_file, "w") as file:
+            file.write(str(samples))
```
pauljxtan: Just thinking out loud, but I wonder now if there might be a better or more standard way to serialize the output here. Maybe we could write to an `.npy` file if the specified filename has the extension, or maybe that's trying to be too clever :thinking:

tests/frontend/test_starship.py
```diff
@@ -0,0 +1,486 @@
...
+class TestPing:
+    """Tests for the pinging mechanism of the CLI."""
+
+    def test_success(self, monkeypatch, capsys):
+        """Test that pinging was successful."""
+        with monkeypatch.context() as m:
+            m.setattr(cli, "Connection", MockSuccessfulConnection)
+            cli.ping()
+
+        out, _ = capsys.readouterr()
```
pauljxtan: This is cool, learned something new today :+1:

antalszava: Was new for me as well, @josh146 pointed me to it! :blush:

requirements.txt
```diff
@@ -1,3 +1,4 @@
+antlr4-python3-runtime==4.7.2
```
antalszava: The reason for this is e.g. Travis installing `antlr` version 4.8, but SF is only compatible with `antlr` 4.7.3 (creates a problem for the integration tests on Python 3.6).

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+def create_parser():
+    """Creates a parser to process the commands and arguments passed to the
+    command line interface.
+
+    Returns:
+        ArgumentParser: an argument parser object that defines the related
+        options
+    """
+    parser = argparse.ArgumentParser(
+        usage="starship <command> [<args>]",
```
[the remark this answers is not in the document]
antalszava: That's a great question! I've mostly redefined what is outputted as a help menu, such that it appears nice (but this is entirely personal taste, can change it). Also, in this regard, I sort of tried to follow `pip`'s example:
```shell
$ pip
Usage:   
  pip <command> [options]
```
Specifically, without this, the usage would appear like
```shell
usage: starship [-h] [--ping] {configure,run} ...
```

Which one do you think would make more sense overall? :)

strawberryfields/cli/__init__.py
```diff
@@ -0,0 +1,311 @@
...
+        help="Configure Strawberry Fields with your Xanadu cloud platform API token.",
+    )
+    configure_parser.add_argument(
+        "--local", "-l", action="store_true", help="Create a local configuration file in the current directory."
+    )
+
+    # Adding the input subparser
+    run_parser = subparsers.add_parser("run", help="Run a blackbird script.")
+    run_parser.add_argument(
+        "input", type=str, help="The input blackbird script to run.",
```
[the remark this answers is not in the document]
antalszava: Oh great, thanks for that! Updated it

strawberryfields/cli/__init__.py
antalszava: I think that's a good idea, we could provide a way for that! Maybe using `np.save` could aid in this. For now, however, I'd rather keep it as is and see if a need for such a change (or potentially an extra option) would be needed.

strawberryfields/cli/__init__.py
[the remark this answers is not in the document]
pauljxtan: I think I actually prefer the second one as it seems a bit more explicit, and not too cluttered since we don't have too many options or commands; but as you point out this is quite subjective, and I think either would be pretty good!

strawberryfields/cli/__init__.py
pauljxtan: Yes, I agree it's best to leave it as is to avoid the possibility of breaking users' current workflows. :+1:

strawberryfields/cli/__init__.py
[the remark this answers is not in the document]
antalszava: Oh, that's great feedback! As I don't have a strong preference here myself, I'll delete the `usage` part for now (and apply your suggestion :slightly_smiling_face: )

PR #317, opened 2020-03-12
measure_homodyne returns complex (tfbackend)

ziofil:
Fixes the issue where the `sympy.lambdify` function failed to multiply a complex value by a float eager tensor. Solved by having measure_homodyne cast to tf.complex64 (in the tfbackend)


strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -794,7 +794,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
 
             self._update_state(new_state)
 
-        return meas_result
+        return tf.cast(meas_result, dtype=tf.complex64)
```
co9olguy:
```suggestion
        return tf.cast(meas_result, dtype=ops.def_type)
```
for compatibility if we change things in future

co9olguy:
Seems to be 2 failing tests :thinking:

josh146:
Note: @smite and I had a conversation about this in #softwaredev, and it looks like this isn't the cause of the issue --- instead, a quickfix is required within the parameter evaluation code

josh146:
Closing this for now, as this is superceded by #320

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -794,7 +794,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
 
             self._update_state(new_state)
 
-        return meas_result
+        return tf.cast(meas_result, dtype=ops.def_type)
```
co9olguy: @josh146 @ziofil Please make sure this suggestion makes it to the branch that is taking over

josh146: Done!

PR #318, opened 2020-03-13
Correct CLI docstring and test typos

antalszava:
Correct the typos in the CLI docstring for the `usage` and add missing asserts in the general test.

The correction for the `usage` is needed, as in its final form #312 added the CLI with the `usage` being generated by default. In previous commits, a custom `usage` was defined, but this was eventually removed.


tests/frontend/test_starship.py
```diff
@@ -39,9 +39,8 @@ class TestCreateParser:
     def test_general_details(self):
         """Test the general details of the parser created."""
         parser = cli.create_parser()
-        parser._optionals.title = "General Options"
-        parser.usage="starship <command> [<args>]",
-        parser.description="These are common options when working on the Xanadu cloud platform."
+        assert parser._optionals.title == "General Options"
+        assert parser.description == "These are common options when working on the Xanadu cloud platform."
```
josh146: @antalszava, how does `sf --help` look like in the terminal? I'm having trouble visualizing where this line would appear :)

antalszava: It appears before the options in the help menu (still `starship` is being used here, will update any conflicts after the rename PR!):
```shell
usage: starship [-h] [--ping] {configure,run} ...

These are common options when working on the Xanadu cloud platform.

General Options:
  -h, --help       show this help message and exit
  --ping, -p       Tests the connection to the remote backend.

Commands:
  {configure,run}
    configure      Configure each detail of the API connection.
    run            Run a blackbird script.
```

josh146: Thanks Antal 🙂 In that case, I would suggest the following change:
```suggestion
        assert parser.description == "See below for available options and commands for working with the Xanadu cloud platform."
```

PR #320, opened 2020-03-17
[TF2] Fix casting of TensorFlow tensors in the frontend

josh146:
**Context:**

The new parameter handling in the Strawberry Fields frontend processes gate parameters symbolically using SymPy. When ready to evaluate, the symbolic expression is evaluated by lambdifying it, via the provided NumPy or TensorFlow printers.

However, there is a subtlety; while most gates simply take the gate parameters and pass them directly to the backend, some of the operations transform two real parameters `(r, phi)` to a single complex value `z = r*exp(1j*phi)`.

One example is the `Dgate`:

```python
p = self.p[0] * pf.exp(1j * self.p[1])
z = par_evaluate(p, dtype=np.complex128)
backend.displacement(z, *reg)
```

While this works fine when `self.p[0]` and `self.p[1]` are NumPy real arrays, issues arise when they are instead *real TensorFlow tensors*, as TensorFlow does not perform explicit type casting:

```pycon
>>> tf.Variable(1.0)*1j
InvalidArgumentError: cannot compute Mul as input #1(zero-based)
was expected to be a float tensor but is a complex128 tensor [Op:Mul] name: mul/
```

This propagates all the way from the frontend:

```pycon
>>> eng = sf.Engine('tf', backend_options={'cutoff_dim':5})
>>> prog = sf.Program(1)
>>> r = prog.params('r')
>>> phi = prog.params('phi')
>>> with prog as q:
...     sf.ops.Dgate(r, phi) | q[0]
>>> result= eng.run(prog, args={'r': tf.Variable(0.1), 'phi': tf.Variable(0.1)})
InvalidArgumentError: cannot compute Mul as input #1(zero-based)
was expected to be a float tensor but is a complex128 tensor [Op:Mul] name: mul/
```

There are two solutions to this:

1. **Quick fix:** simply update the SymPy expression evaluation, so that TensorFlow tensors are correctly cast to `tf.complex128` if the evaluated expression returns a complex value.
  
   - Pros: very minimal change

   - Cons: unnecessary casting

2. **Better long term fix:** the backend API should _not_ be passed complex values `z` that differ from the gates `(r, phi)` arguments. This avoids the casting altogether, and reduces redundant parameter processing.

   - Pros: no unnecessary casting, removes redundant processing, less complex logic

   - Cons: much larger change; frontend, plus all backends, must be updated. 

Solution (2) should be implemented regardless, the question is whether now, or later. Note that we have previously discussed implementing solution (2) solely to reduce complicated logic, and avoid issues with unclear conventions.

**Description of the change:**

* `par_evaluate` function now accepts a `dtype` argument, specifying whether the free parameters in the expression should be cast to a specific NumPy dtype _before_ the expression is evaluated. If not provided, no casting occurs. If the free parameters are TensorFlow tensors, the TensorFlow dtype equivalent is used for casting.

* `Coherent`, `Sgate`, `Dgate`, `BSgate` (just the `r` parameter), `Catstate`, and `S2gate` now cast all arguments to complex before the parameter transformation, _if_ the phase is non-zero.

* The `xfail` has been removed from `test_teleportation_fidelity`

**Benefits:**

The following now works, as expected:

```pycon
>>> eng = sf.Engine('tf', backend_options={'cutoff_dim':5})
>>> prog = sf.Program(1)
>>> r = prog.params('r')
>>> phi = prog.params('phi')
>>> with prog as q:
...     sf.ops.Dgate(r, phi) | q[0]
>>> result= eng.run(prog, args={'r': tf.Variable(0.1), 'phi': tf.Variable(0.1)})
>>> result
Result: 1 subsystems, state: <FockStateTF: num_modes=1, cutoff=5, pure=True, batched=False, hbar=2>
```

**Drawbacks:**

* Unnecessary casting for the above gates.

* Solution (2) should still be implemented.


tests/integration/test_tf_symbolic.py
```diff
@@ -565,7 +565,13 @@ def test_eval_true_state_all_fock_probs(self, setup_eng, cutoff, batch_size, tol
             Dgate(-ALPHA) | q[1]
 
         state = eng.run(prog).state
-        probs = state.all_fock_probs(eval=True).flatten()
+
+        if eng.backend_name == 'tf':
+            import tensorflow as tf
+            probs = tf.reshape(state.all_fock_probs(eval=True), [-1])
```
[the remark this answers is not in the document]
smite: I think currently the backend API measurement methods do not return consistent types anyway (#300 converts the results to a consistent type only later, in `Result.__init__`). I think it'd be better if all the backend API measurement methods returned an `array` of a consistent shape. This was one of the changes implemented in #111, along with the multishot capability.

strawberryfields/ops.py
```diff
@@ -534,7 +534,12 @@ def __init__(self, a=0., p=0.):
 
     def _apply(self, reg, backend, **kwargs):
         p = self.p[0] * pf.exp(1j * self.p[1])
-        z = par_evaluate(p)
+
+        if np.any(self.p[1] != 0):
```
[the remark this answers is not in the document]
josh146: Yep. Originally I didn't have it, but the batching tests then fail. I think this is another 'quick fix' on top of the casting 'quick fix'.

strawberryfields/parameters.py
```diff
@@ -161,6 +161,9 @@ def par_evaluate(params):
 
     Args:
         params (Sequence[Any]): parameters to evaluate
+        dtype (np.dtype): NumPy data type to optionally cast parameters
+            to *before* evaluating the symbolic expression. If the parameter
+            is a TensorFlow tensor, the corresponding TensorFlow dtype is used.
```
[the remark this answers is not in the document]
josh146: > Only real TF data types require casting to complex, so you're using NumPy dtypes as placeholders for TF dtypes, so the caller does not have to import tensorflow?

Yep. Another option would be to go with strings, but I thought that might be more prone to bugs in the short term? Unless you have another suggestion?

tests/integration/test_engine_integration.py
```diff
@@ -126,7 +126,7 @@ def test_return_samples(self, setup_eng):
         assert [r.val for r in prog.register] == res.samples[0].tolist()
         # first mode was measured
         if eng.backend_name == 'tf':
-            assert isinstance(res.samples[0], tf.Tensor)
+            assert isinstance(res.samples[0][0], tf.Tensor)
```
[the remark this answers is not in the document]
josh146: `res.samples[0]` is returning the following in this specific test:
```python
>>> res.samples[0]
array([[<tf.Tensor: id=520, shape=(), dtype=float64, numpy=0.02130021300212981>
  None]], dtype=np.object)
```

Is this unexpected?

tests/integration/test_tf_symbolic.py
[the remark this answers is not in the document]
josh146: >  I think it'd be better if all the backend API measurement methods returned an array of a consistent shape. This was one of the changes implemented in #111, along with the multishot capability.

I agree. Are the changes in #111 portable to the latest master?

strawberryfields/parameters.py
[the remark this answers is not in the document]
antalszava: A boolean could work here as well, as the only case seems to be real TF data types. If there would be new cases identified (and transitioning to the logic envisioned by solution 2 is still out of reach), then this could be changed to strings perhaps.

Right now the code in the other two files give the impression that `np.complex128` `dtypes` are going to be used, which without prior knowledge of the change in the PR, might lead to confusion.

strawberryfields/parameters.py
```diff
@@ -187,6 +190,17 @@ def do_evaluate(p):
         is_tf = (type(v).__module__.startswith('tensorflow') for v in vals)
         printer = 'tensorflow' if any(is_tf) else 'numpy'
         func = sympy.lambdify(atoms, p, printer)
+
+        if dtype is not None:
```
antalszava: With the boolean solution from my other comment, this could read as e.g. `if casting:` and then the value of `printer` could determine which casting should take place.

strawberryfields/ops.py
```diff
@@ -959,9 +973,9 @@ def __init__(self, p):
 
     def _decompose(self, reg, **kwargs):
         # into a displacement
-        z = self.p[0] * 1j/np.sqrt(2 * sf.hbar)
+        z = self.p[0] / np.sqrt(2 * sf.hbar)
```
smite:
```suggestion
        r = self.p[0] / np.sqrt(2 * sf.hbar)
```

strawberryfields/ops.py
```diff
@@ -959,9 +973,9 @@ def __init__(self, p):
...
         return [
-            Command(Dgate(z, 0), reg)
+            Command(Dgate(z, np.pi/2), reg)
```
smite:
```suggestion
            Command(Dgate(r, np.pi/2), reg)
```

strawberryfields/parameters.py
smite:
```suggestion
        dtype (None, np.dtype): NumPy data type to optionally cast atomic symbols
            to *before* evaluating the parameter expression. If the atom
            is a TensorFlow tensor, the corresponding TensorFlow dtype is used.
```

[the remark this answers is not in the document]
smite: Since this is just a temporary fix I think NumPy types are fine, there's no need to overthink this.

tests/integration/test_engine_integration.py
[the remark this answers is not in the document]
smite: I think all the backend methods should return consistent results, i.e. the dimensionality should not depend on the backend, but this is clearly a question for another PR.

tests/integration/test_tf_symbolic.py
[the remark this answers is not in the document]
smite: I can fix them so they are, but `tf2` should probably be merged to master first.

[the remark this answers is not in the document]
co9olguy: It keeps getting forgotten in all these discussions, but we specifically want the tf backend to return tf.tensors. If we convert these to arrays, then we are breaking the computational graph, which may negate the point of using tf backend to begin with

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -794,7 +794,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
 
             self._update_state(new_state)
 
-        return meas_result
+        return tf.cast(meas_result, dtype=ops.def_type)
```
josh146: @co9olguy

co9olguy: :+1:

tests/integration/test_tf_symbolic.py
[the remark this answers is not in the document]
smite: I guess returning tf.Tensors could be a special property of the tf backend, but the Tensors should always have the same shape (and similar dtype) as the arrays returned by the other backends. Previously we had the optional `eval` argument that decided whether the tf backend returned Tensors or arrays, right?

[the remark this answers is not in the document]
josh146: Oh, I'm not actually sure. I thought `eval` was deprecated now that we support eager mode.

tests/integration/test_engine_integration.py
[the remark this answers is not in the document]
josh146: @thisac, perhaps this is a bug in the new results shape processing when receiving TensorFlow tensors(?)

tests/integration/test_tf_symbolic.py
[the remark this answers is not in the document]
co9olguy: That's right @smite, TF backend should return tensor objects whose data has the same shape/dtype as the numpy arrays returned by the other backends.

`eval` keyword argument is no longer necessary, since TF2 has moved to eager mode and deprecated all the logic around `session`s. Now there's less need to distinguish unevaluated vs evaluated tensors. However, we still need to return tf eager tensors, not numpy arrays, in order to properly support backpropagation

tests/integration/test_engine_integration.py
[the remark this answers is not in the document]
thisac: That could definitely be the case. I'll look into it tomorrow!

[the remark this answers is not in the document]
co9olguy: Everyone always seems to forget, but the TF backend supports a special case that the other backends do not support, namely batching. The tests were originally written so that they pass with or without batching, which sometimes results in unusual-looking code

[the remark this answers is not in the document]
thisac: I looked into this and it seems like, in `engine.py`, the samples are transformed into an array and sent to the `Result` class where the samples are reshaped into `(shots, modes)` by adding `np.array([self.samples])`.

Each value in `self.samples` is either a float or a `tf.Tensor` or `None`. So, what the test above is comparing is:

1. the first value of the first shot in `self.samples` if using the `tf` backend (which should be a `tf.Tensor` if using `tf` and a float otherwise) or

2. the first shot in `self.samples` if using other backends (which should _always_ be an `ndarray`, even for the `tf` backend).

So, I don't think there's any consistency faults here really, rather the `tf` test checks the type of _one_ value in `self.samples` while the non-`tf` test checks the type of one shot (i.e. an array several values).

Should the `tf` backend return samples with the type `ndarray[tf.Tensor]` and shape (shots, modes)? In that case I think that this is actually OK. I guess the alternative would be to return a `tf.Tensor[float]` with shape `(shots, modes)`, but I'm not sure how/where that should be changed or if that is actually wanted.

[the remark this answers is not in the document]
co9olguy: Definitely the second option is preferred. The tensorflow backend should return a tensor tensor with the same size/shape as the comparable numpy array. In the case of batched mode, there should be an extra first axis (but batched mode is not co-compatible with many-shots measurement, so that shouldn't be a problem in this specific case)

Wrapping a tensorflow tensor in a numpy array is unnecessary, and may also break the computational graph unnecessarily.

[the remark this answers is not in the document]
thisac: I agree that the second option sounds nicer. Although right now the samples are set to `None` whenever there are no measurements, meaning that the samples might be an `ndarray[object]` with some floats or `tf.Tensor`s (depending on backend) and some `Nones`. Looking at the comment above from @josh146, `res.samples` is just that (an `ndarray[object]` with one `tf.Tensor` and one `None`) and, since I don't think that `tf.Tensor` supports `None` objects further changes might need to be done in the process to make everything work.

Actually, I also noticed that right now, if the tensorflow backend is used with a single shot (where all modes are measured), `engine.py` will actually cast `self.samples` into a numpy array (i.e. removing the tensorflow tensor altogether). I'm guessing this is not desirable.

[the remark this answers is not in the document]
co9olguy: The `None`s were just a placeholder and may not make sense anymore. Feel free to suggest a strategy to replace them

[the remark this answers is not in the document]
co9olguy: > Actually, I also noticed that right now, if the tensorflow backend is used with a single shot (where all modes are measured), `engine.py` will actually cast `self.samples` into a numpy array (i.e. removing the tensorflow tensor altogether). I'm guessing this is not desirable.

Again, I think this is an artifact of how we used to do things with the TF1 days. The default was for tensorflow tensors to be "evaluated", which turned them into numpy arrays. Users could specify a kwarg to force the tensors to remain "unevaluated" (i.e., abstract tensors objects with no set data value). Now that tensorflow has thrown out sessions, tensors are somehow always evaluated. When we need them to be specifically numpy objects, we use `.numpy()`

antalszava:
Just checking in here. This PR continues to stand for solution 1 outlined in the original description, correct?

josh146:
Yes, thanks @antalszava. Now that the v0.13.0rc is out, focus should be returned to this PR. It should be close to merging.

tests/frontend/test_parameters.py
```diff
@@ -206,3 +229,124 @@ def test_parameter_unary_negation(self, p):
...
+
+        # create numeric values
+        mapping = {'p': tf.Variable(0.1)}
+        prog.bind_params(mapping)
+
+        # verify bound parameters are correct
+        assert prog.free_params['p'].val is mapping['p']
+
+        # assert executed program is constructed correctly
+        global applied
```
co9olguy, line 340: how come a `global` is used here?

tests/frontend/test_parameters.py
```diff
@@ -206,3 +225,127 @@ def test_parameter_unary_negation(self, p):
...
+        applied = []
+
+        with monkeypatch.context() as m:
+            m.setattr(sf.LocalEngine, "_run_program", mock_run_prog)
+            eng = sf.LocalEngine(backend)
+            result = eng.run(prog, args=mapping)
+
+        assert isinstance(applied[0].op, gate)
+        assert applied[0].op.p[0].val == mapping["r"]
+        assert applied[0].op.p[1].val == mapping["phi"]
```
antalszava, line 284: Should there be a tear-down for `applied`? Although each test seems to have a build-up where they "reset" it, it might still be worth adding one for safety

tests/frontend/test_parameters.py
```diff
@@ -206,3 +225,127 @@ def test_parameter_unary_negation(self, p):
...
+
+        # create numeric values
+        mapping = {'r': tf.Variable(0.1), 'phi': tf.Variable(0.1)}
+        prog.bind_params(mapping)
+
+        # verify bound parameters are correct
+        assert prog.free_params['r'].val is mapping['r']
+        assert prog.free_params['phi'].val is mapping['phi']
+
+        # assert executed program is constructed correctly
```
antalszava, line 272: From this on the test case could be split up into two: one for checking the parameters and one for checking that the executed program is constructed correctly.

tests/frontend/test_parameters.py
```diff
@@ -206,3 +225,127 @@ def test_parameter_unary_negation(self, p):
...
+        global applied
+        cmds = []
+        applied = []
+
+        with monkeypatch.context() as m:
+            m.setattr(sf.LocalEngine, "_run_program", mock_run_prog)
+            eng = sf.LocalEngine(backend)
+            result = eng.run(prog, args=mapping)
+
+        assert isinstance(applied[0].op, sf.ops.Dgate)
```
antalszava, line 349: So this would be an example for decomposition, correct? Worth adding this to the docstring

tests/frontend/test_parameters.py
```diff
@@ -206,3 +225,127 @@ def test_parameter_unary_negation(self, p):
...
+        global applied
+        cmds = []
+        applied = []
+
+        with monkeypatch.context() as m:
+            m.setattr(sf.LocalEngine, "_run_program", mock_run_prog)
+            eng = sf.LocalEngine(backend)
+            result = eng.run(prog, args=mapping)
+
+        assert isinstance(applied[0].op, gate)
```
antalszava, line 282: Just to make sure that there are no leftover operations which were not tested for (this is reoccurring in the other test cases as well)
```suggestion
        assert isinstance(applied[0].op, gate)
        assert len(applied) == 1
```

tests/frontend/test_parameters.py
```diff
@@ -172,6 +172,25 @@ def test_par_evaluate(self, p):
...
+    @pytest.mark.parametrize("p", TEST_VALUES)
+    def test_par_evaluate_dtype_TF(self, p):
+        """Test the TF parameter evaluation works when a dtype is provided"""
+        pytest.importorskip("tensorflow", minversion="2.0")
+        import tensorflow as tf
+
+        x = FreeParameter("x")
+        x.val = tf.Variable(p)
+        res = par_evaluate(x, dtype=np.complex128)
+        assert res.dtype is tf.complex128
```
antalszava, line 192: Cool! :100: :blush:

tests/frontend/test_ops_decompositions.py
```diff
@@ -798,8 +798,10 @@ def test_Zgate_decomposition(self, hbar, tol):
         Z = ops.Zgate(p)
         cmds = Z.decompose(prog.register)
         assert isinstance(cmds[0].op, ops.Dgate)
-        assert np.allclose(cmds[0].op.p[0], alpha, atol=tol, rtol=0)
-        assert np.allclose(cmds[0].op.p[1], 0, atol=tol, rtol=0)
+
+        r = cmds[0].op.p[0]
```
antalszava, line 802: Just for safety reasons
```suggestion
       assert len(cmds) == 1
        r = cmds[0].op.p[0]
```

tests/frontend/test_parameters.py
antalszava: Was wondering the same thing myself! It essentially seems to help with having a module-level object, so the `applied` objects within the tests that are defined as `global applied` will be referencing the same object as the one that is used in this mock function. It is a way of switching from a local scope, though I'd imagine that this might have caveats in between tests if not torn down correctly.

antalszava:
Left a couple of comments/questions related to testing :slightly_smiling_face:

tests/frontend/test_parameters.py
josh146: This is a bit of a hack from my end; while coding the tests, I realised that a global variable would allow me to 'extract' the internal engine command list with some monkeypatching 😆 I'll rewrite this as a fixture, which will hopefully avoid issues with set up and tear down.

josh146: Have replaced this with a fixture :)

tests/frontend/test_parameters.py
josh146: Have updated the docstring and test name!

tests/frontend/test_parameters.py
josh146: I've streamlined the tests a bit --- let me know if it looks better!

I didn't separate out the program and engine run into separate tests, however. This is partly because the program mapping is already being tested earlier in the file; this bug/intergration test only affects the engine run. The previous assertion on `prog.free_params` is just a sanity check.

tests/frontend/test_parameters.py
```diff
@@ -206,3 +225,125 @@ def test_parameter_unary_negation(self, p):
...
+    Simply include this fixture in a test; any local calls to
+    ``eng.run`` will subsequently be reflected in the fixture.
+
+    Returns:
+        list (Command): commands that were applied to the engine
+    """
+    applied = []
+
+    def mock_run_prog(self, prog, **kwargs):
+        """Mock function that uses for extracting
```
antalszava, line 244:
```suggestion
        """Mock function that is used for extracting
```

tests/frontend/test_parameters.py
```diff
@@ -227,126 +227,123 @@ def test_parameter_unary_negation(self, p):
...
+                    c.op.apply(c.reg, self.backend, **kwargs)
+                    applied.append(c)
+        return applied
+
+    with monkeypatch.context() as m:
+        m.setattr(sf.LocalEngine, "_run_program", mock_run_prog)
+        yield applied
+
+    # tear down
+    applied = []
```
antalszava, line 261: Neat!! :100:

tests/frontend/test_parameters.py
antalszava: Looks better, thanks! :100:

antalszava:
Thanks for resolving my comments, looks great! :1st_place_medal:

tests/integration/test_tf_symbolic.py
```diff
@@ -565,6 +565,7 @@ def test_eval_true_state_all_fock_probs(self, setup_eng, cutoff, batch_size, tol
             Dgate(-ALPHA) | q[1]
 
         state = eng.run(prog).state
+
```
josh146, line 568:
```suggestion
```

strawberryfields/parameters.py
```diff
@@ -168,6 +168,9 @@ def par_evaluate(params):
 
     Args:
         params (Sequence[Any]): parameters to evaluate
+        dtype (None, np.dtype): NumPy data type to optionally cast atomic symbols
```
co9olguy, line 171: To be more accurate, should bracketed part include `tf.dtype` (or equivalent)?

strawberryfields/ops.py
```diff
@@ -542,7 +542,12 @@ def __init__(self, a=0.0, p=0.0):
 
     def _apply(self, reg, backend, **kwargs):
         p = self.p[0] * pf.exp(1j * self.p[1])
-        z = par_evaluate(p)
+
+        if np.any(self.p[1] != 0):
```
co9olguy, line 546: Since this pattern is repeated several times, and the only clear explanation for what it's doing is in a resolved github comment, I think it would be worthwhile to pull out into a separate helper function (with docstring which explains behaviour under non-batched and batched operation)

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -888,7 +888,7 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
...
```
co9olguy, line 891: Should we update the functions that return the state tensor to do this casting as well?

strawberryfields/ops.py
```diff
@@ -542,7 +542,12 @@ def __init__(self, a=0.0, p=0.0):
...
+            z = par_evaluate(p, dtype=np.complex128)
```
co9olguy, line 547: We should set this dtype once in a single spot, rather than independently in each function

tests/frontend/test_parameters.py
```diff
@@ -172,6 +172,25 @@ def test_par_evaluate(self, p):
...
+
+    @pytest.mark.parametrize("p", TEST_VALUES)
+    def test_par_evaluate_dtype_TF(self, p):
+        """Test the TF parameter evaluation works when a dtype is provided"""
+        pytest.importorskip("tensorflow", minversion="2.0")
+        import tensorflow as tf
+
+        x = FreeParameter("x")
+        x.val = tf.Variable(p)
+        res = par_evaluate(x, dtype=np.complex128)
```
co9olguy, line 191: same here

tests/frontend/test_parameters.py
```diff
@@ -206,3 +225,125 @@ def test_parameter_unary_negation(self, p):
...
+            gate(*params) | q
+
+        # bind numeric values
+        prog.bind_params(mapping)
+        return prog
+
+    @pytest.mark.parametrize("gate", [sf.ops.Dgate, sf.ops.Sgate, sf.ops.Coherent])
+    def test_single_mode_gate_complex_phase(self, backend, gate, applied_cmds):
+        """Test non-decomposed single mode gates with complex phase arguments."""
+        import tensorflow as tf
```
co9olguy, line 296: How come the above additions had `pytest.importorskip("tensorflow", minversion="2.0")`, but it's not used here?

strawberryfields/backends/tfbackend/circuit.py
josh146: Potentially; lets keep this in mind when we review #323.

strawberryfields/ops.py
josh146: done!

strawberryfields/ops.py
josh146: Have done!

strawberryfields/parameters.py
josh146: Good point, done.

tests/frontend/test_parameters.py
josh146: It's a class attribute for the `TestParameterTFIntegration` class, so it will be applied to every test method in this class

tests/frontend/test_parameters.py
```diff
@@ -172,6 +172,25 @@ def test_par_evaluate(self, p):
         x.default = 0.0
         assert np.all(par_evaluate(x) == p)
 
+    @pytest.mark.parametrize("p", TEST_VALUES)
+    def test_par_evaluate_dtype_numpy(self, p):
+        """Test the numpy parameter evaluation works when a dtype is provided"""
+        x = FreeParameter("x")
+        x.val = p
+        res = par_evaluate(x, dtype=np.complex128)
```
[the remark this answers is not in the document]
josh146: done

tests/frontend/test_parameters.py
josh146: done

PR #334, opened 2020-03-25
Polishing the hardware documentation

josh146:
**Context:**

The existing hardware-focused documentation consists of two pages:

* The photonic hardware quickstart page. This is top level in the table of contents, and the purpose is to provide users a _general_ and _quick_ overview for using Strawberry Fields with photonic hardware. For more advanced or specific details, the user is instead directed to docstrings and tutorials.

* A tutorial on submitting jobs to a remote chip. This page is not top level, and is instead embedded on both the hardware quickstart page above, as well as the overall 'Tutorials' page. Its aim is to provide a reproducible walkthrough of submitting remote jobs.

However, there are a couple of issues with both pages:

* The quickstart does not provide any mention of remote job management (i.e., cancelling jobs).

* The tutorial is neither reproducible _nor_ specific --- it contains a collection of loosely connected sections, and attempts to generalize to arbitrary photonic chips, rather than focusing purely on X8. Furthermore, by attempting to be generic, it becomes much more unmaintainable.

* Various typos.

**Description of the Change:**

* Have added a warning about API keys to the quickstart page, echoing the warning on the tutorial page (as requested by @co9olguy).

* The job management details are summarised on the photonic hardware quickstart page. This involved updating the docstrings in the `job.py` module.

* The tutorial has been lightly edited to (a) make it specific to the X8 chip, and (b) change the flow to be more reproducible and more closely match the existing tutorial structure.

**Benefits:**

* Tutorial is specific to the X8 chip, so is now more maintainable. No need to constantly update it as chip architectures change.

* The quickstart page does a better job summarising job management,.

* Several typos/incorrect code blocks were fixed.

**Possible Drawbacks:** n/a

**Related GitHub Issues:** n/a

PR #337, opened 2020-03-27
Unify RemoteEngine.run() and LocalEngine.run() signatures, and allow for default chip specs for each device family.

### LINKED ISSUE #335: `run()` has inconsistent signatures for the local and remote engines
While working on #334, I noticed that the `LocalEngine` and `RemoteEngine` have inconsistent signatures. In particular:

* `LocalEngine.run()`:

  ```python3
  run(self, program, *, args=None, compile_options=None, run_options=None):
  ```

* `RemoteEngine.run()`:

  ```python3
  run(self, program: Program, shots: Optional[int] = None):
  ```

This leads to the odd case where running a program for 50 shots looks like
```python
eng.run(prog, run_options={"shots": 50})
```
on the local engine (here, `run_options` passes *runtime specific* options directly to the backend `Operation.apply()` methods), but looks like
```python
eng.run(prog, shots=50)
```
on the remote engine.

josh146:
**Context:**

Currently in Strawberry Fields:

* The user must initialize an engine or a compile a program with a specific chip target, e.g., `X8_01`, `X12_01`, etc. However, there should be a way for the user to initialize an engine or compile a program for a chip _family_ (e.g., `eng.run("X8")`, `prog.compile("X8")`), with the Remote engine delegating the job to a specified default for each family.

* `LocalEngine.run()` and `RemoteEngine.run()` have inconsistent signatures:

  ```python3
  LocalEngine.run(self, program, *, args=None, compile_options=None, run_options=None)
  RemoteEngine.run(self, program: Program, shots: Optional[int] = None)
  ```

While these appear as two separate issues, they are quite interconnected. Since `eng.run()` internally calls the `prog.compile()` method, both engines must keep track and pass on any `run_options` that the user specifies so that they can be taken into account during compilation/execution.

* The base `X8Spec` and `X12Spec` specifications are now included in the Strawberry Fields circuit spec database, with short names `X8` and `X12` respectively. Programs that are compiled against these specs, e.g., `prog.compile("X8")` have `prog.target == "X8"`.

* The `RemoteEngine` can now be instantiated with targets `"X8"` and `"X12"`.

* The `RemoteEngine` has a class attribute
  ```python
  DEFAULT_TARGETS = {"X8": "X8_01", "X12": "X12_01"}
  ```
  which specifies the default chip to target for each device family. Hardcoding this is a temporary fix, and will be removed once the chip API is online.

* For now, since only chip family compilation is available (i.e., there is no chip specific compilation), compiling against either `"X8"` and `"X8_01"` makes no difference. Therefore, `eng=RemoteEngine("X8")` will only recompile a program if `program.target[:2] != "X8"`.

* Old references to `REMOTE` in the `BaseEngine` have been removed, since `RemoteEngine` no longer inherits from `BaseEngine`.

* `RemoteEngine.run()` and `LocalEngine.run()` signatures have both been standardized to the following:

  ```python
  run(run(self, program, *, args=None, compile_options=None, **run_options)
  ```

  This is a compromise that aims to balance a nicer, more consistent UI against internal behaviour. In particular:

  - The `run_options={}` dictionary argument from the LocalEngine has been changed to `**run_options`. This allows for the syntax `eng.run(prog, shots=100)`, which is cleaner and more intutive than `eng.run(prog, run_options={"shots": 100})`.

  - `compile_options={}` dictionary argument has been added to the RemoteEngine, as well as compilation (previously the remote engine was not validating/compiling programs).

* `RunEngine.__init__()` now takes a `backend_options={}` argument, allowing backend options to be set. This is not needed currently, but will be needed for remote simulators (e.g., passing cutoff dimension).

* The `Connection.create_job()` method has been updated to send _all_ run and backend options, rather than just `shots` (useful for future compatibility with remote simulators).

**Benefits:** see above.

**Possible Drawbacks:**

* Currently, compiling a program against `"X8"` and `"X8_01"` makes no difference. However, in the future, compiling against `"X8"` will compile to match the (static) family architecture, while `X8_01` will compile to match (dynamic) specific hardware constraints of chip 01 from the API service.

* Hardcoding of the default chip specs is temporary, and will be removed once a chip API is online.

**Related GitHub Issues:**

- Fixes #335 
- Will allow implementation of suggested changes from #334

### DIFF
--- a/strawberryfields/circuitspecs/X12.py
+++ b/strawberryfields/circuitspecs/X12.py
@@ -29,7 +29,7 @@ from .gbs import GBSSpecs
 
 # Supporting multiple string formatting, such that the target can be replaced
 # first followed by squeezing amplitude and phase values
-DEFAULT_CIRCUIT = textwrap.dedent(
+X12_CIRCUIT = textwrap.dedent(
     """\
     name template_6x2_X12
     version 1.0
@@ -113,7 +113,7 @@ class X12Specs(CircuitSpecs):
     remote = True
     local = True
     interactive = False
-    circuit = DEFAULT_CIRCUIT.format(target=short_name)
+    circuit = X12_CIRCUIT.format(target=short_name)
 
     sq_amplitude = 1.0
 
@@ -122,6 +122,8 @@ class X12Specs(CircuitSpecs):
         "Interferometer": {"mesh": "rectangular_symmetric", "drop_identity": False},
         "BipartiteGraphEmbed": {"mesh": "rectangular_symmetric", "drop_identity": False},
     }
+
+    circuit = X12_CIRCUIT.format(target=short_name)
 
     def compile(self, seq, registers):
         """Try to arrange a quantum circuit into a form suitable for X12.
@@ -275,11 +277,11 @@ class X12_01(X12Specs):
     """Circuit specifications for the first X12 chip."""
 
     short_name = "X12_01"
-    circuit = DEFAULT_CIRCUIT.format(target=short_name)
+    circuit = X12_CIRCUIT.format(target=short_name)
 
 
 class X12_02(X12Specs):
     """Circuit specifications for the second X12 chip."""
 
     short_name = "X12_02"
-    circuit = DEFAULT_CIRCUIT.format(target=short_name)
+    circuit = X12_CIRCUIT.format(target=short_name)
--- a/strawberryfields/circuitspecs/X12.py
+++ b/strawberryfields/circuitspecs/X12.py
@@ -123,8 +123,6 @@ class X12Specs(CircuitSpecs):
         "BipartiteGraphEmbed": {"mesh": "rectangular_symmetric", "drop_identity": False},
     }
 
-    circuit = X12_CIRCUIT.format(target=short_name)
-
     def compile(self, seq, registers):
         """Try to arrange a quantum circuit into a form suitable for X12.
 
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -481,6 +481,7 @@ class RemoteEngine:
         target (str): the target device
         connection (strawberryfields.api.Connection): a connection to the remote job
             execution platform
+        backend_options (Dict[str, Any]): keyword arguments for the backend
     """
 
     POLLING_INTERVAL_SECONDS = 1
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -364,16 +364,13 @@ class LocalEngine(BaseEngine):
                 ) from None
         return applied
 
-    def run(self, program, *, args=None, compile_options=None, **run_options):
+    def run(self, program, *, args=None, compile_options=None, **kwargs):
         """Execute quantum programs by sending them to the backend.
 
         Args:
             program (Program, Sequence[Program]): quantum programs to run
             args (dict[str, Any]): values for the free parameters in the program(s) (if any)
             compile_options (None, Dict[str, Any]): keyword arguments for :meth:`.Program.compile`
-
-        Returns:
-            Result: results of the computation
 
         Keyword Args:
             shots (int): number of times the program measurement evaluation is repeated
@@ -385,6 +382,9 @@ class LocalEngine(BaseEngine):
                 TF backend only.
             feed_dict (dict[str, Any]): TensorFlow feed dictionary, used when evaluating returned measurement results
                 and state. TF backend only.
+
+        Returns:
+            Result: results of the computation
         """
         args = args or {}
         compile_options = compile_options or {}
@@ -403,7 +403,7 @@ class LocalEngine(BaseEngine):
             program_lst = [program]
             temp_run_options.update(program.run_options)
 
-        temp_run_options.update(run_options or {})
+        temp_run_options.update(kwargs or {})
         temp_run_options.setdefault("shots", 1)
         temp_run_options.setdefault("modes", None)
 
@@ -489,11 +489,15 @@ class RemoteEngine:
     DEFAULT_TARGETS = {"X8": "X8_01", "X12": "X12_01"}
 
     def __init__(self, target: str, connection: Connection = Connection(), backend_options: dict = None):
-        if target not in self.VALID_TARGETS:
+        self._target = self.DEFAULT_TARGETS.get(target, target)
+
+        if self._target not in self.VALID_TARGETS:
             raise ValueError(
-                "Invalid engine target: {} (valid targets: {})".format(target, self.VALID_TARGETS)
+                "Invalid engine target: {} (valid targets: {})".format(
+                    target, tuple(self.DEFAULT_TARGETS.keys()) + self.VALID_TARGETS
+                )
             )
-        self._target = target
+
         self._connection = connection
         self._backend_options = backend_options or {}
 
@@ -515,7 +519,7 @@ class RemoteEngine:
         """
         return self._connection
 
-    def run(self, program: Program, *, compile_options=None, **run_options) -> Optional[Result]:
+    def run(self, program: Program, *, compile_options=None, **kwargs) -> Optional[Result]:
         """Runs a blocking job.
 
         In the blocking mode, the engine blocks until the job is completed, failed, or
@@ -533,9 +537,9 @@ class RemoteEngine:
 
         Returns:
             [strawberryfields.api.Result, None]: the job result if successful, and
-                ``None`` otherwise
+            ``None`` otherwise
         """
-        job = self.run_async(program, compile_options=compile_options, **run_options)
+        job = self.run_async(program, compile_options=compile_options, **kwargs)
         try:
             while True:
                 job.refresh()
@@ -552,7 +556,7 @@ class RemoteEngine:
             self._connection.cancel_job(job.id)
             return None
 
-    def run_async(self, program: Program, *, compile_options=None, **run_options) -> Job:
+    def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
         """Runs a non-blocking remote job.
 
         In the non-blocking mode, a ``Job`` object is returned immediately, and the user can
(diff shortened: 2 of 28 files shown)


strawberryfields/circuitspecs/X12.py
```diff
@@ -123,6 +123,8 @@ class X12Specs(CircuitSpecs):
         "BipartiteGraphEmbed": {"mesh": "rectangular_symmetric", "drop_identity": False},
     }
 
+    circuit = X12_CIRCUIT.format(target=short_name)
+
```
antalszava: Seems to be redundant (see line 116)
```suggestion
```

strawberryfields/engine.py
```diff
@@ -369,21 +364,17 @@ def _run_program(self, prog, **kwargs):
                 ) from None
         return applied
 
-    def run(self, program, *, args=None, compile_options=None, run_options=None):
+    def run(self, program, *, args=None, compile_options=None, **run_options):
```
antalszava: Wondering here: should the `run_options` = keyword arguments be mentioned in the docstring?

strawberryfields/engine.py
```diff
@@ -556,21 +551,40 @@ def run(self, program: Program, shots: Optional[int] = None) -> Optional[Result]
...
                 argument is not provided, the shots are derived from the given ``program``.
 
         Returns:
             strawberryfields.api.Job: the created remote job
         """
-        return self._connection.create_job(self.target, program, shots)
+        # get the specific chip to submit the program to
+        # TODO: this should be provided by the chip API, rather
+        # than built-in to Strawberry Fields.
+        target = self.DEFAULT_TARGETS.get(self.target, self.target)
```
antalszava: Just for clarification here: if a chip family was defined (in `self.target`), then get the default specific chip, else a specific chip was defined (in `self.target`) so we simply need to use that, correct?

strawberryfields/engine.py
```diff
@@ -556,21 +551,40 @@ def run(self, program: Program, shots: Optional[int] = None) -> Optional[Result]
...
+        run_options = run_options.update(self._backend_options)
+
+        if program.target is None or (program.target.split("_")[0] != target.split("_")[0]):
+            # Program is either:
+            #
+            # * uncompiled (program.target is None)
+            # * compiled to a different chip family to the engine target
+            #
+            # In both cases, recompile the program to match the intended target.
+            program = program.compile(target, **compile_options)
```
antalszava: Cool!

strawberryfields/engine.py
```diff
@@ -494,14 +485,16 @@ class RemoteEngine:
 
     POLLING_INTERVAL_SECONDS = 1
     VALID_TARGETS = ("X8_01", "X12_01", "X12_02")
+    DEFAULT_TARGETS = {"X8": "X8_01", "X12": "X12_01"}
 
-    def __init__(self, target: str, connection: Connection = Connection()):
+    def __init__(self, target: str, connection: Connection = Connection(), backend_options: dict = None):
```
antalszava: Could be nice to reflect this in the docstring, also, adding the fact that this overrides any `run_options`, should there be options in both (if I understand correctly the change in `io.p`)

strawberryfields/program.py
```diff
@@ -500,7 +500,7 @@ def compile(self, target, **kwargs):
         # create the compiled Program
         compiled = self._linked_copy()
         compiled.circuit = seq
-        compiled._target = target
+        compiled._target = db.short_name
```
antalszava: Nice catch!

strawberryfields/circuitspecs/X12.py
josh146: Good catch!

strawberryfields/engine.py
josh146: I'm wondering as well. `Keyword Args:` is part of the sphinx.napoleon extension though, so I left is as that in the end 🤔

antalszava: Sure I think that makes sense. I think it's fine to leave it as is or to add a (slightly unconventional) note to the beginning of the section on this. For me either works fine :)

strawberryfields/engine.py
josh146: Added it to the docstring!

The distinction between backend options and run options actually comes from the LocalEngine:

* `backend_options`: global options for the entire backend, required when initializing a backend (e.g., `cutoff_dim`)

* `run_options`: options specified during the backend execution; these are passed directly to the operation `apply` methods (`shots`, etc.).

The blackbird program however doesn't distinguish; backend options and run options are merged into a single dictionary, hence why it looks like they are being overwritten.

For the remote engine, distinguishing between backend options and run options is more semantics than the local engine, since both options apply remotely during `run`. We might revisit this in an ADR. For now, though, since this is a bugfix, I think it's enough to align remote engine with the local engine.

strawberryfields/engine.py
josh146: Yep! It does look a bit weird, but essentially:

* If `self.target` is a DEFAULT_TARGET key, replace it with the corresponding default value.

* Else, just use the value of `self.target`.

Note that validation of `self.target` is already done in `__init__`, hence why it is not repeated here.

strawberryfields/engine.py
```diff
@@ -280,15 +278,12 @@ def _broadcast_nones(val, dim):
...
-                self._run_program(p, **kwargs)
-                shots = kwargs.get("shots", 1)
-                self.samples = [
-                    _broadcast_nones(p.reg_refs[k].val, shots) for k in sorted(p.reg_refs)
-                ]
-                self.run_progs.append(p)
+            self._run_program(p, **kwargs)
+            shots = kwargs.get("shots", 1)
+            self.samples = [
+                _broadcast_nones(p.reg_refs[k].val, shots) for k in sorted(p.reg_refs)
```
co9olguy: @josh146 @thisac `_broadcast_nones` is candidate for removal in future PR

strawberryfields/engine.py
```diff
@@ -532,14 +526,16 @@ def run(self, program: Program, shots: Optional[int] = None) -> Optional[Result]
 
         Args:
             program (strawberryfields.Program): the quantum circuit
+
+        Keyword Args:
             shots (Optional[int]): The number of shots for which to run the job. If this
```
co9olguy: How does this show up on rendered docs? Does the name `run_options` appear in signature, but here `shots` is listed as a kwarg (essentially just a renamed `kwargs`)

strawberryfields/engine.py
thisac: I'm not sure, but I guess so if the `None`s should be removed from samples somehow. `_broadcast_nones` merely takes a single `None` in a mode and reshapes it into a list of `None`s with length `shots`, right?

strawberryfields/engine.py
antalszava: Exactly!

strawberryfields/engine.py
thisac: What should the placeholder be instead of `None` when there are no results in that case? @josh146 can probably give a better answer to this.

strawberryfields/engine.py
antalszava: image

strawberryfields/engine.py
thisac: Maybe I'm missing something, but I cannot find `self.target` as an attribute, only `self._target`. Is this a typo?

Also, shouldn't the validation of `self._target` also allow for the `DEFAULT_TARGET` keys as well (otherwise an error will be raised when initializing with `X8` or `X12`, since they are not in `VALID_TARGETS`).

strawberryfields/engine.py
```diff
@@ -556,21 +552,40 @@ def run(self, program: Program, shots: Optional[int] = None) -> Optional[Result]
...
                 argument is not provided, the shots are derived from the given ``program``.
 
         Returns:
             strawberryfields.api.Job: the created remote job
         """
-        return self._connection.create_job(self.target, program, shots)
+        # get the specific chip to submit the program to
+        # TODO: this should be provided by the chip API, rather
+        # than built-in to Strawberry Fields.
+        target = self.DEFAULT_TARGETS.get(self.target, self.target)
```
thisac:
```suggestion
        target = self.DEFAULT_TARGETS.get(self._target, self._target)
```

strawberryfields/engine.py
```diff
@@ -556,21 +552,40 @@ def run(self, program: Program, shots: Optional[int] = None) -> Optional[Result]
...
+        compile_options = compile_options or {}
+        run_options = run_options.update(self._backend_options)
+
+        if program.target is None or (program.target.split("_")[0] != target.split("_")[0]):
```
thisac: Seems to be `self._target` in the `Program` class as well.

```suggestion
        if program._target is None or (program._target.split("_")[0] != target.split("_")[0]):
```

strawberryfields/engine.py
```diff
@@ -556,21 +552,40 @@ def run(self, program: Program, shots: Optional[int] = None) -> Optional[Result]
...
+            # Program is either:
+            #
+            # * uncompiled (program.target is None)
```
thisac:
```suggestion
            # * uncompiled (program._target is None)
```

strawberryfields/engine.py
```diff
@@ -490,18 +481,21 @@ class RemoteEngine:
...
 
     POLLING_INTERVAL_SECONDS = 1
     VALID_TARGETS = ("X8_01", "X12_01", "X12_02")
+    DEFAULT_TARGETS = {"X8": "X8_01", "X12": "X12_01"}
 
-    def __init__(self, target: str, connection: Connection = Connection()):
+    def __init__(self, target: str, connection: Connection = Connection(), backend_options: dict = None):
         if target not in self.VALID_TARGETS:
             raise ValueError(
                 "Invalid engine target: {} (valid targets: {})".format(target, self.VALID_TARGETS)
```
thisac: What if `target` is a key in `DEFAULT_TARGETS` but not in `VALID_TARGETS`, like `X8` and `X12`? Shouldn't these also be valid targets (since they are changed to the default target further down)?

strawberryfields/engine.py
thisac: Ignore my first comment. I just missed the `@property` being defined. Sorry.

thisac:
The tests fail because of the new `run_options` convention. I created a new branch where I've updated all of the instances of `run_options` as dictionaries, e.g. `run_options={"shots": shots}`, to keyword arguments instead. I'm not sure if this is the easiest way to fix this, though. Please tell me if I should rather create a separate PR for this.

The branch with all the changes is called [shots_as_kwargs](https://github.com/XanaduAI/strawberryfields/tree/shots_as_kwargs), and should fix the `run_options` in all files (including tutorials and documentation).

strawberryfields/engine.py
josh146: Oops, looks like there is a formatting issue in the `Returns:` line!

strawberryfields/engine.py
josh146: Ah, good catch! I've fixed this by moving the line
```python
self._target = self.DEFAULT_TARGETS.get(target, target)
```
here instead of in `eng.run()`.

It also makes a lot more sense to have it here, since that logic is independent of program execution, and didn't need to be called each time.

strawberryfields/engine.py
josh146: @co9olguy, how come? Currently `_broadcast_nones` is used a lot in the local simulators; when a mode is not terminated in measurement, the same result is currently `None`.

strawberryfields/engine.py
josh146: Thanks for catching that @thisac, have fixed it in https://github.com/XanaduAI/strawberryfields/pull/337/files#r399648997 and added tests!

strawberryfields/engine.py
josh146: Unfortunately, attempting to change `Keyword Args` to anything not recognised by `sphinx.napoleon` leads to formatting/parsing issues:

image

The other option is to just change `**runtime_options` to `**kwargs`?

josh146: Have updated it to `**kwargs` in all places it appeared

PR #343, opened 2020-03-31
Fix RemoteEngine.run_async kwargs bug, and add test

josh146:
**Context:**

The `RemoteEngine.run_async()` method had the erroneous line
```python
kwargs = kwargs.update(self._backend_options)
```
The `dict.update` method is an in-place method, and returns `None`, resulting in an incorrectly compiled blackbird program.

* The line above has been modified to remove the assignment.

* A tests has been added to catch this behaviour in future.

* The logic in determining the run options has been moved to `RemoteEngine.run_async`, as it makes more logical sense here than as part of the `Connection` object.

* I noticed that SF was not providing a default value for the number of shots in the remote engine. Local engine has `shots=1` by default, but this is because the `fock` and `tf` backends don't support multishot sampling yet, so isn't as applicable to the remote engine. The `RemoteEngine.run_async` method now raises an error if number of shots is not provided.

**Benefits:**

* Fixes a bug

* More robust remote engine tests

**Possible Drawbacks:** n/a

### DIFF
--- a/tests/api/test_remote_engine.py
+++ b/tests/api/test_remote_engine.py
@@ -70,7 +70,7 @@ class TestRemoteEngine:
     def test_run_complete(self, connection, prog, job_to_complete):
         """Tests a successful blocking job execution."""
         engine = RemoteEngine("X8_01", connection=connection)
-        result = engine.run(prog)
+        result = engine.run(prog, shots=10)
 
         assert np.array_equal(result.samples, np.array([[1, 2], [3, 4]]))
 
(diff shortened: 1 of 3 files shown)


tests/api/test_remote_engine.py
```diff
@@ -102,3 +103,62 @@ def test_device_class_target(self):
         target = "X8"
         engine = RemoteEngine(target)
         assert engine.target == engine.DEFAULT_TARGETS[target]
+
```
antalszava: Would it make sense to have some of these in a separate `class TestRemoteEngineIntegration`?

josh146: I think `test_compilation` definitely! The others are not testing integration due to mocking, and are solely testing internal engine logic, so are probably fine to leave as is.

Will move `test_compilation`, and then merge.

PR #345, opened 2020-04-01
Fix image in hardware tutorial

trbromley:
The "SU(4)" text is currently not displaying correctly in the chip diagram. It is overlapping the text of the following word.

This PR adds a `.png` version of the image and uses that.

Currently:
image

With this PR:
image


josh146:
Oh, thanks for this Tom! For some reason, it displays fine on my PC (using Windows), so I never noticed this... Perhaps it's a linux vs. windows rendering problem?

trbromley:
Yes maybe! I think @co9olguy reported the same. Thanks

PR #348, opened 2020-04-01
Number expectation

nquesada:
Adds the method `number_expectation` in `BaseGaussianState` and `BaseFockState`.

### DIFF
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -823,6 +823,10 @@ class BaseFockState(BaseState):
         Return:
             (float): the expectation value.
         """
+        if len(modes) != len(set(modes)):
+            raise ValueError("there can be no duplicates in modes")
+
+
         state = self._data #representation of the quantum state either as tensor of probability amplitudes
         # or as
         pure = self._pure #if pure or as a tensor of density matrix elements if not pure.
@@ -847,7 +851,7 @@ class BaseFockState(BaseState):
             ps = np.tensordot(np.identity(cutoff), ps, axes=((0, 1), (2 * mode, 2 * mode + 1)))
         for mode in range(len(modes)):
             ps = np.tensordot(np.diag(values), ps, axes=((0, 1), (0, 1)))
-        return ps
+        return float(ps)
 
 class BaseGaussianState(BaseState):
     r"""Class for the representation of quantum states using the Gaussian formalism.
@@ -1152,6 +1156,8 @@ class BaseGaussianState(BaseState):
         Return:
             (float): the expectation value.
         """
+        if len(modes) != len(set(modes)):
+            raise ValueError("there can be no duplicates in modes")
         mu = self._mu
         cov = self._cov
         if len(modes) == 1:
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -816,7 +816,7 @@ class BaseFockState(BaseState):
         return mean, var
 
     def number_expectation(self, modes):
-        """
+        r"""
         Calculates the expectation value of the product of the number operators of the modes.
 
         This method computes the analytic expectation value
@@ -828,28 +828,51 @@ class BaseFockState(BaseState):
 
         Return:
             (float): the expectation value.
+
+        **Example**
+
+        Consider the following program:
+
+        .. code-block:: python
+
+            prog = sf.Program(3)
+
+            with prog.context as q:
+                ops.Sgate(0.5) | q[0]
+                ops.Sgate(0.5) | q[1]
+                ops.Sgate(0.5) | q[2]
+                ops.BSgate(np.pi/3, 0.1) |  (q[0], q[1])
+                ops.BSgate(np.pi/3, 0.1) |  (q[1], q[2])
+
+        Executing this on the Fock backend,
+
+        >>> eng = sf.Engine("fock", backend_options={"cutoff_dim": 10})
+        >>> state = eng.run(prog).state
+
+        we can compute the expectation value :math:`\langle \hat{n}_0\hat{n}_2\rangle`:
+
+        >>> state.number_expectation([0, 2])
+        0.07252895071309405
         """
         if len(modes) != len(set(modes)):
-            raise ValueError("there can be no duplicates in modes")
+            raise ValueError("There can be no duplicates in the modes specified.")
 
         state = self._data #representation of the quantum state either as tensor of probability amplitudes
-        # or as
-        pure = self._pure #if pure or as a tensor of density matrix elements if not pure.
         cutoff = self._cutoff #Fock space cutoff.
         num_modes = self._modes #number of modes in the state.
 
         values = np.arange(cutoff)
         traced_modes = tuple(item for item in range(num_modes) if item not in modes)
-        if pure is True:
+        if self.is_pure:
             # state is a tensor of probability amplitudes
-            ps = np.abs(state)**2
+            ps = np.abs(self.ket())**2
             ps = ps.sum(axis=traced_modes)
             for _ in modes:
                 ps = np.tensordot(values, ps, axes=1)
-            return ps
+            return float(ps)
 
         # state is a tensor of density matrix elements in the SF convention
-        ps = state.real
+        ps = np.real(self.dm())
         traced_modes = list(traced_modes)
         traced_modes.sort(reverse=True)
         for mode in traced_modes:
@@ -1154,14 +1177,41 @@ class BaseGaussianState(BaseState):
         return mean, var
 
     def number_expectation(self, modes):
-        """
+        r"""
         Calculates the expectation value of the product of the number operators of the modes.
+
+        .. warning:: This method only supports at most two modes in the Gaussian backend.
 
         Args:
             modes (list): list of modes for which one wants the expectation of the product of their number operator.
 
         Return:
             (float): the expectation value.
+
+        **Example**
+
+        Consider the following program:
+
+        .. code-block:: python
+
+            prog = sf.Program(3)
+
+            with prog.context as q:
+                ops.Sgate(0.5) | q[0]
+                ops.Sgate(0.5) | q[1]
+                ops.Sgate(0.5) | q[2]
+                ops.BSgate(np.pi/3, 0.1) |  (q[0], q[1])
+                ops.BSgate(np.pi/3, 0.1) |  (q[1], q[2])
+
+        Executing this on the Fock backend,
+
+        >>> eng = sf.Engine("gaussian")
+        >>> state = eng.run(prog).state
+
+        we can compute the expectation value :math:`\langle \hat{n}_0\hat{n}_2\rangle`:
+
+        >>> state.number_expectation([0, 2])
+        0.07566984755267293
         """
         if len(modes) != len(set(modes)):
             raise ValueError("There can be no duplicates in the modes specified.")
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -25,6 +25,7 @@ from scipy.linalg import block_diag
 from scipy.stats import multivariate_normal
 from scipy.special import factorial
 from scipy.integrate import simps
+from thewalrus.quantum import photon_number_mean, photon_number_covar
 
 import strawberryfields as sf
 from .shared_ops import rotation_matrix as _R
@@ -823,7 +824,7 @@ class BaseFockState(BaseState):
             (float): the expectation value.
         """
         state = self._data #representation of the quantum state either as tensor of probability amplitudes
-        # or as 
+        # or as
         pure = self._pure #if pure or as a tensor of density matrix elements if not pure.
         cutoff = self._cutoff #Fock space cutoff.
         num_modes = self._modes #number of modes in the state.
@@ -833,20 +834,20 @@ class BaseFockState(BaseState):
         if pure is True:
             # state is a tensor of probability amplitudes
             ps = np.abs(state)**2
-            ps = ps.sum(axis = traced_modes)
+            ps = ps.sum(axis=traced_modes)
             for mode in modes:
                 ps = np.tensordot(values, ps, axes=1)
             return ps
-        else:
-            # state is a tensor of density matrix elements in the SF convention
-            ps = state.real
-            traced_modes = list(traced_modes)
-            traced_modes.sort(reverse=True)
-            for mode in traced_modes:
-                ps = np.tensordot(np.identity(cutoff), ps, axes=((0,1),(2*mode,2*mode+1)))
-            for mode in range(len(modes)):
-                ps = np.tensordot(np.diag(values), ps, axes=((0,1),(0,1)))
-            return ps
+
+        # state is a tensor of density matrix elements in the SF convention
+        ps = state.real
+        traced_modes = list(traced_modes)
+        traced_modes.sort(reverse=True)
+        for mode in traced_modes:
+            ps = np.tensordot(np.identity(cutoff), ps, axes=((0, 1), (2 * mode, 2 * mode + 1)))
+        for mode in range(len(modes)):
+            ps = np.tensordot(np.diag(values), ps, axes=((0, 1), (0, 1)))
+        return ps
 
 class BaseGaussianState(BaseState):
     r"""Class for the representation of quantum states using the Gaussian formalism.
@@ -1142,6 +1143,28 @@ class BaseGaussianState(BaseState):
         var -= np.sum([np.linalg.det(self._hbar*quad_coeffs[:, m][n]) for m in modes for n in modes])
 
         return mean, var
+
+    def number_expectation(self, modes):
+        """
+        Calculates the expectation value of the product of the number operators of the modes.
+        Args:
+            modes (list): list of modes for which one wants the expectation of the product of their number operator.
+        Return:
+            (float): the expectation value.
+        """
+        mu = self._mu
+        cov = self._cov
+        if len(modes) == 1:
+            return photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
+
+        if len(modes) == 2:
+            ni = photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
+            nj = photon_number_mean(mu, cov, modes[1], hbar=self._hbar)
+            return photon_number_covar(mu, cov, modes[1], modes[0], hbar=self._hbar) + ni * nj
+
+        raise ValueError("number_expectation only supports one or two modes")
+
+
 
     @abc.abstractmethod
     def reduced_dm(self, modes, **kwargs):
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -839,7 +839,7 @@ class BaseFockState(BaseState):
 
         This method computes the analytic expectation value
         :math:`\langle \hat{n}_{i_0} \hat{n}_{i_1}\dots \hat{n}_{i_{N-1}}\rangle`
-        for a (sub)set of modes `[i_0, i_1, \dots, i_{N-1}]` of the system.
+        for a (sub)set of modes :math:`[i_0, i_1, \dots, i_{N-1}]` of the system.
 
         Args:
             modes (list): list of modes for which one wants the expectation of the product of their number operator.
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -357,6 +357,10 @@ class BaseState(abc.ABC):
         >>> state = eng.run(prog).state
         >>> state.number_expectation([0, 2])
         0.07566984755267293
+        
+        This slight difference in value compared to the result from the Fock backend above
+        is due to the finite Fock basis truncation when using the Fock backend; this can be
+        avoided by increasing the value of ``cutoff_dim``.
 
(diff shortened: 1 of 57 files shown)


nquesada:
For some reason CodeFactor is complaining about things I did not do. Not sure how to fix it. Other than that this PR is ready for review.

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
 
         return mean, var
 
+    def number_expectation(self, modes):
+        """
+        Calculates the expectation value of the product of the number operators of the modes.
```
antalszava, line 820:
```suggestion
        Calculates the expectation value of the product of the number operators of the modes.

```

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
...
+        Return:
+            (float): the expectation value.
+        """
+        if len(modes) != len(set(modes)):
+            raise ValueError("there can be no duplicates in modes")
+
+
+        state = self._data #representation of the quantum state either as tensor of probability amplitudes
+        # or as
+        pure = self._pure #if pure or as a tensor of density matrix elements if not pure.
```
antalszava, line 832: The comments seem to be slightly off here, I assume they would be one single comment?

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
...
+        cutoff = self._cutoff #Fock space cutoff.
+        num_modes = self._modes #number of modes in the state.
+
+        values = np.arange(cutoff)
+        traced_modes = tuple(item for item in range(num_modes) if item not in modes)
+        if pure is True:
```
antalszava, line 838: (Small suggestion): this could be simplified as follows
```suggestion
        if pure:
```

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
...
+            # state is a tensor of probability amplitudes
+            ps = np.abs(state)**2
+            ps = ps.sum(axis=traced_modes)
+            for mode in modes:
```
antalszava, line 842: (Small suggestion): as `mode` is not used here, the name can be removed
```suggestion
            for _ in modes:
```

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
 
         return mean, var
 
+    def number_expectation(self, modes):
```
antalszava, line 818: Might be worth renaming `modes` such it somehow describes the difference between modes of the state and modes for the expectation value (maybe `active_modes` or so). I got a bit confused with grasping the difference between `range(len(modes))` and `num_modes` later in the method.

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
...
+                ps = np.tensordot(values, ps, axes=1)
+            return ps
+
+        # state is a tensor of density matrix elements in the SF convention
+        ps = state.real
+        traced_modes = list(traced_modes)
+        traced_modes.sort(reverse=True)
+        for mode in traced_modes:
+            ps = np.tensordot(np.identity(cutoff), ps, axes=((0, 1), (2 * mode, 2 * mode + 1)))
+        for mode in range(len(modes)):
```
antalszava, line 852: (Small suggestion): as `mode` is not used here, the name can be removed
```suggestion
        for _ in range(len(modes)):
```

strawberryfields/backends/states.py
```diff
@@ -1110,6 +1148,30 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
 
         return mean, var
 
+    def number_expectation(self, modes):
+        """
+        Calculates the expectation value of the product of the number operators of the modes.
```
antalszava, line 1153:
```suggestion
        Calculates the expectation value of the product of the number operators of the modes.

```

strawberryfields/backends/states.py
```diff
@@ -1110,6 +1148,30 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
+        Args:
+            modes (list): list of modes for which one wants the expectation of the product of their number operator.
+        Return:
+            (float): the expectation value.
+        """
+        if len(modes) != len(set(modes)):
+            raise ValueError("there can be no duplicates in modes")
```
antalszava, line 1160: (Small suggestion)
```suggestion
            raise ValueError("There can be no duplicates in the modes specified.")
```

strawberryfields/backends/states.py
```diff
@@ -1110,6 +1148,30 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
+        cov = self._cov
+        if len(modes) == 1:
+            return photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
+
+        if len(modes) == 2:
+            ni = photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
+            nj = photon_number_mean(mu, cov, modes[1], hbar=self._hbar)
+            return photon_number_covar(mu, cov, modes[1], modes[0], hbar=self._hbar) + ni * nj
+
+        raise ValueError("number_expectation only supports one or two modes")
```
antalszava, line 1171:
```suggestion
        raise ValueError("The number_expectation method only supports one or two modes for Gaussian states.")
```

tests/backend/test_states.py
```diff
@@ -282,6 +282,68 @@ def test_squeezed_coherent(self, setup_backend, hbar, batch_size, tol):
 
         assert np.allclose(res.flatten(), res_exact.flatten(), atol=tol, rtol=0)
 
+@pytest.mark.backends("fock", "gaussian")
+class TestNumberExpectation:
+    """Multimode photon-number expectation value tests"""
```
antalszava, line 287: Neat! :1st_place_medal: :blush:

tests/backend/test_states.py
```diff
@@ -282,6 +282,68 @@ def test_squeezed_coherent(self, setup_backend, hbar, batch_size, tol):
...
+        state = backend.state()
+        nbar = np.sinh(r) ** 2
+        assert np.allclose(
+            state.number_expectation([0, 1]), 2 * nbar ** 2 + nbar, atol=tol, rtol=0
+        )
+        assert np.allclose(state.number_expectation([0]), nbar, atol=tol, rtol=0)
+        assert np.allclose(state.number_expectation([1]), nbar, atol=tol, rtol=0)
+
+    def test_number_expectation_repeated_modes(self, setup_backend, tol):
+        """Tests that the correct exception is raised when repeated modes"""
```
antalszava, line 334:
```suggestion
        """Tests that the correct exception is raised for repeated modes"""
```

tests/backend/test_states.py
```diff
@@ -282,6 +282,68 @@ def test_squeezed_coherent(self, setup_backend, hbar, batch_size, tol):
...
+        backend = setup_backend(2)
+        state = backend.state()
+        with pytest.raises(ValueError, match="there can be no duplicates in modes"):
+            state.number_expectation([0, 0])
+
+    @pytest.mark.backends("gaussian")
+    def test_number_expectation_only_two_modes_gaussian(self, setup_backend, tol):
+        """Tests that the correct exception is raised when repeated modes"""
```
antalszava, line 342:
```suggestion
        """Tests that the correct exception is raised when there are more than two modes specified for Gaussian states"""
```

tests/backend/test_states.py
```diff
@@ -282,6 +282,68 @@ def test_squeezed_coherent(self, setup_backend, hbar, batch_size, tol):
...
+        backend.prepare_displaced_squeezed_state(a0, r0, phi0, 0)
+        backend.prepare_displaced_squeezed_state(a1, r1, phi1, 1)
+        state = backend.state()
+        n0 = np.sinh(r0) ** 2 + np.abs(a0) ** 2
+        n1 = np.sinh(r1) ** 2 + np.abs(a1) ** 2
+        assert np.allclose(state.number_expectation([0, 1]), n0 * n1, atol=tol, rtol=0)
+        assert np.allclose(state.number_expectation([0]), n0, atol=tol, rtol=0)
+        assert np.allclose(state.number_expectation([1]), n1, atol=tol, rtol=0)
+
+    def test_number_expectation_two_mode_squeezed(self, setup_backend, tol):
```
antalszava, line 316: It might be worth adding 1-1 test cases for when there are:
* more than 2 modes in the state, `number_expectation` is called with **consecutive** modes
* more than 2 modes in the state, `number_expectation` **non-consecutive** modes

Was wondering if it was useful to also test having more than two modes in the `Fock` case.

strawberryfields/backends/states.py
josh146: Even better, could just be `if self.is_pure:` avoiding the need to assign `pure` as a local variable.

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
 
         return mean, var
 
+    def number_expectation(self, modes):
+        """
+        Calculates the expectation value of the product of the number operators of the modes.
+        Args:
```
josh146, line 821: Perhaps good to add more details here:
```suggestion
        This method computes the analytic expectation value
        :math:`\langle \hat{n}_{i_0} \hat{n}_{i_1}\dots \hat{n}_{i_{N-1}}\rangle`
        for a (sub)set of modes `[i_0, i_1, \dots, i_{N-1}]` of the system.

        Args:
```

strawberryfields/backends/states.py
```diff
@@ -814,6 +815,43 @@ def expand_dims(op, n, modes):
...
+            modes (list): list of modes for which one wants the expectation of the product of their number operator.
+        Return:
+            (float): the expectation value.
+        """
```
josh146, line 825: Note: it could be good to add a small example here. For instance:
```rest
**Example**

Consider the following program:

.. code-block:: python

    prog = sf.Program(3)

    with prog.context as q:
        ops.Sgate(0.5) | q[0]
        ops.Sgate(0.5) | q[1]
        ops.Sgate(0.5) | q[2]
        ops.BSgate(np.pi/3, 0.1) |  (q[0], q[1])
        ops.BSgate(np.pi/3, 0.1) |  (q[1], q[2])

Executing this on the Fock backend,

>>> eng = sf.Engine("fock", backend_options={"cutoff_dim": 10})
>>> state = eng.run(prog).state

we can compute the expectation value :math:`\langle \hat{n}_0\hat{n}_2\rangle`:

>>> state.number_expectation([0, 2])
2.423234
```
```

File: strawberryfields/backends/states.py (hunk 2 above)
Comment (josh146) [reply strawberryfields/backends/states.py:832]: Or, even better: just use `self.ket()` and `self.dm()` directly inside the if statements below, to avoid creating a `state` variable that could represent either.

File: strawberryfields/backends/states.py [hunk 16]
```diff
@@ -1110,6 +1148,30 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
 
         return mean, var
 
+    def number_expectation(self, modes):
+        """
+        Calculates the expectation value of the product of the number operators of the modes.
+        Args:
```
josh146, line 1154:
```suggestion
        This method computes the analytic expectation value
        :math:`\langle \hat{n}_{i_0} \hat{n}_{i_1}\dots \hat{n}_{i_{N-1}}\rangle`
        for a (sub)set of modes `[i_0, i_1, \dots, i_{N-1}]` of the system.
        
        Args:
```

strawberryfields/backends/states.py
```diff
@@ -1110,6 +1148,30 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
+            modes (list): list of modes for which one wants the expectation of the product of their number operator.
+        Return:
```
josh146, line 1156:
```suggestion

        Return:
```

strawberryfields/backends/states.py
josh146: This should also be specified in the docstring, either using a `.. warning::` or in the `Args:` description.

tests/backend/test_states.py
josh146: > Was wondering if it was useful to also test having more than two modes in the Fock case.

Would definitely be useful!

strawberryfields/backends/states.py
nquesada: I don't like the word analytical, if it is not analytical then what is it?

strawberryfields/backends/states.py
nquesada: by parsimony I prefer Josh's answer.

strawberryfields/backends/states.py
nquesada: Why should I use methods instead of grabbing directly the data? Seems like wanting to go through the middle man instead of getting directly to the substance.

strawberryfields/backends/states.py
josh146: Good point, have edited the above to remove it!

tests/backend/test_states.py
nquesada: Don't understand the first two bullet points in Antals suggestion.

nquesada: In particular why is it interesting to test consecutive or non consecutive modes.

strawberryfields/backends/states.py
josh146: `self.ket()` and `self.dm()` simply return the `self.data` directly. However it's

* More future proof (changes to how the data is stored means we just need to update `self.ket()` and `self.dm()`, not `number_expectation` as well)

* Leads to cleaner code

strawberryfields/backends/states.py
nquesada: Done.

nquesada:
Any idea why Code Factor is complaining about things I did not do? @antalszava @josh146 ?

antalszava:
I'd think that the Code Factor things come up since those files were edited. But they seem to have been resolved now. :slightly_smiling_face:

strawberryfields/backends/states.py
```diff
@@ -832,6 +833,74 @@ def expand_dims(op, n, modes):
 
         return mean, var
 
+    def number_expectation(self, modes):
+        r"""
+        Calculates the expectation value of the product of the number operators of the modes.
+
+        This method computes the analytic expectation value
+        :math:`\langle \hat{n}_{i_0} \hat{n}_{i_1}\dots \hat{n}_{i_{N-1}}\rangle`
+        for a (sub)set of modes `[i_0, i_1, \dots, i_{N-1}]` of the system.
```
antalszava, line 842: (Needed for rendering purposes)
```suggestion
        for a (sub)set of modes :math:`[i_0, i_1, \dots, i_{N-1}]` of the system.
```

strawberryfields/backends/states.py
```diff
@@ -832,6 +833,74 @@ def expand_dims(op, n, modes):
...
         return mean, var
 
+    def number_expectation(self, modes):
+        r"""
+        Calculates the expectation value of the product of the number operators of the modes.
+
+        This method computes the analytic expectation value
+        :math:`\langle \hat{n}_{i_0} \hat{n}_{i_1}\dots \hat{n}_{i_{N-1}}\rangle`
+        for a (sub)set of modes :math:`[i_0, i_1, \dots, i_{N-1}]` of the system.
+
```
antalszava, line 843: Now I got myself wondering about the advantages of having that class which is "empty" :blush:  It might make sense to create this facade-like abstract method in `BaseState` and include docstring containing the formula there. The reason is, that in the documentation it would appear for both the `BaseFockState` and for the `BaseGaussianState`. The example that is included could stay specific to either, we could just have a comment that users should be replaced in the other case. This would also just help us to make sure that we have only one docstring that we need to maintain overall. Let me know if I can help with the details on this.

strawberryfields/backends/states.py
```diff
@@ -832,6 +833,74 @@ def expand_dims(op, n, modes):
...
+        >>> state.number_expectation([0, 2])
+        0.07252895071309405
+        """
+        if len(modes) != len(set(modes)):
+            raise ValueError("There can be no duplicates in the modes specified.")
+
+        state = (
+            self._data
+        )  # representation of the quantum state either as tensor of probability amplitudes
```
antalszava, line 880:
```suggestion
        )  # representation of the quantum state as a tensor of probability amplitudes
```

antalszava, line 880: If all's true, this part was replaced by other calls
```suggestion
```

strawberryfields/backends/states.py
nquesada: Yup, maybe you can implement this. Is it appropriate to have it in this PR?

tests/backend/test_states.py
antalszava: With both ones, the first thing that is interesting is that we have **more modes (>2)** in our system than we are requesting the expectation for. This basically tests that tracing out works really the way we expect it to work.

Then, **consecutive** modes are the basic test. We'd expect that the user usually just requests these.

However, there could potentially be something funky going on with  **non-consecutive** modes, that's why I was suggesting to include that as well. It's simply just a potential edge case that users could get into.

strawberryfields/backends/states.py
antalszava: Sure, I think so! I'll get around to it :)

strawberryfields/backends/states.py
```diff
@@ -311,6 +312,56 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
         """
         raise NotImplementedError
 
+    def number_expectation(self, modes):
+        r"""
```
antalszava, line 316: Renders like this:
image

antalszava: For all of `BaseState`, `BaseFockState` and `BaseGaussianState`

nquesada: Super cool! Thanks Antal.

tests/backend/test_states.py
```diff
@@ -282,6 +282,92 @@ def test_squeezed_coherent(self, setup_backend, hbar, batch_size, tol):
 
         assert np.allclose(res.flatten(), res_exact.flatten(), atol=tol, rtol=0)
 
+#@pytest.mark.backends("fock", "gaussian")
```
antalszava, line 285: ?

nquesada: removed

strawberryfields/backends/states.py
josh146: This looks really nice!

strawberryfields/backends/states.py
```diff
@@ -311,6 +312,56 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
+
+        >>> state.number_expectation([0, 2])
+        0.07252895071309405
+
+        Executing the same program on the Gaussian backend,
+
+        >>> eng = sf.Engine("gaussian")
+        >>> state = eng.run(prog).state
+        >>> state.number_expectation([0, 2])
+        0.07566984755267293
```
josh146, line 359: Might be worth mentioning after the example that the slight difference in value is due to the Fock basis truncation when using the Fock backend:
```suggestion
        0.07566984755267293
        
        This slight difference in value compared to the result from the Fock backend above
        is due to the finite Fock basis truncation when using the Fock backend; this can be
        avoided by increasing the value of ``cutoff_dim``.
```

nquesada:
seems like the test failed for the batched mode of TF. I guess the code we wrote cannot work for TF batched. Is there a way to mark this?

josh146:
@nquesada: You can include the `batch_size` fixture:

```python
def test(batch_size):
    if batch_size is not None:
        pytest.skip("Does not support batch mode")
```

nquesada:
Where am I supposed to add this? I don't have a way to test where it sould go locally since likely my laptop will die a slow dead doing batched TF computations.

antalszava:
If you've identified the tests which would need it, you could include `batch_size` in the signature of the test case. What this means, is the following.

Take the `test_number_expectation_displaced_squeezed(self, setup_backend, tol)` as an example.

If you would like this to be skipped for the batched TF case, then this becomes:
`test_number_expectation_displaced_squeezed(self, setup_backend, tol, batch_size)`, and you start the test case with the two lines Josh linked. This will ensure that the test will be skipped if `batch_size` was specified for the test.

nquesada:
By the way, I think the method `state.mean_photon` should be removes since this can now be done by passing a list with a single mode to `number_expectation`. I guess the only problem is that we still don't have a way to do this in batch mode. Although I suspect supporting batching is not too difficult.

josh146:
> By the way, I think the method state.mean_photon should be removed since this can now be done by passing a list with a single mode to number_expectation. 

I recommend keeping `mean_photon` for now, for two reasons:

1. a lot of external code uses it

2. The TF backend overwrites it with a batched implementation in `FockStateTF.mean_photon`.

A partial solution could be:

* `BaseState.mean_photon` has a default implementation that simply calls `number_expectation`.

* Backends are free to override `mean_photon` if they have a better/different implementation.

josh146:
@nquesada, this is now ready to be merged!

Issue #349, opened 2020-04-02
Bug in compile/decompose functions in gaussian backend


AroosaIjaz:
#### Issue description

I called the Monte Carlo event probability function which has some issue with the decomposition functions for gaussian backend. I tried to solve the error by changing the tolerance level in the `rectangular` function in `decompositions.py`. This did not help and on printing the 'diffn'  in the `rectangular` function, values >1 were also seen. The error looks like this:

```
----> 7     fv = np.array(similarity.feature_vector_mc(nx.Graph(mol), events, 2, n_mean=8, samples=1000))
      8     t2 = time.time()
      9     fv = fv*(1/np.linalg.norm(fv))
~/Desktop/xanadu/strawberryfields/strawberryfields/apps/similarity.py in feature_vector_mc(graph, event_photon_numbers, max_count_per_mode, n_mean, samples, loss)
    551     return [
    552         prob_event_mc(graph, photons, max_count_per_mode, n_mean, samples, loss)
--> 553         for photons in event_photon_numbers
    554     ]
~/Desktop/xanadu/strawberryfields/strawberryfields/apps/similarity.py in <listcomp>(.0)
    551     return [
    552         prob_event_mc(graph, photons, max_count_per_mode, n_mean, samples, loss)
--> 553         for photons in event_photon_numbers
    554     ]
~/Desktop/xanadu/strawberryfields/strawberryfields/apps/similarity.py in prob_event_mc(graph, photon_number, max_count_per_mode, n_mean, samples, loss)
    451 
    452     eng = sf.LocalEngine(backend="gaussian")
--> 453     result = eng.run(p)
    454 
    455     prob = 0
~/Desktop/xanadu/strawberryfields/strawberryfields/engine.py in run(self, program, args, compile_options, run_options)
    388         eng_run_options = {key: temp_run_options[key] for key in temp_run_options.keys() & eng_run_keys}
    389 
--> 390         result = super()._run(program, args=args, compile_options=compile_options, **eng_run_options)
    391 
    392         modes = temp_run_options["modes"]
~/Desktop/xanadu/strawberryfields/strawberryfields/engine.py in _run(self, program, args, compile_options, **kwargs)
    275             target = self.backend.circuit_spec
    276             if target is not None and p.target != target:
--> 277                 p = p.compile(target, **compile_options)
    278             p.lock()
    279 
~/Desktop/xanadu/strawberryfields/strawberryfields/program.py in compile(self, target, **kwargs)
    509                 )
[...]

PR #352, opened 2020-04-03
Takagi for real

nquesada:
**Context:**
Make `takagi` better by being clever when the input matrices are real.
Adds an extra `if` branch in the case where the matrices are real.
**Benefits:**
Uses extra information and is faster than the current `takagi` which does not use the extra information of when the input matrix is real.
**Possible Drawbacks:**
None?
**Related GitHub Issues:**
closes https://github.com/XanaduAI/strawberryfields/issues/349

### DIFF
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -56,7 +56,7 @@ def takagi(N, tol=1e-13, rounding=13):
         # If the matrix N is real one can be more clever and use its eigendecomposition
         l, U = np.linalg.eigh(N)
         vals = np.abs(l)  # These are the Takagi eigenvalues
-        phases = np.sqrt(np.complex128(np.sign(l)))
+        phases = np.sqrt(np.complex128([1 if i>0 else -1 for i in l]))
         Uc = U @ np.diag(phases)  # One needs to readjust the phases
         list_vals = [(vals[i], i) for i in range(len(vals))]
         list_vals.sort(reverse=True)
--- a/strawberryfields/decompositions.py
+++ b/strawberryfields/decompositions.py
@@ -61,7 +61,6 @@ def takagi(N, tol=1e-13, rounding=13):
         # And also rearrange the unitary and values so that they are decreasingly ordered
         return np.array(sorted_l), Uc
 
-
     v, l, ws = np.linalg.svd(N)
     w = np.transpose(np.conjugate(ws))
     rl = np.round(l, rounding)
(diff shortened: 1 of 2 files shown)


strawberryfields/decompositions.py
```diff
@@ -47,6 +47,25 @@ def takagi(N, tol=1e-13, rounding=13):
     if np.linalg.norm(N - np.transpose(N)) >= tol:
         raise ValueError("The input matrix is not symmetric")
 
+    N = np.real_if_close(N)
+
+    if np.allclose(N, np.zeros(N.shape)):
```
antalszava, line 52: (Let me know if I'm missing something here!)
```suggestion
    if np.allclose(N, 0):
```

strawberryfields/decompositions.py
```diff
@@ -47,6 +47,25 @@ def takagi(N, tol=1e-13, rounding=13):
...
+        return np.zeros(n), np.eye(n)
+
+    if np.isrealobj(N):
+        # If the matrix N is real one can be more clever and use its eigendecomposition
+        l, U = np.linalg.eigh(N)
+        vals = np.abs(l)  # This are the Takagi eigenvalues
```
antalszava, line 58:
```suggestion
        vals = np.abs(l)  # These are the Takagi eigenvalues
```

strawberryfields/decompositions.py
nquesada: Here `N` is a matrix, you want to compare it with the zero matrix of the same dimension.

nquesada: I think this is correct as is.

antalszava: Right! The reason was, that with the suggestion, numpy's broadcasting will happen anyways, essentially comparing a zero matrix even if writing `0`. My initial thought was that this would perform significantly better.

```python
import numpy as np

shape1 = 1000

a = np.zeros((shape1, shape1))
b = np.zeros((shape1, shape1))
%timeit -n 500 np.allclose(a, b)
6.8 ms ± 400 µs per loop (mean ± std. dev. of 7 runs, 500 loops each)
```
```python
%timeit -n 500 np.allclose(a, 0)
1.88 ms ± 125 µs per loop (mean ± std. dev. of 7 runs, 500 loops each)
```
I had to, however, increasing the size of `shape1` to get a real difference.

antalszava: So overall, I'm fine with whichever way! Having `0` reads slightly better and seems to perform better for large matrices, but I have no strong preference.

trbromley:
:thinking: Some tests are not failing in the apps layer. Need to check this out.

nquesada:
@trbromley can you try with the latest version, the one without sign functions?

trbromley:
All good, tests pass locally with the latest commits. :100:

PR #358, opened 2020-04-08
X series compiler

nquesada:
Implements a Gaussian compiler for Xn


strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Convert the unitary into a sequence of MZgate and Rgate commands on the signal modes
+        U1 = ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=True)._decompose(
+            registers[:half_n_modes]
+        )
+
+        # Convert the unitary into a sequence of MZgate and Rgate commands on the idler modes
+        U2 = ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=True)._decompose(
+            registers[half_n_modes:]
+        )
```
josh146, line 254: There is a potential optimization here: we could only perform the optimization _once_, and then shift the registers on the second decomposition:

```python
U1 = ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=True)._decompose(
    registers[:half_n_modes]
)

import copy
U2 = copy.deepcopy(U1)

for i in range(len(U2)):
    U2[i].reg = [registers[r.ind+half_n_modes] for r in U2[i].reg]
```
or something like that

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,256 @@
...
+            # the programs registers. We must expand the symplectic
+            # matrix to one that acts on all registers.
+            # simply extract the computed symplectic matrix
+            S = expand(seq[0].op.p[0], used_modes, n_modes)
+
+        half_n_modes = n_modes // 2
+
+        # Construct the covariance matrix of the state.
+        # Note that hbar is a global variable that is set by the user
+        cov = (sf.hbar / 2) * S @ S.T
```
josh146, line 189: @nquesada worth double checking this: I wasn't sure if this should be the user-set hbar value or not.

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def __repr__(self):
+        return ", ".join([str(i) for i in self.ranges])
+
+
+class XSpecs(CircuitSpecs):
+    """Circuit specifications for the X class of circuits."""
+
+    short_name = "X"
+    modes = None
```
josh146, line 128: The compiler now works for arbitrary number of modes (as long as `modes` is even)

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,256 @@
...
+        used_modes = [x.ind for x in seq[0].reg]
+
+        # extract the compiled symplectic matrix
+        S = seq[0].op.p[0]
+
+        if len(used_modes) != n_modes:
+            # The symplectic transformation acts on a subset of
+            # the programs registers. We must expand the symplectic
+            # matrix to one that acts on all registers.
+            # simply extract the computed symplectic matrix
```
josh146, line 182:
```suggestion
            # matrix to one that acts on all registers.
```

strawberryfields/circuitspecs/x.py
nquesada: I like this idea very much!

strawberryfields/circuitspecs/x.py
nquesada: You can do any `hbar`, since that `hbar` is local and it is explicitly passe to `Amat`

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Now that the unitary has been validated, perform the Takagi decomposition
+        # to determine the constituent two-mode squeezing and interferometer
+        # parameters.
+        sqs, U = takagi(B01)
+        sqs = np.arctanh(sqs)
+
+        # ensure provided S2gates all have the allowed squeezing values
+        if not all(s in self.allowed_sq_ranges for s in sqs):
+            wrong_sq_values = [s for s in sqs if s not in self.allowed_sq_ranges]
```
josh146, line 231: Minor optimization: only need to perform the range check once.
```suggestion
        wrong_sq_values = [s for s in sqs if s not in self.allowed_sq_ranges]
        if wrong_sq_values :
```

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,260 @@
...
+from strawberryfields.decompositions import takagi
+from strawberryfields.program_utils import CircuitError, Command
+import strawberryfields.ops as ops
+
+from .circuit_specs import CircuitSpecs
+from .gbs import GBSSpecs
+from .gaussian_unitary import GaussianUnitary
+
+
+class Range:
```
josh146, line 30: I think we should move `Range` and `Ranges` to the base `circuitspecs/circuit_specs.py` file, since they are quite general and might be used in other compilers.

Then, we can just do `from .circuit_specs import CircuitSpecs, Range, Ranges` here.

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def __repr__(self):
+        return ", ".join([str(i) for i in self.ranges])
+
+
+class XSpecs(CircuitSpecs):
+    """Circuit specifications for the X class of circuits.
+    An important property of this compilation routines is that it is done at the covariance matrix level.
+    This implies that one should not use to compare the interferometers of a given circuit since they may
+    differ by permutations in the unitary and the squeezing parameters.
```
josh146, line 128:
```suggestion

    An important property of this compilation routine is that it is done at the covariance matrix level.
    This implies that one should not use it to compare the interferometers of a given circuit since they may
    differ by permutations in the unitary and the squeezing parameters.
```

strawberryfields/circuitspecs/x.py
josh146: @nquesada, do you still want to implement this? 🙂

tests/frontend/test_circuitspecs_x.py
```diff
@@ -0,0 +1,481 @@
...
+        U = random_interferometer(4)
+
+        with prog.context as q:
+            ops.S2gate(SQ_AMPLITUDE/2) | (q[0], q[4])
+            ops.S2gate(SQ_AMPLITUDE/2) | (q[0], q[4])
+            ops.Interferometer(U) | (q[0], q[1], q[2], q[3])
+            ops.Interferometer(U) | (q[4], q[5], q[6], q[7])
+            ops.MeasureFock() | q
+
+        res = prog.compile("X")
```
josh146, line 318: This test doesn't seem to be testing anything :/ Should the S2gates not be repeated, and instead it's just testing that half squeezing is not allowed?

tests/frontend/test_circuitspecs_x.py
```diff
@@ -0,0 +1,481 @@
...
+        with prog.context as q:
+            ops.S2gate(SQ_AMPLITUDE, 0) | (q[0], q[4])
+            ops.S2gate(SQ_AMPLITUDE, 0) | (q[1], q[5])
+            ops.S2gate(SQ_AMPLITUDE, 0) | (q[2], q[6])
+            ops.S2gate(SQ_AMPLITUDE, 0) | (q[3], q[7])
+            ops.Interferometer(U) | q
+            ops.MeasureFock() | q
+
+        with pytest.raises(CircuitError, match="The applied unitary cannot mix between the modes"):
+            res = prog.compile("X")
```
josh146, line 481: Nice tests! However, perhaps there should also be some tests for `num_modes != 8` (e.g., `num_modes = 4`), just to make sure it works. There should also likely be a test that an exception is raised if the number of modes is not even

tests/frontend/test_circuitspecs_x.py
nquesada: Quite the opposite, this tests shows the utility of the compiler, in particular by knowing that those two Squeezing gates by half the allowed amplitude can compiled into a single gate with parameter equal to allowed amplitude.

josh146: Oh that's really cool! You could add in an assert to show that the compiled program has squeezing of 1.0

strawberryfields/circuitspecs/x.py
nquesada: Done :)

tests/frontend/test_circuitspecs_x.py
nquesada: All the ones that can maid to work for arbitrary number of modes now take an extra argument for this quantity.

josh146: awesome

strawberryfields/circuitspecs/x.py
```diff
@@ -0,0 +1,169 @@
...
+        ]
+
+        # NOTE: at some point, it might make sense to add a keyword argument to this method,
+        # to allow the user to specify if they want the interferometers decomposed or not.
+
+        # Convert the unitary into a sequence of MZgate and Rgate commands on the signal modes
+        U1 = ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=True)._decompose(
+            registers[:half_n_modes]
+        )
+        U2 = copy.deepcopy(U1)
```
josh146, line 164: Awesome that this works, I never tested it!

tests/frontend/test_circuitspecs_x.py
```diff
@@ -0,0 +1,478 @@
...
+            ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=False) | tuple(q[i] for i in range(num_pairs))
+            ops.Interferometer(U, mesh="rectangular_symmetric", drop_identity=False) | tuple(q[i] for i in range(num_pairs, 2 * num_pairs))
+            ops.MeasureFock() | q
+
+        expected = expected.compile(DummyCircuit())
+        # Note that since DummyCircuit() has a hard coded limit of 8 modes we only check for this number
+        assert program_equivalence(res, expected, atol=tol, compare_params=False)
+
+    @pytest.mark.parametrize("num_pairs", [4, 5, 6, 7])
+    def test_unitaries_do_not_match(self, num_pairs):
```
josh146, line 447: 💯

strawberryfields/circuitspecs/xstrict.py
```diff
@@ -0,0 +1,183 @@
...
+from .gbs import GBSSpecs
+from .gaussian_unitary import GaussianUnitary
+
+
+class XstrictSpecs(CircuitSpecs):
+    """Circuit specifications for the X class of circuits.
+
+    An important property of this compilation routine is that it is done at the covariance matrix level.
+    This implies that one should not use it to compare the interferometers of a given circuit since they may
+    differ by permutations in the unitary and the squeezing parameters.
```
josh146, line 38: I think this docstring needs to be updated 🤔

strawberryfields/circuitspecs/xstrict.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # ensure that all modes are measured
+        if len(seq[-1].reg) != n_modes:
+            raise CircuitError("All modes must be measured.")
+
+        # Check circuit begins with two-mode squeezers
+        # --------------------------------------------
+        A, B, C = group_operations(seq, lambda x: isinstance(x, ops.S2gate))
+        # If there are no two-mode squeezers add squeezers at the beginning with squeezing param equal to zero.
+        if B == []:
```
josh146, line 84:
```suggestion
        if not B:
```

josh146: What happens if there are _some_ TMS gates, but not all? Should we be filling in all missing S2gates with `S2gate(0)`?

strawberryfields/circuitspecs/xstrict.py
```diff
@@ -0,0 +1,183 @@
...
+            initS2 = [
+                Command(ops.S2gate(0, 0), [registers[i], registers[i + half_n_modes]])
+                for i in range(half_n_modes)
+            ]
+            seq = initS2 + seq
+            A, B, C = group_operations(seq, lambda x: isinstance(x, ops.S2gate))
+
+        if A != []:
```
josh146, line 92:
```suggestion
        if A:
```

strawberryfields/circuitspecs/xstrict.py
```diff
@@ -0,0 +1,183 @@
...
+
+        if not regrefs.issubset(allowed_modes):
+            raise CircuitError("S2gates do not appear on the correct modes.")
+
+        # determine which modes do not have input S2gates specified
+        missing = allowed_modes - regrefs
+
+        for i, j in missing:
+            # insert S2gates with 0 squeezing
+            B.insert(0, Command(ops.S2gate(0, 0), [registers[i], registers[j]]))
```
josh146, line 112: This answers my question above!

strawberryfields/circuitspecs/xstrict.py
josh146: Edit: nevermind, I see you are already doing this below :)

strawberryfields/circuitspecs/xstrict.py
```diff
@@ -0,0 +1,183 @@
...
+        sqs = [cmd.op.p[0] for cmd in B]
+
+        # ensure provided S2gates all have the allowed squeezing values
+        if not all(s in self.allowed_sq_ranges for s in sqs):
+            wrong_sq_values = [np.round(s, 4) for s in sqs if s not in self.allowed_sq_ranges]
+            raise CircuitError(
+                "Incorrect squeezing value(s) r={}. Allowed squeezing "
+                "value(s) are {}.".format(wrong_sq_values, self.allowed_sq_ranges)
+            )
+        # This could in principle be changed
```
josh146, line 123: What do you mean by this comment? As in, we should change this to allow for any phases, and just absorb it into the unitary?

strawberryfields/circuitspecs/xstrict.py
```diff
@@ -0,0 +1,183 @@
...
+        phases = [cmd.op.p[1] for cmd in B]
+        if not np.allclose(phases, 0):
+            raise CircuitError(
+                "Incorrect phase value(s) phi={}. Allowed squeezing "
+                "value(s) are 0.0.".format(phases)
+            )
+
+        meas_seq = [C[-1]]
+        seq = GaussianUnitary().compile(C[:-1], registers)
```
josh146, line 132: Nice 🙂

josh146: Can you remove the `[:-1]` simply by adding `MeasureFock` to the primitives of the `GaussianCompiler`?

tests/frontend/test_circuitspecs_xstrict.py
```diff
@@ -0,0 +1,507 @@
...
+
+    def test_wrong_squeezing_phase(self):
+        """Test error is raised when the phase of S2gate is not zero"""
+        prog = sf.Program(2)
+        with prog.context as q:
+            ops.S2gate(SQ_AMPLITUDE, 137) | (q[0], q[1])
+            ops.MeasureFock() | q
+
+        with pytest.raises(CircuitError, match="Incorrect phase value"):
+            res = prog.compile("Xstrict")
```
josh146, line 507:
```suggestion
            res = prog.compile("Xstrict")

```

strawberryfields/circuitspecs/xstrict.py
nquesada: But maybe we can do that later

nquesada:
@josh146 : I like `Xcov` and `Xunitary`. Should I also rename the files?

josh146:
> @josh146 : I like Xcov and Xunitary. Should I also rename the files?
Yep!

PR #359, opened 2020-04-09
View/delete existing configurations

thisac:
**Context:**
There is currently no simple way of viewing what configuration file a user is using, and where this is located on the user's computer. The ability to view active configuration files as well as to be able to delete specific configuration files and to reset all configurations to default, would be very nice; particularly when it comes to debugging and issue-solving for the user.

The following main functions are added to `configuration.py`:

- `active_configurations()`: prints the current configuration files, marking the active configuration path with `(active)`.

- `delete_config(filename="config.toml", directory=None)`: removes a configuration file with name `filename` in `directory`. If no arguments are passed it removes the config file with name `config.toml` in the default directory.

- `reset_config()`: removes all configuration files in directories potentially containing SF configuration files.

Furthermore, the `Connection` class now directly calls `load_config()` in `configuration.py` instead of simply loading a module attribute called `configuration` (instantiated at import), letting a new config file be loaded automatically when creating a `Connection` object after removing or resetting the config file(s).

**Benefits:**
The configuration file(s) can now be deleted and completely reset to default if needed, and it's now possible to view what configurations files are active as well as their location.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -111,6 +111,23 @@ def create_config(authentication_token=None, **kwargs):
     return config
 
 
+def delete_config(filename="config.toml"):
+    """Delete a configuration file
+
+    Keyword Args:
+        filename (str): the configuration file to delete
+    """
+    file_path = get_config_filepath(filename)
+    if file_path is not None:
+        os.remove(file_path)
+
+
+def reset_config():
+    """Delete all active configuration files"""
+    for config in active_configs():
+        delete_config(config)
+
+
 def get_config_filepath(filename="config.toml"):
     """Get the filepath of the first configuration file found from the defined
     configuration directories (if any).
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -314,8 +314,34 @@ def active_configs(filename="config.toml"):
         filename (str): the name of the configuration files to look for
     """
     active_configs_list = get_available_config_paths(filename)
-    print_active_configs(active_configs_list, filename)
-    print_directories_checked()
+
+    # print the active configurations found based on the filename specified
+    if active_configs_list:
+        active = True
+
+        print(
+            "\nThe following Strawberry Fields configuration files were found "
+            'with the name "{}":\n'.format(filename)
+        )
+
+        for config in active_configs_list:
+            if active:
+                config += " (active)"
+                active = False
+
+            print("* " + config)
+    else:
+        print(
+            "\nNo Strawberry Fields configuration files were found with the "
+            'name "{}".\n'.format(filename)
+        )
+
+    # print the directores that are being checked for a configuration file
+    directories = directories_to_check()
+
+    print("\nThe following directories were checked:\n")
+    for directory in directories:
+        print("* " + directory)
 
 
 def get_available_config_paths(filename="config.toml"):
@@ -337,46 +363,6 @@ def get_available_config_paths(filename="config.toml"):
             active_configs_list.append(filepath)
 
     return active_configs_list
-
-
-def print_active_configs(active_configs_list, filename):
-    """Prints the active configurations found based on the filename specified.
-
-    This function relies on the precedence ordering of directories to mark the
-    active configuration.
-
-    Args:
-        active_configs_list (list[str]): the filepaths for the active configurations
-        filename (str): the name of the configuration file
-    """
-    if active_configs_list:
-        active = True
-
-        print(
-            "\nThe following Strawberry Fields configuration files were found "
-            'with the name "{}":\n'.format(filename)
-        )
-
-        for config in active_configs_list:
-            if active:
-                config += " (active)"
-                active = False
-
-            print("* " + config)
-    else:
-        print(
-            "\nNo Strawberry Fields configuration files were found with the "
-            'name "{}".\n'.format(filename)
-        )
-
-
-def print_directories_checked():
-    """Prints the directores that are being checked for a configuration file."""
-    directories = directories_to_check()
-
-    print("\nThe following directories were checked:\n")
-    for directory in directories:
-        print("* " + directory)
 
 
 def store_account(authentication_token, filename="config.toml", location="user_config", **kwargs):
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -303,8 +303,8 @@ def active_configs(filename="config.toml"):
     Args:
         filename (str): the name of the configuration files to look for
     """
-    active_configs = get_active_configs(filename)
-    print_active_configs(active_configs, filename)
+    active_configs_list = get_active_configs(filename)
+    print_active_configs(active_configs_list, filename)
     print_directories_checked()
 
 def get_active_configs(filename="config.toml"):
@@ -315,34 +315,34 @@ def get_active_configs(filename="config.toml"):
     Returns:
         list[str]: the filepaths for the active configurations
     """
-    active_configs = []
+    active_configs_list = []
 
     directories = directories_to_check()
 
     for directory in directories:
         filepath = os.path.join(directory, filename)
         if os.path.exists(filepath):
-            active_configs.append(filepath)
+            active_configs_list.append(filepath)
 
-    return active_configs
+    return active_configs_list
 
-def print_active_configs(active_configs, filename):
+def print_active_configs(active_configs_list, filename):
     """Prints the active configurations found based on the filename specified.
 
     This function relies on the precedence ordering of directories to mark the
     active configuration.
 
     Args:
-        active_configs (list[str]): the filepaths for the active configurations
+        active_configs_list (list[str]): the filepaths for the active configurations
         filename (str): the name of the configuration file
     """
-    if active_configs:
+    if active_configs_list:
         active = True
 
         print("\nThe following Strawberry Fields configuration files were found "
               "with the name \"{}\":\n".format(filename))
 
-        for config in active_configs:
+        for config in active_configs_list:
             if active:
                 config += " (active)"
                 active = False
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -117,7 +117,7 @@ def delete_config(filename="config.toml", directory=None):
     """Delete a configuration file
 
     Keyword Args:
-        filename (str): the configuration file to delete
+        filename (str): the filename of the configuration file to delete
         directory (str): the directory of the configuration file if None, use
             default directory
     """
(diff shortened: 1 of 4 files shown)


strawberryfields/configuration.py
```diff
@@ -111,7 +113,29 @@ def create_config(authentication_token=None, **kwargs):
...
+    else:
+        file_path = os.path.join(directory, filename)
+
+    os.remove(file_path)
+
+
+def reset_config():
+    """Delete all active configuration files"""
+    for config in get_active_configs():
+        delete_config(os.path.basename(config), os.path.dirname(config))
```
josh146, line 135: Maybe best to add a `.. warning::` here in the docstring? this could be destructive if called accidentally.

strawberryfields/configuration.py
```diff
@@ -158,6 +206,31 @@ def load_config_file(filepath):
     return config_from_file
 
 
+def get_api_section_safely(loaded_config, filepath):
```
josh146, line 209: Since the safety should be implicit, can we rename this `get_api_config`?

strawberryfields/configuration.py
```diff
@@ -158,6 +206,31 @@ def load_config_file(filepath):
...
+    """Gets the API section from the loaded configuration.
+
+    Args:
+        loaded_config (dict): the configuration that was loaded from the toml
```
josh146, line 213:
```suggestion
        loaded_config (dict): the configuration that was loaded from the TOML config
```

strawberryfields/configuration.py
```diff
@@ -158,6 +206,31 @@ def load_config_file(filepath):
...
+            configuration
+    """
+    try:
+        return loaded_config["api"]
+    except KeyError:
+        log = create_logger(__name__)
+        log.error(
+            "The configuration from the %s file does not 'contain an \"api\" section.'", filepath
+        )
+        raise ConfigurationError()
```
josh146, line 231: Should this have the `()`?

strawberryfields/configuration.py
josh146: I'm curious the reason behind adding this function 🙂 I would think that a `KeyError` should be a sufficient warning?

strawberryfields/configuration.py
```diff
@@ -222,6 +295,81 @@ def parse_environment_variable(key, value):
...
+
+    Args:
+        filename (str): the name of the configuration files to look for
+    """
+    active_configs_list = get_active_configs(filename)
+    print_active_configs(active_configs_list, filename)
+    print_directories_checked()
+
+
+def get_active_configs(filename="config.toml"):
```
josh146, line 313: How about `get_available_config_paths`? Also, this function is missing an initial docstring summary!

strawberryfields/configuration.py
```diff
@@ -222,6 +295,81 @@ def parse_environment_variable(key, value):
...
+        )
+
+
+def print_directories_checked():
+    """Prints the directores that are being checked for a configuration file."""
+    directories = directories_to_check()
+
+    print("\nThe following directories were checked:\n")
+    for directory in directories:
+        print("* " + directory)
```
josh146, line 370: Nitpick: I see the logic behind splitting out the two print into separate functions (functions should do one thing, easier testing). In this case, however, I feel like the re-use capability for these two subfunctions is pretty low (since they do not return anything, just print), and find it a bit easier to follow the logic if they were part of `active_configs`.

tests/api/test_connection.py
```diff
@@ -217,3 +237,58 @@ def test_ping_failure(self, connection, monkeypatch):
...
+        The logic of the test goes as follows:
+        1. Two temporary paths and files are being created
+        2. The directories to be checked are mocked out to the temporary paths
+        3. A connection object is created, using the configuration from the
+        first config
+        4. The first configuration is deleted, leaving the second config as
+        default
+        5. Another connection object is created, using the configuration from the
+        second config
+        6. Checks for the configuration for each Connection instances
```
josh146, line 257: Really nice test and docstring! Makes it really clear to read and follow

tests/api/test_connection.py
```diff
@@ -217,3 +237,58 @@ def test_ping_failure(self, connection, monkeypatch):
...
+        with open(file1, "w") as f:
+            f.write(TEST_CONFIG_FILE_1)
+
+        with open(file2, "w") as f:
+            f.write(TEST_CONFIG_FILE_2)
+
+        with monkeypatch.context() as m:
+            m.setattr(os, "getcwd", lambda: path1)
+            m.delenv("SF_CONF", raising=False)
+            m.setattr(conf, "user_config_dir", lambda *args: path2)
```
josh146, line 277: 🔥

tests/frontend/test_configuration.py
```diff
@@ -131,6 +131,148 @@ def test_conf_file_loads_well(self, monkeypatch, tmpdir):
...
+        active_configs = ["first_path"]
+        conf.print_active_configs(active_configs, test_file_name)
+
+        captured = capsys.readouterr()
+
+        general_message = "\nThe following Strawberry Fields configuration files were found "\
+                          "with the name \"{}\":\n".format(test_file_name)
+        single_config = "\n* " + active_configs[0] + " (active)\n"
+
+        assert general_message + single_config == captured.out
```
josh146, line 229: 💯

tests/frontend/test_configuration.py
```diff
@@ -250,7 +454,7 @@ def raise_wrapper(ex):
             m.setenv("SF_CONF", "NoConfigFileHere")
             m.setattr(conf, "user_config_dir", lambda *args: "NoConfigFileHere")
 
-            config_filepath = conf.get_config_filepath(filename="config.toml")
+            config_filepath = conf.get_default_config_path(filename="config.toml")
 
         assert config_filepath is None
```
josh146, line 459: Really nice tests, and really nice use of monkeypatching/tmpdir!

tests/api/test_connection.py
```diff
@@ -217,3 +237,58 @@ def test_ping_failure(self, connection, monkeypatch):
...
+
+            a = Connection()
+            assert os.path.exists(file1)
+
+            assert a.token == "DummyToken"
+            assert a.host == "DummyToken"
```
josh146, line 283:
```suggestion
            assert a.host == "DummyHost"
```

tests/frontend/test_configuration.py
```diff
@@ -131,6 +131,148 @@ def test_conf_file_loads_well(self, monkeypatch, tmpdir):
...
+
+        with open(filename, "w") as f:
+            f.write(empty_file)
+
+        with monkeypatch.context() as m:
+            with pytest.raises(conf.ConfigurationError, match=""):
+                m.setattr(os, "getcwd", lambda: tmpdir)
+                configuration = conf.load_config()
+
+        assert "does not contain an \"api\" section" in caplog.text
```
josh146, line 149: I think there's an issue with string escaping in the error message generated. The error message ends with
```python
"file does not \'contain an "api" section.\'\n'"
```

josh146: (note the additional escaped inverted commas)

strawberryfields/configuration.py
```diff
@@ -111,7 +113,29 @@ def create_config(authentication_token=None, **kwargs):
...
+    os.remove(file_path)
+
+
+def reset_config():
+    """Delete all active configuration files"""
+    for config in get_active_configs():
+        delete_config(os.path.basename(config), os.path.dirname(config))
+
+
+def get_default_config_path(filename="config.toml"):
```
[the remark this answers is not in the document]
antalszava: The renaming here was done since a call to `get_config_filepath(filename)` might indicate that this function will find the `filepath` of `filename`.

Cool, `find_config_file` sounds great! :blush:

strawberryfields/configuration.py
antalszava: Just added it as extra care, as I was trying out the new functionalities had an empty configuration file where `KeyError` was not too descriptive.

strawberryfields/configuration.py
antalszava: Hmm perhaps not, works with me without it! :)

strawberryfields/configuration.py
antalszava: Good catch, thanks!

Something along (?):
`Get the paths for the configuration files available in Strawberry Fields.`

strawberryfields/configuration.py
antalszava: It seems something like:
```python
    active_configs = get_active_configs(filename)

    if active_configs:
        active = True

        print("\nThe following Strawberry Fields configuration files were found "
              "with the name \"{}\":\n".format(filename))

        for config in active_configs:
            if active:
                config += " (active)"
                active = False

            print("* " + config)
    else:
        print("\nNo Strawberry Fields configuration files were found with the "
              "name \"{}\".\n".format(filename))

    directories = directories_to_check()

    print("\nThe following directories were checked:\n")
    for directory in directories:
        print("* " + directory)
```

I'd be in for having the transition, as it is still not entirely difficult to parse. @thisac what are your thoughts on this?

thisac: Sure, I think that it makes sense to have them in the same function actually, even though it's sort of the opposite of what was told during the clean-code seminar. I would consider adding some comments though detailing what's being done in the different parts of the function.

strawberryfields/configuration.py
```diff
@@ -111,7 +113,39 @@ def create_config(authentication_token=None, **kwargs):
...
 
-def get_config_filepath(filename="config.toml"):
+def delete_config(filename="config.toml", directory=None):
+    """Delete a configuration file.
+
+    If called with no arguments, the currently active configuration file is deleted.
+
+    Keyword Args:
+        filename (str): the filename of the configuration file to delete
+        directory (str): The directory of the configuration file to delete.
```
antalszava, line 123:
```suggestion
        directory (str): the directory of the configuration file to delete
```

tests/frontend/test_configuration.py
```diff
@@ -159,9 +301,71 @@ def test_config_object_every_keyword_argument(self):
...
+        with monkeypatch.context() as m:
+            m.setattr(conf, "get_available_config_paths", lambda *args: active_configs)
+
+            assert os.path.exists(filename_1)
+            assert os.path.exists(filename_2)
+
+            conf.reset_config()
+
+            assert not os.path.exists(filename_1)
+            assert not os.path.exists(filename_2)
```
antalszava, line 365: Cool tests! :confetti_ball: :1st_place_medal:

PR #360, opened 2020-04-12
[WIP] Add ability to configure logging for debugging

josh146:
**Context:**

When debugging configuration and API access, it's useful to be able to output debugging logging information to a file. This can then be shared when requesting help.

Currently, Strawberry Fields performs logging of internal state and warning, but requires the user have knowledge of the Python logging module in order to modify the logging level/output the log information to a file.

* Adds a new section to the configuration pertaining to logging:

  ```toml
  [logging]
  # default logger output on the terminal
  level = "info"
  # optional logfile output
  logfile = "sf.log"
  ```

  By changing the `level` option, the user can specify the types of log messages that are displayed in the terminal when using the default SF logger. For instance, by setting `level = "debug"`, more detailed information is output during execution.

  By setting the `logfile` option, the user can specify a logfile that is appended to during SF execution. Currently, the logfile contains all logged messages, of level DEBUG and up.

  <details>
    <summary>Click here to see an example logfile created when submitting a remote program.  </summary>

  ```console
  2020-04-12 21:12:26,640 - DEBUG - Configuration file /home/josh/.config/strawberryfields/config.toml loaded
  2020-04-12 21:12:26,641 - DEBUG - Loaded configuration: {'api': {'authentication_token': 'm8Tox***********************************', 'hostname': 'platform.strawberryfields.ai', 'use_ssl': True, 'port': 443}, 'logging': {'level': 'debug'}}
  2020-04-12 21:12:26,641 - DEBUG - Compiling program for target X8_01 with compile options {}
  2020-04-12 21:12:26,964 - DEBUG - Submitting job
  name test
  version 1.0
  target X8_01 (shots=10)

  S2gate(0, 0) | [2, 6]
  S2gate(1, 0.0) | [1, 5]
  S2gate(0.0, 0.0) | [3, 7]
  S2gate(1.0, 0.0) | [0, 4]
  MZgate(5.3074357926500895, 1.5739876247864792) | [0, 1]
  MZgate(4.279290161703878, 4.202979995136468) | [2, 3]
  MZgate(5.824121814032434, 5.033017799971674) | [1, 2]
  MZgate(5.198882953351497, 3.715902833866422) | [0, 1]
  MZgate(4.712954779003035, 3.107023958221405) | [2, 3]
  MZgate(4.83853796968657, 5.988756943006127) | [1, 2]
  Rgate(0.6651640322699326) | 0
  Rgate(0.06106261020227405) | 1
  Rgate(2.94378117796741) | 2
  Rgate(1.820738947340938) | 3
  MZgate(5.3074357926500895, 1.5739876247864792) | [4, 5]
  MZgate(4.279290161703878, 4.202979995136468) | [6, 7]
  MZgate(5.824121814032434, 5.033017799971674) | [5, 6]
  MZgate(5.198882953351497, 3.715902833866422) | [4, 5]
  MZgate(4.712954779003035, 3.107023958221405) | [6, 7]
  MZgate(4.83853796968657, 5.988756943006127) | [5, 6]
  Rgate(0.6651640322699326) | 4
  Rgate(0.06106261020227405) | 5
  Rgate(2.94378117796741) | 6
  Rgate(1.820738947340938) | 7
  MeasureFock() | [0, 1, 2, 3, 4, 5, 6, 7]

  2020-04-12 21:12:27,608 - INFO - Job d101fc18-289f-442b-a899-3d5a6d40fffc was successfully submitted.
  ```
  </details>

  *Note: these options only affect the built-in SF logger handlers. If the user creates their own custom logging handler, this overrides the SF built-in handlers.*

* Adds a DEBUG log message on job submission, logging the submitted blackbird script.

* Adds a DEBUG log message on `RemoteEngine.run_async()`, providing details on program compilation if it occurs.

* Adds an INFO log message if the job is complete after a refresh

* Adds a `sf.configuration.SESSION_CONFIG` global variable that stores the configuration of Strawberry Fields as first import. This is used to configure the logger --- the configuration of the logger should not change after this.

* `load_config()` now logs debugging information, including the loaded configuration. Note that loaded API tokens are automatically masked.

* Minor refactor of the `sf.configuration` module:

  - The logging section has been added to `DEFAULT_CONFIG_SPEC`.

  - Some existing functions had hardcoded dependence on the `"api"` section of the configuration file. This has been removed; now no functions, except for `store_account()`, depend on the structure of the `"api"` section. As a result, `create_config` has been replaced with `_generate_config`, which automatically generates a default configuration object given a configuration spec.

  - `find_config_file` seemed to duplicate the logic in `get_available_config_paths`, so now depends on the former.

  - `get_api_section` was currently used only in `load_config`. Since `load_config` is now agnostic of the configuration structure, this function is redundant.

  - `keep_valid_options` was removed; in testing, it tended to lead to bad user experience to filter out unknown options. Typos (e.g., `prot` instead of `port`) would be pruned from the loaded configuration, making debugging difficult.

* Since there is now more than just an `api` section in the configuration, `store_account` now loads the existing configuration, modifies it as requested, and then re-writes it to disk. This avoids non-API configuration from being overwritten.

**Benefits:**

* Provides an interface for easy logging configuration for the user.

**Possible Drawbacks:**

* Currently, `create_logger` requires a loaded configuration to configure the logger. However, `load_config` creates a logger in order to log details about finding/loading the configuration! This leads to a circular dependency. Currently, this is avoided by passing `logging=False` to `load_config` on first load, to turn off any logging (since the logger has not been created yet). However, this is ugly, and bad form. If something goes wrong with the users logging configuration, it won't be logged!

* We could consider having a function `sf.configure_logger()` in order to configure the logging section of the config file within SF, similar to how `sf.store_account()` configures the API section. E.g.:

  ```python
  >>> sf.configure_logger(level="debug", logfile="sf.log")
   ```

  However, since the logger is configured on import, these settings would not take affect until the next SF import. 

### DIFF
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -154,6 +154,8 @@ class Connection:
 
         circuit = bb.serialize()
 
+        self.log.debug("Submitting job\n%s", circuit)
+
         path = "/jobs"
         response = requests.post(self._url(path), headers=self._headers, json={"circuit": circuit})
         if response.status_code == 201:
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -30,8 +30,27 @@ DEFAULT_CONFIG_SPEC = {
         "hostname": (str, "platform.strawberryfields.ai"),
         "use_ssl": (bool, True),
         "port": (int, 443),
-    }
+    },
+    "logging": {"level": (str, "info"), "logfile": ((str, type(None)), None)},
 }
+"""dict: Nested dictionary representing the allowed configuration
+sections, options, default values, and allowed types for Strawberry
+Fields configurations. For each configuration option key, the
+corresponding value is a length-2 tuple, containing:
+
+* A type or tuple of types, representing the allowed type
+  for that configuration option.
+
+* The default value for that configuration option.
+
+.. note::
+
+    By TOML convention, keys with a default value of ``None``
+    will **not** be present in the generated/loaded configuration
+    file. This is because TOML has no concept of ``NoneType`` or ``Null``,
+    instead, the non-presence of a key indicates that the configuration
+    value is not set.
+"""
 
 
 class ConfigurationError(Exception):
@@ -39,37 +58,66 @@ class ConfigurationError(Exception):
 
 
 def _deep_update(source, overrides):
-    """Update a nested dictionary."""
+    """Recursively update a nested dictionary.
+
+    This function is a generalization of Python's built in
+    ``dict.update`` method, modified to recursively update
+    keys with nested dictionaries.
+    """
     for key, value in overrides.items():
         if isinstance(value, collections.Mapping) and value:
+            # Override value is a non-empty dictionary.
+            # Update the source key with the override dictionary.
             returned = _deep_update(source.get(key, {}), value)
             source[key] = returned
         elif value != {}:
+            # Override value is not an empty dictionary.
             source[key] = overrides[key]
     return source
 
 
 def _generate_config(config_spec, **kwargs):
-    """Generates a configuration, given a configuration specification
-    and optional keyword arguments.
+    """Generates a configuration, given a Strawberry Fields configuration
+    specification.
+
+    See :attr:`~.DEFAULT_CONFIG_SPEC` for an example of a valid configuration
+    specification.
+
+    Optional keyword arguments may be provided to override default values
+    in the cofiguration specification. If the provided override values
+    do not match the expected type defined in the configuration spec,
+    a ``ConfigurationError`` is raised.
+
+    **Example**
+
+    >>> _generate_config(DEFAULT_CONFIG_SPEC, api={"port": 54})
+    {
+        "api": {
+            "authentication_token": "",
+            "hostname": "platform.strawberryfields.ai",
+            "use_ssl": True,
+            "port": 54,
+        }
+    }
 
     Args:
-        config_spec (dict): Nested dictionary representing the
-            configuration specification. Keys in the dictionary
-            represent allowed configuration keys. Corresponding
-            values must be a tuple, with the first element representing
-            the type, and the second representing the default value.
+        config_spec (dict): nested dictionary representing the
+            configuration specification
 
     Keyword Args:
         Provided keyword arguments may overwrite default values of
-        matching (nested) keys.
+        matching keys.
 
     Returns:
         dict: the default configuration defined by the input config spec
+
+    Raises:
+        ConfigurationError: if provided keyword argument overrides do not
+        match the expected type defined in the configuration spec.
     """
     res = {}
     for k, v in config_spec.items():
-        if isinstance(v, tuple) and isinstance(v[0], type):
+        if isinstance(v, tuple):
             # config spec value v represents the allowed type and default value
 
             if k in kwargs:
@@ -82,9 +130,15 @@ def _generate_config(config_spec, **kwargs):
                         )
                     )
 
-                res[k] = kwargs[k]
+                if kwargs[k] is not None:
+                    # Only add the key to the configuration object
+                    # if the provided override is not None.
+                    res[k] = kwargs[k]
             else:
-                res[k] = v[1]
+                if v[1] is not None:
+                    # Only add the key to the configuration object
+                    # if the default value is not None.
+                    res[k] = v[1]
 
         elif isinstance(v, dict):
             # config spec value is a dictionary of more options
@@ -92,7 +146,7 @@ def _generate_config(config_spec, **kwargs):
     return res
 
 
-def load_config(filename="config.toml", **kwargs):
+def load_config(filename="config.toml", verbose=True, **kwargs):
     """Load configuration from keyword arguments, configuration file or
     environment variables.
 
@@ -107,7 +161,8 @@ def load_config(filename="config.toml", **kwargs):
         3. data contained in a configuration file (if exists)
 
     Args:
-        filename (str): the name of the configuration file to look for.
+        filename (str): the name of the configuration file to look for
+        verbose (bool): whether or not to log warnings and errors
 
     Keyword Args:
         Additional configuration options are detailed in
@@ -123,13 +178,15 @@ def load_config(filename="config.toml", **kwargs):
         with open(filepath, "r") as f:
             config = toml.load(f)
 
-        if "api" not in config:
+        if "api" not in config and verbose:
             # Raise a warning if the configuration doesn't contain
             # an API section.
             log = create_logger(__name__)
-            log.warning('The configuration from the %s file does not contain an "api" section.', filepath)
+            log.warning(
+                'The configuration from the %s file does not contain an "api" section.', filepath
+            )
 
-    else:
+    elif verbose:
         config = {}
         log = create_logger(__name__)
         log.warning("No Strawberry Fields configuration file found.")
@@ -138,10 +195,8 @@ def load_config(filename="config.toml", **kwargs):
     update_from_environment_variables(config)
 
     # update the configuration from keyword arguments
-    # NOTE: currently the configuration keyword arguments are specific
-    # only to the API section. Once we have more configuration sections,
-    # they will likely need to be passed via separate keyword arguments.
-    _deep_update(config, {"api": kwargs})
+    for config_section, section_options in kwargs.items():
+        _deep_update(config, {config_section: section_options})
 
     # generate the configuration object by using the defined
     # configuration specification at the top of the file
@@ -436,10 +491,24 @@ def store_account(authentication_token, filename="config.toml", location="user_c
 
     filepath = os.path.join(directory, filename)
 
+    config = {}
+
+    # load the existing config if it already exists
+    if os.path.isfile(filepath):
+        with open(filepath, "r") as f:
+            config = toml.load(f)
+
+    # update the loaded configuration file with the specified
+    # authentication token
+    kwargs.update({"authentication_token": authentication_token})
+
+    # update the loaded configuration with any
+    # provided API options passed as keyword arguments
+    _deep_update(config, {"api": kwargs})
+
     # generate the configuration object by using the defined
     # configuration specification at the top of the file
-    kwargs.update({"authentication_token": authentication_token})
-    config = _generate_config(DEFAULT_CONFIG_SPEC, api=kwargs)
+    config = _generate_config(DEFAULT_CONFIG_SPEC, **config)
 
     with open(filepath, "w") as f:
         toml.dump(config, f)
--- a/strawberryfields/configuration.py
+++ b/strawberryfields/configuration.py
@@ -21,8 +21,6 @@ import os
 import toml
 from appdirs import user_config_dir
 
-from strawberryfields.logger import create_logger
-
 
 DEFAULT_CONFIG_SPEC = {
     "api": {
@@ -146,7 +144,7 @@ def _generate_config(config_spec, **kwargs):
     return res
 
 
-def load_config(filename="config.toml", verbose=True, **kwargs):
+def load_config(filename="config.toml", logging=True, **kwargs):
     """Load configuration from keyword arguments, configuration file or
     environment variables.
 
@@ -162,7 +160,7 @@ def load_config(filename="config.toml", verbose=True, **kwargs):
 
     Args:
         filename (str): the name of the configuration file to look for
-        verbose (bool): whether or not to log warnings and errors
+        logging (bool): whether or not to log details
 
     Keyword Args:
         Additional configuration options are detailed in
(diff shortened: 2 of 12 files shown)


strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+    value is not set.
+"""
 
 
 class ConfigurationError(Exception):
     """Exception used for configuration errors"""
 
 
-def load_config(filename="config.toml", **kwargs):
+def _deep_update(source, overrides):
```
antalszava, line 58: Or comparable such that within the function it's more apparent when read
```suggestion
def _deep_update(old_dict, new_dict):
```

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+    """Recursively update a nested dictionary.
+
```
antalszava, line 60: Here the `Args` part seems to be missing

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+    This function is a generalization of Python's built in
+    ``dict.update`` method, modified to recursively update
+    keys with nested dictionaries.
+    """
+    for key, value in overrides.items():
+        if isinstance(value, collections.Mapping) and value:
+            # Override value is a non-empty dictionary.
+            # Update the source key with the override dictionary.
+            returned = _deep_update(source.get(key, {}), value)
```
antalszava, line 69: Splitting this up might make it easier to grasp the small steps here
```suggestion
            source_key = source.get(key, {})
            returned = _deep_update(source_key, value)
```

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+* The default value for that configuration option.
+
+.. note::
+
+    By TOML convention, keys with a default value of ``None``
+    will **not** be present in the generated/loaded configuration
+    file. This is because TOML has no concept of ``NoneType`` or ``Null``,
+    instead, the non-presence of a key indicates that the configuration
+    value is not set.
+"""
```
antalszava, line 51: This is great! :confetti_ball:

strawberryfields/logger.py
```diff
@@ -108,7 +110,23 @@ def create_logger(name, level=logging.INFO):
     no_handlers = not logging_handler_defined(logger)
 
     if effective_level_inherited and level_not_set and no_handlers:
-        logger.setLevel(level)
-        logger.addHandler(default_handler)
+        # The root logger should pass all log message levels
+        # to the handlers.
+        logger.setLevel(logging.DEBUG)
+        level = level or getattr(logging, SESSION_CONFIG["logging"]["level"].upper())
```
antalszava, line 116: `SESSION_CONFIG["logging"]["level"]` will result in a `KeyError` if the `level` option is not included in the configuration file or there is no configuration file at all

strawberryfields/api/connection.py
```diff
@@ -154,6 +154,8 @@ def create_job(self, target: str, program: Program, run_options: dict = None) ->
 
         circuit = bb.serialize()
 
+        self.log.debug("Submitting job\n%s", circuit)
```
antalszava, line 157: Would be good to keep the level consistent here and on line 164: `self.log.info("Job %s was successfully submitted.", job_id)`. Either both could be `DEBUG` or both could be `INFO`.

One thing I was wondering about: for errors, how should we handle logging? Would it make sense to have the same message when instantiating the error and also log that at the same time?

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+                    # if the provided override is not None.
+                    res[k] = kwargs[k]
+            else:
+                if v[1] is not None:
+                    # Only add the key to the configuration object
+                    # if the default value is not None.
+                    res[k] = v[1]
+
+        elif isinstance(v, dict):
+            # config spec value is a dictionary of more options
```
antalszava, line 142: (Minor suggestion): Just for consistency in names that we are using
```suggestion
            # config spec value is a configuration section
```

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+
+    Returns:
+        dict: the default configuration defined by the input config spec
+
+    Raises:
+        ConfigurationError: if provided keyword argument overrides do not
+        match the expected type defined in the configuration spec.
+    """
+    res = {}
+    for k, v in config_spec.items():
```
antalszava, line 117: Is there any use case for being able to pass `config_spec` to this function? Most of the times one would be using `DEFAULT_CONFIG_SPEC`, but now the function allows any nested dictionary structure. Not sure if the complexity arising from this is worth having.

To me, it seems that the block within the `isinstance(v, tuple)` case could be renamed and organized into a new function. Since we know that the maximum depth of nesting is 2, using recursion here seems like a general solution to a specific problem which can still be avoided to yield a simpler line of logic.

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+
+    Optional keyword arguments may be provided to override default values
+    in the cofiguration specification. If the provided override values
+    do not match the expected type defined in the configuration spec,
+    a ``ConfigurationError`` is raised.
+
+    **Example**
+
+    >>> _generate_config(DEFAULT_CONFIG_SPEC, api={"port": 54})
+    {
```
antalszava, line 92: How come that `logging` is not included when using `DEFAULT_CONFIG_SPEC`?

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
 
 DEFAULT_CONFIG_SPEC = {
     "api": {
         "authentication_token": (str, ""),
         "hostname": (str, "platform.strawberryfields.ai"),
         "use_ssl": (bool, True),
         "port": (int, 443),
-    }
+    },
+    "logging": {"level": (str, "info"), "logfile": ((str, type(None)), None)},
```
antalszava, line 32: How about this one to go with the same as for `authentication_token`?
```suggestion
    "logging": {"level": (str, "info"), "logfile": (str, "")},
```

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+    **Example**
+
+    >>> _generate_config(DEFAULT_CONFIG_SPEC, api={"port": 54})
+    {
+        "api": {
+            "authentication_token": "",
+            "hostname": "platform.strawberryfields.ai",
+            "use_ssl": True,
+            "port": 54,
+        }
```
antalszava, line 98: (Guess `black` will take care of the formatting anyways)
```suggestion
        },
 'logging': {'level': 'info'}
```

antalszava:
Seems like the circular dependency could be resolved by having module imports: https://stackoverflow.com/a/22210807/12899253

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
+            else:
+                if v[1] is not None:
+                    # Only add the key to the configuration object
+                    # if the default value is not None.
+                    res[k] = v[1]
+
+        elif isinstance(v, dict):
+            # config spec value is a dictionary of more options
+            res[k] = _generate_config(v, **kwargs.get(k, {}))
+    return res
```
thisac, line 144: (apart from maybe not needing to use recursion) really nice! :tada:

strawberryfields/configuration.py
```diff
@@ -50,66 +158,65 @@ def load_config(filename="config.toml", **kwargs):
...
-    authentication_token = authentication_token or ""
-    hostname = kwargs.get("hostname", DEFAULT_CONFIG_SPEC["api"]["hostname"][1])
-    use_ssl = kwargs.get("use_ssl", DEFAULT_CONFIG_SPEC["api"]["use_ssl"][1])
-    port = kwargs.get("port", DEFAULT_CONFIG_SPEC["api"]["port"][1])
+    # Log the loaded configuration details, masking out the API key.
+    if logging:
+        config_details = "Loaded configuration: {}".format(config)
+        auth_token = config.get("api", {}).get("authentication_token", "")
+        config_details = config_details.replace(auth_token[5:], "*" * len(auth_token[5:]))
+        log.debug(config_details)
```
thisac, line 218: Nice! :+1:

tests/frontend/test_configuration.py
```diff
@@ -46,7 +47,8 @@
         "hostname": "platform.strawberryfields.ai",
         "use_ssl": True,
         "port": 443,
-    }
+    },
+    'logging': {'level': 'info'}
```
thisac, line 51:
```suggestion
    "logging": {"level": "info"}
```

tests/frontend/test_configuration.py
```diff
@@ -55,7 +57,8 @@
         "hostname": "SomeHost",
         "use_ssl": False,
         "port": 56,
-    }
+    },
+    'logging': {'level': 'info'}
```
thisac, line 61:
```suggestion
    "logging": {"level": "info"}
```

thisac:
@josh146 There seem to be an issue with updating from environment variables. `test_environment_variables_take_precedence_over_conf_file` is failing, and by quickly looking at it it seems like the environment variables never overwrite the config (haven't run it yet though, only briefly looked at the code). :bug:

strawberryfields/api/connection.py
josh146: Good question! Errors should definitely be logged, the generated log files should have a running commentary of what happens during program execution.

One way of doing this could be using [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception). 

This article seems to provide a nice overview of potential logging/exception patterns: https://www.loggly.com/blog/exceptional-logging-of-exceptions-in-python/

josh146: I originally had this as `info`, but I'm not so sure they both belong at the same level. From the Python docs:

Level | Description
-- | --
`DEBUG` | Designates fine-grained informational events that are most useful to debug an application.
`INFO` | Designates informational messages that highlight the progress of the application at coarse-grained level.



* Job submission provides information about the progress of the job submission at a very course-grained level. Seems particularly well suited, given that `info` is the default level, and will be displayed on the output by default.

* Logging the submitted circuit is instead more fine-grained debugging data. In most cases, users won't want this displayed as 'info', since the submitted circuit should be evident from their Python code (or determined via `prog.compile()`). However, this will be useful information when debugging.

strawberryfields/configuration.py
```diff
@@ -15,28 +15,136 @@
...
-from strawberryfields.logger import create_logger
 
 DEFAULT_CONFIG_SPEC = {
     "api": {
         "authentication_token": (str, ""),
         "hostname": (str, "platform.strawberryfields.ai"),
         "use_ssl": (bool, True),
         "port": (int, 443),
-    }
+    },
```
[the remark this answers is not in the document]
josh146: Nothing Pythonic, just a way of having a single 'source of truth' for the allowed configuration at the top of the file 🙂 This means that changes to the configuration/default values/allowed types requires simply modifying this variable --- all subsequent functions are independent of the allowed keys/options/types.

strawberryfields/configuration.py
josh146: Oh, I think I just forgot to update the docstring!

strawberryfields/configuration.py
josh146: Ah, good point 🙂 I tried to keep this as general as possible, because TOML allows for nested sections. For example, in the future, we might have something like this:

```toml
[api]
hostname = "platform.strawberryfields.ai"

  [default]
  token= "personal_token"

  [org1]
  token= "organization_token"
```

When loaded via `toml.load`, it would look like this:

```python
{
  "api": {
    "hostname": "platform.strawberryfields.ai",
    "default": {"token": "personal_token"},
    "org1": {"token": "organization_token"}
  }
}
```

I suppose it's a question of do we want to disallow nested TOML config files (in which case it makes sense to restrict to simply using `DEFAULT_CONFIG_SPEC` here), or if we want to keep this function general.

strawberryfields/configuration.py
josh146: A bit hacky, though 😆

strawberryfields/configuration.py
antalszava: That's a good point, yes :) I'd be leaning towards keeping it simple for now and updating it whenever a specific nested section is introduced for the configuration (but this is subjective point of view). Overall, this would come with the

- Advantage: the configuration related code is tailored onto the current needs, is slightly simpler and easier to be picked up
- Disadvantage: once a nested section is being added, the logic here will have to change to this more general solution

strawberryfields/configuration.py
antalszava: Oh right that makes sense :)

strawberryfields/api/connection.py
antalszava: Sounds good! Happy if these are kept as they are right now :)

PR #362, opened 2020-04-13
Updating changelog with new configuration functions

antalszava:


.github/CHANGELOG.md
```diff
@@ -102,6 +102,10 @@
   For more details, see the
   [configuration documentation](https://strawberryfields.readthedocs.io/en/stable/introduction/configuration.html)
 
+* Further configuration functions for resetting, deleting configurations, as
+  well as displaying available configuration files.
+  [(#359)](https://github.com/XanaduAI/strawberryfields/pull/359)
```
thisac, line 107: I'm not sure if any specific convention is used for changelog messages, but I think `adds` seems to fit better in with the rest (even though a mix of different changelog messages seem to be used). Feel free to ignore this suggestion if you think it looks better the way it is now. :slightly_smiling_face: 

```suggestion
* Adds configuration functions for resetting, deleting configurations, as
  well as displaying available configuration files.
  [(#359)](https://github.com/XanaduAI/strawberryfields/pull/359)
```

antalszava: I was also wondering about this one! @josh146 would you have any suggestions? Also, was not entirely sure if this should be the place to put this (given that it's technically the release candidate description).

josh146: I think this sounds good, I might update this section of the changelog to instead have the final release version 🙂

PR #363, opened 2020-04-14
Hardware-friendly decomposition

shreyapkumar:
**Context:**
This PR attempts to remove needless complications in the code arising from the difference between how theorists define T matrices and the equivalent transformations that the chip implements.

Currently, the unitary is decomposed into Clements T matrices which are later converted to phases on the chip by `rectangular_symmetric` function. Decomposing the unitary directly into Mach Zehnder units would make the code simpler. It would also make it easier to modify the code and decompositions as planned in order to account for imperfect beamsplitters and also for power reduction, etc.

- Adds `MZ` matrices to eventually replace `T` matrices to now represent the units on the chip. 
- Removes need for function `rectangular_symmetric` by modifying `rectangular_phase_end_MZ` to  directly calculate internal and external phases on chip.
- Tests very similar to the ones already present are added.

**Possible Drawbacks:**
- Doesn't match the theoretical conventions in Clements and Xanadu modular decompositions.


shreyapkumar:
Possible reviewers could be @lneuhaus , @nquesada  and/or @ishdhand

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
...
+    BS1 = BS_mat(theta)
+    BS2 = BS_mat(theta)
+    T = BS1 @ R(phi_int) @ BS2 @ R(phi_ext)
+    
+    mat = np.identity(nmax, dtype=np.complex128)
+    mat[m, m] = T[0][0]
+    mat[m, n] = T[0][1]
+    mat[n, m] = T[1][0]
+    mat[n, n] = T[1][1]
+    return mat
```
josh146, line 506: This seems like a duplicate of the `mach_zehnder` function above (line 383); if this (plus `BS` and `R`) replace that function, perhaps best to delete the old `mach_zehnder` function and rename this one to `mach_zehnder` (and make sure that the signature remains the same!)

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
     return new_tlist, new_diags, None
 
 
+def BS_mat(theta):
+    r"""Beam splitter matrix as implemented in hardware
+    """
+    mat = np.array([[np.cos(theta), 1j*np.sin(theta)], [1j*np.sin(theta), np.cos(theta)]])
+    return mat
+
+def R(phi):
```
josh146, line 485: For consistency with `BS_mat` above, best to call this `R_mat`?

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
     return new_tlist, new_diags, None
 
 
+def BS_mat(theta):
+    r"""Beam splitter matrix as implemented in hardware
+    """
```
josh146, line 481: Best to add more detail to this docstring for future developers :) e.g.
```suggestion
    r"""Beamsplitter matrix as implemented in hardware.
    
    Args:
        theta (float): Transmittivity angle :math:`\theta`. The transmission amplitude of
            the beamsplitter is :math:`t = \cos(\theta)`.
    """
```

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
...
+    mat = np.array([[np.cos(theta), 1j*np.sin(theta)], [1j*np.sin(theta), np.cos(theta)]])
+    return mat
+
+def R(phi):
+    r"""Phase as implemented in hardware
+    """
```
josh146, line 487: Same comment here re: expanding out the docstring :)

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
...
+    mat = np.array([[np.exp(1j*phi), 0], [0, 1]])
+    return mat
+
+def MZ(m, n, phi_int, phi_ext, nmax):
+    r"""One unit as implemented on the chip. 
+    An external phase followed by a beamsplitter followed by 
+    the internal phase followed by another beamsplitter
+    """
```
josh146, line 495: Generally, its best for the first sentence in the docstring to summarise the function concisely, e.g.,

```suggestion
    r"""One Mach-Zehnder unit as implemented on hardware. 
   
    This consists of an external phase, followed by a beamsplitter, followed by 
    the internal phase, followed by another beamsplitter.
    """
```

It could be good in this case to include math rendering in the docstring, either inline or display mode:

```rest
Inline math: :math:`x^2`.

Display math:

.. math::

    BS = \begin{bmatrix}0&1\\1&0\end{matrix}
```

Also, don't forget to add details on the arguments and return type!

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
     return new_tlist, new_diags, None
 
 
+def BS_mat(theta):
+    r"""Beam splitter matrix as implemented in hardware
+    """
+    mat = np.array([[np.cos(theta), 1j*np.sin(theta)], [1j*np.sin(theta), np.cos(theta)]])
```
josh146, line 482: Just curious, off the top of my head, the default SF beamsplitter is the following?
```python
np.array([[np.cos(theta), -np.exp(-1j*phi)*np.sin(theta)], [np.exp(1j*phi)*np.sin(theta), np.cos(theta)]])
```
So isn't this the same as the default SF beamsplitter, just with phi->pi/2? Although I might be misremembering the SF default.

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
...
+        # numpy.identity(20) yields an external_phase of exactly 2 * pi
+        new_phi_i %= (2 * np.pi)
+        new_phi_e %= (2 * np.pi)
+        
+        new_i = [i[0], i[1], new_phi_i, new_phi_e, i[4]]
+        new_diags[em], new_diags[en] = np.exp(1j*new_alpha), np.exp(1j*new_beta)
+
+        new_tlist = new_tlist + [new_i]
+
+    return new_tlist, new_diags, None
```
josh146, line 660: Nice! From my understanding, this avoids the added complication from rectangular_symmetric in that it avoids decomposing into SF BSgates, instead decomposing directly into hardware MZgates, and therefore avoiding the process whereby the BSgates and Rgates are modified to transform them back into MZgates?

strawberryfields/decompositions.py
```diff
@@ -476,6 +476,190 @@ def rectangular_symmetric(V, tol=1e-11):
...
+        phi_i = 0
+        phi_e = 0
+    else:
+        r = U[n-1, m]/ U[n, m] 
+        phi_i = 2*np.arctan(np.abs(r))
+        phi_e = - np.angle(r)
+
+    return [n-1, n, phi_i, phi_e, nmax]
+
+def rectangular_MZ(V, tol=1e-11):
```
josh146, line 554: Should this function be named `rectangular_symmetric`, and fully replace the existing `rectangular_symmetric` function? As, as far as I can tell, the functionality here is exactly the same as the existing `rectangular_symmetric`, but more efficient.

By renaming this function to `rectangular_symmetric` (and deleting the old one), the advantage is that the existing rectangular symmetric tests will also run on this function.

josh146:
[CodeFactor](https://www.codefactor.io/repository/github/xanaduai/strawberryfields) found multiple issues last seen at 2fa0b43f300d92bae13c9424d1add9346de1e570:

#### Redefining name &#39;T&#39; from outer scope (line 233)
[strawberryfields\decompositions.py:499
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L499)
#### No space allowed before :
    if U[n, m] == 0 :
                    ^
[strawberryfields\decompositions.py:541
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L541)
#### Trailing whitespace
[strawberryfields\decompositions.py:493
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L493)[strawberryfields\decompositions.py:654
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L654)[strawberryfields\decompositions.py:661
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L661)[strawberryfields\decompositions.py:492
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L492)[strawberryfields\decompositions.py:548
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L548)[strawberryfields\decompositions.py:500
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L500)
#### No space allowed before :
    if U[m, n] == 0 :
                    ^
[strawberryfields\decompositions.py:520
](https://github.com/XanaduAI/strawberryfields/blob/2fa0b43f300d92bae13c9424d1add9346de1e570/strawberryfields/decompositions.py#L520)

strawberryfields/decompositions.py
shreyapkumar: This is spot on except that it avoids decomposing into the `T` matrices from the Clements paper (which are neither SF BS gates nor hardware MZgates).

strawberryfields/decompositions.py
```diff
@@ -400,6 +400,114 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
     return BS @ Rinternal @ BS @ Rexternal
 
 
+def mach_zehnder_inv(m, n, phi_int, phi_ext, nmax):
+    r"""The inverse of the mach_zehnder matrix."""
+    return np.transpose(np.conj(mach_zehnder(m, n, phi_int, phi_ext, nmax)))
```
nquesada, line 405:
```suggestion
    return mach_zehnder(m, n, phi_int, phi_ext, nmax).T.conj()
```

strawberryfields/decompositions.py
```diff
@@ -400,6 +400,114 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+    if nmax != mmax:
+        raise ValueError("U must be a square matrix")
+
+    if U[m, n] == 0:
+        phi_i = np.pi
+        phi_e = 0
+    elif U[m, n + 1] == 0:
+        phi_i = 0
+        phi_e = 0
+    else:
```
nquesada, line 421: Are these the checks that take care of edge cases like the identity?

strawberryfields/decompositions.py
```diff
@@ -400,6 +400,114 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+            where:
+
+            * ``tilist``: list containing ``[n,m,phi_int,phi_ext,n_size]`` of the MZi unitaries needed
+            * ``tlist``: list containing ``[n,m,phi_int,phi_ext,n_size]`` of the MZ unitaries needed
+            * ``localV``: Diagonal unitary sitting sandwiched by MZi's and the MZ's
+    """
+    localV = V
+    (nsize, _) = localV.shape
+
+    diffn = np.linalg.norm(V @ V.conj().T - np.identity(nsize))
```
nquesada, line 480: Wonder if it would be good to compare closeness to the identity using `np.allclose`

tests/frontend/test_decompositions.py
```diff
@@ -288,6 +288,59 @@ def test_random_unitary_phase_end(self, tol):
...
+
+    def test_random_unitary_MZ(self, tol):
+        """This test checks the rectangular decomposition for a random unitary.
+
+        A random unitary is drawn from the Haar measure, then is decomposed via
+        the rectangular decomposition of Clements et al., and the resulting
+        beamsplitters are multiplied together. Test passes if the product
+        matches the drawn unitary.
+        """
+        # TODO: this test currently uses the T and Ti functions used to compute
```
nquesada, line 325: Is this meant to fixed in this PR?

strawberryfields/decompositions.py
shreyapkumar: Yes, exactly! One if for identity-like cases, another one for divide by zero.

strawberryfields/decompositions.py
shreyapkumar: I think the test here is a bit more strict than using `np.allclose`. 
For example `V = U + tol * np.identity(n)`fails the above test, but passes the `np.allclose`

shreyapkumar: That said, good to move this and other decompositions to `np.allclose` as that is easier to read and might be faster? Do you (or @josh146) have any idea about any potential downsides?

co9olguy: Good to use `allclose` if possible to keep things consistent (all the tests use the `np.allclose(result, expected, atol=tol, rtol=0)` pattern

tests/frontend/test_decompositions.py
shreyapkumar: Oops, this was copied from other tests. I don't really know what this todo means, so removed this from the docstrings of `test_random_unitary_MZ` and `test_identity_MZ`

shreyapkumar:
Documentation and formatting changes are fixed now. Comments about this or anything else are welcome!

strawberryfields/decompositions.py
```diff
@@ -387,6 +384,23 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
     m, a symmetric beamsplitter combining modes m and n, an internal phase
     shifter on mode m, and another symmetric beamsplitter combining modes m
     and n.
+
+    The resulting matrix is
+    .. math::
```
josh146, line 389:
```suggestion

    .. math::
```

strawberryfields/decompositions.py
josh146: @shreyapkumar: maybe worth adding short comments to the if-else statement, specifying which branch is for the identity-like case, which one is for the divide by zero case :)

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+    if U[n, m] == 0:
+        phi_i = np.pi
+        phi_e = 0
+    elif U[n - 1, m] == 0:
+        phi_i = 0
+        phi_e = 0
+    else:
+        r = U[n - 1, m] / U[n, m]
+        phi_i = 2 * np.arctan(np.abs(r))
+        phi_e = -np.angle(r)
```
josh146, line 479: Same here re: adding brief comments explaining each branch. Just a simple 1-liner is fine, just helps future developers if they need to track down bugs etc :)

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
-    shifts at the end of the interferometer are added to the ones from the
-    diagonal unitary at the end of the interferometer obtained from :func:`~.rectangular_phase_end`.
+
+def rectangular_MZ(V, tol=1e-11):
+    r"""Rectangular decomposition of a unitary matrix, with local
+    phase shifts applied between two interferometers.
+
+    Is similar to :func:`rectangular` except that it uses mach_zehner matrices to null elements of V
+    using the :func:`null_MZ` and :func:`null_MZi` instead of :func:`T` matrices and corresponding :func:`nullT`
+    and :func:`nullTi` functions.
```
josh146, line 490:
```suggestion
    Is similar to :func:`~.rectangular` except that it uses Mach Zehnder matrices to null elements of V
    using the :func:`~.null_MZ` and :func:`~.null_MZi` instead of :func:`~.T` matrices and corresponding :func:`~.nullT`
    and :func:`~.nullTi` functions.
```

Unless otherwise specified, Sphinx will assume the function/class name is an absolute path. If no function/class exists at that absolute path, sphinx will just silently fail to resolve the link.

We can use two syntactical tricks here to avoid this:

* Pattern matching (`.`): resolve to a class or function at any path matching a particular name.

  Note: in the case where there exists two objects with the same name in different places, parts of the path can be specified to avoid ambiguity. For example:

  ```rest
  :func:`.utils.catstate` and :func:`.ops.catstate`.
  ```

* When using `.` to pattern match, Sphinx will display the absolute path in the rendered link. For example, the code block above will render as

  `strawberryfields.utils.catstate` and `strawberryfields.ops.catstate`.

  To avoid this, we can use the prefix `~` to indicate to Sphinx to only use the last component of the target in the link text:

  ```rest
  :func:`~.utils.catstate` and :func:`~.ops.catstate`.
  ```

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+        V (array[complex]): unitary matrix of size n_size
+        tol (float): the tolerance used when checking if the matrix is unitary
+
+    Returns:
+        tuple[array]: tuple of the form ``(tilist,np.diag(localV),tlist)``
+            where:
+
+            * ``tilist``: list containing ``[n,m,phi_int,phi_ext,n_size]`` of the ``mach_zehnder_inv`` unitaries needed
+            * ``tlist``: list containing ``[n,m,phi_int,phi_ext,n_size]`` of the ``mach_zehnder`` unitaries needed
+            * ``localV``: Diagonal unitary sitting sandwiched by ``mach_zehnder_inv``'s and the ``mach_zehnder``'s
```
josh146, line 502:
```suggestion
            * ``localV``: diagonal unitary sandwiched between the ``mach_zehnder_inv`` unitaries and the ``mach_zehnder`` unitaries
```

A quirk of Sphinx/ReST markup is that you must always have spaces after backticks 🙁

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Decomposition of a unitary into an array of symmetric beamsplitters.
+
+    This decomposition starts with the output from :func:`rectangular_MZ`
+    and performs the equivalent of :func:`rectangular_phase_end` by placing all the
+    local phase shifts after the interferometers.
+
+    If the mach_zehnder's are represented as M and the local phase shifts as D, the new
+    parameters to shift the local phases to the end are calculated such that
```
josh146, line 545: Awesome 🔥

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+                localV = mach_zehnder(*tlist[-1]) @ localV
+
+    return tilist, np.diag(localV), tlist
+
+
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Decomposition of a unitary into an array of symmetric beamsplitters.
+
+    This decomposition starts with the output from :func:`rectangular_MZ`
+    and performs the equivalent of :func:`rectangular_phase_end` by placing all the
```
josh146, line 541:
```suggestion
    This decomposition starts with the output from :func:`~.rectangular_MZ`
    and performs the equivalent of :func:`~.rectangular_phase_end` by placing all the
```

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+                # numpy.identity(20) yields an external_phase of exactly 2 * pi
+                tilist[-1][2] %= 2 * np.pi
+                tilist[-1][3] %= 2 * np.pi
+                localV = localV @ mach_zehnder_inv(*tilist[-1])
+        else:
+            for j in range(nsize - 1 - i):
+                tlist.append(nullMZ(i + j + 1, j, localV))
+                tlist[-1][2] %= 2 * np.pi
+                tlist[-1][3] %= 2 * np.pi
+                # repeat modulo operations , otherwise the input unitary
```
josh146, line 528:
```suggestion
                # repeat modulo operations, otherwise the input unitary
```

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+
+    tilist = []
+    tlist = []
+    for k, i in enumerate(range(nsize - 2, -1, -1)):
+        if k % 2 == 0:
+            for j in reversed(range(nsize - 1 - i)):
+                tilist.append(nullMZi(i + j + 1, j, localV))
+                tilist[-1][2] %= 2 * np.pi
+                tilist[-1][3] %= 2 * np.pi
+                # repeat modulo operations , otherwise the input unitary
```
josh146, line 518:
```suggestion
                # repeat modulo operations, otherwise the input unitary
```

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+
+
+def rectangular_symmetric(V, tol=1e-11):
+    r"""Decomposition of a unitary into an array of symmetric beamsplitters.
+
+    This decomposition starts with the output from :func:`rectangular_MZ`
+    and performs the equivalent of :func:`rectangular_phase_end` by placing all the
+    local phase shifts after the interferometers.
+
+    If the mach_zehnder's are represented as M and the local phase shifts as D, the new
```
josh146, line 544:
```suggestion
    If the Mach-Zehnder unitaries are represented as M and the local phase shifts as D, the new
```

strawberryfields/decompositions.py
```diff
@@ -400,44 +414,139 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
     return BS @ Rinternal @ BS @ Rexternal
 
 
-def rectangular_symmetric(V, tol=1e-11):
-    r"""Decomposition of a unitary into an array of symmetric beamsplitters.
+def mach_zehnder_inv(m, n, phi_int, phi_ext, nmax):
+    r"""The inverse of the mach_zehnder matrix."""
```
josh146, line 418:
```suggestion
    r"""The inverse of the Mach-Zehnder unitary matrix.
    
    See :func:`~.mach_zehnder` for more details on the Mach-Zehnder unitary.
    """
```

strawberryfields/decompositions.py
```diff
@@ -387,6 +384,23 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
...
+    Args:
+        m (int): mode number on which the phase shifters act
+        n (int): mode number which is combined with mode m by the beamsplitters
+        internal_phase (float): phase in between the symmetric beamsplitters
+        external_phase (float): phase acting before the first beamsplitter
+        nmax (int): maximum number of modes in the circuit
+
+    Returns:
+        array: unitary matrix of the effective transformation the series of phaseshifters
+            and beamspliiters.
```
josh146, line 403: Note: since `Returns:` can only have one item, line continuations are not indented.

```suggestion
        array: unitary matrix of the effective transformation the series of phaseshifters
        and beamsplitters.
```

shreyapkumar:
Thanks so much, @josh146 for the very helpful comments! I learnt a lot :)

strawberryfields/decompositions.py
shreyapkumar: Sounds good!

tests/frontend/test_circuitspecs_X8.py
```diff
@@ -375,10 +375,10 @@ def test_no_unitary(self, tol):
             ops.MZgate(np.pi, 0) | (q[0], q[1])
             ops.MZgate(np.pi, 0) | (q[2], q[3])
             ops.MZgate(np.pi, np.pi) | (q[1], q[2])
-            ops.MZgate(np.pi, np.pi) | (q[0], q[1])
+            ops.MZgate(np.pi, 0) | (q[0], q[1])
```
shreyapkumar, line 378: Oops, looks like this file's changes came in from some merge from upstream. Not sure what to do about it!

josh146: Oh, that was me when I merged master into this branch! I noticed that this was the test that was failing

strawberryfields/decompositions.py
```diff
@@ -400,44 +419,146 @@ def mach_zehnder(m, n, internal_phase, external_phase, nmax):
     return BS @ Rinternal @ BS @ Rexternal
 
 
-def rectangular_symmetric(V, tol=1e-11):
-    r"""Decomposition of a unitary into an array of symmetric beamsplitters.
+def mach_zehnder_inv(m, n, phi_int, phi_ext, nmax):
+    r"""The inverse of the Mach-Zehnder unitary matrix.
+
```
josh146, line 424:
```suggestion

```

josh146:
@shreyapkumar, is this ready to be merged in?

shreyapkumar:
Yes, nothing more to add from my side!

PR #364, opened 2020-04-14
Fixes bug where negative probabilities can be calculated during Homodyne measurement in the Fock backend

### LINKED ISSUE #354: `measure_homodyne` sometimes constructs a probability vector with p<0, p>1 in the Fock backend, resulting in an exception.
```
pvals < 0, pvals > 1 or pvals contains NaNs
```

When I try to run the circuit (mentioned below) I occasionally end up with the error mentioned above.

```python
prog = sf.Program(2)

with prog.context as q:
    Catstate(complex(a,0)) | q[0]
    Catstate(complex(0,a)) | q[1]
    
    BSgate() | (q[0], q[1])
    MeasureX | q[0]
    MeasureX | q[1]

result = eng.run(prog)
 x, y = result.samples
```

 *Expected behaviour: Ideally when it works, it is supposed to give me the two quadrature values.

* *Reproduces how often:* 30-40% of the time.

* *System information:
```
Copyright 2018-2019 Xanadu Quantum Technologies Inc.

Python version:            3.7.4
Platform info:             Darwin-19.4.0-x86_64-i386-64bit
Installation path:         /Users/biveenshajilal/opt/anaconda3/lib/python3.7/site-packages/strawberryfields
Strawberry Fields version: 0.12.1
Numpy version:             1.18.2
Scipy version:             1.3.1
The Walrus version:           0.12.0
Blackbird version:         0.2.3
TensorFlow version:        None
```
```python
ValueError                                Traceback (most recent call last)
<ipython-input-27-d8f9402683f5> in <module>
      2 
      3 for x in pbar(flag):
----> 4     dummy = xaxb()
      5 
      6     x_vals[i]=dummy[0]

<ipython-input-22-64fa305c29f6> in xaxb()
     12         MeasureX | q[1]
     13 
---> 14     result = eng.run(prog)
     15 
     16     x, y = result.samples

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/engine.py in run(self, program, args, compile_options, run_options)
    479         eng_run_options = {key: temp_run_options[key] for key in temp_run_options.keys() & eng_run_keys}
    480 
--> 481         result = super()._run(program, args=args, compile_options=compile_options, **eng_run_options)
    482 
    483         modes = temp_run_options["modes"]

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/engine.py in _run(self, program, args, compile_options, **kwargs)
    347             p.lock()
    348 
--> 349             self._run_program(p, **kwargs)
    350             self.run_progs.append(p)
    351             # store the latest measurement results

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/engine.py in _run_program(self, prog, **kwargs)
    421             try:
    422                 # try to apply it to the backend
--> 423                 cmd.op.apply(cmd.reg, self.backend, **kwargs)  # NOTE we could also handle storing measured vals here
    424                 applied.append(cmd)
    425             except NotApplicableError:

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/ops.py in apply(self, reg, backend, **kwargs)
    285                 Only applies to Measurements.
    286         """
--> 287         values = super().apply(reg, backend, **kwargs)
    288         # convert the returned values into an iterable with the measured modes indexed along
    289         # the first axis and shots along second axis (if larger than 1), so that we can assign

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/ops.py in apply(self, reg, backend, **kwargs)
    217         temp = [rr.ind for rr in reg]
    218         # call the child class specialized _apply method
--> 219         return self._apply(temp, backend, **kwargs)
    220 
    221 

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/ops.py in _apply(self, reg, backend, shots, **kwargs)
    806             select = select / s
    807 
--> 808         return s * backend.measure_homodyne(p[0], *reg, shots=shots, select=select, **kwargs)
    809 
    810     def __str__(self):

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/backends/fockbackend/backend.py in measure_homodyne(self, phi, mode, shots, select, **kwargs)
    187             raise NotImplementedError("fock backend currently does not support "
    188                                       "shots != 1 for homodyne measurement")
--> 189         return self.circuit.measure_homodyne(phi, self._remap_modes(mode), select=select, **kwargs)
    190 
    191     def loss(self, T, mode):

~/opt/anaconda3/lib/python3.7/site-packages/strawberryfields/backends/fockbackend/circuit.py in measure_homodyne(self, phi, mode, select, **kwargs)
    498             probs = rho_dist.flatten().real
    499             probs /= np.sum(probs)
--> 500             sample_hist = np.random.multinomial(1, probs)
    501             sample_idx = list(sample_hist).index(1)
    502             homodyne_sample = q_tensor[sample_idx]

mtrand.pyx in numpy.random.mtrand.RandomState.multinomial()

_common.pyx in numpy.random._common.check_array_constraint()

ValueError: pvals < 0, pvals > 1 or pvals contains NaNs
```
#### Additional information

To reproduce this, I didn't have to particularly do anything. I just have to repeat the process multiple times. I tried it on multiple devices, both on Linux and macOS. I get the same error.

josh146:
**Context:**

To calculate homodyne measurements in the Fock backend, a Hermite polynomial probability distribution is computed. However, due to floating point precision error, values in the probability distribution may have a small negative value (e.g., `-2e-20`), causing issues when `np.random.multinomial` is subsequently called.

Small negative values are set to zero in the probability distribution using NumPy fancy indexing.

**Benefits:** Fixes #354 

**Possible Drawbacks:** n/a

**Related GitHub Issues:** #354

### DIFF
--- a/strawberryfields/backends/fockbackend/circuit.py
+++ b/strawberryfields/backends/fockbackend/circuit.py
@@ -764,7 +764,7 @@ class Circuit:
             probs = rho_dist.flatten().real
             probs /= np.sum(probs)
 
-            # Due to numerical error, values in the calculated probability distribution
+            # Due to floating point precision error, values in the calculated probability distribution
             # may have a very small negative value of -epsilon. The following sets
             # these small negative values to 0.
             probs[np.abs(probs) < 1e-16] = 0
--- a/strawberryfields/backends/fockbackend/circuit.py
+++ b/strawberryfields/backends/fockbackend/circuit.py
@@ -767,7 +767,7 @@ class Circuit:
             # Due to floating point precision error, values in the calculated probability distribution
             # may have a very small negative value of -epsilon. The following sets
             # these small negative values to 0.
-            probs[np.abs(probs) < 1e-16] = 0
+            probs[np.abs(probs) < 1e-10] = 0
 
             sample_hist = np.random.multinomial(1, probs)
             sample_idx = list(sample_hist).index(1)
--- a/strawberryfields/backends/fockbackend/circuit.py
+++ b/strawberryfields/backends/fockbackend/circuit.py
@@ -763,6 +763,12 @@ class Circuit:
             # Numpy also does not use the log probabilities
             probs = rho_dist.flatten().real
             probs /= np.sum(probs)
+
+            # Due to numerical error, values in the calculated probability distribution
+            # may have a very small negative value of -epsilon. The following sets
+            # these small negative values to 0.
+            probs[np.abs(probs) < 1e-16] = 0
+
             sample_hist = np.random.multinomial(1, probs)
             sample_idx = list(sample_hist).index(1)
             homodyne_sample = q_tensor[sample_idx]
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -185,6 +185,11 @@
   from the Blackbird script or `Program` if not explicitly specified.
   [(#327)](https://github.com/XanaduAI/strawberryfields/pull/327)
 
+* Fixed a bug in homodyne measurements in the Fock backend, where computed
+  probability values could occasionally include small negative values
+  due to floating point precision error.
+  [(#364)](https://github.com/XanaduAI/strawberryfields/pull/364)
+
 * Improves the Takagi decomposition, by making explicit use of the eigendecomposition of real symmetric matrices. [(#352)](https://github.com/XanaduAI/strawberryfields/pull/352)
 
 <h3>Contributors</h3>


strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -763,6 +763,12 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
             # Numpy also does not use the log probabilities
             probs = rho_dist.flatten().real
             probs /= np.sum(probs)
+
+            # Due to floating point precision error, values in the calculated probability distribution
+            # may have a very small negative value of -epsilon. The following sets
+            # these small negative values to 0.
+            probs[np.abs(probs) < 1e-16] = 0
```
thisac, line 770: If e.g. a value in `probs` is `-1e-14` then this won't set it to zero, since you're just comparing the size of absolute values. Is this actually what you want? I would think that something like `probs[probs < 0] = 0` instead, but maybe I'm missing something.

josh146: Ah, this is because I don't want to set all negative values to 0, I only want to set very small values (negative or positive) to zero.

This fix above will catch issues caused by floating point imprecision (where we expect the negative value to be in the range `-2*epsilon < p < 0`, where `epsilon ~ np.finfo(float).eps`), but if there is ever the case where `p = -0.2`, for example, than this indicates a systematic bug and I would prefer an exception is still raised 🙂

thisac: Ok, I see. That makes sense. Thank you! :smile:

nquesada: This is more in line with that @thisac is suggesting: https://github.com/XanaduAI/thewalrus/blob/078837ebe86a4d39edb6d06b06b67ed6c8a1b606/thewalrus/quantum.py#L1091

I am happy with @josh146 solution, although am not sure if `1e-16` is the correct threshold value.

antalszava: The syntax here is really neat! :fire: 

`1e-16` seems pretty small indeed. Not sure about precision errors arising in these cases, but e.g. in PennyLane the tolerance being used is mostly `1e-10`.

josh146: Oops, I might have made a typo here. I noticed when reproducing the bug that the negative probability values were always `2.220446049250313e-16`, which is machine epsilon. I can increase this to 1e-10, but it seems a bit arbitrary.

antalszava: That makes sense! Happy if the threshold is adjusted that way then :)

josh146:
```suggestion
            probs[np.abs(probs) < 1e-10] = 0
```

PR #365, opened 2020-04-14
Bump version number for 0.13.0 release

josh146:

* Bump version number to v0.13.0 for release

* Fix documentation build errors, by skipping functions that are being imported in multiple places

* Fix small backdoor for older versions of NumPy, by enforcing `allow_pickle=False` when loading sampled data.

**Benefits:** n/a
 
**Possible Drawbacks:** n/a

### DIFF
--- a/doc/code/sf_configuration.rst
+++ b/doc/code/sf_configuration.rst
@@ -36,6 +36,12 @@ and has the following format:
 Configuration options
 ---------------------
 
+``[api]``
+^^^^^^^^^
+
+Settings for the Xanadu cloud platform.
+
+
 **authentication_token (str)** (*required*)
     API token for authentication to the Xanadu cloud platform. This is required
     for submitting remote jobs using :class:`~.RemoteEngine`.
@@ -67,5 +73,5 @@ Functions
 
 .. automodapi:: strawberryfields.configuration
     :no-heading:
-    :skip: user_config_dir
+    :skip: user_config_dir, store_account, active_configs, reset_config, create_logger
     :no-inheritance-diagram:
--- a/strawberryfields/__init__.py
+++ b/strawberryfields/__init__.py
@@ -24,7 +24,7 @@ and backend components (all found within the :mod:`strawberryfields.backends` su
 from . import apps
 from ._version import __version__
 from .cli import ping
-from .configuration import store_account, active_configs, reset_config
+from .configuration import store_account, active_configs, reset_config, delete_config
 from .engine import Engine, LocalEngine, RemoteEngine
 from .io import load, save
 from .parameters import par_funcs as math
@@ -43,6 +43,7 @@ __all__ = [
     "store_account",
     "active_configs",
     "reset_config",
+    "delete_config",
 ]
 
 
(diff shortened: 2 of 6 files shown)


strawberryfields/__init__.py
```diff
@@ -24,13 +24,26 @@
 from . import apps
 from ._version import __version__
 from .cli import ping
-from .configuration import store_account
+from .configuration import store_account, active_configs, reset_config
```
antalszava, line 27: Oh, thanks for this!!

antalszava, line 27:
```suggestion
from .configuration import store_account, active_configs, reset_config, delete_config
```

strawberryfields/__init__.py
```diff
@@ -24,13 +24,26 @@
...
+    "Program",
+    "version",
+    "save",
+    "load",
+    "about",
+    "cite",
+    "ping",
+    "store_account",
+    "active_configs",
+    "reset_config",
```
antalszava, line 45:
```suggestion
    "reset_config",
    "delete_config",
```

strawberryfields/api/connection.py
```diff
@@ -227,7 +227,7 @@ def get_job_result(self, job_id: str) -> Result:
             with io.BytesIO() as buf:
                 buf.write(response.content)
                 buf.seek(0)
-                samples = np.load(buf)
+                samples = np.load(buf, allow_pickle=False)
```
antalszava, line 230: This is great! What sort of error did this raise before?

doc/code/sf_configuration.rst
```diff
@@ -36,6 +36,12 @@ and has the following format:
 Configuration options
 ---------------------
 
+``[api]``
+^^^^^^^^^
+
+Settings for the Xanadu cloud platform.
+
+
```
antalszava, line 44: This renders like this right now:
image

(Minor suggestion): How about indenting the options? (Such that "Settings for the..." is as is now, and "**authentication_token**" is the first to be indented one level)

strawberryfields/api/connection.py
thisac: It seems like `allow_pickle` is `False` by default. Have you added this as a precaution/to be explicit regarding not accepting anything other than not loading object arrays, or is there another reason for this?

doc/code/sf_configuration.rst
josh146: Good question, I'm not sure if that is possible? In ReST, indenting blocks of text is how you markup block quotations, so it would affect the formatting

strawberryfields/api/connection.py
josh146: Spot on @thisac. The issue is more that `allow_pickle=False` was only made the default in NumPy v1.17; in <=1.16, it's True by default!

While SF pins `numpy==1.17.4`, `pip` is very bad at dependency resolving, and it's possible that a subsequent installation could downgrade numpy --- this just plugs this as a potential security hole.

thisac: That makes perfect sense. Thanks for the explanation (and a really nice security fix).

PR #367, opened 2020-04-15
Fix printing of results object when it contains no state

### LINKED ISSUE #356: Error when printing Result for stateless engine
#### Issue description
* *Expected behavior:*
The result (samples) is printed without the state being included.

* *Actual behavior:*
The following error raised:
```python
  File "remote_engine_example.py", line 27, in <module>
    print(result)
  File "/strawberryfields/strawberryfields/api/result.py", line 113, in __str__
    len(self.samples), self.state, self.samples
  File "/strawberryfields/strawberryfields/api/result.py", line 107, in state
    raise AttributeError("The state is undefined for a stateless computation.")
AttributeError: The state is undefined for a stateless computation.
```

* *Reproduces how often:*
Every time.

* *System information:*
```python

Python version:            3.7.6
Platform info:             Linux-4.19.11-041911-generic-x86_64-with-debian-buster-sid
Installation path:        /strawberryfields/strawberryfields
Strawberry Fields version: 0.13.0.rc0
Numpy version:             1.18.1
Scipy version:             1.3.2
SymPy version:             1.5.1
NetworkX version:          2.3
The Walrus version:        0.12.0
Blackbird version:         0.2.3
TensorFlow version:        2.0.1
```

```python
import strawberryfields as sf 

con = sf.api.Connection() 
eng = sf.RemoteEngine("X8", connection=con) 
prog = sf.Program(8) 
U = random_interferometer(4) 

with prog.context as q: 
    ops.S2gate(1.0) | (q[0], q[4]) 
    ops.S2gate(1.0) | (q[1], q[5]) 
    ops.S2gate(1.0) | (q[2], q[6]) 
    ops.S2gate(1.0) | (q[3], q[7]) 
    ops.Interferometer(U) | q[:4] 
    ops.Interferometer(U) | q[4:] 
    ops.MeasureFock() | q 

result = eng.run(prog, shots=1000)
print(result)
```

#### Additional information

The error simply comes from the fact that the `state` property raises an error in the stateless case. It is, however, also used in the definition for string representation in `Result`, therefore `print(result)` will indeed result in an error for stateless engines.

A possible solution could be creating a branching logic inside of `__str__` without calling `state` in the stateless case.

josh146:
**Context:** The `Results.__str__()` method attempted to print out the contained state and samples, even if the backend didn't return a state.

**Description of the Change:** Removes the printing of contained data from the class string representation.

```python
>>> results
<Result: num_modes=3, shots=1, contains state=True>
>>> results.state
<FockState: num_modes=3, cutoff=5, pure=True, hbar=2>
>>>  results.samples
array([[0, None, -0.1783017830178295]], dtype=object)
```

**Benefits:** No longer errors if the backend does not return a state object. Provides metadata about the results object, rather than the underlying state and sample data, which is a better separation of logic.

**Possible Drawbacks:** We should add more metadata to the results object, including:

- The program used to generate the result
- The backend and backend options used
- If remote, details about the submitted remote job.

However, this is significantly more work.

**Related GitHub Issues:** Fixes #356

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -191,7 +191,7 @@
   [(#364)](https://github.com/XanaduAI/strawberryfields/pull/364)
 
 * Fixed a bug that caused an exception when printing results with no state.
-  [(#368)](https://github.com/XanaduAI/strawberryfields/pull/368)
+  [(#367)](https://github.com/XanaduAI/strawberryfields/pull/367)
 
 * Improves the Takagi decomposition, by making explicit use of the eigendecomposition of real symmetric matrices. [(#352)](https://github.com/XanaduAI/strawberryfields/pull/352)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -190,6 +190,9 @@
   due to floating point precision error.
   [(#364)](https://github.com/XanaduAI/strawberryfields/pull/364)
 
+* Fixed a bug that caused an exception when printing results with no state.
+  [(#368)](https://github.com/XanaduAI/strawberryfields/pull/368)
+
 * Improves the Takagi decomposition, by making explicit use of the eigendecomposition of real symmetric matrices. [(#352)](https://github.com/XanaduAI/strawberryfields/pull/352)
 
 <h3>Contributors</h3>
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -230,7 +230,7 @@ class Connection:
 
                 samples = np.load(buf, allow_pickle=False)
 
-                if samples.dtype.type is np.int8:
+                if np.issubdtype(samples.dtype, np.integer):
                     # Samples represent photon numbers.
                     # Convert to int64, to avoid unexpected behaviour
                     # when users postprocess these samples.
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -227,7 +227,15 @@ class Connection:
             with io.BytesIO() as buf:
                 buf.write(response.content)
                 buf.seek(0)
-                samples = np.load(buf, allow_pickle=False).astype(np.int64)
+
+                samples = np.load(buf, allow_pickle=False)
+
+                if samples.dtype.type is np.int8:
+                    # Samples represent photon numbers.
+                    # Convert to int64, to avoid unexpected behaviour
+                    # when users postprocess these samples.
+                    samples = samples.astype(np.float64)
+
             return Result(samples, is_stateful=False)
         raise RequestFailedError(
             "Failed to get job result: {}".format(self._format_error_message(response))
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -227,7 +227,7 @@ class Connection:
             with io.BytesIO() as buf:
                 buf.write(response.content)
                 buf.seek(0)
-                samples = np.load(buf, allow_pickle=False)
+                samples = np.load(buf, allow_pickle=False).astype(np.int64)
             return Result(samples, is_stateful=False)
         raise RequestFailedError(
             "Failed to get job result: {}".format(self._format_error_message(response))
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -234,7 +234,7 @@ class Connection:
                     # Samples represent photon numbers.
                     # Convert to int64, to avoid unexpected behaviour
                     # when users postprocess these samples.
-                    samples = samples.astype(np.float64)
+                    samples = samples.astype(np.int64)
 
             return Result(samples, is_stateful=False)
         raise RequestFailedError(
--- a/strawberryfields/api/result.py
+++ b/strawberryfields/api/result.py
@@ -109,6 +109,7 @@ class Result:
 
     def __str__(self):
         """String representation."""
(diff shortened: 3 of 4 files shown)


.github/CHANGELOG.md
```diff
@@ -190,6 +190,9 @@
   due to floating point precision error.
   [(#364)](https://github.com/XanaduAI/strawberryfields/pull/364)
 
+* Fixed a bug that caused an exception when printing results with no state.
+  [(#368)](https://github.com/XanaduAI/strawberryfields/pull/368)
```
antalszava, line 194:
```suggestion
  [(#367)](https://github.com/XanaduAI/strawberryfields/pull/367)
```

strawberryfields/api/connection.py
```diff
@@ -227,7 +227,15 @@ def get_job_result(self, job_id: str) -> Result:
             with io.BytesIO() as buf:
                 buf.write(response.content)
                 buf.seek(0)
+
                 samples = np.load(buf, allow_pickle=False)
+
+                if samples.dtype.type is np.int8:
+                    # Samples represent photon numbers.
+                    # Convert to int64, to avoid unexpected behaviour
+                    # when users postprocess these samples.
```
antalszava, line 236: Curious about this one: what type of errors arose for the users? :slightly_smiling_face:

tests/api/test_result.py
```diff
@@ -29,9 +30,27 @@ class TestResult:
     def test_stateless_result_raises_on_state_access(self):
         """Tests that `result.state` raises an error for a stateless result.
         """
-        result = Result([[1, 2], [3, 4]], is_stateful=False)
+        result = Result(np.array([[1, 2], [3, 4]]), is_stateful=False)
```
antalszava, line 33: How come this one was needed?

tests/api/test_result.py
```diff
@@ -29,9 +30,27 @@ class TestResult:
...
+        assert "contains state=False" in out
+
+    def test_state_print(self, capfd):
+        """Test that printing a result object with a state provides the correct output."""
+        result = Result(np.array([[1, 2], [3, 4], [5, 6]]), is_stateful=True)
+        print(result)
+        out, err = capfd.readouterr()
+        assert "modes=2" in out
+        assert "shots=3" in out
+        assert "contains state=True" in out
```
antalszava, line 56: Great! :confetti_ball: :100:

strawberryfields/api/result.py
```diff
@@ -107,8 +107,9 @@ def state(self):
             raise AttributeError("The state is undefined for a stateless computation.")
         return self._state
 
-    def __str__(self):
+    def __repr__(self):
         """String representation."""
-        return "Result: {} subsystems, state: {}\n samples: {}".format(
-            len(self.samples), self.state, self.samples
+        shots, modes = self.samples.shape
+        return "<Result: num_modes={}, shots={}, contains state={}>".format(
```
antalszava, line 113: So when working in a Jupyter Notebook, one could first have `print(Result)` which would then include the information about e.g. `Result.is_stateful==True`. And then this could be followed simply just by returning `Result.state`, right?

.github/CHANGELOG.md
josh146: Oops!

strawberryfields/api/connection.py
josh146: Users were performing numerical post-processing on the returned samples, and hitting overflow 🙂 

```python
>>> samples = np.array([1, 0, 2, 7], dtype=np.int8)
>>> samples**6
array([   1,    0,   64, -111], dtype=int8)
```

strawberryfields/api/result.py
josh146: Yep!

tests/api/test_result.py
josh146: Currently, the results object always receives the samples as a NumPy array, so the tests should be consistent and test the same.

strawberryfields/api/connection.py
antalszava: Great catch! :blush:

strawberryfields/api/connection.py
```diff
@@ -227,7 +227,15 @@ def get_job_result(self, job_id: str) -> Result:
             with io.BytesIO() as buf:
                 buf.write(response.content)
                 buf.seek(0)
+
                 samples = np.load(buf, allow_pickle=False)
+
+                if samples.dtype.type is np.int8:
```
lneuhaus, line 233: just wondering whether we shouldn't be more conservative here, and cast of the low-byte integer data type to int64, to avoid future issues (e.g. in case we'd switch to uint8 one day, or int16)?

```suggestion
                if samples.dtype in [np.int8, np.uint8, np.int16, np.uint16, np.int32, np.uint32]:
```

josh146:
> Nice fix! I was just wondering whether you want to add a test for the type casting or not, but I tested that the check for int8 works on my machine anyways.

Thanks @lneuhaus, I tested this as well locally. We will need to build up a set of integration tests for the platform in SF; the test for this conversion ends up being trivial due to mocking!

strawberryfields/api/connection.py
josh146: How about the following?
```suggestion
                if np.issubdtype(samples.dtype, np.integer):
```

This is much more general, and will cast any integral type.

PR #370, opened 2020-04-17
Add error details for failed remote job

pauljxtan:
**Context:**

Currently, a failed remote job produces a generic error message that is difficult to diagnose. The job info response from the remote platform contains a `meta` field that is currently unused, but has some information that may be useful, e.g.,

```python
{
    'error_code': 'hardware-message-error',
    'error_detail': 'The job failed because the message received from the hardware could not be understood.'
}
```

- Add a `meta` property to the `Job` class. In the event of job failure, this dictionary will contain an `error-code` and `error-detail`.
- Refresh the `meta` attribute (along with `status`) when `Job.refresh()` is called.
- Add the contents of `meta` to the error message for a failed job.

Example:
- Previous error message: `"The remote job 928ee53e-677a-48c0-8ba5-a0566ba3e295 failed due to an internal server error. Please try again."`
- New error message: `"The remote job 928ee53e-677a-48c0-8ba5-a0566ba3e295 failed due to an internal server error. Please try again. {'error_code': 'hardware-message-error', 'error_detail': 'The job failed because the message received from the hardware could not be understood.'}"`

**Benefits:**

The added context in a copy-pastable error message is likely to make it easier to diagnose and debug job failures.

**Possible Drawbacks:**

The new error message is slightly verbose.

**Related GitHub Issues:**

N/A

### DIFF
--- a/strawberryfields/api/job.py
+++ b/strawberryfields/api/job.py
@@ -73,7 +73,7 @@ class Job:
         status (strawberryfields.api.JobStatus): the job status
         connection (strawberryfields.api.Connection): the connection over which the
             job is managed
-        meta (dict[str, str]): auxiliary information related to job execution
+        meta (dict[str, str]): metadata related to job execution
     """
 
     def __init__(self, id_: str, status: JobStatus, connection, meta: dict = None):
@@ -123,7 +123,7 @@ class Job:
 
     @property
     def meta(self) -> dict:
-        """Returns a dictionary of auxiliary information on job execution, such as error
+        """Returns a dictionary of metadata on job execution, such as error
         details.
 
         Returns:
@@ -132,7 +132,7 @@ class Job:
         return self._meta
 
     def refresh(self):
-        """Refreshes the status and meta information of an open or queued job,
+        """Refreshes the status and metadata of an open or queued job,
         along with the job result if the job is newly completed.
         """
         if self._status.is_final:
@@ -140,7 +140,7 @@ class Job:
             return
         job_info = self._connection.get_job(self.id)
         self._status = JobStatus(job_info.status)
-        self._meta = JobStatus(job_info.meta)
+        self._meta = job_info.meta
         if self._status == JobStatus.COMPLETED:
             self._result = self._connection.get_job_result(self.id)
 
--- a/strawberryfields/api/job.py
+++ b/strawberryfields/api/job.py
@@ -81,7 +81,7 @@ class Job:
         self._status = status
         self._connection = connection
         self._result = None
-        self._meta = meta if meta is not None else {}
+        self._meta = meta or {}
 
         self.log = create_logger(__name__)
 
--- a/strawberryfields/api/job.py
+++ b/strawberryfields/api/job.py
@@ -141,6 +141,7 @@ class Job:
         job_info = self._connection.get_job(self.id)
         self._status = JobStatus(job_info.status)
         self._meta = job_info.meta
+        self.log.debug("Job {} metadata: {}".format(self.id, job_info.meta))
         if self._status == JobStatus.COMPLETED:
             self._result = self._connection.get_job_result(self.id)
 
--- a/strawberryfields/api/job.py
+++ b/strawberryfields/api/job.py
@@ -73,14 +73,15 @@ class Job:
         status (strawberryfields.api.JobStatus): the job status
         connection (strawberryfields.api.Connection): the connection over which the
             job is managed
+        meta (dict[str, str]): auxiliary information related to job execution
     """
 
-    def __init__(self, id_: str, status: JobStatus, connection):
+    def __init__(self, id_: str, status: JobStatus, meta: dict, connection):
         self._id = id_
         self._status = status
+        self._meta = meta
         self._connection = connection
         self._result = None
-        self._meta = None
 
         self.log = create_logger(__name__)
 
@@ -122,7 +123,8 @@ class Job:
 
     @property
     def meta(self) -> dict:
-        """Returns a dictionary of meta information on job execution.
+        """Returns a dictionary of auxiliary information on job execution, such as error
+        details.
 
         Returns:
             dict[str, str]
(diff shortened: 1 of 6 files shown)


strawberryfields/engine.py
```diff
@@ -548,7 +548,7 @@ def run(self, program: Program, *, compile_options=None, **kwargs) -> Optional[R
                 if job.status == "failed":
                     message = (
                         "The remote job %s failed due to an internal "
-                        "server error. Please try again." % job.id
+                        "server error. Please try again. %s" % (job.id, job.meta)
```
jswinarton, line 551: Any reason not to use new-style formatting here?

josh146: Oh, I think this is an artifact of [Python's logger expecting `printf` string formatting](https://docs.python.org/3/library/logging.html#formatter-objects). At one point, this line must have been inside the `self.log.error()` function.

Now that it's been refactored out, it's best to update this to use new-style formatting.

strawberryfields/api/job.py
```diff
@@ -119,14 +121,26 @@ def result(self) -> Result:
...
+        """Refreshes the status and metadata of an open or queued job,
         along with the job result if the job is newly completed.
         """
         if self._status.is_final:
             self.log.warning("A %s job cannot be refreshed", self._status.value)
             return
-        self._status = JobStatus(self._connection.get_job_status(self.id))
+        job_info = self._connection.get_job(self.id)
+        self._status = JobStatus(job_info.status)
+        self._meta = job_info.meta
```
josh146, line 143: Would it be a good idea to log the retrieved metadata at the DEBUG level here? Combined with #360, this will allow retrieved job metadata to appear within the users logs (if they decide to configure it).

pauljxtan: Good idea, this has been added! Let me know if you think the message formatting needs to be changed.

strawberryfields/engine.py
pauljxtan: Done!

pauljxtan:
Just FYI, @lneuhaus and I coordinated an end-to-end test for a deliberately failing job and didn't find any problems.

strawberryfields/api/job.py
```diff
@@ -119,14 +121,27 @@ def result(self) -> Result:
...
+        self.log.debug("Job {} metadata: {}".format(self.id, job_info.meta))
```
antalszava, line 144: Thanks for adding this! Usually using the old string formatting syntax is suggested for logging, I also looked up just now [a couple of further reasons to it](https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html). 
```suggestion
        self.log.debug("Job %s metadata: %s", self.id, job_info.meta)
```

strawberryfields/api/job.py
```diff
@@ -119,14 +121,27 @@ def result(self) -> Result:
...
-        """Refreshes the status of an open or queued job,
+        """Refreshes the status and metadata of an open or queued job,
         along with the job result if the job is newly completed.
         """
         if self._status.is_final:
             self.log.warning("A %s job cannot be refreshed", self._status.value)
             return
-        self._status = JobStatus(self._connection.get_job_status(self.id))
+        job_info = self._connection.get_job(self.id)
+        self._status = JobStatus(job_info.status)
```
antalszava, line 142: Logically for me, the `meta` data belongs to the status of the `Job`. At least this attribute will change the similarly to the change in `status` and as far as I understand it, is meant to provide more details about the status of the `Job`.

strawberryfields/engine.py
```diff
@@ -547,8 +547,8 @@ def run(self, program: Program, *, compile_options=None, **kwargs) -> Optional[R
 
                 if job.status == "failed":
                     message = (
-                        "The remote job %s failed due to an internal "
-                        "server error. Please try again." % job.id
+                        "The remote job {} failed due to an internal "
+                        "server error. Please try again. {}".format(job.id, job.meta)
```
antalszava, line 551: These were previously made so such that they can be used for logging, but here the message is already pre-defined, so I guess `format` should work just fine as well?

tests/api/test_remote_engine.py
```diff
@@ -35,16 +35,17 @@ class MockServer:
...
         threshold, beyond which a 'complete' job status is returned.
         """
         self.request_count += 1
-        return (
+        status = (
             JobStatus.COMPLETED
             if self.request_count >= self.REQUESTS_BEFORE_COMPLETED
             else JobStatus.QUEUED
         )
+        return Job(id_="123", status=status, connection=None)
```
antalszava, line 48: A small dummy meta dictionary could be added here, and then later checked similarly to the `status`.

antalszava:
Thanks so much for this @pauljxtan, looks really good! :100: Had a couple of suggestions which might be improvements to add, but I'm also happy with the current state.

strawberryfields/api/job.py
pauljxtan: Good catch, thanks!

tests/api/test_remote_engine.py
pauljxtan: Good idea, this has been added!

strawberryfields/api/job.py
pauljxtan: Yes, I agree with the general idea, and also that it would be a relatively big change, so it seems best to tackle that in the future :slightly_smiling_face:

strawberryfields/api/job.py
```diff
@@ -73,13 +73,15 @@ class Job:
...
+        meta (dict[str, str]): metadata related to job execution
     """
 
-    def __init__(self, id_: str, status: JobStatus, connection):
+    def __init__(self, id_: str, status: JobStatus, connection, meta: dict = None):
         self._id = id_
         self._status = status
         self._connection = connection
         self._result = None
+        self._meta = meta if meta is not None else {}
```
thisac, line 84: Any reason for not using the syntax below instead?

```suggestion
        self._meta = meta or {}
```

pauljxtan: No reason, just didn't think of it :slightly_smiling_face:

PR #371, opened 2020-04-19
Update gates.rst

nquesada:
Fixes wrong sign of two-mode squeezing gate in docs

### DIFF
--- a/doc/introduction/intro_photonics/conventions/gates.rst
+++ b/doc/introduction/intro_photonics/conventions/gates.rst
@@ -290,7 +290,7 @@ Two-mode squeezing
    :class: defn
 
    .. math::
-      S_2(z) = \exp\left(z \ad_1 \ad_2 - z^* \a_1\a_2 \right) = \exp\left(r (e^{-i\phi} \ad_1\ad_2 -e^{i\phi} \a_1 \a_2 \right)
+      S_2(z) = \exp\left(z \ad_1 \ad_2 - z^* \a_1\a_2 \right) = \exp\left(r (e^{i\phi} \ad_1\ad_2 -e^{-i\phi} \a_1 \a_2 \right)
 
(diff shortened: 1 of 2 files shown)


co9olguy:
So the one on the conventions page is the correct convention employed by SF?
https://strawberryfields.readthedocs.io/en/stable/code/api/strawberryfields.ops.S2gate.html

nquesada:
The correct convention is S^2(z) = \exp( z a_1^\dagger a_2^\dagger - H.c.) with z = r  e^{i \theta}. 
To verify this is the case one can calculate <1,1|S^2(z)|0,0> = tanh(r) e^{i \theta} /cosh(r)

Here is a simple code that checks that:

```python
eng = sf.LocalEngine(backend="fock",backend_options={"cutoff_dim": 6})
width = 2
circuit = sf.Program(width)
z = 0.5*np.exp(1j*np.pi/6)
r = np.abs(z)
theta = np.angle(z)
with circuit.context as q:
    ops.S2gate(r,theta)|(q[0],q[1])
x = eng.run(circuit)
state = x.state
state.ket()[1,1], np.tanh(r)*np.exp(1j*theta)/np.cosh(r)
```
giving
```
((0.3549095267938162+0.20490711083237245j),
 (0.35490952679381627+0.20490711083237248j))
```
as expected. 

The one in https://strawberryfields.readthedocs.io/en/stable/code/api/strawberryfields.ops.S2gate.html is incorrect. A likely cause of confusion is that we define the S gate as S(z) = \exp( -z a^\dagger/2 -H.c.)  and thus one would hope that S^(2)(z) is defined as \exp(-z a_1^\dagger a_^\dagger - H.c.). All of this in turn I think goes back to how Eqs. 31 and 33 of here https://arxiv.org/pdf/1110.3234.pdf are incosistent.

co9olguy:
Thanks. Can you update the error also in `sf.ops.S2gate` docstring?

nquesada:
No problem.

PR #374, opened 2020-04-29
Updates the gate synthesis tutorial for TensorFlow 2.0

antalszava:
**Context:** The existing TensorFlow tutorial was out of date (requiring the old TF1.3 backend), and in the form of a Jupyter notebook.

* Ports the tutorial to Sphinx-Gallery, by using the [converter script available here](https://gist.github.com/chsasank/7218ca16f8d022e02a9c0deb94a310fe).

* Ports the tutorial to use the new TF2 backend

**Benefits:** As the tutorial has been ported to Sphinx gallery, it is now executable, making it easy to test to see whether it provides the correct output.

**Possible Drawbacks:** n/a


examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,453 @@
...
+    mapping = {p.name: w for p, w in zip(sf_params.flatten(), tf.reshape(weights, [-1]))}
+
+    # Run engine
+    state = eng.run(prog, args=mapping).state
+
+    # Extract the statevector
+    ket = state.ket()
+
+    # # overlaps
+    overlaps = tf.math.real(tf.einsum('bi,bi->b', tf.math.conj(target_kets), ket))
```
trbromley, line 297: Note the presence of einsum here, and that its gradient is incorrect for complex numbers.
See https://github.com/XanaduAI/strawberryfields/pull/377.

antalszava: Thanks @trbromley !

antalszava:
The [rendered version looks like this](https://user-images.githubusercontent.com/24476053/80644530-c9b2e400-8a37-11ea-9101-7979eb25ea34.png).

When training had a mean overlap value of `0.89` after 100 reps with a final of `0.9628` at 300 reps.

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+
+# We attempt to import the legacy version of tf.einsum to make sure that no
+# issues arise when computing with complex tensors (this is required due to a
+# known issue in TF version 2.1)
+try:
+    from tensorflow.python.ops.special_math_ops import _einsum_v1
+
+    tf.einsum = _einsum_v1
+except ImportError:
+    pass
```
antalszava, line 146: Tried skipping the use of `tf.einsum` by using a solution with `tf.matmul`, however, got no convergence on TF 2.1.

The attempted solution is a part of a former commit https://github.com/XanaduAI/strawberryfields/pull/374/commits/3ea096bf403aeefc3373d65ecd5a960cee92bb6a .

josh146: @antalszava, I think this can be removed now. I've merge #377 into this branch, so it should converge with 2.1 (the state learning tutorial converges correctly with 2.1 after #377 was merged in).

examples_apps/run_gate_synthesis.py
josh146: Now that #377 has been merged in, this call to `tf.einsum` will automatically call `_einsum_v1`, due to a quirk of how Python imports work :)

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+we choose must consist of layers with each layer containing
+parameterized Gaussian *and* non-Gaussian gates. **The non-Gaussian
+gates provide both the nonlinearity and the universality of the model.**
+To this end, we employ the CV quantum neural network architecture
+described below:
+
+.. figure:: https://i.imgur.com/NEsaVIX.png
+   :alt: layer
+
+   layer
```
trbromley, line 47:
```suggestion
```

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+# the shape of ``weights``.
+
+sf_params = np.array(sf_params)
+print(sf_params.shape)
+
+######################################################################
+# Now, we can create a function to define the :math:`i`\ th layer, acting
+# on qumode ``q``. We add the :class:`~.utils.operation` decorator so that the layer can be used
+# as a single operation when constructing our circuit within the usual
+# Strawberry Fields Program context
```
trbromley, line 212:
```suggestion
# Strawberry Fields Program context.
```

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+# relation.
+#
+# For our target unitary, let's use Strawberry Fields to generate a 4x4
+# random unitary:
+#
+
+from strawberryfields.utils import random_interferometer
+target_unitary = np.identity(cutoff, dtype=np.complex128)
+target_unitary[:gate_cutoff, :gate_cutoff] = random_interferometer(4)
+print(target_unitary)
```
trbromley, line 269:
```suggestion
from strawberryfields.utils import random_interferometer

# define unitary up to gate_cutoff
random_unitary = random_interferometer(gate_cutoff)
print(random_unitary)

# extend unitary up to cutoff
target_unitary = np.identity(cutoff, dtype=np.complex128)
target_unitary[:gate_cutoff, :gate_cutoff] = random_unitary
```

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+
+plt.plot(cost_progress)
+plt.ylabel('Cost')
+plt.xlabel('Step')
+plt.show()
+
+
+######################################################################
+# We can use matrix plots to plot the real and imaginary components of the
+# target and learnt unitary.
```
trbromley, line 387:
```suggestion
# target unitary :math:`V` and learnt unitary :math:`U`.
```

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+    overlap_progress.append(overlaps_val)
+
+    # one repetition of the optimization
+    gradients = tape.gradient(loss, weights)
+    opt.apply_gradients(zip([gradients], [weights]))
+
+    # Prints progress at every 100 reps
+    if i % 1 == 0:
+        # print progress
+        print("Rep: {} Cost: {:.4f} Mean overlap: {:.4f}".format(i, loss, mean_overlap_val))
```
trbromley, line 363: Note that after 300 steps the I got a mean overlap of 90% - whereas in the paper we were looking at 99%+ and in the previous Jupyter notebook version we got to 98%+.

I think this is due to the number of iterations dropping to 300 from 1000 and the depth also going from 25 to 20. Is the motivation to speed things up? I guess it's ok, perhaps we could add a note saying that higher mean overlaps are possible if you keep optimizing?

antalszava: Thanks for that @trbromley! Yes, the motivation was speeding up, but these are fair points. Added a note on this for increasing the number of reps at the end of the tutorial. Also added certain updates and now training seems to be faster for me locally.

antalszava:
[Rendered version](https://user-images.githubusercontent.com/24476053/80711958-31f8d880-8abf-11ea-8979-158bb8d6e8bf.png) with updates after #377 

`Rep: 199 Cost: 0.1077 Mean overlap: 0.9731`

examples_apps/run_gate_synthesis.py
josh146: Yep, this is just to reduce the time taken. We have two options:

* Leave it as is, and at the bottom of the tutorial, have a note box letting users know they can download the tutorial, and increase the number of steps for even better results!

* Turn it into a non-executable tutorial, and hardcode in the results.

As we're about to move the tutorials out to their own repository, I'm somewhat leaning towards option 1. Once it's moved over to it's own repo and building on CircleCI, we can just increase the number of reps.

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,469 @@
...
+a quantum circuit with a layer structure that is **universal** - that
+is, by ‘stacking’ the layers, we can guarantee that we can produce *any*
+CV state with at-most polynomial overhead. Therefore, the architecture
+we choose must consist of layers with each layer containing
+parameterized Gaussian *and* non-Gaussian gates. **The non-Gaussian
+gates provide both the nonlinearity and the universality of the model.**
+To this end, we employ the CV quantum neural network architecture
+described below:
+
+.. figure:: https://i.imgur.com/NEsaVIX.png
```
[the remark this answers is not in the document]
josh146: That was me 🙂 Reason was because this was originally a standalone Jupyter notebook which I was distributing at a workshop, so relative image paths wouldn't work.

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,470 @@
+r"""
+Quantum gate synthesis
+======================
+
+This demonstration works through the process used to produce the gate
+synthesis results presented in `“Machine learning method for state
+preparation and gate synthesis on photonic quantum
+computers” <https://arxiv.org/abs/1807.10781>`__.
```
josh146, line 8:
```suggestion
synthesis results presented in `"Machine learning method for state
preparation and gate synthesis on photonic quantum
computers" <https://arxiv.org/abs/1807.10781>`__.
```

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,470 @@
...
+unitary.
+
+.. note::
+
+    For more details on the TensorFlow backend in Strawberry Fields, please see
+    :ref:`machine_learning_tutorial`.
+
+For arbitrary gate synthesis using optimization, we need to make use of
+a quantum circuit with a layer structure that is **universal** - that
+is, by ‘stacking’ the layers, we can guarantee that we can produce *any*
```
josh146, line 36:
```suggestion
is, by 'stacking' the layers, we can guarantee that we can produce *any*
```

examples_apps/run_gate_synthesis.py
```diff
@@ -0,0 +1,470 @@
...
+#
+
+print(np.max(np.abs(weights[:, 0])))
+
+######################################################################
+# Further results
+# ---------------
+#
+# Even more refined results can be obtained by increasing the number of
+# repetitions (``reps``) after downloading the tutorial!
```
josh146, line 452: Nice! Could even suggest playing around with the number of layers and the gate cutoff

antalszava: Added!

PR #377, opened 2020-04-29
Fix einsum gradient by using old TF einsum

trbromley:
TF2.1+ introduces a bug into `einsum` that results in an incorrect gradient for complex numbers: https://github.com/tensorflow/tensorflow/issues/37307.

This PR adds a quick fix where the old version of `einsum` will be used for TF2.1+.

### DIFF
--- a/strawberryfields/backends/tfbackend/circuit.py
+++ b/strawberryfields/backends/tfbackend/circuit.py
@@ -39,9 +39,10 @@ from scipy.special import factorial
 import tensorflow as tf
 
 try:
-    from tensorflow.python.ops.special_math_ops import _einsum_v1 as tf_einsum
+    from tensorflow.python.ops.special_math_ops import _einsum_v1
+    tf.einsum = _einsum_v1
 except ImportError:
-    from tensorflow import einsum as tf_einsum
+    pass
 
 from . import ops
 
@@ -65,7 +66,7 @@ class Circuit:
         one = tf.cast([1.0], ops.def_type)
         v = tf.scatter_nd([[0]], one, [cutoff_dim])
         self._single_mode_pure_vac = v
-        self._single_mode_mixed_vac = tf_einsum("i,j->ij", v, v)
+        self._single_mode_mixed_vac = tf.einsum("i,j->ij", v, v)
         if self._batched:
             self._single_mode_pure_vac = tf.stack([self._single_mode_pure_vac] * self._batch_size)
             self._single_mode_mixed_vac = tf.stack([self._single_mode_mixed_vac] * self._batch_size)
@@ -690,7 +691,7 @@ class Circuit:
                     ]
                     cond_ctr += 1
             eqn = eqn_lhs + "->" + batch_index + eqn_rhs
-            new_state = tf_einsum(eqn, meas_modes_vac, normalized_conditional_state)
+            new_state = tf.einsum(eqn, meas_modes_vac, normalized_conditional_state)
 
             self._update_state(new_state)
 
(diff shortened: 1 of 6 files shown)


strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -38,6 +38,13 @@
 from scipy.special import factorial
 import tensorflow as tf
 
+try:
+    from tensorflow.python.ops.special_math_ops import _einsum_v1
+
+    tf.einsum = _einsum_v1
+except ImportError:
+    pass
```
josh146, line 46: With TF 2.1, the legacy `tf.einsum` was renamed to `_einsum_v1`, while the replacement `tf.einsum` introduced the bug. This `try-except` block will dynamically patch versions where `_einsum_v1` exists, to make it the default einsum implementation _within this file only_.

tests/integration/test_tf_integration.py
```diff
@@ -422,3 +422,25 @@ def test_coherent_dm_gradient(self, setup_eng, cutoff, tol, batch_size):
         grad = tape.gradient(res, [a])
         grad_ex = -a ** 3 * (a ** 2 - 2) * np.exp(-a ** 2)
         assert np.allclose(grad, grad_ex, atol=tol, rtol=0)
+
+
+def test_TODO():
```
co9olguy, line 427: I'm curious about this test name :thinking:

tests/integration/test_tf_integration.py
```diff
@@ -422,3 +422,25 @@ def test_coherent_dm_gradient(self, setup_eng, cutoff, tol, batch_size):
...
+        return tf.abs(tf.reduce_sum(_einsum_v1('ab,bc->ac', h, h)))
+
+    A = tf.Variable(tf.random.uniform(shape=[2, 2]))
+    B = tf.random.uniform(shape=[2, 2])
+    for f in (f1, f2, f3):
+        with tf.GradientTape() as tape:
+            C = tf.cast(A, dtype=tf.complex64) + 1j * tf.cast(B, dtype=tf.complex64)
+            loss = f(C)
+        grad = tape.gradient(loss, A)
+        print(grad)
```
co9olguy, line 446: shouldn't a test conclude with an `assert`, not a `print`?

tests/integration/test_tf_integration.py
co9olguy: No docstring either

co9olguy:
I'm glad to have this hotfix in, but can we give the commensurate test a bit more attention?

josh146:
@co9olguy, the test was updated and available here: https://github.com/XanaduAI/strawberryfields/blob/c7dd6f58fb6b8a9cb306de4b2e4e9ef13824d2a1/tests/integration/test_tf_integration.py#L484

tests/integration/test_tf_integration.py
trbromley: Yes this shouldn't be here (yet), I think it was a bit of a miscommunication on the PR being ready to merge.

The idea of this test is to compare gradients using old and new versions of einsum. The plan is to assert that the gradients should be the same and add an xfail to show that we expect failure - when this stops being the case we know they've fixed the bug.

Unfortunately I couldn't get it to work and was going to come back to this today, hence the TODO and the print statements - it's just in an early prototype form.

co9olguy:
Thanks

tests/integration/test_tf_integration.py
trbromley: Since this is merged into the `tf2` branch and has not yet made it to `master`, we can sort this out by doing a follow up PR to properly add the test.

trbromley: Oh wait, @josh146 already fixed it! https://github.com/XanaduAI/strawberryfields/commit/932856b9e399331be7966c7d962964f8d750f361 :100:

PR #379, opened 2020-04-29
Add dark counts to MeasureFock

thisac:
**Context:**
It would be a nice addition to have dark counts added to `MeasureFock` to make simulations more realistic by the adding the ability to add noise to the measurements.

Dark counts are added as an attribute in `MeasureFock` at initialisation and then applied to the samples returned from all backends when calling `MeasureFock._apply`. If no dark counts are supplied (or `dark_counts=None`), the samples are returned without any dark counts added to the samples. The dark counts are sampled from a [Poisson distribution ](https://docs.scipy.org/doc/numpy-1.14.1/reference/generated/numpy.random.poisson.html).

These can be added either as a float, when only measuring one mode:

```python
with prog.context as q:
    MeasureFock(dark_counts=0.2) | q
```

or as a list for both one or more modes:

```python
with prog.context as q:
    MeasureFock(dark_counts=[0.2, 1.3]) | q
```

Note: 
- The number of dark counts _must_ be equal to the number of modes measured; otherwise an exception is raised.

- If both the `dark_counts` and `select` keywords are supplied, a `NotImplementedError` is raised.

- The `io.to_blackbird` function is updated to support `dark_counts`; thus saving with `sf.save` works (`sf.load`, however, does not; see comment in **Possible Drawbacks:** below).

**Benefits:**
Simulations which use `MeasureFock` can now be run with dark counts added, providing more realistic, noisy, simulations.

**Possible Drawbacks:**
1. Since the `dark_counts` keyword is only part of `MeasureFock` (and not its super-class `Measurements`) there is now added complexity both in the `MeasureFock` class, especially in `MeasureFock.__init__`, and in the `io.to_blackbird` function, which could be a bit confusing.

2. Blackbird doesn't seem to support `[` nor `]` causing en error when trying to load a blackbird script that contains square brackets. This is the case when saving a circuit containing either post-selection or dark counts (since both of these are saved as lists).

3. The way that the TensorFlow backend returns results as a list of `tf.Tensor` and `None` currently causes issues with this implementation of dark counts. This is deliberately not tested for here, and I believe that it's something that might be fixed fairly soon (i.e. results _always_ having the type `tf.Tensor`). As soon as this is fixed everything should work with the `tf` backend, including the tests.

**Related GitHub Issues:**
N/A


strawberryfields/ops.py
```diff
@@ -757,13 +757,24 @@ class MeasureFock(Measurement):
...
+        self.dark_counts = dark_counts
+
         if select is not None and not isinstance(select, Sequence):
             select = [select]
+
         super().__init__([], select)
 
     def _apply(self, reg, backend, shots=1, **kwargs):
-        return backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
+        samples = backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
```
nquesada, line 769: wonder if the additions can be vectorized? Other than that this looks good!

strawberryfields/ops.py
```diff
@@ -757,13 +757,24 @@ class MeasureFock(Measurement):
 
     ns = None
 
-    def __init__(self, select=None):
+    def __init__(self, dark_counts=None, select=None):
```
nquesada, line 760: I guess this means the user either specifies the dark count rates for all the modes or for none of them, which I guess is fair.

thisac: Yep, it wouldn't be difficult _per se_ to add the option to specify dark counts for some modes (or e.g. supply just one value that is applied to all modes) but I'm not sure the extra complexity is worth it.

strawberryfields/ops.py
thisac: Not sure how this could be vectorised. Do you mean the list comprehension below?

nquesada: are the samples a list or a numpy array?

strawberryfields/ops.py
```diff
@@ -757,13 +757,24 @@ class MeasureFock(Measurement):
...
 
     def _apply(self, reg, backend, shots=1, **kwargs):
-        return backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
+        samples = backend.measure_fock(reg, shots=shots, select=self.select, **kwargs)
+
+        if self.dark_counts is not None:
+            return [
+                [s[i] + np.random.poisson(j) for i, j in enumerate(self.dark_counts)]
+                for s in samples
+            ]
```
thisac, line 775: Actually, I guess this is actually a much better, vectorised, solution. Thanks @nquesada for pointing it out.

```suggestion
            return samples + np.random.possion(self.dark_counts, samples.shape)
```

nquesada: fantastisk!

thisac: But only if `samples` is returned as `ndarray`.

strawberryfields/ops.py
josh146: I suppose one use-case could be if the hardware at some point provides per-detector dark count info. Summoning @lneuhaus to see if this is likely 🙂 

If so, it should be relatively easy to do the following, right?

```python
MeasureFock(dark_counts=[0.1, 0.2]) | (q[0], q[1])
```

lneuhaus: > if the hardware at some point provides per-detector dark count info

We plan to provide information about coherent (=pump leakage) and incoherent (=Raman photons, blackbody radiation on the detector) "dark" count rates in the hardware certificate. This is relatively easy to measure. 

However, I don't see what the user can do about this by setting the dark count rate to a desired value (if we could control it on the hardware, we would always have it zero of course). What do you have in mind @josh146 ?

lneuhaus: Yes, this can be done. And I think it makes sense to be able to simulate hardware imperfections in this way. And passing the arguments in this way (e.g. rather than as a backend kwarg) makes sense as well, as this is really a detector property.

thisac: Now it should be possible to add a dark-counts value to each measured mode (if `len(dark_counts)` is *not* equal to the number of measured modes, a `ValueError` is raised). Also, the samples are now always returned as a `np.ndarray`; before it depended on the number of shots.

josh146: Awesome!

strawberryfields/io.py
```diff
@@ -69,6 +69,11 @@ def to_blackbird(prog, version="1.0"):
                 # argument is quadrature phase
                 op["kwargs"]["phi"] = cmd.op.p[0]
 
+            if op["op"] == "MeasureFock":
+                # special case to take into account 'dark_counts' keyword argument
+                if cmd.op.dark_counts is not None:
+                    op["kwargs"]["dark_counts"] = cmd.op.dark_counts
```
josh146, line 75: Nice!

strawberryfields/ops.py
```diff
@@ -782,13 +787,46 @@ class MeasureFock(Measurement):
 
     ns = None
 
-    def __init__(self, select=None):
+    def __init__(self, select=None, dark_counts=None):
+        if dark_counts and select:
+            raise NotImplementedError("Post-selection cannot be used together with dark counts.")
```
josh146, line 792: :+1:

tests/integration/test_measurement_integration.py
```diff
@@ -145,3 +145,85 @@ def test_coherent_state_has_photons(self, setup_eng, hbar):
...
+
+
+class TestDarkCounts:
+    @pytest.mark.backends("fock")
+    @pytest.mark.parametrize("dark_counts", [[4, 2], [3, 0], [0, 6]])
+    def test_fock_darkcounts_two_modes(self, dark_counts, setup_eng, monkeypatch):
+        """Test Fock measurements return expected results with dark counts on all detectors"""
+        with monkeypatch.context() as m:
+            # add the number of dark counts instead of sampling from a poisson distribution
+            m.setattr(np.random, "poisson", lambda dc, shape: dc * np.ones(shape, dtype=int))
```
josh146, line 157: This is a great way to avoid the stochasticity!

tests/integration/test_measurement_integration.py
```diff
@@ -145,3 +145,85 @@ def test_coherent_state_has_photons(self, setup_eng, hbar):
...
+        n = [2, 1]
+
+        with pytest.raises(
+                NotImplementedError,
+                match="Post-selection cannot be used together with dark counts",
+            ):
+            with prog.context as q:
+                ops.Fock(n[0]) | q[0]
+                ops.Fock(n[1]) | q[1]
+                ops.MeasureFock(select=1, dark_counts=2) | q
```
josh146, line 228: Great integration tests! Only change I would make here is swapping the order of the with contexts, so its easier to see which line should raise the error:
```suggestion
        with prog.context as q:
            ops.Fock(n[0]) | q[0]
            ops.Fock(n[1]) | q[1]
            with pytest.raises(
                    NotImplementedError,
                    match="Post-selection cannot be used together with dark counts",
                ):
                ops.MeasureFock(select=1, dark_counts=2) | q
```

PR #380, opened 2020-04-30
Increasing required version for The Walrus

antalszava:
Increasing the required version for The Walrus such that the newest features are available.


antalszava:
@josh146 thanks for getting that! How come we are pinning in `setup.py`?

josh146:
> @josh146 thanks for getting that! How come we are pinning in setup.py?

While not essential, this is mostly for users upgrading Strawberry Fields versions; it ensures that The Walrus is also upgraded.

PR #383, opened 2020-05-02
Norun tutorial

josh146:
**Context:** The new TF2.0 tutorials are too computational expensive to execute to completion on read the docs.

**Description of the Change:** Makes the new tutorials non-executable by (a) removing the `run_` from their filenames, and (b) hardcoding in the output as ReST-formatted comment blocks.

**Benefits:**

* Now executes on ReadTheDocs, and tested to confirm.

* Since the output of the tutorials are hardcoded in as _comments_, they are still executable as Python scripts.

**Possible Drawbacks:** n/a


josh146:
The RTDs build for this PR is available here: https://strawberryfields.readthedocs.io/en/norun_tutorial/

doc/conf.py
```diff
@@ -84,7 +94,7 @@
     # execute files that match the following filename pattern,
     # and skip those that don't. If the following option is not provided,
     # all example scripts in the 'examples_dirs' folder will be skiped.
-    'filename_pattern': r'run',
+    'filename_pattern': r'/run_',
```
antalszava, line 97: Just our of curiosity: was this change needed even after renaming the tutorials?

josh146:
Thanks @antalszava!

> Might be worth removing the Total running time of the script: ( 0 minutes 0.000 seconds) text appearing at the end of each tutorial

This is a built-in part of Sphinx-gallery and cannot be removed on a case-by-case basis. Since this is just a temporary fix, we'll leave it for now.

> For the minimizing correlations tutorial: the author section got into the main text

Good catch, will fix this!

> Numbers for citations sometimes are in the same flow with the main text (end of the sentence at this example). Perhaps they could be enclosed by []? (though this might be something coming from sphinx)

Yes, this seems to be due to a change in Sphinx :( We are slowly removing sphinx-bibtex from our documentation to avoid this issue

PR #387, opened 2020-05-08
Add GBS parametrization for training variational GBS circuits

trbromley:
This PR adds to the `train` module of `sf.apps`. A `parametrization` module is added which contains a `VGBS` class allowing users to vary an adjacency matrix given an embedding and trainable parameters. The embedding is provided in https://github.com/XanaduAI/strawberryfields/pull/388.


strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+        self,
+        A: np.ndarray,
+        n_mean: float,
+        embedding,
+        threshold: bool,
+        samples: Optional[np.ndarray] = None,
+    ):
+        if not np.allclose(A, A.T):
+            raise ValueError("Input must be a NumPy array corresponding to a symmetric matrix")
+        self.A_init = A
```
ixfoduap, line 146: is `A_init` vs `A` really helpful? I understand it's meant to clarify that it's the input `A`, but I get confused with many `A` flying around

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+        self.embedding = embedding
+        self.threshold = threshold
+        self.A_init_scaled = rescale_adjacency(A, n_mean, threshold)
+        self.n_modes = len(A)
+        self._A_init_samples = None
+        if samples:
+            self._add_A_init_samples(samples)
+
+    def W(self, params: np.ndarray) -> np.ndarray:
```
ixfoduap, line 155: Here we have to make a decision regarding convention: the diagonal entries of `W` can be either `w` or `sqrt(w)`. In the paper, we use `sqrt(w)` and so in the gradient formulas only the `w` comes out. If we use `w` instead, gradient formulas will appear with `w^2`

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+            params (array): the trainable parameters :math:`\theta`
+
+        Returns:
+            array: the diagonal matrix of weights
+        """
+        return np.diag(self.embedding(params))
+
+    @staticmethod
+    def _WAW(A: np.ndarray, W: np.ndarray) -> np.ndarray:
+        """Calculate the :math:`_WAW` parametrization.
```
ixfoduap, line 177: Maybe "calculate the WAW matrix?" I'm not sure a parametrization is something you calculate

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+
+        Args:
+            params (array): the trainable parameters :math:`\theta`
+
+        Returns:
+            array: the trained adjacency matrix
+        """
+        return self._WAW(self.A_init_scaled, self.W(params))
+
+    def _generate_samples(self, A: np.ndarray, n_samples: int, **kwargs) -> np.ndarray:
```
ixfoduap, line 218: I think it would be best to have a single `generate_samples()` function and simply specify the corresponding `A` matrix depending on the intended sampling. Having three separate sample functions, even if some are internal, seems unnecessary.

strawberryfields/apps/train/parametrization.py
ixfoduap: This actually touches open a point we need to get right: the role of `A` and `A_scaled`. I feel like there should only be one `A`, the rescaled one. So for example, if we initialize the class with an unnormalized `A` and then call `vgbs.A(np.ones(dim))` (so `W` is the identity), we should just get back the rescaled `A`.

So essentially the rescaled `A` is always the point of reference; we just make it easy for the user to input even an unnormalized one. Alternatively, we could force the user to always input a rescaled matrix

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+        shape = samples.shape
+        if shape[1] != self.n_modes:
+            raise ValueError("Must input samples of shape (number, {})".format(self.n_modes))
+
+        if self._A_init_samples is None:
+            self._A_init_samples = samples
+        else:
+            self._A_init_samples = np.vstack([self._A_init_samples, samples])
+
+    def mean_photons_by_mode(self, params: np.ndarray) -> np.ndarray:
```
ixfoduap, line 301: clever function name!

tests/apps/test_train.py
```diff
@@ -0,0 +1,213 @@
...
+            m.setattr(thewalrus.samples, "torontonian_sample_state", lambda *args, **kwargs: 0)
+            s_threshold = gbs._generate_samples(gbs.A_init_scaled, 10)
+
+        gbs.threshold = False
+
+        with monkeypatch.context() as m:
+            m.setattr(thewalrus.samples, "hafnian_sample_state", lambda *args, **kwargs: 1)
+            s_pnr = gbs._generate_samples(gbs.A_init_scaled, 10)
+
+        assert s_threshold == 0
```
ixfoduap, line 122: I'm confused by this test. Is the idea that we're testing whether we are making the right call to the walrus function? I thought the output of `_generate_samples` was a sample, so how come we're testing if it's `0` or `1`? Not saying there's a problem with the test, only that it's hard to follow and therefore to review

tests/apps/test_train.py
```diff
@@ -0,0 +1,213 @@
...
+            gbs._add_A_init_samples(s)
+
+    def test_add_A_init_samples_none_there(self, adj, n_mean, dim):
+        """Test that _add_A_init_samples correctly adds samples"""
+        gbs = train.VGBS(adj, n_mean, embedding, True)
+        s = np.ones((2, dim))
+        gbs._add_A_init_samples(s)
+        assert np.allclose(gbs._A_init_samples, s)
+
+    def test_add_A_init_samples_already_there(self, adj, n_mean, dim):
```
ixfoduap, line 140: I'm being picky here, but technically we could be adding the wrong samples or adding them in the wrong order and this test would still pass, right? We could make initial samples to be all zeros and new samples be all ones and check that they are added as intended

tests/apps/test_train.py
```diff
@@ -0,0 +1,213 @@
...
+        assert gbs._A_init_samples.shape == (4, dim)
+
+    def test_mean_photons_per_mode(self, n_mean, dim):
+        """Test that mean_photons_by_mode is correct when given a simple fully connected
+        adjacency matrix and an identity W. We expect each mode to have the same mean photon number
+        and for that to add up to n_mean."""
+        adj = np.ones((dim, dim))
+        params = np.zeros(dim)
+        gbs = train.VGBS(adj, n_mean, embedding, False)
+        n_mean_vec = np.round(gbs.mean_photons_by_mode(params), 5)
```
ixfoduap, line 155: Does the test fail without `round`?

tests/apps/test_train.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def test_add_A_init_samples_already_there(self, adj, n_mean, dim):
+        """Test that _add_A_init_samples correctly adds more samples when some are already there"""
+        gbs = train.VGBS(adj, n_mean, embedding, True)
+        s = np.ones((2, dim))
+        gbs._A_init_samples = s
+        gbs._add_A_init_samples(s)
+        assert gbs._A_init_samples.shape == (4, dim)
+
+    def test_mean_photons_per_mode(self, n_mean, dim):
```
ixfoduap, line 148: Great test!

tests/apps/test_train.py
```diff
@@ -0,0 +1,213 @@
...
+        adj = np.ones((dim, dim))
+        params = np.zeros(dim)
+        gbs = train.VGBS(adj, n_mean, embedding, True)
+        n_mean_vec = np.round(gbs.mean_clicks_by_mode(params), 5)
+
+        assert np.allclose(np.sum(n_mean_vec), n_mean)  # check that the vector sums to n_mean
+        assert len(np.unique(n_mean_vec)) == 1  # check that the vector is constant
+        assert np.allclose(n_mean_vec[0], n_mean / dim)  # check that elements have correct values
+
+    def test_photons_clicks_comparison(self, n_mean, dim):
```
ixfoduap, line 174: This test seems a bit dangerous if we ever want to change `n_mean` or the number of modes. Maybe better to test that `n_mean_vec_click[0] < 1` and `n_mean_vec_photon[0] >= n_mean_vec_click[0]` ? Note the second inequality is true by construction

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+    """
+    n = len(A)
+    A_big = np.block([[A, 0 * A], [0 * A, np.conj(A)]])
+    I = np.identity(2 * n)
+    X = Xmat(n)
+    return np.linalg.inv(I - X @ A_big) - I / 2
+
+
+# TODO: update embedding class in type hint
+class VGBS:
```
ixfoduap, line 86: I'm not 100% behind the name `VGBS`, yet I can't think of a better one. Does anyone else have other ideas?

tests/apps/test_train.py
trbromley: Yes, it is used for checking that the vector is constant. But good spot, I don't like it and have changed to
`assert np.allclose(n_mean_vec - n_mean_vec[0], np.zeros(dim))`

tests/apps/test_train.py
trbromley: Nice! I have changed the test to check that all elements of n_mean_vec_photon are larger than n_mean_vec_click and that all elements of n_mean_vec_click do not exceed one.

tests/apps/test_train.py
trbromley: Nice that's a good point! As well as the shape, the test now checks that the first two samples are ones and the second two are zeros.

strawberryfields/apps/train/parametrization.py
trbromley: Great! I have updated to:

- Remove the `_WAW` method
- Make the `A` method be the one that gives the scaled WAW
- Remove the `A_scaled` method as this is now `A`

One thing to note is that there is that there is still an `A_init` attribute - it would be great to have an `A` attribute _and_ and `A` method, but this is not possible in Python(?).

strawberryfields/apps/train/parametrization.py
trbromley: Oh nice, I have changed to use square root - which also makes this method more nontrivial :laughing:

strawberryfields/apps/train/parametrization.py
trbromley: Great, we now just have one `generate_samples()` that takes an adjacency matrix as input!

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,312 @@
...
+            array:  the generated samples
+        """
+        cov = A_to_cov(A)
+
+        if self.threshold:
+            samples = thewalrus.samples.torontonian_sample_state(
+                cov, n_samples, hbar=sf.hbar, **kwargs
+            )
+        else:
+            samples = thewalrus.samples.hafnian_sample_state(cov, n_samples, hbar=sf.hbar, **kwargs)
```
trbromley, line 221: @nquesada we are using `hbar=sf.hbar`, do you think that will work ok (tests are passing!)

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,312 @@
+# Copyright 2020 Xanadu Quantum Technologies Inc.
```
trbromley, line 1: What about changing the name of the module? It won't be front facing anyway as the init above imports `VGBS` directly, but still good to choose a name that works.

I feel `parametrization` is quite long. What about `variational`? or just `var`?

ixfoduap: `var` is probably not good as it resembles variable. `variational` is an option but still long. How about `param`?

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,363 @@
...
+            photons
+
+    Returns:
+        array: the rescaled adjacency matrix
+    """
+    scale = rescale_tor(A, n_mean) if threshold else rescale(A, n_mean)
+    return A * scale
+
+
+def A_to_cov(A: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
ixfoduap: Ok, leaving as is sounds good. Just wanted to make sure we were aware of existing functions

strawberryfields/apps/train/parametrization.py
ixfoduap: Let's keep VGBS

tests/apps/test_train.py
```diff
@@ -0,0 +1,213 @@
...
+@pytest.mark.parametrize("n_mean", [4, 6])
+class TestVGBS:
+    """Tests for the class ``train.VGBS``"""
+
+    def test_W(self, adj, n_mean, params):
+        """Test that the W method correctly gives the diagonal matrix of embedded parameters"""
+        gbs = train.VGBS(adj, n_mean, embedding, True)
+        W = gbs.W(params)
+        assert np.allclose(np.diag(W), embedding(params))  # check that diagonal equals embedding
+        assert np.allclose(np.diag(np.diag(W)), W)  # check that matrix is diagonal
```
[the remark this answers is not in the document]
ixfoduap: `W` is literally using `np.diag()` and I can't imagine a situation where we do that differently, so I would prefer to remove this test. No big deal of course if we want to keep

strawberryfields/apps/train/parametrization.py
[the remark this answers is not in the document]
nquesada: Yup, I'd think this function makes more sense in TW.

strawberryfields/apps/train/parametrization.py
trbromley: Ok, I like it!

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+def rescale_adjacency(A: np.ndarray, n_mean: float, threshold: bool) -> np.ndarray:
+    """Rescale an adjacency matrix so that it can be mapped to GBS.
+
+    An adjacency matrix must have singular values not exceeding one if it can be mapped to GBS.
+    Arbitrary adjacency matrices must first be rescaled to satisfy this condition.
+
+    This function rescales an input adjacency matrix :math:`A` so that the corresponding gaussian
+    state has:
```
antalszava, line 39: Would it be worth mentioning that this involves an optimization task ([which might potentially fail](https://github.com/XanaduAI/thewalrus/blob/cf7a19e71f6df5e8017bc21625fa3fb8af6cc5c8/thewalrus/quantum.py#L661))?

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+    A_big = np.block([[A, 0 * A], [0 * A, np.conj(A)]])
+    X = Xmat(n)
+    return X @ A_big
+
+
+def prob_click(A: np.ndarray, sample: np.ndarray):
+    """Calculate the probability of a click pattern.
+
+    The input adjacency matrix must have singular values not exceeding one. This can be achieved
+    by rescaling an arbitrary adjacency matrix using :func:`rescale_adjacency`.
```
antalszava, line 84: If the adjacency matrix is not rescaled, how would this function behave? Would it be worth handling that case within (e.g. with an error)?

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+    """Find the :math:`O` matrix of an adjacency matrix.
+
+    Args:
+        A (array): the adjacency matrix
+
+    Returns:
+        array: the :math:`O` matrix of :math:`A`
+    """
+    n = len(A)
+    A_big = np.block([[A, 0 * A], [0 * A, np.conj(A)]])
```
antalszava, line 75: (Minor suggestion): would this be better for readability? (let me know if I missed something)
```suggestion
    A_big = np.block([[A, np.zeros_like(A)], [np.zeros_like(A), np.conj(A)]])
```

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+    X = Xmat(n)
+    return X @ A_big
```
antalszava, line 77: Perhaps I'm missing something here, would this not mean `np.block([[0 * A, np.conj(A)], [A, 0 * A]])`?

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Returns:
+        float: the probability
+    """
+    n = len(A)
+    O = _Omat(A)
+    sample_big = np.hstack([sample, sample]).astype("int")
+    O_sub = reduction(O, sample_big)
+    scale = np.sqrt(np.linalg.det(np.identity(2 * n) - O))
+    return scale * tor(O_sub)
```
antalszava, line 98: Could be useful to include something similar to [this one](https://the-walrus.readthedocs.io/en/latest/gbs.html#gaussian-states-and-threshold-detection) at a later stage as part of an explanation here

nquesada:
@antalszava : I think that error message in The Walrus is unnecesary. One can prove that the root one is looking for a) exists and b) will be found using something like Newton-Raphson. That error message has never been displayed.

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+    stochastic optimization and unsupervised machine learning problems.
+
+    The above variational model can be used in both the threshold and PNR modes of GBS, which
+    is specified using the ``threshold`` flag. An initial value for the mean number of clicks
+    (if ``threshold=True``) or mean number of photons (if ``threshold=False``) is also required.
+    The initial value ``n_mean`` should be set high since varying :math:`\theta` can only lower
+    the mean number of clicks or photons.
+
+    The mapping from :math:`\theta` to :math:`W(\theta)` is specified using the ``embedding``
+    argument. TODO mention embedding module
```
antalszava, line 165: Small TODO left here, might be left for later stage

nquesada:
> Perhaps I'm missing something here, would this not mean np.block([[0 * A, np.conj(A)], [A, 0 * A]])?

This is because in principle the matrix A could be complex.

tests/apps/test_train.py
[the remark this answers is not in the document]
trbromley: Ok, removed this line!

strawberryfields/apps/train/parametrization.py
trbromley: Thanks! Have removed the TODOs!

strawberryfields/apps/train/parametrization.py
trbromley: Thanks! Good point, we'll revisit the docstrings again once all the features are in.

strawberryfields/apps/train/parametrization.py
trbromley: > @antalszava : I think that error message in The Walrus is unnecesary. One can prove that the root one is looking for a) exists and b) will be found using something like Newton-Raphson. That error message has never been displayed.

Adding this quote here from @nquesada - I think it's ok here not to mention it.

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+
+    **Example usage:**
+
+    >>> g = nx.erdos_renyi_graph(4, 0.7, seed=1967)
+    >>> A = nx.to_numpy_array(g)
+    >>> embedding = Exp(4)  # TODO does this need updating?
```
antalszava, line 171: as per the comment, will need an update alike to `embedding = ExpFeatures([[1,2,3,4]])` (`[[1,2,3,4]]` is just example features)

strawberryfields/apps/train/parametrization.py
trbromley: Awesome, have simplified the function but still kept as a separate function for now.

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+
+        Args:
+            params (array): the trainable parameters :math:`\theta`
+
+        Returns:
+            array: the diagonal matrix of weights
+        """
+        return np.sqrt(np.diag(self.embedding(params)))
+
+    def A(self, params: np.ndarray) -> np.ndarray:
```
antalszava, line 232: (Minor suggestion): wondering here: could the `WAW` name work? :thinking: as later used, could `A(params)` lead to some confusion when seen out of context?

strawberryfields/apps/train/parametrization.py
trbromley: Done

strawberryfields/apps/train/parametrization.py
```diff
@@ -0,0 +1,406 @@
...
+        Args:
+            n_samples (int): number of samples to get
+
+        Returns:
+            array: samples from the initial adjacency matrix
+        """
+        try:
+            current_n_samples = self.A_init_samples.shape[0]
+        except AttributeError:
+            current_n_samples = 0
```
antalszava, line 322: Curious, which cases does this cover?
a) `self.A_init_samples is None`
b) `not isinstance(self.A_init_samples, np.ndarray)`
c) there or something else too

If it's either a) or b) perhaps it's good to go with the explicit case. If it's c), then curious what I've missed :smile:

strawberryfields/apps/train/parametrization.py
trbromley: Thanks! I'm on the fence for whether this should be a private function, we don't expect it to be used much directly by the user :thinking:

strawberryfields/apps/train/parametrization.py
trbromley: I updated to `>>> embedding = train.embed.Exp(4)` as it's simpler to go with the `Exp` embedding in general. Do you think `train.embed` is the best way to point to it though?

strawberryfields/apps/train/parametrization.py
trbromley: That is a very good question! We've discussed this a bit and are so far sticking with the idea that we should not overburden users with more technical namings, and in the end will present it as `A(\theta)`, simply as a matrix that is a function of the params (which is the main thing a user needs to know).

strawberryfields/apps/train/parametrization.py
trbromley: Ohh yes this is a good question for code quality.

So yeah generally I expect `A_init_samples` to be either `None` or `np.ndarray`. I want to check the shape, and of course if `A_init_samples = None` then assume there are no samples present.

One options is `isinstance(A_init_samples, np.ndarray)`, but this requires `A_init_samples` to be exactly an `np.ndarray`.

Another option I guess is to check if `A_init_samples` is None :thinking: 

But then I thought to follow the EAFP principle and just try and query the shape.

So I guess this option allows for the rare occasion that a user puts in something like a numpy array but not exactly (e.g. if they subclass it).

But interested to hear your thoughts!

tests/apps/test_train.py
antalszava: (Minor suggestion): How about creating two small test cases by splitting this one up for further separation? `test_generate_samples_torontonian`, `test_generate_samples_hafnian`

PR #389, opened 2020-05-09
Added Parity Expectation

Lucaman99:
**Context:**

Implements changes outlined in Strawberry Fields such that https://github.com/XanaduAI/pennylane/issues/193 can be eventually resolved.

Added a `parity_expectation` function, which takes the index of one qumode as an argument and outputs <(-1)^N> for the given mode (where N is the number operator). Currently is implemented in the `BaseFockState` class (not the `BaseGaussianState` class yet). It also only currently supports pure states and only has one "basic" unit test associated with it. 

I was hoping to get a bit of feedback (to make sure I'm not making any big mistakes) before I proceed, hence the WIP pull request. 🙂 

**Benefits:**

Benefits are outlined in https://github.com/XanaduAI/pennylane/issues/193.

**Possible Drawbacks:**

None that I can think of.

**Related GitHub Issues:**

https://github.com/XanaduAI/pennylane/issues/193


strawberryfields/backends/states.py
```diff
@@ -914,6 +914,26 @@ def number_expectation(self, modes):
...
+        if (type(mode) != int):
+            raise ValueError("Specified mode index must be an integer.")
+
+        cutoff = self._cutoff  # Fock space cutoff.
+        num_modes = self._modes # number of modes in the state.
+        values = np.arange(cutoff)
+
+        print(self.is_pure)
+
+        if self.is_pure:
```
nquesada, line 928: @Lucaman99 this would probably make a better addition if you also allow for mixed state for which you can also get the Fock probabilities.

Lucaman99:
@nquesada I just realized that`.all_fock_probs` works for mixed states as well (for some reason I thought it didn't), so it should support mixed states now!

strawberryfields/backends/states.py
```diff
@@ -25,11 +25,11 @@
 from scipy.stats import multivariate_normal
 from scipy.special import factorial
 from scipy.integrate import simps
-from thewalrus.quantum import photon_number_mean, photon_number_covar
+#from thewalrus.quantum import photon_number_mean, photon_number_covar
```
josh146, line 28: @Lucaman99, how come these are commented out?

strawberryfields/backends/states.py
```diff
@@ -914,6 +914,23 @@ def number_expectation(self, modes):
...
+        values = np.arange(cutoff)
+
+        ps = self.all_fock_probs()
+        sum_axes = list(range(num_modes))
+        del sum_axes[mode]
+        ps = np.sum(ps, axis=tuple(sum_axes))
+        vals = [(-1)**i for i in values]
+        ev = np.dot(ps, vals)
+        
+        return float(ev)
```
josh146, line 933: This is looking good! @nquesada has some ideas on how to extend and vectorize this :)

strawberryfields/backends/states.py
Lucaman99: Yep, sorry, I meant to un-comment these! For some reason when I tried to test the code, these imports weren't working, but it's probably a problem with my local copy of my fork.

josh146: 🤔 You may have to upgrade your version of the walrus: `pip install thewalrus --upgrade`

Lucaman99: Thanks, I'll try updating it 🙂

nquesada:
Hey @Lucaman99, adding expectations of parity operators is a great idea. I would like to suggest the following: Instead of implementing it from scratch we can re-purpose code already in SF.
In this file https://github.com/XanaduAI/strawberryfields/blob/master/strawberryfields/backends/states.py we define the function `number_expectation`. If you have a look at this function you quickly realize that if in line https://github.com/XanaduAI/strawberryfields/blob/1698b95902a149fced9f3a19783a3f9e8019185f/strawberryfields/backends/states.py#L897 you change the value of the parameter `values` to be `(-1)**np.arange(cutoff)` you can use the function to calculate the expectation values of the product of the parity operator over as many modes as you wish, with the advantage that this function already works for Numpy and TensorFlow.

This I'd suggest the following 

1. Take the logic of the function `number_expectation` and put it in a new function called `diagonal_expectation` that takes `values` as a parameter. This function will be useful to calculate expectation values of *any* operator that is diagonal in the number basis.
2. Have `number_expectation` call `diagonal_expectation` with `values = np.arange(cutoff)`.
3. Create a new function called `parity_expectation` that calls `diagonal_expectation` with `values = (-1)**np.arange(cutoff).
4. Write tests, similar to the ones here https://github.com/XanaduAI/strawberryfields/blob/1698b95902a149fced9f3a19783a3f9e8019185f/tests/backend/test_states.py#L85

Finally, I think one can also write in a reasonably simple form the expectation of the parity for a Gaussian state. Maybe after you do it for the Fock representation I can add the parity expectation for Gaussian states.

Hope this makes sense, and let me know if you have questions!

nquesada:
Quick update: For Gaussian states the expectation value of the parity is just the value of the Wigner function at the origin,  you just need to obtain the vector of means `r`  and the covariance matrix `V` of the modes that you are interested in and calculate `np.exp(-0.5*r@V@r)/np.linalg.det(V)` and you will get what you want.

Lucaman99:
@nquesada Thanks for the suggestions. This all makes sense, I'll get to work on implementing these changes right away!

Lucaman99:
@nquesada In addition to the parity operator, do you think it would make sense to add some other functions that return expectation values of other operators that are diagonal in the Fock basis (for instance, a specified order-two polynomial of the form aN^2 + bN + c, with N being the number operator) by utilizing the functionality of the `diagonal_expectation`?

Lucaman99:
Also, relating to this, do you think it would be better to take a function that acts on `np.arrange(cutoff)` rather than an array of values as an argument so that it might be more intuitive for a user to use `diagonal_expectation` to calculate the expectation of some operator of the form f(N)?

nquesada:
Great! @Lucaman99 ,that is an interesting design question, I'd be tempted to keep the values, but maybe there are reasons to make it "functional".

Lucaman99:
@nquesada Yeah, I'm definitely no expert at code design, so if you think it's best to keep it as `values` then let's do that!

Lucaman99:
@nquesada Just made the changes to the Fock backend version, will add tests and Gaussian version tomorrow!

strawberryfields/backends/states.py
```diff
@@ -914,6 +914,38 @@ def number_expectation(self, modes):
...
+        ps = np.real(self.dm())
+        traced_modes = list(traced_modes)
+        traced_modes.sort(reverse=True)
+        for mode in traced_modes:
+            ps = np.tensordot(np.identity(cutoff), ps, axes=((0, 1), (2 * mode, 2 * mode + 1)))
+        for _ in range(len(modes)):
+            ps = np.tensordot(np.diag(values), ps, axes=((0, 1), (0, 1)))
+        return float(ps)
+
+    def parity_expectation(self, modes):
```
nquesada, line 943: This looks good. You will also need to write a function for the number expectation, with a very similar logic to this one.

strawberryfields/backends/states.py
```diff
@@ -914,6 +914,38 @@ def number_expectation(self, modes):
...
+
+        cutoff = self._cutoff
+        values = [(-1)**i for i in np.arange(cutoff)]
```
nquesada, line 946:
```suggestion
        values = (-1)**np.arange(cutoff)
```

strawberryfields/backends/states.py
Lucaman99: Does that mean you want me to delete the old `number_expectation` method?

nquesada: Yes. As far as I can seen you move the functionality into `diagonal_expectation`, now you need to write a wrapper `number_expectation` so that we still have access to that function.

Lucaman99:
@nquesada Just pushed changes the re-define `number_expectation`. Working on the tests right now. Also, I added a function that calculates "quadratic expectation" (EV of an operator of the form aN^2 + bN + c), as per my comment from a couple of days ago. If you don't think it is useful, let me know and I'll delete it. 👍

nquesada:
Here are a few tests you could add:

1. For a coherent state with parameter \alpha and defining its mean photon number as m = |\alpha|^2 the expectation of the parity is \exp(-2m)
2. For a single mode squeezed vacuum the parity is always 1.
3. For two-mode squeezed vacuum the expectation of the product of the two parities is also 1.

nquesada:
Also, for a thermal state with mean photon number m, the parity is 1/(2m+1)

Lucaman99:
@nquesada Awesome thanks

Lucaman99:
@nquesada Done with the tests, let me know what you think.

nquesada:
Before review Jack please make sure to remove these files:

image

and, then solve any issues detected by code factor

image

Lucaman99:
@nquesada Will do 👍

Lucaman99:
@nquesada Ok, CodeFactor seems to be working, I don't really know why the Formatting check is failing.

nquesada:
You need to check the formatting in `states.py`:

image

Also try to update to the latest version of `black`

Lucaman99:
@nquesada Ok, I think I've finally figured out why the Travis tests are failing: the cutoff dimension of the states created by the backend in `test_states.py` is way too low to be compared to the exact values. Is there any way I can change this for the one particular class in which the parity tests are located?

Lucaman99:
Nevermind, just realized I was setting the parameters in the tests to absurdly high values (relative to the set cutoff).

josh146:
> Nevermind, just realized I was setting the parameters in the tests to absurdly high values (relative to the set cutoff).

Yep, it's a bit of a tradeoff 🙂 We want a low cutoff so that the tests complete in a meaningful time, which means we need to choose relatively small values of squeezing and displacement

tests/backend/test_states.py
```diff
@@ -377,11 +385,118 @@ def test_number_expectation_four_modes(self, setup_backend, tol, batch_size):
...
+    def test_coherent(self, setup_backend, tol):
+
+        backend = setup_backend(1)
+        state = backend.state()
+        alpha = 0.2
+        backend.prepare_coherent_state(alpha, 0)
+        state = backend.state()
+
+        assert np.allclose(
+            state.parity_expectation([0]), np.exp(-2 * (alpha**2)), atol=tol, rtol=0
```
nquesada, line 444:
```suggestion
            state.parity_expectation([0]), np.exp(-2 * (np.abs(alpha)**2)), atol=tol, rtol=0
```

strawberryfields/backends/states.py
```diff
@@ -914,6 +914,25 @@ def number_expectation(self, modes):
...
+        values = np.arange(cutoff)
+        return self.diagonal_expectation(modes, values)
+
+    def parity_expectation(self, modes):
+        """Calculates the expectation value of a product of parity operators acting on given modes"""
+        cutoff = self._cutoff
+        values = (-1) ** np.arange(cutoff)
+        return self.diagonal_expectation(modes, values)
+
+    def quadratic_expectation(self, modes, a, b, c):
```
nquesada, line 929: I am not sure this function has a very strong use case.

Lucaman99: Yeah no problem, I'll remove it

nquesada:
Hi @Lucaman99 , almost there! The test are failing because Travis is trying to run them in the Gaussian backend for which we have not implemented parity. You need to mark the tests you wrote for the parity so that they are only run in Fock and TF backends, like it is  currently done for number operator. Also, it might be simpler if you run pytest on the test file you have locally, that way you can see how the test are failing before having to wait for Travis.

nquesada:
@Lucaman99 you also need to update the CHANGELOG adding an entry for this new functionality and your name to the author list.

Lucaman99:
@nquesada Alright, done! Travis worked for the most recent changes to the tests file, so all the tests should pass!

.github/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 <h3>New features since last release</h3>
 
+* Adds `diaongal_expectation` method for the `BaseFockState` class, which returns
```
nquesada, line 5:
```suggestion
* Adds `diagonal_expectation` method for the `BaseFockState` class, which returns
```

.github/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
+  the expectation value of any operator that is diagonal in the number basis.
+  [(#389)](https://github.com/XanaduAI/strawberryfields/pull/389)
+
+* Adds `parity_expectation` method as an instance of `diaongal_expectation` for
```
nquesada, line 9:
```suggestion
* Adds `parity_expectation` method as an instance of `diagonal_expectation` for 
```

tests/backend/test_states.py
```diff
@@ -377,11 +385,112 @@ def test_number_expectation_four_modes(self, setup_backend, tol, batch_size):
...
+            nbar * (2 * nbar ** 2 + nbar),
+            atol=tol,
+            rtol=0,
+        )
+
+
+class TestParityExpectation:
+
+    @pytest.mark.backends("fock", "tf")
+    def test_parity_fock(self, setup_backend, tol, batch_size):
```
antalszava, line 410: Hi @Lucaman99, thanks so much for this PR! :blush: just leaving a small suggestion: could you add a one-line description as docstring for each test case?

Lucaman99: Will do!

Lucaman99:
@antalszava Done!

Lucaman99:
@antalszava Done!

PR #390, opened 2020-05-14
Updating the similarity module functions in the Apps layer

AroosaIjaz:

This PR redesigns the similarity module such that :
- feature vector functions based on direct sampling are now provided for both orbits and events

- one function is used for all other methods of obtaining feature vectors; Monte Carlo and exact calculation. The number of MC samples is used as the keyword argument for this purpose. This function also comes in both orbits and events "flavor".

- smaller functions used in the functions mentioned above are turned into internal functions.

- All functions have unit tests

- corresponding landing page for the module is updated

- corresponding tutorial on graph similarity is updated

**Benefits:**
- flexibility of using a wider range of methods when making feature vectors

- flexibility of using different constituent units (events or orbits) when making feature vectors

- compact functions instead of many different smaller functions

**Possible Drawbacks:**
NA
**Related GitHub Issues:**
NA


strawberryfields/apps/similarity.py
```diff
@@ -306,33 +320,148 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
 
 
-def prob_orbit_mc(
-    graph: nx.Graph, orbit: list, n_mean: float = 5, samples: int = 1000, loss: float = 0.0
+def _prob_orbit_exact(
+    graph: nx.Graph,
+    orbit: list,
+    n_mean: float = 5,
+    loss: float = 0.0
 ) -> float:
```
trbromley, line 328: When running `black`, are you doing it with a line width of `100`?

`black -l 100 strawberryfields/apps/similarity.py` should be the right command.

strawberryfields/apps/similarity.py
```diff
@@ -306,33 +320,148 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+       P(O) = \sum_{n \in O} P(n) = \frac{{\rm haf}(A_{n})^2}{n! \sqrt{\det(Q)}}
+
+    where :math:`n` represents one GBS output pattern, :math:`A_{n}` is its induced
+    sub-matrix and :math:`Q` is the Q-matrix obtained from the adjacency matrix.
+
+    **Example usage:**
+
+    >>> graph = nx.complete_graph(8)
+    >>> _prob_orbit_exact(graph, [2, 1, 1])
+    0.03744
```
trbromley, line 345: Note that since this is a private function we probably don't need to do example code blocks.
But I take the view that it can only help, so :+1: from me!

strawberryfields/apps/similarity.py
```diff
@@ -306,33 +320,148 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+
+    with p.context as q:
+        sf.ops.GraphEmbed(A, mean_photon_per_mode=mean_photon_per_mode) | q
+
+        if loss:
+            for _q in q:
+                sf.ops.LossChannel(1 - loss) | _q
+
+    eng = sf.LocalEngine(backend="gaussian")
+    result = eng.run(p)
```
trbromley, line 377: This sort of logic seems to crop up a few times now. I'd recommend factoring out to make a private function that encodes the problem into SF:
```python
def _get_state(graph, n_mean, loss):
    modes = graph.order()
    photons = sum(orbit)
    A = nx.to_numpy_array(graph)
    mean_photon_per_mode = n_mean / float(modes)
    
    p = sf.Program(modes)
    
    with p.context as q:
        sf.ops.GraphEmbed(A, mean_photon_per_mode=mean_photon_per_mode) | q
    
        if loss:
            for _q in q:
                sf.ops.LossChannel(1 - loss) | _q
    
    eng = sf.LocalEngine(backend="gaussian")
    return eng.run(p).state
```
This would help separate the two functionalities of (i) embedding into SF, (ii) using the SF state to calculate the probability.

strawberryfields/apps/similarity.py
```diff
@@ -306,33 +320,148 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+        if max(orbit) <= max_count_per_mode:
+            prob += _prob_orbit_exact(graph, orbit, n_mean, loss)
+    return prob
+
+
+def _prob_orbit_mc(
+    graph: nx.Graph,
+    orbit: list,
+    n_mean: float = 5,
+    mc_samples: int = 1000,
```
trbromley, line 438: What is the motivation for changing the `mc_samples` argument name (before it was `samples`)?

strawberryfields/apps/similarity.py
```diff
@@ -430,101 +559,205 @@ def prob_event_mc(
...
+        raise ValueError("Loss parameter must take a value between zero and one")
 
-    e = (sample_to_event(s, max_count_per_mode) for s in samples)
-    count = Counter(e)
+    if mc_samples == None:
+        return [_prob_orbit_exact(graph, orbit, n_mean, loss) for orbit in list_of_orbits]
 
-    return [count[p] / n_samples for p in event_photon_numbers]
+    else:
+        return [_prob_orbit_mc(graph, orbit, n_mean, mc_samples, loss) for orbit in list_of_orbits]
```
trbromley, line 628: You could have:
```python
    if mc_samples:
        return [_prob_orbit_mc(graph, orbit, n_mean, mc_samples, loss) for orbit in list_of_orbits]

    return [_prob_orbit_exact(graph, orbit, n_mean, loss) for orbit in list_of_orbits]
```
In this way, `mc_samples=None` and `mc_samples=0` will result in exact sampling, otherwise for `samples` being an `int` it will do MC.

strawberryfields/apps/similarity.py
```diff
@@ -359,21 +488,21 @@ def prob_orbit_mc(
...
         prob += result.state.fock_prob(sample, cutoff=photons + 1)
 
-    prob = prob * orbit_cardinality(orbit, modes) / samples
+    prob = prob * orbit_cardinality(orbit, modes) / mc_samples
 
     return prob
 
 
-def prob_event_mc(
+def _prob_event_mc(
```
trbromley, line 500: What is the motivation for making these `_prob_...` functions private? I think they can be left public (without the first underscore)

strawberryfields/apps/similarity.py
trbromley: Note that if we have `prob_orbit_mc` be public (i.e. without the first underscore), then changing the argument name from `samples` to `mc_samples` could break some user's code, e.g., if before they had:
`prob_orbit_mc(g, orb, samples=10)`

strawberryfields/apps/similarity.py
AroosaIjaz: Thanks. I was not running with this line width. I have implemented it now. :)

strawberryfields/apps/similarity.py
AroosaIjaz: I had the same thought :) It might help if someone looks into the source code.

strawberryfields/apps/similarity.py
AroosaIjaz: The motivation is that in the sampling-based feature vector functions, we use ``samples`` to represent the list of all pre-produced samples. I wanted to differentiate between the argument ``samples`` used in MC functions where it refers to the number of samples to be used in the approximation.

AroosaIjaz: I dont understand your second comment. Will there not be an error raised even if the function is public. ``sample`` is not a defined argument for ``prob_orbit_mc``.

strawberryfields/apps/similarity.py
AroosaIjaz: The idea is to provide only high-level functions that are compact and to avoid clutter. We decided this in the ADR but we can reconsider this. It is possible that the user has cases where it wants to calculate just a single orbit/event probability rather than the whole feature vector.

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
-    each mode but a range of total photon numbers specified by ``event_photon_numbers``.
+    These probabilities can be either exact or estimated using Monte Carlo estimation.
+    The argument ``mc_samples`` is set to ``None`` to get exact feature vector by default.
+    To use Monte Carlo estimation, ``mc_samples`` can be set to the number of samples desired
+    to be used in the estimation.
 
-    Probabilities are reconstructed by measuring the occurrence of events in the input ``samples``.
+    Computing exact probabilities for a large number of orbits especially for orbits with high
+    total photon numbers can be quite time-consuming. For example, calculating the exact
+    probabilities of observing 8 total photons in a 25 mode graph can take on the order of a
```
ixfoduap, line 559:
```suggestion
    probabilities of observing 8 total photons in a 25-mode graph can take on the order of a

```

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    all can take many days. Monte Carlo estimation, although less precise, can be much quicker.
 
     **Example usage:**
 
-    >>> from strawberryfields.apps import data
-    >>> samples = data.Mutag0()
-    >>> feature_vector_sampling(samples, [2, 4, 6])
-    [0.19035, 0.2047, 0.1539]
+    >>> graph = nx.complete_graph(8)
+    >>> feature_vector_orbits(graph, [ [1,1], [2], [2,1,1], [1,1,1,1] ])
```
ixfoduap, line 566: I suppose that the probability for the orbit `[2]` is 0 in this example, which come out awkwardly as ~1e-31. Can we choose another orbit with non-zero probability?

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    [0.18423136914879445,
+     7.843627302119571e-31,
+     0.03744399092424445,
+     0.07020748298295838]
+
+    >>> feature_vector_orbits(graph, [ [1,1], [2] , [2,1,1], [1,1,1,1] ], mc_samples = 1000)
```
ixfoduap, line 572: This is very surprising, do the MC estimates really agree so precisely with the exact calculations? I find it hard to believe. I can only spot differences in the 16-th decimal place. No way that's correct

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
     loss: float = 0.0,
 ) -> list:
-    r"""Calculates feature vector using Monte Carlo estimation of event probabilities according to the
-    input graph.
+    r"""Calculates feature vector made up of either exact or MC-approximated event probabilities for the input graph.
 
-    The feature vector is composed of event probabilities with a fixed maximum photon count in
-    each mode but a range of total photon numbers specified by ``event_photon_numbers``.
+    These probabilities can be either exact or estimated using Monte Carlo estimation.
+    The argument ``mc_samples`` is set to ``None`` to get exact feature vector by default.
```
ixfoduap, line 616:
```suggestion
    The argument ``mc_samples`` is set to ``None`` to get exact feature vectors by default.

```

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    To use Monte Carlo estimation, ``mc_samples`` can be set to the number of samples desired
+    to be used in the estimation.
 
-    Probabilities are reconstructed using Monte Carlo estimation.
+    Computing exact probabilities for a large number of events especially for events with high
+    total photon numbers can be quite time-consuming. For example, calculating the exact
+    probabilities of observing 8 total photons in a 25 mode graph can take on the order of a
+    few minutes. This means that if the dataset has 1000 graphs, obtaining feature vectors for
```
ixfoduap, line 623: Same as before, remove sentence "This means..."

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+        raise ValueError("Cannot request events with photon number below zero")
+    if max_count_per_mode < 0:
+        raise ValueError("Maximum number of photons per mode must be non-negative")
     if n_mean < 0:
         raise ValueError("Mean photon number must be non-negative")
     if not 0 <= loss <= 1:
         raise ValueError("Loss parameter must take a value between zero and one")
+
+    if mc_samples:
+        return [
```
ixfoduap, line 660: Check formatting here and below; the return shouldn't span so many lines

strawberryfields/apps/similarity.py
AroosaIjaz: okay I will make them all public.

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
     event_photon_numbers: list,
     max_count_per_mode: int = 2,
     n_mean: float = 5,
-    samples: int = 1000,
+    mc_samples: int = None,
     loss: float = 0.0,
 ) -> list:
-    r"""Calculates feature vector using Monte Carlo estimation of event probabilities according to the
-    input graph.
+    r"""Calculates feature vector made up of either exact or MC-approximated event probabilities for the input graph.
```
trbromley, line 613: I'd actually not mention exact vs MC in the one liner intro sentence, it can be mentioned in the main paragraph and this kept short and sweet.

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
-    samples: list, event_photon_numbers: list, max_count_per_mode: int = 2
+def feature_vector_orbits(
+    graph: nx.Graph,
+    list_of_orbits: list,
+    n_mean: float = 5,
+    mc_samples: int = None,
+    loss: float = 0.0,
 ) -> list:
-    r"""Calculates feature vector with respect to input samples.
+    r"""Calculates feature vector made up of either exact or MC-approximated orbit probabilities for the input graph.
```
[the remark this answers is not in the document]
trbromley: Alternatively, you can omit mentioning exact vs MC here and instead do it in the main part.

strawberryfields/apps/similarity.py
trbromley: (similar to a previous comment, where @ixfoduap suggested not to abbreviate MC)

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
 
-    The feature vector is composed of event probabilities with a fixed maximum photon count in
-    each mode but a range of total photon numbers specified by ``event_photon_numbers``.
+    These probabilities can be either exact or estimated using Monte Carlo estimation.
+    The argument ``mc_samples`` is set to ``None`` to get exact feature vector by default.
+    To use Monte Carlo estimation, ``mc_samples`` can be set to the number of samples desired
+    to be used in the estimation.
 
-    Probabilities are reconstructed using Monte Carlo estimation.
+    Computing exact probabilities for a large number of events especially for events with high
```
trbromley, line 620: This can also be in a warning box

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    probabilities of observing 8 total photons in a 25 mode graph can take on the order of a
+    few minutes. This means that if the dataset has 1000 graphs, obtaining feature vectors for
+    all can take many days. Monte Carlo estimation, although less precise, can be much quicker.
 
     **Example usage:**
 
     >>> graph = nx.complete_graph(8)
-    >>> feature_vector_mc(graph, [2, 4, 6], 2)
-    [0.2115, 0.1457, 0.09085]
+    >>> feature_vector_events(graph, [2,3,4,6], 1)
```
trbromley, line 629:
```suggestion
    >>> feature_vector_events(graph, [2, 3, 4, 6], 1)
```

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    [0.18423136914879445, 0.0, 0.07020748298295838, 0.011891063791931982]
+
+    >>> feature_vector_events(graph, [2,3,4,6], 1, mc_samples = 1000)
```
trbromley, line 632:
```suggestion
    >>> feature_vector_events(graph, [2, 3, 4, 6], 1, mc_samples = 1000)
```

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    [0.18423136914879246, 0.0, 0.07020748298295802, 0.011891063791932176]
```
trbromley, line 633: Similar to a comment before about the mc and non mc being so similar, is it true?

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
 
     Args:
         graph (nx.Graph): input graph
-        event_photon_numbers (list[int]): a list of events described by their total photon number
-        max_count_per_mode (int): maximum number of photons per mode for all events
+        event_photon_number (int): number of photons in the event
```
trbromley, line 637: Why is this line changed from before? I.e.., shouldn't it be `event_photon_numbers (list[int]): a list of events described by their total photon number`

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+        max_count_per_mode (int): maximum number of photons per mode in the event
```
trbromley, line 638: Also this line, did it need to change?

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
 
     Returns:
-        list[float]: a feature vector of event probabilities in the same order as
-        ``event_photon_numbers``
+        list[float]: a feature vector of orbit probabilities
     """
-    if samples < 1:
-        raise ValueError("Number of samples must be at least one")
+    if type(event_photon_numbers) != list:
+        raise ValueError("Event photon numbers must be given as a list of integers")
```
trbromley, line 647: This is too strong, `event_photon_numbers` could be a tuple and this would fail.

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+        samples (list[list[int]]): a list of samples
+        event_photon_numbers (list[int]): a list of events described by their total photon number
+        max_count_per_mode (int): maximum number of photons per mode for all events
+
+    Returns:
+        list[float]: a feature vector made up of estimated event probabilities in the
+        same order as ``event_photon_numbers``
+    """
+    if type(event_photon_numbers) != list:
+        raise ValueError("Event photon numbers must be given as a list of integers")
```
trbromley, line 731: This check is too strong, recommend to remove.

Also, you can use `isinstance(event_photon_numbers, list)`

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    r"""Calculates feature vector made of given orbits with respect to input samples.
+
+    The feature vector is composed of orbit probabilities reconstructed by measuring the
+    occurrence of given orbits in the input ``samples``.
+
+    **Example usage:**
+
+    >>> from strawberryfields.apps import data
+    >>> samples = data.Mutag0()
+    >>> feature_vector_orbits_sampling(samples, [ [1,1], [2], [2,1,1], [1,1,1,1] ])
```
trbromley, line 681:
```suggestion
    >>> feature_vector_orbits_sampling(samples, [[1,1], [2], [2,1,1], [1,1,1,1]])

```

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
-    if max_count_per_mode < 0:
-        raise ValueError("Maximum number of photons per mode must be non-negative")
-
-    n_samples = len(samples)
-
-    e = (sample_to_event(s, max_count_per_mode) for s in samples)
-    count = Counter(e)
+    if len(list_of_orbits) <= 0:
+        raise ValueError("List of orbits must have at least one orbit")
+    if all(isinstance(elem, list) for elem in list_of_orbits) == False:
```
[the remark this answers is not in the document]
trbromley: Actually, now I think about it this check might also be too strong, e.g., if someone has a tuple of tuples. I think we can just leave this out.

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+        samples (list[list[int]]): a list of samples
+        list_of_orbits (list[list[int]]): a list of orbits
+
+    Returns:
+        list[float]: a feature vector made up of estimated orbit probabilities
+    """
+    if len(list_of_orbits) <= 0:
+        raise ValueError("List of orbits must have at least one orbit")
+    if all(isinstance(elem, list) for elem in list_of_orbits) == False:
+        raise ValueError("List of orbits should only consist of orbits")
```
trbromley, line 694: Recommend to remove this check

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
     assert similarity.event_cardinality(photons, max_count, modes) == expected
 
 
+class TestGetState:
+    """Tests for the function ``strawberryfields.apps.similarity._get_state.``"""
```
trbromley, line 227:
```suggestion
    """Tests for the function ``strawberryfields.apps.similarity._get_state``."""
```

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+            similarity._get_state(graph)
+            p_func = mock_eng_run.call_args[0][0]
+
+        assert not all([isinstance(op, sf.ops.LossChannel) for op in p_func.circuit])
+
+    def test_max_loss(self):
+        """Test if function samples from the vacuum when maximum loss is applied."""
+        dim = 5
+        graph = nx.complete_graph(dim)
+        State = similarity._get_state(graph, 0, 1)
```
trbromley, line 257: Also `State` to `state` here

Classes should be capitalized, but this is an _instance_ of a class.

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
     assert similarity.event_cardinality(photons, max_count, modes) == expected
 
 
+class TestGetState:
```
trbromley, line 226: Nice tests!

tests/apps/test_similarity.py
trbromley: Actually this seems to happen a lot in this module, even before this PR, so not a big deal :shrug:

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+            similarity._prob_orbit_exact(g, [1, 1, 1, 1], loss=2)
+
+    def test_prob_vacuum_orbit(self):
+        """Tests if the function gives the right probability for the empty orbit when the GBS
+        device has been configured to have zero mean photon number."""
+        graph = nx.complete_graph(5)
+        assert similarity._prob_orbit_exact(graph, [], 0) == 1.0
+
+    @pytest.mark.parametrize("k", [1, 3, 5, 7, 9])
+    def test_odd_photon_numbers(self, k):
```
trbromley, line 291: Nice!

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,106 @@
...
-# 2. Using a Monte Carlo estimate of the probability.
+# 2. Using exact probability calculations.
+# 3. Using a Monte Carlo estimate of the probability.
 #
 # In the first method, all one needs to do is generate some GBS samples from the graph of
 # interest and fix the composition of the feature vector. For example, for a feature vector
 # :math:`f_{\mathbf{k} = (2, 4, 6), n_{\max}=2}` we use:
 
-print(similarity.feature_vector_sampling(m0, event_photon_numbers=[2, 4, 6], max_count_per_mode=2))
+print(
```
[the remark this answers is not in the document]
AroosaIjaz: Yes, running with black renders the text in this fashion. I will change it back.

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
 
     def test_invalid_loss(self):
         """Test if function raises a ``ValueError`` when the loss parameter is specified outside
         of range."""
-        g = nx.complete_graph(10)
-        with pytest.raises(ValueError, match="Loss parameter must take a value between zero and"):
-            similarity.prob_orbit_mc(g, [1, 1, 1, 1], loss=2)
+        g = nx.complete_graph(5)
+        with pytest.raises(
+            ValueError, match="Loss parameter must take a value between zero and one"
```
trbromley, line 366: Note that the match doesn't have to be exact, it can be just the first few words - it should be enough to uniquely identify the error.

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,106 @@
...
+# .. math::
+#     p(O) = \sum_{n \in O} p(n) = \frac{{\rm haf}(A_{n})^2}{n! \sqrt{\det(Q)}}
+#
+# where :math:`n` represents one GBS output click pattern, :math:`A_{n}` is its induced sub-matrix
+# and :math:`Q` is the :math:`Q`-matrix obtained from the adjacency matrix of the graph. Exact probability of
+# an event :math:`p_{k,n_{\max}}` can be calculated similarly or by summing over its constituent orbit
+# probabilities.
+#
+# Strawberry Fields provides built-in methods to get exact feature vectors of any desired
+# composition with :func:`~.feature_vector_orbits` and :func:`~.feature_vector_events`. These functions
```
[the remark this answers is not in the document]
AroosaIjaz: composed of either events or orbits. can try and reword this sentence if it is not clear

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,106 @@
...
+# use a keyword argument ``mc_samples`` to signal producing either exact or approximate probabilities.
+# ``mc_samples`` is set to ``None`` to get exact feature vector by default. To use Monte Carlo estimation,
+# ``mc_samples`` can be set to the number of samples desired to be used in the estimation as shown later.
+# For example, to get exact feature vector :math:`f_{\mathbf{k} = (2, 4, 6), n_{\max}=2}`, we use:
```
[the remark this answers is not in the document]
AroosaIjaz: yes, I use the same feature vector example in all three sampling, exact and MC code snippets for easier learning.

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,106 @@
...
+##############################################################################
+# For the second method, suppose we want to calculate the orbit probabilities exactly rather than
+# through sampling. This can be useful as it does not require the pre-generation of GBS samples for
+# a given graph which can be quite time consuming. Considering a feature vector of orbit probabilities,
+# the probability of a single orbit :math:`p(O)` is composed of the sum over probabilities of all possible
+# GBS sample permutations that belong to it. In GBS, the probability of any GBS output click pattern
+# is determined by the hafnian of a relevant sub-adjacency matrix. Hence,
+#
+# .. math::
+#     p(O) = \sum_{n \in O} p(n) = \frac{{\rm haf}(A_{n})^2}{n! \sqrt{\det(Q)}}
```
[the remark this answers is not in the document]
AroosaIjaz: ok. I removed it and added a link to the walrus page on ``hafnian`` instead.

strawberryfields/apps/similarity.py
```diff
@@ -306,74 +321,174 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+
+    where :math:`n` represents one GBS output pattern, :math:`A_{n}` is its induced
+    sub-matrix and :math:`Q` is the Q-matrix obtained from the adjacency matrix.
 
     **Example usage:**
 
     >>> graph = nx.complete_graph(8)
-    >>> prob_orbit_mc(graph, [2, 1, 1])
+    >>> _prob_orbit_exact(graph, [2, 1, 1])
     0.03744
```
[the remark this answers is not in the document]
AroosaIjaz: Yes, I checked. The output differs in the 8th decimal place or so. Should we give the whole number then?

strawberryfields/apps/similarity.py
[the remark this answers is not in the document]
AroosaIjaz: I will mention it at the start as well. But in case, someone directly goes to the function without reading the module introduction, I think I will keep it in the docstring too.

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
+    mc_samples: int = None,
+    loss: float = 0.0,
 ) -> list:
-    r"""Calculates feature vector with respect to input samples.
+    r"""Calculates feature vector made up of either exact or MC-approximated orbit probabilities for the input graph.
 
-    The feature vector is composed of event probabilities with a fixed maximum photon count in
-    each mode but a range of total photon numbers specified by ``event_photon_numbers``.
+    These probabilities can be either exact or estimated using Monte Carlo estimation.
+    The argument ``mc_samples`` is set to ``None`` to get exact feature vector by default.
```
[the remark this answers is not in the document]
AroosaIjaz: I don't agree with this one, JM. The function returns a single feature vector at a time.

strawberryfields/apps/similarity.py
AroosaIjaz: I updated this now.

strawberryfields/apps/similarity.py
AroosaIjaz: I checked the code of ``prob_orbit_exact`` again. I am not sure where the error could be.

examples_apps/run_tutorial_similarity.py
```diff
@@ -262,9 +311,10 @@
 print(R)
 
 ##############################################################################
-# There is freedom in the choice of ``events`` composing the feature vectors and we encourage the
-# reader to explore different combinations. Note, however, that odd photon-numbered events have
-# zero probability because ideal GBS only generates and outputs pairs of photons.
+# The choice of what ``events`` to use for the feature vectors can be significant and we encourage the
+# reader to explore different combinations. We can also use any orbits of our choice above instead of events.
+# Note, however, that GBS samples with odd total number of photons have zero probability because ideal
+# GBS only generates and outputs pairs of photons.
```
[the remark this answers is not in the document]
AroosaIjaz: Yes, it builds and the docs were running smoothly on my side. It takes a minute or so to build all print calls and the last part with the SVM.

strawberryfields/apps/similarity.py
```diff
@@ -306,74 +321,174 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
 
-    for _ in range(samples):
+    for orbit in orbits(photon_number):
+        if max(orbit) <= max_count_per_mode:
+            prob += _prob_orbit_exact(graph, orbit, n_mean, loss)
+    return prob
+
+
+def _prob_orbit_mc(
+    graph: nx.Graph, orbit: list, n_mean: float = 5, mc_samples: int = 1000, loss: float = 0.0
```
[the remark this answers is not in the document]
AroosaIjaz: ok I will update this :)

strawberryfields/apps/similarity.py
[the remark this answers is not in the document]
AroosaIjaz: ok I will remove this test :)

strawberryfields/apps/similarity.py
```diff
@@ -412,119 +527,218 @@ def prob_event_mc(
...
     return prob
 
 
-def feature_vector_sampling(
-    samples: list, event_photon_numbers: list, max_count_per_mode: int = 2
+def feature_vector_orbits(
+    graph: nx.Graph,
+    list_of_orbits: list,
+    n_mean: float = 5,
+    mc_samples: int = None,
```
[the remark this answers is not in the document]
AroosaIjaz: Yes, we did. we agreed on making it default with warnings about large graphs. Do you think we should change the default to MC?

strawberryfields/apps/similarity.py
AroosaIjaz: see my comment on JM's comment

strawberryfields/apps/similarity.py
AroosaIjaz: As far as I have checked, yes.

tests/apps/test_similarity.py
AroosaIjaz: oh nice! I did not know this :)

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,79 @@
...
 #
-# This method can be accessed using the :func:`~.prob_event_mc` function. The 4-photon event is
-# approximated as:
+# This method can be accessed using the :func:`~.feature_vector_events` function
+# with ``samples`` set to the number of samples desired to be used in the estimation.
+# For example, to get MC-estimated probabilities for our example feature vector
+# :math:`f_{\mathbf{k} = (2, 4, 6), n_{\max}=2}`, we use:
 
-print(similarity.prob_event_mc(nx.Graph(m0_a), 4, max_count_per_mode=2, n_mean=6))
+print(similarity.feature_vector_events(nx.Graph(m0_a), [2, 4, 6], 2, 6, samples=1000))
```
trbromley, line 238: I just tracked down why the tutorials take so long to build and it is this line - expecially for the 6 total photon number case this takes a long time. I'd recommend we work a way to instead have `prob_event_exact` for the 2 or 4 photon number case.

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,79 @@
...
+#
+# Built-in functions :func:`~.feature_vector_orbits` and :func:`~.feature_vector_events`
+# can be used to get exact feature vectors. These functions
+# use a keyword argument ``samples`` to signal producing either exact or approximate probabilities.
+# ``samples`` is set to ``None`` to get exact feature vector by default. To use Monte Carlo estimation,
+# ``samples`` can be set to the number of samples desired to be used in the estimation, as shown later.
+# For example, to get the exact event probabilities in the feature vector example
+# :math:`f_{\mathbf{k} = (2, 4, 6), n_{\max}=2}` seen previously, we use:
+
+print(similarity.feature_vector_events(nx.Graph(m0_a), [2, 4, 6], 2, 6))
```
trbromley, line 211: I just tracked down why the tutorials take so long to build and it is this line - expecially for the 6 total photon number case this takes a long time. I'd recommend we work a way to instead have prob_event_exact for the 2 or 4 photon number case.

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+
+    @pytest.mark.parametrize("k", [3, 5, 7, 9])
+    @pytest.mark.parametrize("nmax", [1, 2])
+    def test_odd_photon_numbers(self, k, nmax):
+        """Test if function returns zero probability for odd number of total photons."""
+        graph = nx.complete_graph(10)
+        assert similarity.prob_event_exact(graph, k, nmax) == 0.0
+
+
 class TestProbOrbitMC:
```
ixfoduap, line 344: Also needs to test that correct values are calculated for non-trivial specific cases

tests/apps/test_similarity.py
```diff
@@ -222,142 +223,224 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
-        cov = state.cov()
-        disp = state.displacement()
-
-        assert np.allclose(cov, 0.5 * state.hbar * np.eye(2 * dim))
-        assert np.allclose(disp, np.zeros(dim))
+        graph = nx.complete_graph(5)
+        assert similarity.prob_orbit_mc(graph, [1, 1, 1, 1], samples=1, loss=1) == 0.0
 
 
 class TestProbEventMC:
```
ixfoduap, line 394: Also needs to test that correct values are calculated for non-trivial specific cases

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
         with monkeypatch.context() as m:
-            m.setattr(sf.LocalEngine, "run", mock_eng_run)
-            similarity.prob_event_mc(graph, 6, 3, samples=1)
-            p_func = mock_eng_run.call_args[0][0]
+            m.setattr(
+                similarity,
+                "prob_orbit_mc",
+                lambda _graph, orbit, n_mean, samples, loss: 1.0 / sum(orbit),
+            )
+            graph = nx.complete_graph(8)
```
ixfoduap, line 510: Too many lines

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
+            similarity.feature_vector_orbits(g, [[1, 1], [2]], loss=2)
+
+    def test_calls_exact_for_zero_samples(self):
+        """Test if function calls the exact function for zero samples"""
+        g = nx.complete_graph(5)
+        assert similarity.feature_vector_orbits(
+            g, [[1, 1]], samples=0
+        ) == similarity.prob_orbit_exact(g, [1, 1])
+
+    def test_correct_vector_returned(self, monkeypatch):
```
ixfoduap, line 499: This is a good test

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
+        graph = nx.complete_graph(6)
+        assert similarity.prob_event_mc(graph, 6, 3, samples=1, loss=1) == 0.0
 
-        with monkeypatch.context() as m:
-            m.setattr(sf.LocalEngine, "run", mock_eng_run)
-            similarity.prob_event_mc(graph, 6, 3, samples=1, loss=0.5)
-            p_func = mock_eng_run.call_args[0][0]
 
-        assert isinstance(p_func.circuit[1].op, sf.ops.LossChannel)
+class TestFeatureVectorOrbits:
```
ixfoduap, line 460: Also needs to test that correct values are calculated for non-trivial specific cases

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
+                similarity,
+                "prob_orbit_exact",
+                lambda _graph, orbit, n_mean, loss: 0.5 * (1.0 / sum(orbit)),
+            )
+            graph = nx.complete_graph(8)
+
+            assert similarity.feature_vector_orbits(graph, [[1, 1], [2, 1, 1]]) == [0.25, 0.125]
+
+
+class TestFeatureVectorEvents:
```
ixfoduap, line 528: Also needs to test that correct values are calculated for non-trivial specific cases

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
+            ValueError, match="Maximum number of photons per mode must be non-negative"
+        ):
+            similarity.feature_vector_events(g, [2, 4], max_count_per_mode=-1)
+
+    def test_calls_exact_for_zero_samples(self):
+        """Test if function calls the exact function for zero samples"""
+        g = nx.complete_graph(5)
+        assert similarity.feature_vector_events(
+            g, [2], 1, samples=0
+        ) == similarity.prob_event_exact(g, 2, 1)
```
ixfoduap, line 576: Too many lines

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
 
         with monkeypatch.context() as m:
-            m.setattr(sf.LocalEngine, "run", mock_eng_run)
-            similarity.prob_event_mc(graph, 6, 3, samples=1, loss=1)
-            p_func = mock_eng_run.call_args[0][0]
+            m.setattr(
+                similarity,
+                "prob_event_mc",
+                lambda _graph, photons, max_count, n_mean, samples, loss: 1.0 / photons,
+            )
```
ixfoduap, line 588: Too many lines

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
+            g = nx.complete_graph(8)
+            assert similarity.feature_vector_events(g, [2, 4, 8], 1, samples=1) == [
+                0.5,
+                0.25,
+                0.125,
+            ]
```
ixfoduap, line 594: Too many lines

ixfoduap: Basically, important to check the entire set of tests to make sure there aren't any parts that could be shortened to fewer lines

tests/apps/test_similarity.py
```diff
@@ -368,71 +451,230 @@ def test_mean_event(self, monkeypatch):
...
 
-class TestFeatureVectorSampling:
-    """Tests for the function ``strawberryfields.apps.graph.similarity.feature_vector_sampling``"""
+    def test_bad_orbit_photon_numbers(self):
+        """Test if function raises a ``ValueError`` when input is an orbit with numbers
+        below zero."""
+        with pytest.raises(ValueError, match="Cannot request orbits with photon number below zero"):
+            similarity.feature_vector_orbits_sampling([[1, 1, 0], [1, 0, 1]], [[-1, 1]])
+
+    def test_correct_vector_returned(self, monkeypatch):
```
ixfoduap, line 620: This is great! These are the most important tests: are we getting the values we should from the function?

AroosaIjaz:
I have added some non-trivial tests to all classes. Thanks Juan Miguel for the suggestion.

strawberryfields/apps/similarity.py
```diff
@@ -510,21 +652,98 @@ def feature_vector_mc(
         loss (float): fraction of photons lost in GBS
```
trbromley, line 652: On line above, would also mention that the samples are optional like I suggested in `feature_vector_orbits`.

strawberryfields/apps/similarity.py
```diff
@@ -510,21 +652,98 @@ def feature_vector_mc(
         loss (float): fraction of photons lost in GBS
 
     Returns:
-        list[float]: a feature vector of event probabilities in the same order as
-        ``event_photon_numbers``
+        list[float]: a feature vector of orbit probabilities
```
trbromley, line 655:
```suggestion
        list[float]: a feature vector of event probabilities
```

strawberryfields/apps/similarity.py
```diff
@@ -510,21 +652,98 @@ def feature_vector_mc(
...
+    >>> samples = data.Mutag0()
+    >>> feature_vector_orbits_sampling(samples, [[1,1], [2], [2,1,1], [1,1,1,1]])
+    [0.19035, 0.0, 0.05175, 0.1352]
+
+    Args:
+        samples (list[list[int]]): a list of samples
+        list_of_orbits (list[list[int]]): a list of orbits
+
+    Returns:
+        list[float]: a feature vector made up of estimated orbit probabilities
```
trbromley, line 698: In the return part of `feature_vector_events_sampling`, we mention the order - would be good to also do it here.

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+        assert similarity.prob_orbit_exact(graph, [k]) == 0.0
+
+    def test_correct_result_returned(self, monkeypatch):
+        """Tests if the call to _get_state function is performed correctly and probabilities over all
+        permutations of the given orbit are summed correctly. The test monkeypatches the fock_prob
+        function so that the probability is the same for each sample permutation and
+        is equal to 1/8. For a 4-mode graph, [1, 1] has 6 possible permutations. """
+        graph = nx.complete_graph(4)
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob", lambda *args, **kwargs: 1/8)
```
trbromley, line 301: This line looks too long

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+            assert similarity.prob_orbit_exact(graph, [1, 1]) == 6/8
+
+    def test_known_result(self):
+        """Tests if the probability of detecting photons in nearly all modes of a
+        graph with a small mean photon number is correctly reproduced as
+        almost zero."""
+        graph = nx.complete_graph(4)
+        assert np.allclose(similarity.prob_orbit_exact(graph, [4], 1), 0)
```
trbromley, line 309: We could probably raise even higher than 4. But since the test passes I don't mind.

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+        permutations of the given orbit are summed correctly. The test monkeypatches the fock_prob
+        function so that the probability is the same for each sample permutation and
+        is equal to 1/8. For a 4-mode graph, [1, 1] has 6 possible permutations. """
+        graph = nx.complete_graph(4)
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob", lambda *args, **kwargs: 1/8)
+            assert similarity.prob_orbit_exact(graph, [1, 1]) == 6/8
+
+    def test_known_result(self):
+        """Tests if the probability of detecting photons in nearly all modes of a
```
trbromley, line 305: Is the orbit `[4]` the case of "photons in nearly all modes"? Isn't it the case of having 4 photons appear at once in only one of the modes?

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob", lambda *args, **kwargs: 1/8)
+            assert similarity.prob_event_exact(graph, 2, 1) == 6/8
+
+    def test_known_result(self):
+        """Tests if the probability for detecting multiple photons in one mode, orbit [2], is
+        correctly reproduced to be close to zero for a small graph and a small mean photon number.
+        To get orbit [2], we use ``p(E_{2,2}) - p(E_{2,1})``. """
+        graph = nx.complete_graph(4)
+        assert np.allclose((similarity.prob_event_exact(graph, 2, 2, 1) - similarity.prob_event_exact(graph, 2, 1, 1)), 0)
+
```
trbromley, line 371: Would be good to run black on `test_similarity.py` before merging. It's not required but just keeps things tidy.

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob", lambda *args, **kwargs: 1/8)
+            assert similarity.prob_event_exact(graph, 2, 1) == 6/8
+
+    def test_known_result(self):
+        """Tests if the probability for detecting multiple photons in one mode, orbit [2], is
+        correctly reproduced to be close to zero for a small graph and a small mean photon number.
+        To get orbit [2], we use ``p(E_{2,2}) - p(E_{2,1})``. """
+        graph = nx.complete_graph(4)
+        assert np.allclose((similarity.prob_event_exact(graph, 2, 2, 1) - similarity.prob_event_exact(graph, 2, 1, 1)), 0)
```
trbromley, line 370: Nice test! One question, is the probability of getting orbit `[2]` always zero, independent of the mean photon number? To me that seems likely, as I don't think generating 2 photons is that rare with a mean photon number of 1.

What about a test where we check that `similarity.prob_event_exact(graph, 2, 1, 1)` (which I think is nonzero) is equal to `similarity.prob_orbit_exact(graph, [1, 1], 1)` ?

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
-        disp = state.displacement()
+        graph = nx.complete_graph(5)
+        assert similarity.prob_orbit_mc(graph, [1, 1, 1, 1], samples=1, loss=1) == 0.0
 
-        assert np.allclose(cov, 0.5 * state.hbar * np.eye(2 * dim))
-        assert np.allclose(disp, np.zeros(dim))
+    def test_known_result(self):
+        """Tests if the probability of detecting photons in nearly all modes of a
+        graph with a small mean photon number is correctly reproduced as
+        almost zero."""
```
trbromley, line 426: Also not sure in this case if the docstring corresponds to the setting below.

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
-        g = nx.complete_graph(10)
-        with pytest.raises(ValueError, match="Loss parameter must take a value between zero and"):
+        g = nx.complete_graph(5)
+        with pytest.raises(ValueError, match="Loss parameter must take a value between zero and one"):
             similarity.prob_event_mc(g, 2, 2, loss=2)
 
     def test_invalid_photon_number(self):
         """Test if function raises a ``ValueError`` when a photon number below zero is specified"""
-        g = nx.complete_graph(10)
+        g = nx.complete_graph(5)
```
trbromley, line 457: It's not a big problem, but it would be good to minimize doing changes like these to reduce the overhead of code review. Unless of course the changes are needed/improve something!

tests/apps/test_similarity.py
```diff
@@ -222,217 +223,474 @@ def test_event_cardinality(photons, max_count, modes, expected):
...
 
-    def test_all_loss(self, monkeypatch):
-        """Test if function samples from the vacuum when maximum loss is applied."""
-        dim = 5
-        graph = nx.complete_graph(dim)
-        mock_eng_run = mock.MagicMock()
+    def test_known_result(self):
+        """Tests if the probability of detecting photons in nearly all modes of a
+        graph with a small mean photon number is correctly reproduced as
+        almost zero."""
```
trbromley, line 556: Is this another one where the docstring needs updating?

strawberryfields/apps/similarity.py
```diff
@@ -306,21 +335,132 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
 
+def _get_state(graph: nx.Graph, n_mean: float = 5, loss: float = 0.0) -> GaussianState:
+    r"""Embeds the input graph into a GBS device and returns the corresponding Gaussian state
+    """
+    modes = graph.order()
+    A = nx.to_numpy_array(graph)
+    mean_photon_per_mode = n_mean / float(modes)
+
+    p = sf.Program(modes)
+
```
trbromley, line 346:
```suggestion

# pylint: disable=expression-not-assigned
```
The following lines are causing the codefactor/pylint problems: https://www.codefactor.io/repository/github/xanaduai/strawberryfields/pull/390. We're not doing anything wrong here, it's just the style used to define circuits in SF.

We can hence disable this pylint warning. This should only be done in special cases, and is done using
`# pylint: disable=error-to-disable`.

strawberryfields/apps/similarity.py
```diff
@@ -47,50 +47,77 @@
 
 A feature vector of a graph can be created by choosing a collection of orbits or events and
 evaluating their probabilities with respect to GBS with the embedded graph. These
-probabilities are then selected to be elements of the feature vector. Evaluating the
-probabilities of orbits or events can be achieved through two approaches:
+probabilities are then selected to be elements of the feature vector. For example, event
```
trbromley, line 50: Note, can also update line 1 of this file to read:
`Copyright 2019-2020 Xanadu Quantum Technologies Inc.`

strawberryfields/apps/similarity.py
```diff
@@ -306,74 +321,174 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+    r"""Gives the exact GBS probability of a given orbit according to the input
     graph.
 
-    To make this estimate, several samples from the orbit are drawn uniformly at random using
-    :func:`orbit_to_sample`. The GBS probabilities of these samples are then calculated and the
-    sum is used to create an estimate of the orbit probability.
+    The exact probability of an orbit is made up of the sum of probabilities of
+    all possible GBS output patterns that belong to it:
+
+    .. math::
```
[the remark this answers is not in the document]
AroosaIjaz: Should I remove it completely?

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,79 @@
...
+# exponentially difficult with increasing photon number. Exact probability of an event
+# :math:`p_{k,n_{\max}}` can be calculated similarly or by summing over its constituent orbit probabilities.
+#
+# Built-in functions :func:`~.feature_vector_orbits` and :func:`~.feature_vector_events`
+# can be used to get exact feature vectors. These functions
+# use a keyword argument ``samples`` to signal producing either exact or approximate probabilities.
+# ``samples`` is set to ``None`` to get exact feature vector by default. To use Monte Carlo estimation,
+# ``samples`` can be set to the number of samples desired to be used in the estimation, as shown later.
+# For example, to get the exact event probabilities in the feature vector example
+# :math:`f_{\mathbf{k} = (2, 4, 6), n_{\max}=2}` seen previously, we use:
```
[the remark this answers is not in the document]
AroosaIjaz: should we just keep using [2, 4, 6] like before? I changed it only for faster documentation building but it is good to be consistent; either all examples use [2,4] or [2,4,6]

examples_apps/run_tutorial_similarity.py
```diff
@@ -170,57 +170,79 @@
...
 #
-# This method can be accessed using the :func:`~.prob_event_mc` function. The 4-photon event is
-# approximated as:
+# This method can be accessed using the :func:`~.feature_vector_events` function
+# with ``samples`` set to the number of samples desired to be used in the estimation.
+# For example, to get MC-estimated probabilities for our example feature vector
+# :math:`f_{\mathbf{k} = (2, 4, 6), n_{\max}=2}`, we use:
 
-print(similarity.prob_event_mc(nx.Graph(m0_a), 4, max_count_per_mode=2, n_mean=6))
+print(similarity.feature_vector_events(nx.Graph(m0_a), [2, 4], 2, samples=1000))
```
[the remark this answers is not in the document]
AroosaIjaz: Yes, I don't have issues with using 6 photon number for any method. MC is faster than exact but neither are too slow for me. :)

tests/apps/test_similarity.py
AroosaIjaz: ah yes, you are right! :) I will update this.

tests/apps/test_similarity.py
AroosaIjaz: running black spreads some lines of code over many lines that JM did not like. so, I am manually doing the formatting here.

tests/apps/test_similarity.py
AroosaIjaz: yes, the probability of orbit [2] is usually ~0. 

And yes, I can add this test too :)

AroosaIjaz:
Thank you @trbromley for the in-depth review. I incorporated the changes :)

tests/apps/test_similarity.py
trbromley: I recommend splitting up into a few separate lines rather than going over line width:
```python
p1 = similarity.prob_event_exact(graph, 2, 2, 1)
p2 = similarity.prob_event_exact(graph, 2, 1, 1)
assert np.allclose(p1 - p2, 0)
```
Let's keep the standard of running `black` for `test_similarity.py`. In general, we shouldn't worry about second guessing the formatting of `black` (the idea is that it saves you thinking about that). However, in some cases the result will be brackets split over quite a few lines - in that case we can define some intermediate variables as above for readability.

examples_apps/run_tutorial_similarity.py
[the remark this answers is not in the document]
trbromley: I'm cautious about switching to [2, 4, 6] for the exact calculation because it was really slow on my machine. Does it change the story much if we change everything to [2, 4]? This is a bit more work though.

strawberryfields/apps/similarity.py
```diff
@@ -306,21 +335,133 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+def _get_state(graph: nx.Graph, n_mean: float = 5, loss: float = 0.0) -> GaussianState:
+    r"""Embeds the input graph into a GBS device and returns the corresponding Gaussian state.
+    """
+    modes = graph.order()
+    A = nx.to_numpy_array(graph)
+    mean_photon_per_mode = n_mean / float(modes)
+
+    p = sf.Program(modes)
+
+    # pylint: disable=error-to-disable.
```
trbromley, line 347:
```suggestion
    # pylint: disable=expression-not-assigned
```

strawberryfields/apps/similarity.py
```diff
@@ -591,11 +591,12 @@ def feature_vector_orbits(
         graph (nx.Graph): input graph
         list_of_orbits (list[list[int]]): a list of orbits
         n_mean (float): total mean photon number of the GBS device
-        samples (int): number of samples used in the Monte Carlo estimation
+        samples (int): optional number of samples used in the Monte Carlo estimation. Defaults to exact calculation if ``samples`` is unspecified.
```
ixfoduap, line 594: This line may be too long

strawberryfields/apps/similarity.py
```diff
@@ -648,11 +648,12 @@ def feature_vector_events(
         event_photon_numbers (list[int]): a list of events described by their total photon number
         max_count_per_mode (int): maximum number of photons per mode for all events
         n_mean (float): total mean photon number of the GBS device
-        samples (int): number of samples used in the Monte Carlo estimation
+        samples (int): optional number of samples used in the Monte Carlo estimation. Defaults to exact calculation if ``samples`` is unspecified.
```
ixfoduap, line 651: Again, I think this line is too long

tests/apps/test_similarity.py
```diff
@@ -293,6 +291,24 @@ def test_odd_photon_numbers(self, k):
...
+        permutations of the given orbit are summed correctly. The test monkeypatches the fock_prob
+        function so that the probability is the same for each sample permutation and
+        is equal to 1/8. For a 4-mode graph, [1, 1] has 6 possible permutations. """
+        graph = nx.complete_graph(4)
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob",
+                      lambda *args, **kwargs: 1/8)
+            assert similarity.prob_orbit_exact(graph, [1, 1]) == 6/8
+
+    def test_known_result(self):
```
ixfoduap, line 305: This test seems dangerous: how small exactly is the orbit probability? Also, it seems you can pass this test with a wrong implementation of the function, e.g., if it's not summing correctly over all samples in the orbit -- if the probabilities are small it doesn't matter if you miss a few in the sum.

I think a better version of this test would be to set a lower number of photons, say the orbit `[2]`, a higher mean photon number, say `n_mean = 2`, then calculate the probability by hand, hard-code it into the test, and check that we get the right answer

tests/apps/test_similarity.py
```diff
@@ -340,6 +352,26 @@ def test_odd_photon_numbers(self, k, nmax):
...
+        function so that the probability is the same for each sample permutation of all constituent orbits
+        and is equal to 1/8. For a 4-mode graph, an event with ``photon_number = 2``, and
+        ``max_count_per_mode = 1`` contains orbit [1, 1] which has 6 possible sample permutations."""
+        graph = nx.complete_graph(4)
+        with monkeypatch.context() as m:
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob",
+                      lambda *args, **kwargs: 1/8)
+            assert similarity.prob_event_exact(graph, 2, 1) == 6/8
+
+    def test_known_result(self):
```
ixfoduap, line 367: Same comment as with the test above: better to test that we reproduce a hard-coded value when the mean photon number is larger

tests/apps/test_similarity.py
```diff
@@ -373,12 +403,16 @@ def test_mean_computation_orbit(self, monkeypatch):
...
         with monkeypatch.context() as m:
-            m.setattr(
-                "strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob",
-                lambda *args, **kwargs: 0.2,
-            )
+            m.setattr("strawberryfields.backends.gaussianbackend.states.GaussianState.fock_prob",
+                      lambda *args, **kwargs: 0.2)
             assert np.allclose(similarity.prob_orbit_mc(graph, [1, 1, 1, 1]), 1.0)
 
+    @pytest.mark.parametrize("k", [1, 3, 5, 7, 9])
```
ixfoduap, line 410: Nice!

tests/apps/test_similarity.py
```diff
@@ -390,6 +424,13 @@ def test_max_loss(self):
         graph = nx.complete_graph(5)
         assert similarity.prob_orbit_mc(graph, [1, 1, 1, 1], samples=1, loss=1) == 0.0
 
+    def test_known_result(self):
```
ixfoduap, line 427: Same comment as before

tests/apps/test_similarity.py
```diff
@@ -444,18 +481,29 @@ def test_mean_computation_event(self, monkeypatch):
...
+        """Test if function returns zero probability for odd number of total photons."""
+        graph = nx.complete_graph(10)
+        assert similarity.prob_event_mc(graph, k, nmax) == 0.0
+
     def test_max_loss(self):
         """Test if function samples from the vacuum when maximum loss is applied."""
         graph = nx.complete_graph(6)
         assert similarity.prob_event_mc(graph, 6, 3, samples=1, loss=1) == 0.0
 
+    def test_known_result(self):
```
ixfoduap, line 500: Same comment as before

tests/apps/test_similarity.py
```diff
@@ -484,46 +532,38 @@ def test_invalid_loss(self):
...
-                "prob_orbit_exact",
-                lambda _graph, orbit, n_mean, loss: 0.5 * (1.0 / sum(orbit)),
-            )
+            m.setattr(similarity, "prob_orbit_exact",
+                      lambda _graph, orbit, n_mean, loss: 0.5 * (1.0 / sum(orbit)))
             graph = nx.complete_graph(8)
-
             assert similarity.feature_vector_orbits(graph, [[1, 1], [2, 1, 1]]) == [0.25, 0.125]
 
+    def test_known_result(self):
```
ixfoduap, line 560: Same comment as before

tests/apps/test_similarity.py
```diff
@@ -563,45 +599,38 @@ def test_bad_event_photon_numbers(self):
...
-                similarity,
-                "prob_event_exact",
-                lambda _graph, photons, max_count, n_mean, loss: 0.5 * (1.0 / photons),
-            )
+            m.setattr(similarity, "prob_event_exact",
+                      lambda _graph, photons, max_count, n_mean, loss: 0.5 * (1.0 / photons))
             g = nx.complete_graph(8)
             assert similarity.feature_vector_events(g, [2, 4, 8], 1) == [0.25, 0.125, 0.0625]
 
+    def test_known_result(self):
```
ixfoduap, line 627: Same

examples_apps/run_tutorial_similarity.py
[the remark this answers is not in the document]
AroosaIjaz: no worries. it is not too much. I can change it to [2,4] everywhere :)

strawberryfields/apps/similarity.py
AroosaIjaz: I made it two lines and checked that the documents are building correctly

tests/apps/test_similarity.py
AroosaIjaz: It is hard to encode a test for monte carlo functions as outputs change on every run.

AroosaIjaz:
@ixfoduap I added some more tests as you suggested :)

strawberryfields/apps/similarity.py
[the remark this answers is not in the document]
trbromley: Removing `\frac{{\rm haf}(A_{S})^2}{S! \sqrt{\det(Q)}}` would be my preference, but also good to see what @ixfoduap thinks.

strawberryfields/apps/similarity.py
```diff
@@ -306,21 +335,133 @@ def event_cardinality(photon_number: int, max_count_per_mode: int, modes: int) -
...
+    r"""Gives the exact probability of a given orbit for the input graph.
+
+    The exact probability of an orbit is the sum of probabilities of
+    all possible GBS output patterns that belong to it:
+
+    .. math::
+       p(O) = \sum_{S \in O} p(S) = \frac{{\rm haf}(A_{S})^2}{S! \sqrt{\det(Q)}}
+
+    where :math:`S` represents one GBS output pattern, :math:`A_{S}` is its induced
+    sub-matrix and :math:`Q` is the :math:`Q` covariance matrix obtained from the adjacency matrix.
```
trbromley, line 369:
```suggestion
       p(O) = \sum_{S \in O} p(S)

    where :math:`S` are samples belonging to :math:`O`.
```

trbromley: Just to emphasize that the last equality is wrong in its current form, it would need the sum over S in O to be correct and also ideally a clarification that S! means the product of factorials of the elements.

^ Hence I think we should keep things simple and just delete.

co9olguy:
:fireworks:

PR #391, opened 2020-05-15
Add KL divergence to cost module

ixfoduap:
As part of the `train` module of the apps layer, the `cost` module of the SF apps layer contains methods for evaluating and computing gradients of cost functions. 

This PR adds a class to the `cost` module dedicated to the Kullback-Liebler cost function, for which gradients can be calculated efficiently using classical methods.

PR #393, opened 2020-05-19
Add stochastic cost function to GBS cost module in apps layer

trbromley:
This PR adds a stochastic optimization based cost function to the `train.cost` module of the applications layer.

A similar PR for the KL cost function can be found here: https://github.com/XanaduAI/strawberryfields/pull/391.

### DIFF
--- a/strawberryfields/apps/train/__init__.py
+++ b/strawberryfields/apps/train/__init__.py
@@ -8,6 +8,10 @@ Tools for training variational GBS devices.
 
     embed
     Stochastic
+    VGBS
+    param
 """
+from strawberryfields.apps.train.param import VGBS
+import strawberryfields.apps.train.param
 import strawberryfields.apps.train.embed
 from strawberryfields.apps.train.cost import Stochastic
--- a/strawberryfields/apps/train/__init__.py
+++ b/strawberryfields/apps/train/__init__.py
@@ -7,5 +7,7 @@ Tools for training variational GBS devices.
     :toctree: api
 
     embed
+    Stochastic
 """
 import strawberryfields.apps.train.embed
+from strawberryfields.apps.train.cost import Stochastic
--- a/strawberryfields/apps/train/__init__.py
+++ b/strawberryfields/apps/train/__init__.py
@@ -10,9 +10,12 @@ Tools for training variational GBS devices.
     Stochastic
     VGBS
     param
+    KL
     cost
 """
 from strawberryfields.apps.train.param import VGBS
+from strawberryfields.apps.train.cost import KL
+import strawberryfields.apps.train.cost
 import strawberryfields.apps.train.param
 import strawberryfields.apps.train.embed
 from strawberryfields.apps.train.cost import Stochastic
--- a/strawberryfields/apps/train/__init__.py
+++ b/strawberryfields/apps/train/__init__.py
@@ -10,6 +10,7 @@ Tools for training variational GBS devices.
     Stochastic
     VGBS
     param
+    cost
 """
 from strawberryfields.apps.train.param import VGBS
 import strawberryfields.apps.train.param
--- a/strawberryfields/apps/train/cost.py
+++ b/strawberryfields/apps/train/cost.py
@@ -12,12 +12,113 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 r"""
-TODO - this is provided in another PR
+Submodule for computing gradients and evaluating cost functions with respect to GBS circuits
+
+In the context of stochastic optimization, cost functions are expressed as expectation values
+over the GBS distribution. Within the WAW parametrization, gradients of cost functions can be
+expressed as expectation values over the GBS distribution. This module contains methods for
+calculating these gradients and for using gradient-based methods to optimize GBS circuits. In the
+case of optimization with respect to a Kullback-Leibler divergence or log-likelihood cost
+function, gradients can be computed efficiently, leading to fast training.
 """
 from typing import Callable
 
 import numpy as np
 from strawberryfields.apps.train.param import _Omat, VGBS
+
+
+class KL:
+    r"""Kullback-Liebler divergence cost function.
+
+    In a standard unsupervised learning scenario, data are assumed to be sampled from an unknown
+    distribution and a common goal is to learn that distribution. Training of a model
+    distribution can be performed by minimizing the Kullback-Leibler (KL) divergence, which up to
+    additive constants can be written as:
+
+    .. math::
+
+        KL = -\frac{1}{T}\sum_S \log[P(S)],
+
+    where :math:`S` is an element of the data, :math:`P(S)` is the probability of observing that
+    element when sampling from the GBS distribution, and :math:`T` is the total number of elements
+    in the data. For the GBS distribution in the WAW parametrization, the gradient of the KL
+    divergence can be written as
+
+
+    .. math::
+
+        \partial_\theta KL(\theta) = - \sum_{k=1}^m\frac{1}{w_k}(\langle n_k\rangle_{\text{data}}-
+        \langle n_k\rangle_{\text{GBS}})\partial_\theta w_k,
+
+    where :math:`\langle n_k\rangle` denotes the average photon numbers in mode *k*. This class
+    provides methods to compute gradients and evaluate the cost function.
+
+    **Example usage**
+
+    >>> embedding = train.embed.Exp(4)
+    >>> A = np.ones((4, 4))
+    >>> vgbs = train.VGBS(A, 3, embedding, threshold=True)
+    >>> params = np.array([0.05, 0.1, 0.02, 0.01])
+    >>> data = np.zeros((4, 4))
+    >>> kl = cost.KL(data, vgbs)
+    >>> kl.evaluate(params)
+    -0.2866830267216749
+    >>> kl.grad(params)
+    array([-0.52812574, -0.5201932 , -0.53282312, -0.53437824])
+
+    Args:
+        data (array): Array of samples representing the training data
+        vgbs (train.VGBS): Variational GBS class
+
+    """
+
+    def __init__(self, data: np.ndarray, vgbs: VGBS):
+        self.data = data
+        self.vgbs = vgbs
+        self.nr_samples, self.nr_modes = np.shape(self.data)
+        self.mean_n_data = np.mean(self.data, axis=0)
+
+    def grad(self, params: np.ndarray) -> np.ndarray:
+        r"""Calculates the gradient of the Kullback-Liebler cost function with respect to the
+        trainable parameters
+
+        **Example usage**
+
+        >>> kl.grad(params)
+        array([-0.52812574, -0.5201932 , -0.53282312, -0.53437824])
+
+        Args:
+            params (array[float]): the trainable parameters :math:`\theta`
+        Returns:
+            array: the gradient of the KL cost function with respect to :math:`\theta`
+        """
+        weights = self.vgbs.embedding(params)
+        if self.vgbs.threshold:
+            n_diff = self.vgbs.mean_clicks_by_mode(params) - self.mean_n_data
+        else:
+            n_diff = self.vgbs.mean_photons_by_mode(params) - self.mean_n_data
+        return (n_diff / weights) @ self.vgbs.embedding.jacobian(params)
+
+    def evaluate(self, params: np.ndarray) -> float:
+        r"""Computes the value of the Kullback-Liebler divergence cost function.
+
+        **Example usage**
+
+        >>> kl.evaluate(params)
+        -0.2866830267216749
+
+        Args:
+            params (array): the trainable parameters :math:`\theta`
+        Returns:
+            float: the value of the cost function
+        """
+        kl = 0
+        for sample in self.data:
+            kl += np.log(self.vgbs.prob_sample(params, sample))
+        return -kl / self.nr_samples
+
+    def __call__(self, params: np.ndarray) -> float:
+        return self.evaluate(params)
 
 
 class Stochastic:
--- a/strawberryfields/apps/train/cost.py
+++ b/strawberryfields/apps/train/cost.py
@@ -24,7 +24,7 @@ class Stochastic:
     r"""Stochastic cost function given by averaging over samples from a trainable GBS distribution.
 
     A stochastic optimization problem is defined with respect to a function :math:`h(\bar{n})` that
-    assigns a cost to an input sample :math:`\bar{n}`. The cost function is then defined as the
+    assigns a cost to an input sample :math:`\bar{n}`. The cost function is the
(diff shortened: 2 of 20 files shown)


strawberryfields/apps/train/cost.py
```diff
@@ -0,0 +1,183 @@
...
+        diff = self.sample_difference_from_mean(sample, params)
+        return h * (diff / w) @ jac
+
+    def gradient(self, params: np.ndarray, n_samples: int = 100) -> np.ndarray:
+        """Evaluates the gradient of the cost function.
+
+        The gradient is evaluated by finding an average of a function over a number ``n_samples`` of
+        samples generated from the VGBS system using the trainable parameters :math:`\theta`.
+
+        The averaged function is specified in `this paper <https://arxiv.org/abs/2004.04770>`__.
```
ixfoduap, line 165: I think it's good to write down the gradient formula, since it's the key result needed to train GBS distribution. The same is true for the `KL` cost function, that I just realized also does not have an explicit formula

strawberryfields/apps/train/cost.py
```diff
@@ -0,0 +1,183 @@
...
+
+        **Example usage:**
+
+        >>> cost.gradient(params, 100)
+        array([ 0.10880756, -0.1247146 ,  0.12426481, -0.13783342])
+
+        Args:
+            params (array): the trainable parameters :math:`\theta`
+            n_samples (int): the number of GBS samples used to average the gradient
```
ixfoduap, line 174:
```suggestion
            n_samples (int): the number of GBS samples used in the gradient estimation
```
Is it correct that we re averaging the gradient? I think it's more accurate that the gradient is an average, which we estimate by sampling

strawberryfields/apps/train/cost.py
```diff
@@ -0,0 +1,183 @@
...
+            array: the one shot gradient
+        """
+        w = self.vgbs.embedding(params)
+        jac = self.vgbs.embedding.jacobian(params)
+
+        h = self.h_reparametrized(sample, params)
+        diff = self.sample_difference_from_mean(sample, params)
+        return h * (diff / w) @ jac
+
+    def gradient(self, params: np.ndarray, n_samples: int = 100) -> np.ndarray:
```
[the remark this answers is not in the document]
trbromley: Great, following on from the conversation above, I have added more details in the docstrings to explain what is going on.

strawberryfields/apps/train/cost.py
trbromley: Awesome! I added the formula for the gradient and also the reparameterised h function. Still not sure I like exactly how it reads but maybe we can save polishing that until later.

tests/apps/test_cost.py
```diff
@@ -0,0 +1,296 @@
...
+        cost_fn = train.Stochastic(h, vgbs)
+        possible_samples = list(itertools.product([0, 1, 2], repeat=dim))
+
+        probs = [vgbs.prob_sample(params, s) for s in possible_samples]
+        probs_init = [vgbs.prob_sample(np.zeros(dim - 1), s) for s in possible_samples]
+
+        cost = sum([probs[i] * h(s) for i, s in enumerate(possible_samples)])
+        cost_reparam = sum([probs_init[i] * cost_fn.h_reparametrized(s, params) for i, \
+                                                                                    s in enumerate(
+            possible_samples)])
```
ixfoduap, line 84: strange break here

tests/apps/test_cost.py
```diff
@@ -0,0 +1,296 @@
...
+
+        n_mean_by_mode = [n_mean / dim] * dim
+        samples = self.identity_sampler(n_mean_by_mode, n_samples=n_samples)
+        vgbs.add_A_init_samples(samples)
+
+        params = np.linspace(0, 1 / dim, dim)
+        cost_fn = train.Stochastic(h, vgbs)
+        cost = cost_fn(params, n_samples=n_samples)
+
+        # We need to generate new samples and then calculate the cost with respect to them
```
ixfoduap, line 218: Why new samples? Shouldn't we be able to use the reparametrization trick?

tests/apps/test_cost.py
```diff
@@ -0,0 +1,296 @@
...
+        new_n_mean_by_mode = vgbs.mean_photons_by_mode(params)
+        new_samples = self.identity_sampler(new_n_mean_by_mode, n_samples=n_samples)
+        expected_cost = np.mean(np.sum((new_samples - objectives) ** 2, axis=1))
+
+        assert np.allclose(cost, expected_cost, rtol=0.1)
```
ixfoduap, line 223: Are we confident that this tolerance will always be sufficiently large?

tests/apps/test_cost.py
```diff
@@ -0,0 +1,296 @@
...
+        A = np.eye(dim)
+        vgbs = train.VGBS(A, n_mean, simple_embedding, threshold=False)
+
+        n_mean_by_mode = [n_mean / dim] * dim
+        samples = self.identity_sampler(n_mean_by_mode, n_samples=n_samples)
+        vgbs.add_A_init_samples(samples)
+
+        params = np.zeros(dim)
+        cost_fn = train.Stochastic(h, vgbs)
+        reps = 4
```
ixfoduap, line 281: We get close to the analytical minimum with just 4 steps? I find this surprising. Do we start sufficiently far from the analytical minimum that the test would fail without these 4 steps?

trbromley: Ah it's a good point! I think it's a mixture of the learning rate and choice of objective. The objective results in a very small ideal mean photon number, so with that learning rate will quickly just learn to put the mean number of photons in each mode as near-zero.

I tried again with `objectives = np.array([4.75, 4, 4.5, 5])` and things are much more as expected, but it takes many steps to converge. Maybe this isn't so suited as a test and better as part of a tutorial :thinking:

strawberryfields/apps/train/__init__.py
```diff
@@ -7,6 +7,7 @@
     :toctree: api
 
     embed
+    Stochastic
```
antalszava, line 10: very minor thing: what determines capitalization here? (`Stochastic` vs `stochastic`)

strawberryfields/apps/train/cost.py
```diff
@@ -118,3 +120,194 @@ def evaluate(self, params: np.ndarray) -> float:
...
+
+    >>> embedding = train.embed.Exp(4)
+    >>> A = np.ones((4, 4))
+    >>> vgbs = train.VGBS(A, 3, embedding, threshold=True)
+    >>> h = lambda x: sum([x[i] * (-1) ** (i + 1) for i in range(4)])
+    >>> cost = Stochastic(h, vgbs)
+    >>> params = np.array([0.05, 0.1, 0.02, 0.01])
+    >>> cost.evaluate(params, 100)
+    0.03005489236683591
+    >>> cost.gradient(params, 100)
```
antalszava, line 156: :100:

strawberryfields/apps/train/cost.py
```diff
@@ -118,3 +120,194 @@ def evaluate(self, params: np.ndarray) -> float:
...
+    def gradient(self, params: np.ndarray, n_samples: int) -> np.ndarray:
+        r"""Evaluates the gradient of the cost function.
+
+        As shown in `this paper <https://arxiv.org/abs/2004.04770>`__, the gradient can be
+        evaluated by finding an average over samples generated from the input adjacency matrix to
+        the VGBS system:
+
+        .. math::
+
+            \partial_{\theta} C (\theta) = \sum_{\bar{n}} h(\bar{n}, \theta) P(\bar{n})
```
antalszava, line 284: Perhaps I missed something, I just came across that this would be an approximate
```suggestion
            \partial_{\theta} C (\theta) \approx \sum_{\bar{n}} h(\bar{n}, \theta) P(\bar{n})
```

tests/apps/test_cost.py
```diff
@@ -91,7 +94,288 @@ def test_grad_threshold(self, k):
...
+        result out by hand"""
+        cost_fn = train.Stochastic(h, vgbs)
+        sample = np.ones(dim)
+        h_reparam = cost_fn.h_reparametrized(sample, params)
+        h_reparam_expected = -1.088925964188385
+        assert np.allclose(h_reparam, h_reparam_expected)
+
+    @pytest.mark.parametrize("threshold", [False])
+    def test_evaluate(self, vgbs, dim, params):
+        """Test that evaluate returns the expected value when the VGBS class is preloaded with a
```
antalszava, line 179: (Minor suggestion): would be applicable for further docstrings too if agreed on
```suggestion
        """Test that the evaluate method returns the expected value when the VGBS class is preloaded with a
```

tests/apps/test_cost.py
```diff
@@ -91,7 +94,288 @@ def test_grad_threshold(self, k):
...
+        ones = np.ones((n_samples, dim))
+        samples = np.vstack([zeros, ones])
+        vgbs.add_A_init_samples(samples)
+
+        cost_fn = train.Stochastic(h, vgbs)
+        h0 = cost_fn.h_reparametrized(zeros[0], params)
+        h1 = cost_fn.h_reparametrized(ones[0], params)
+        eval_expected = (h0 + h1) * 0.5
+
+        eval = cost_fn(params, 2 * n_samples)
```
antalszava, line 194: could leave a small comment here referring to the fact that calling the instance of `Stochastic` calls on `Stochastic.evaluate`

tests/apps/test_cost.py
```diff
@@ -91,7 +94,288 @@ def test_grad_threshold(self, k):
...
+    the trainable parameters can only control the amount of squeezing in each mode. We then use a
+    mean squared error cost function that simply subtracts a fixed vector from an input sample and
+    squares the result."""
+
+    def identity_sampler(self, n_mean_by_mode, n_samples):
+        """Used for quickly generating samples from a diagonal adjacency matrix. This requires
+        sampling from single mode squeezed states whose photon number distribution is specified
+        by the negative binomial."""
+        ps = 1 / (1 + np.array(n_mean_by_mode))
+        np.random.seed(0)
```
antalszava, line 250: It might be good to move this into a "more global scope" (e.g. `TestStochasticIntegrationPNR.__init__`?). In more specifics, this might lead to unexpected behaviour if another method/function would be added that calls on `np.random` after seeding happened. Right now calling on `identity_sampler` leaves a state when called on by a test case and might affect other test outcomes too. Moving it to a more global scope would perhaps help with spotting that seeding happens.

**Small example**
```python
import numpy as np

class A:
    def __init__(self):
        np.random.seed(0)
        return
    
    def some_random_things(self):
        return np.random.random(55)
    
a = A()
a.some_random_things()
```
will output the same array on every call.

tests/apps/test_cost.py
```diff
@@ -91,7 +94,288 @@ def test_grad_threshold(self, k):
...
+
+        # Consider the problem with respect to a single mode. We want to calculate
+        # E((s - x) ** 2) with s the number of photons in the mode, x the element of the fixed
+        # vector, and with the expectation value calculated with respect to the (twice) negative
+        # binomial distribution. We know that E(s) = 2 * r * (1 - q) / q and
+        # Var(s) = 4 * (1 - q) * r / q ** 2 in terms of the r and q parameters of the negative
+        # binomial distribution. Using q = 1 / (1 + n_mean) with n_mean the mean number of
+        # photons in that mode and r = 0.5, we can calculate
+        # E((s - x) ** 2) = 3 * n_mean ** 2 + 2 * (1 - x) * n_mean + x ** 2,
+        # This can be differentiated to give the derivative below.
```
antalszava, line 346: How about moving this to the docstring with a subheader `Steps of deriving the analytic result` and at the end also having the value of `6 * n_mean_by_mode + 2 * (1 - objectives)`?

It might help with a better overview of the code structure of the test.

tests/apps/test_cost.py
```diff
@@ -91,7 +94,288 @@ def test_grad_threshold(self, k):
...
+        vgbs.add_A_init_samples(samples)
+
+        params = np.zeros(dim)
+        cost_fn = train.Stochastic(h, vgbs)
+        reps = 4
+        lr = 0.01
+
+        # See test_gradient for further context. We want to minimize
+        # E((s - x) ** 2) = 3 * n_mean ** 2 + 2 * (1 - x) * n_mean + x ** 2, which is minimized
+        # by the n_mean given below.
```
antalszava, line 370: this too might be moved into the docstring

strawberryfields/apps/train/cost.py
```diff
@@ -12,12 +12,114 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 r"""
-TODO - this is provided in another PR
+Submodule for computing gradients and evaluating cost functions with respect to GBS circuits
+
+In the context of stochastic optimization, cost functions are expressed as expectation values
+over the GBS distribution. Within the WAW parametrization, gradients of cost functions can be
```
ixfoduap:
```suggestion
over the GBS distribution. Within the WAW parametrization, gradients of cost functions can also be
```

tests/apps/test_cost.py
```diff
@@ -79,9 +155,12 @@ def test_h_reparametrized(self, vgbs, dim, params):
...
         cost = sum([probs[i] * h(s) for i, s in enumerate(possible_samples)])
-        cost_reparam = sum([probs_init[i] * cost_fn.h_reparametrized(s, params) for i, \
-                                                                                    s in enumerate(
-            possible_samples)])
+        cost_reparam = sum(
+            [
+                probs_init[i] * cost_fn.h_reparametrized(s, params)
+                for i, s in enumerate(possible_samples)
+            ]
+        )
```
ixfoduap, line 163: Too many lines?

trbromley: I've tried to simplify by defining the terms inside the sum on a previous line. However, the above is how it looked when running `black -l 100` so I wouldn't worry too much about second guessing the formatting.

strawberryfields/apps/train/cost.py
trbromley: Thanks, done

tests/apps/test_cost.py
antalszava: Hi @trbromley, sorry for misleading you on this! It also seems that `pytest` behaves differently then simple classes making seeding more contrived than I thought so.

I think this could either: 
* be left as is (as seeding, after all, does not seem to matter between cases, this can be queried by having a look at the output of `np.random.get_state()[1][0]`)
* or a fixture could be added in `conftest.py` like
```python
@pytest.fixture
def seed():
    np.random.seed(0)
```
and making `def test_initial_cost(self, dim, n_mean, simple_embedding):` into `def test_initial_cost(self, dim, n_mean, simple_embedding, seed):`

trbromley: Thanks @antalszava, have merged in now, but adding a fixture would be good for the future!

Issue #386, opened 2020-05-08
Adding Parity as an Observable


Lucaman99:
(In reference to this Issue raised over in the Pennylane repository: https://github.com/XanaduAI/pennylane/issues/193)

As @artix41 said in his issue, it might be interesting to try to incorporate a parity observable, 
defined as (-1)^N, into Strawberry Fields, so that it can eventually be incorporated into Pennylane.

I'd love to try to tackle this issue, where would be the best place to start / where in the codebase should a feature like this be added? @josh146

Lucaman99:
I don't think it would be too difficult to take the pre-existing implementation of the number operator in the SF code and calculate (-1)^N by expanding (-1)^x in a power series, with some finite truncation.

josh146:
Hi @Lucaman99! 

> I'd love to try to tackle this issue, where would be the best place to start / where in the codebase should a feature like this be added?

The best place would probably be the `backend/states.py` file; this contains all the methods for manipulating and post-processing state vectors that are returned from the Strawberry Fields backends. 

Something to keep in note is that there are separate state objects, one for the Fock backends, and one for the Gaussian backends! This is because the two states differ in their representation; a Fock-basis density matrix, and a Gaussian distribution respectively.

> so that it can eventually be incorporated into Pennylane.

In addition, if you want to contribute to PennyLane, let me know I can direct you to some 'good first issues' in the PennyLane repository 🙂

Lucaman99:
@josh146 Thanks for your comment, this was super helpful. I just created a method called `parity_expectation` in `backend/state.py` file, inside the `BaseFockState` class. I also created a basic test case. 

I'm going to create a "Work in Progress" pull request, so you can take a look and make sure I'm not making any big mistakes. 😄 

Also, the method I implemented unfortunately only supports pure states right now (I'm not totally sure how to go about implementing this function in the code for mixed states, do you have any suggestions?)

> let me know I can direct you to some 'good first issues' in the PennyLane repositor

That would be great, thanks so much! I'd love to try to tackle a few more issues!

thisac:
Hi @Lucaman99. Thank you very much for contributing to Strawberry Fields. We'll have a closer look at the PR and get back to you with any feedback we might have. :slightly_smiling_face:

PR #394, opened 2020-05-20
Adds the Parity Operator for States in the Gaussian Backend (WIP)

Lucaman99:
**Context:**

https://github.com/XanaduAI/strawberryfields/issues/386

Adds the parity expectation value function to StrawberryFields for Gaussian backend states (previously only implemented for states in the Fock backend).

**Benefits:**

Parity expectations will now work for any state in any backend.

**Possible Drawbacks:**

None

**Related GitHub Issues:**

https://github.com/XanaduAI/strawberryfields/issues/386

### DIFF
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -1257,8 +1257,8 @@ class BaseGaussianState(BaseState):
         if len(modes) != len(set(modes)):
             raise ValueError("There can be no duplicates in the modes specified.")
 
-        mu = self._mu
-        cov = self._cov
+        mu = self.means()
+        cov = self.cov()
         num = np.exp(-(0.5) * (mu @ (np.linalg.inv(cov) @ mu)))
         parity = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
 
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -1259,8 +1259,8 @@ class BaseGaussianState(BaseState):
 
         mu = self._mu
         cov = self._cov
-        new = mu @ (np.linalg.inv(cov) @ mu)
-        num = np.exp(-(0.5) * new)
+        number = mu @ (np.linalg.inv(cov) @ mu)
+        num = np.exp(-(0.5) * number)
         new_num = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
 
         return new_num
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -1259,8 +1259,8 @@ class BaseGaussianState(BaseState):
 
         mu = self._mu
         cov = self._cov
-        number = mu @ (np.linalg.inv(cov) @ mu)
-        num = np.exp(-(0.5) * number)
+        
+        num = np.exp(-(0.5) * (mu @ (np.linalg.inv(cov) @ mu)) )
         parity = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
 
         return parity
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -1263,7 +1263,7 @@ class BaseGaussianState(BaseState):
         num = np.exp(-(0.5) * number)
         new_num = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
 
-        return new_num
+        return parity
 
     @abc.abstractmethod
     def reduced_dm(self, modes, **kwargs):
--- a/strawberryfields/backends/states.py
+++ b/strawberryfields/backends/states.py
@@ -25,9 +25,11 @@ from scipy.linalg import block_diag
 from scipy.stats import multivariate_normal
 from scipy.special import factorial
 from scipy.integrate import simps
+
 from thewalrus.quantum import photon_number_mean, photon_number_covar
 
 import strawberryfields as sf
+
 from .shared_ops import rotation_matrix as _R
 from .shared_ops import changebasis
 
@@ -1249,6 +1251,19 @@ class BaseGaussianState(BaseState):
         raise ValueError(
             "The number_expectation method only supports one or two modes for Gaussian states."
         )
+
+    def parity_expectation(self, modes):
+        """Calculates the expectation value of a product of parity operators acting on given modes"""
+        if len(modes) != len(set(modes)):
+            raise ValueError("There can be no duplicates in the modes specified.")
+
+        mu = self._mu
+        cov = self._cov
+        new = mu @ (np.linalg.inv(cov) @ mu)
+        num = np.exp(-(0.5) * new)
+        new_num = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
+
+        return new_num
 
     @abc.abstractmethod
     def reduced_dm(self, modes, **kwargs):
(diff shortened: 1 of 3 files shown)


Lucaman99:
@josh146 @nquesada @antalszava Tests are passing, let me know if any of you have comments/feedback! 🙂

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1252,19 @@ def number_expectation(self, modes):
             "The number_expectation method only supports one or two modes for Gaussian states."
         )
 
+    def parity_expectation(self, modes):
+        """Calculates the expectation value of a product of parity operators acting on given modes"""
+        if len(modes) != len(set(modes)):
+            raise ValueError("There can be no duplicates in the modes specified.")
+
+        mu = self._mu
```
nquesada, line 1260: I think there a methods to access this hidden attributes; you should try to use them,

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1252,19 @@ def number_expectation(self, modes):
             "The number_expectation method only supports one or two modes for Gaussian states."
```
nquesada, line 1252: This only should be removed.

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1252,19 @@ def number_expectation(self, modes):
...
+    def parity_expectation(self, modes):
+        """Calculates the expectation value of a product of parity operators acting on given modes"""
+        if len(modes) != len(set(modes)):
+            raise ValueError("There can be no duplicates in the modes specified.")
+
+        mu = self._mu
+        cov = self._cov
+        number = mu @ (np.linalg.inv(cov) @ mu)
+        num = np.exp(-(0.5) * number)
+        new_num = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
```
nquesada, line 1264:
```suggestion
        parity = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
```

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1252,19 @@ def number_expectation(self, modes):
...
+
+        return new_num
```
nquesada, line 1266:
```suggestion
        return parity
```

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1252,19 @@ def number_expectation(self, modes):
...
 
+    def parity_expectation(self, modes):
+        """Calculates the expectation value of a product of parity operators acting on given modes"""
+        if len(modes) != len(set(modes)):
+            raise ValueError("There can be no duplicates in the modes specified.")
+
+        mu = self._mu
+        cov = self._cov
+        number = mu @ (np.linalg.inv(cov) @ mu)
+        num = np.exp(-(0.5) * number)
```
nquesada, line 1263: I would suggest to merge this and the previous line.

strawberryfields/backends/states.py
Lucaman99: I didn't add this, it was already there. Do you still want it removed?

Lucaman99: I didn't add this, do you still want it removed?

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1252,18 @@ def number_expectation(self, modes):
             "The number_expectation method only supports one or two modes for Gaussian states."
         )
 
+    def parity_expectation(self, modes):
+        """Calculates the expectation value of a product of parity operators acting on given modes"""
```
antalszava, line 1256: Now that the ParityOperator will be supported by the `BaseGaussianState` class too, a more general `parity_expectation` method can be defined for the parent [`class BaseState(abc.ABC)`](https://github.com/XanaduAI/strawberryfields/blob/a9d12b986689377893f47a07a6fa1763cc1cd821/strawberryfields/backends/states.py#L37) (linking the locations pointing to the source code throughout this comment).

This way a shared docstring can be used here.

**Changes to be done**

For this, the following can be done:
1. Creating a `parity_expectation` method for the `BaseState` class for example [in between the `def number_expectation(self, modes):` and `def p_quad_values(self, mode, xvec, pvec):` definitions](https://github.com/XanaduAI/strawberryfields/blob/a9d12b986689377893f47a07a6fa1763cc1cd821/strawberryfields/backends/states.py#L368):
```python
    def parity_expectation(self, modes):
        """Calculates the expectation value of a product of parity operators acting on given modes"""
        raise NotImplementedError

```
2. Removing the [docstring of the `parity_expectation` method in the `BaseFockState` class](https://github.com/XanaduAI/strawberryfields/blob/a9d12b986689377893f47a07a6fa1763cc1cd821/strawberryfields/backends/states.py#L924) .
3. Removing the docstring of the `parity_expectation` method here too.

**More details on why this is useful**

Each state class inherits from the `BaseState` class (see for example the [`BaseFockState`](https://github.com/XanaduAI/strawberryfields/blob/a9d12b986689377893f47a07a6fa1763cc1cd821/strawberryfields/backends/states.py#L412)  and [`BaseGaussianState`](https://github.com/XanaduAI/strawberryfields/blob/a9d12b986689377893f47a07a6fa1763cc1cd821/strawberryfields/backends/states.py#L930) class which have `BaseState` in their definitions).

Class inheritance can be used not just for inheriting methods, but for inheriting docstrings too. Eventually, any class that inherits from `BaseState` will have a `parity_expectation` method defined (with the definition provided in step 1) this would simply be raising a `NotImplementedError`) and have the same shared docstring for the `parity_expectation` method.

So essentially here the the `parity_expectation` method would be uniquely defined for the `BaseGaussianState` class, but the docstring would be inherited from the `BaseState` class.

Lucaman99: Thanks for the detailed comments @antalszava! Would you prefer that these changes are made in a separate PR?

strawberryfields/backends/states.py
nquesada: I got confused. This line should stay.

strawberryfields/backends/states.py
antalszava: Sure thing! It is up to you, I think it's not too crucial to be done here. A follow-up PR could include making both this change and the change with `number_expectation`.

Lucaman99: Sounds good, I'll make the changes and open a new PR tomorrow 🙂

Lucaman99:
Hey @antalszava! I made the change that you requested, let me know what you think! (I've never come across inheritance of docstrings before, so I may have misinterpreted your comment)

antalszava:
Great! All looks good!

- [base_state_parity](https://user-images.githubusercontent.com/24476053/82681831-578b8480-9c1c-11ea-9b2b-bdf25b72fad4.png)
- [base_fock_parity](https://user-images.githubusercontent.com/24476053/82681783-4478b480-9c1c-11ea-9b1d-be82c6d6839c.png)
- [base_gaussian_parity](https://user-images.githubusercontent.com/24476053/82681819-5195a380-9c1c-11ea-8013-ff98c683becb.png)

strawberryfields/backends/states.py
```diff
@@ -1250,6 +1255,17 @@ def number_expectation(self, modes):
...
+        if len(modes) != len(set(modes)):
+            raise ValueError("There can be no duplicates in the modes specified.")
+
+        mu = self.means()
+        cov = self.cov()
+        num = np.exp(-(0.5) * (mu @ (np.linalg.inv(cov) @ mu)))
+        parity = ((self.hbar / 2) ** len(modes)) * num / (np.sqrt(np.linalg.det(cov)))
+
+        return parity
+
```
nquesada, line 1268: Nice!

PR #398, opened 2020-05-28
Unify backend results

thisac:
**Context:**
Currently, the `result.samples` are being reshaped in the frontend (`engine.py` and `results.py`). The backends save the samples in the registry, which are later picked out and reshaped into an array in the frontend. It would be much nicer if the backends could supply the frontend with samples with the correct shape and type directly.

Three major things have been done here:

1. Only the measured modes are returned via `result.samples`, i.e. all `None` placeholders have been removed. This fixes the issue with the `tf` backend returning a list of `tf.Tensor` objects and `None`, and instead it now returns a large `tf.Tensor` object with shape `(shots, measured_modes)`. The user would thus need to keep track of what modes are being measured, since there will no longer be any `None`-arrays returned.

2. The correct shape and type is handled and returned directly from the backends and no further post-handling is done by the frontend, unless several separate measurements are done in the same circuit. Each measurement is updated to return either an `np.ndarray` (for the `fock` and `gaussian` backends) or a `tf.tensor` (for the `tf` backend) with shape `(shots, measured_modes)`.

3. If several measurements are made in the same circuit, these are returned to the engine as a list of separate measurements. These are then combined by a new function `Results.combine_samples`, which returns the combined samples for all measurements with the correct shape: `(shots, measured_modes)`.

Also, some tests have been updated since the shape of samples has been changed, and further tests have been added for special cases.

**Benefits:**
The results passed from the backends are (hopefully) fully unified, which makes it easier to handle.

**Possible Drawbacks:**
Some of the measurements are simply wrapping the results in an array or a tensor (e.g. `np.array([[res]]`) which should work fine in those particular cases, since they only support a certain type of return; for example, the `measure_homodyne` in the `fock` and `gaussian` backends will only return single values that are now wrapped inside an `np.ndarray` to have shape `(shots, measured_modes)`. This will need to be changed if/when adding multiple-shots support, etc., but since that most likely will entail some further restructuring of the code, it's probably not to hard to update the drawback described above at the same time.

**Related GitHub Issues:**
N/A


strawberryfields/api/result.py
```diff
@@ -64,11 +64,6 @@ class Result:
     def __init__(self, samples, is_stateful=True):
         self._state = None
         self._is_stateful = is_stateful
-
-        # samples arrives as either a 2-D array (for shots > 1) or a 1-D array (for shots = 1)
-        # the latter needs to be converted to a 2-D array with shape (shots, modes)
-        if np.ndim(samples) == 1:
-            samples = np.array([samples])
```
josh146, line 71: Nice to remove this!

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -710,7 +710,8 @@ def measure_fock(self, modes, select=None):
         if select is not None:
             outcome = copy.copy(select)
 
-        return outcome
+        # `outcome` will always be a 1-d sequence since multiple shots is not supported
+        return np.array([outcome])
```
josh146, line 714: 👍

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -796,4 +797,5 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
         # Normalize
         self._state = self._state / self.norm()
 
-        return homodyne_sample
+        # `homodyne_sample` will always be a single value since multiple shots is not supported
+        return np.array([[homodyne_sample]])
```
josh146, line 801: 👍

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -223,11 +225,11 @@ def thermal_loss(self, T, nbar, mode):
         self.circuit.thermal_loss(T, nbar, mode)
 
     def measure_fock(self, modes, shots=1, select=None):
+        if select is not None:
+            raise NotImplementedError(
+                "Gaussian backend currently does not support " "postselection"
+            )
```
josh146, line 231: Can't we support post-selection for Gaussian backend `MeasureFock` if `shots=1`? @nquesada

strawberryfields/backends/gaussianbackend/backend.py
```diff
@@ -242,15 +244,13 @@ def measure_fock(self, modes, shots=1, select=None):
...
+
         # check we are sampling from a gaussian state with zero mean
         if allclose(mu, zeros_like(mu)):
             samples = hafnian_sample_state(reduced_cov, shots)
         else:
             samples = hafnian_sample_state(reduced_cov, shots, mean=reduced_mean)
-        # for backward compatibility with previous measurement behaviour,
-        # if only one shot, then we drop the shots axis
-        if shots == 1:
-            samples = samples.reshape((len(modes),))
```
josh146, line 253: 🔥

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -625,21 +625,14 @@ def measure_fock(self, modes, select=None, **kwargs):
             meas_result = ops.unravel_index(
                 sample_tensor, [self._cutoff_dim] * num_reduced_state_modes
             )
-            if not self._batched:
-                meas_result = meas_result[0]  # no batch index, can get rid of first axis
```
josh146, line 629: @thisac, could this be why the batched tests are failing? Removing this if statement will mean that the TF backend will always return a batched dimension in returned samples, even if `batch_size=1`.

My gut feeling is that we should keep the existing behaviour; when `batch_size=1`, the returned tensors should be of shape `(shots, modes)`, and when `batch_dim>1`, they should be of shape `(batch_size, shots, modes)`.

strawberryfields/engine.py
```diff
@@ -236,12 +236,6 @@ def _run(self, program, *, args, compile_options, **kwargs):
             Result: results of the computation
         """
 
-        def _broadcast_nones(val, dim):
-            """Helper function to ensure register values have same shape, even if not measured"""
-            if val is None and dim > 1:
-                return [None] * dim
-            return val
```
josh146, line 243: Awesome 🔥

strawberryfields/engine.py
```diff
@@ -359,7 +350,11 @@ def _run_program(self, prog, **kwargs):
                         cmd.op, self.backend, kwargs
                     )
                 ) from None
-        return applied
+
+        if values is None:
+            values = np.array([[]])
+
+        return applied, values
```
josh146, line 357: Is the returned `applied` never used? 🤔

strawberryfields/ops.py
```diff
@@ -317,20 +317,13 @@ def apply(self, reg, backend, **kwargs):
...
         values = super().apply(reg, backend, **kwargs)
-        # convert the returned values into an iterable with the measured modes indexed along
-        # the first axis and shots along second axis (if larger than 1), so that we can assign
-        # register values
-        shots = kwargs.get("shots", 1)
-        if self.ns == 1:
-            values = [values]  # values is either a scalar, or has shape (shots,)
-        else:
-            if shots > 1:
-                values = values.T  # shape of values would be (shots, num_meas,)
```
josh146, line 328: So nice that these can all be deleted!

strawberryfields/parameters.py
```diff
@@ -350,6 +350,13 @@ def _eval_evalf(self, prec):
                     self
                 )
             )
+
+        # remove unnecessary dims when returning measurement
+        try:
+            res = np.squeeze(res).item()
+        except ValueError:
+            res = np.squeeze(res)
```
josh146, line 358: As far as I can tell, this will only affect the case where measured registers are used in subsequent ops:

```python
MeasureX | q[0]
Xgate(q[0].par) | q[1]
```

I don't _think_ this affects the TF backend in any way, since the above is not differentiable, but just wanted to double check.

tests/integration/test_tf_integration.py
```diff
@@ -92,9 +92,8 @@ def test_eng_run_measurements_are_tensors(self, setup_eng):
             Dgate(ALPHA) | q
             MeasureX | q
 
-        eng.run(prog)
-        val = q[0].val
-        assert isinstance(val, tf.Tensor)
+        res = eng.run(prog)
+        assert isinstance(res.samples, tf.Tensor)
```
josh146, line 96: 💯

strawberryfields/backends/gaussianbackend/backend.py
thisac: I moved this part outside of `if shots != 1:` since the `select` keyword doesn't actually seem to matter (and post-selection didn't seem to be supported for single-shots with the Gaussian backend).

strawberryfields/parameters.py
thisac: I was thinking the same. Your example would be failing earlier, since e.g. the `Xgate` wouldn't accept a nested `ndarray` as input, which is the reason for this part of the code; to turn `np.array([[3]])` and `tf.Tensor([[3]])` into `3`, and `np.array([[1, 2]])`  and `tf.Tensor([[1, 2]])` into `np.array([1, 2])`, etc.

josh146: When using the TensorFlow backend, will `res` be a TensorFlow tensor? e.g.:

```python
prog = sf.Program(2)

with prog.context as q:
    MeasureX | q[0]
    Xgate(q[0].par) | q[1]

eng = sf.Engine("tf", backend_options={"cutoff_dim":5})
eng.run(prog)
```

Will this fail because `np.squeeze` won't know how to deal with a Tensor?

thisac: I don't think it will, since `np.squeeze` will change the `tf.Tensor` into a `np.ndarray` and then "squeeze" it. At least that's how it seems to work. Your code-snippet works fine locally for me.

josh146: Awesome

strawberryfields/backends/tfbackend/circuit.py
thisac: Thanks @josh146. That was part of the cause. Some more issues popped up in a few tests; should be fixed now. When batched the `tf` backend returns shape `(batch_size, shots, measured_modes)`.

strawberryfields/engine.py
thisac: It doesn't look like it is. :thinking: I guess it could be removed then?

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -893,7 +897,10 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
             self._update_state(new_state)
 
         # `meas_result` will always be a single value since multiple shots is not supported
-        return tf.cast([[meas_result]], dtype=ops.def_type)
+        if self.batched:
+            return tf.cast([[[i]] for i in meas_result], dtype=ops.def_type)
+        else:
+            return tf.cast([[meas_result]], dtype=ops.def_type)
```
josh146, line 903:
```suggestion
        return tf.cast([[meas_result]], dtype=ops.def_type)
```

strawberryfields/engine.py
```diff
@@ -277,15 +271,35 @@ def _broadcast_nones(val, dim):
...
-            self._run_program(p, **kwargs)
-            shots = kwargs.get("shots", 1)
-            self.samples = [_broadcast_nones(p.reg_refs[k].val, shots) for k in sorted(p.reg_refs)]
+            _, values = self._run_program(p, **kwargs)
             self.run_progs.append(p)
 
+            if len(values) > 1:
+                sort_order = [
+                    r.ind for c in p.circuit for r in c.reg if "Measure" in c.op.__str__()
+                ]
```
josh146, line 280: This could be turned into a nested list as per my previous comment:
```python
sort_order = [np.array([r.ind for r in c.reg]) for c in p.circuit if "Measure" in c.op.__str__()]
```

tests/frontend/test_engine.py
```diff
@@ -170,6 +170,40 @@ def inspect():
...
+            ops.MeasureFock() | q[0]
+
+        result = eng.run(prog)
+
+        # check that shape is (shots, measured_modes)
+        assert result.samples.shape == (1, 4)
+
+        # check that MesureFock measures `0` while MeasureX does NOT measure `0`.
+        correct_samples = [0, 0, 1, 0]
+        print(result.samples)
```
josh146, line 190:
```suggestion
```

tests/frontend/test_engine.py
```diff
@@ -170,6 +170,40 @@ def inspect():
...
+        assert [bool(i) for i in result.samples[0]] == correct_samples
```
josh146, line 191: Better to do it like this?
```suggestion
        assert all(i == j for i, j in zip(result.samples[0], correct_samples))
```

tests/integration/test_engine_integration.py
```diff
@@ -304,20 +304,17 @@ def test_measurefock_shots(self, setup_eng):
...
-        assert all(samples[:, 2].astype(int) == expected)
+        assert all(samples[:, 1].astype(int) == expected)
 
         # one mode
         eng, p3 = setup_eng(3)
         with p3.context as q:
             ops.MeasureFock() | q[0]
         samples = eng.run(p3, shots=shots).samples
-        assert samples.shape == (shots, 3)
+        assert samples.shape == (shots, 1)
```
josh146, line 316: Good catch 🤔

strawberryfields/api/result.py
```diff
@@ -107,6 +102,28 @@ def state(self):
...
+        ret = [[] for _ in samples_list[0]]
+        for i, samples in enumerate(samples_list):
+            for j, single_sample in enumerate(samples):
+                for s in single_sample:
+                    ret[j].append(s)
+
+        for i, sam in enumerate(ret):
+            ret[i] = [j for _, j in sorted(zip(mode_order, sam))]
+
+        return np.array(ret)
```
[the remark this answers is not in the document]
josh146: Although, you would have to be smart in the `dtype` used when initializing `ret`; either an int, float, or complex depending on what types of measurements are present.

strawberryfields/api/result.py
```diff
@@ -107,6 +102,28 @@ def state(self):
             raise AttributeError("The state is undefined for a stateless computation.")
         return self._state
 
+    @staticmethod
+    def combine_samples(samples_list, mode_order):
+        """Combine samples measured at different times into one nested array
+
+        Args:
+            samples_list (list): the sample measurements in a list
+            mode_order (list): the mode order for the measured modes
```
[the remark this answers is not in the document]
thisac: > * What is the structure of `samples_list`? It's a bit hard to tell from the docstring.

`samples_list` should be a list with arrays or tensors (depending on the backend), as `list[array, tensor]`. I'll add that to the docstring.

> * Is `mode_order` also nested? Or is it just a flat like, like `[0, 1, 5]` if modes 0, 1, 5 are measured?

`mode_order` shouldn't be nested. It's just a flat list as per your example.

tests/frontend/test_engine.py
thisac: Oh, this wouldn't actually do the same thing. It should only check that only some values (the 1 in `correct_samples` is not equal to 0). Another solution would be  `all(np.array(result.samples[0], dtype=bool) == correct_samples)` or `all(bool(i) == j for i, j in zip(result.samples[0], correct_samples))`. The original solution is actually fastest (1.33 µs) while your suggestion, with the `bool` casting, comes a close second (1.75 µs) and the `bool` casting via `np.array` takes a whooping 4.16 µs.

strawberryfields/api/result.py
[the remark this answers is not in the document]
thisac: Just had another look at it, and there's an issue with the `for m, s in zip(mode_order, samples_list)` since `mode_order` only would include the measured modes, which could include values larger than the number of measured modes `modes`. E.g. if `mode_order = [array([1, 3, 4]), array([2, 6])]` using 7 modes but only measuring 5. Then, `modes = 5` and thus `ret[:, m]` would cause an out-of-bounds error for `m=array([2, 6])`.

Not sure if it's worth having another part re-calculating `mode_order` into, in the example above, `mode_order = [array([0, 2, 3]), array([1, 4])]` is worth it, either in `BaseEngine._run` or 
in `combine_samples`. It could be something like:

```python
sort_order = [
    np.array([r.ind for r in c.reg])
    for c in p.circuit
    if "Measure" in c.op.__str__()
]

s = [tup[0] for tup in sorted(enumerate(sort_order), key=lambda x: x[1])]

res = []
start = 0
for x in sort_order:
    res.append(np.array(s[start:start + len(x)]))
    start = len(x)
sort_order = res
```

but that's quite a lot of _extra_ code. What do you think about this, or do you have another idea for a solution? There's probably a way to to this in a nicer way, but I cannot really think of any solution right now. :sleepy:

strawberryfields/api/result.py
```diff
@@ -107,6 +102,28 @@ def state(self):
...
+    @staticmethod
+    def combine_samples(samples_list, mode_order):
+        """Combine samples measured at different times into one nested array.
+
+        Args:
+            samples_list (list[array, tensor]): the sample measurements in a list
+            mode_order (list[int]): the mode order for the measured modes
+
+        Returns:
+            array: the samples in raising mode order with shape ``(shots, measured_modes)``
```
antalszava, line 114: (minor suggestion)
```suggestion
            array: the samples in ascending mode order with shape ``(shots, measured_modes)``
```

strawberryfields/api/result.py
```diff
@@ -107,6 +102,28 @@ def state(self):
...
+        """
+        ret = [[] for _ in samples_list[0]]
```
antalszava, line 116: (Small suggestion): might be worth adding a small comment e.g. `# Initializing empty lists for each measurement` or renaming `ret` to be more descriptive

strawberryfields/api/result.py
```diff
@@ -107,6 +102,28 @@ def state(self):
...
+        for i, samples in enumerate(samples_list):
```
antalszava, line 117:
```suggestion
        for samples in samples_list:
```

strawberryfields/api/result.py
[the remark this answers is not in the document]
antalszava: this might be quite a complicated task indeed and would need more refined logic I guess! Having a nested `np.argsort` would be cool if the list was flattened, but flattening and unflattering it might also be costly.

I cooked up this one (on first glance I think there's no clear advantage to it at all :smile: just thought of leaving it here if it gives some ideas):
```python
import numpy as np

mode_order = [np.array([1, 3, 4]), np.array([2, 6])]
modes = np.concatenate(mode_order)
all_modes = np.arange(np.max(modes))
remainders = set(all_modes) - set(modes)

for modes in mode_order:
    for i in remainders:
        modes[modes>=i] -= 1
```
```
[array([0, 2, 3]), array([1, 4])]
```

strawberryfields/api/result.py
```diff
@@ -107,6 +102,28 @@ def state(self):
...
+            for j, single_sample in enumerate(samples):
```
antalszava, line 118: (Small suggestion): how about the `single_sample` -> `samples_for_run` renaming or so? `single_sample` sounds like a single outcome on a single mode

strawberryfields/backends/fockbackend/circuit.py
antalszava: hmm, out of scope but find it odd that it's a single mode here whereas it was a list of modes for `measure_fock`. do you perhaps know why this difference is there?

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -577,7 +577,7 @@ def measure_fock(self, modes, select=None, **kwargs):
 
         if select is not None:
             # just use the supplied measurement results
-            meas_result = select
+            meas_result = tf.convert_to_tensor([[i] for i in select], dtype=tf.int64)
```
antalszava, line 580: I've seen `ops.def_type` being used later, just curious if it would be good to have it here too?

strawberryfields/engine.py
antalszava: just curious if `applied` should eventually be removed

strawberryfields/engine.py
```diff
@@ -341,11 +355,13 @@ def _init_backend(self, init_num_subsystems):
...
     def _run_program(self, prog, **kwargs):
         applied = []
+        values = []
         for cmd in prog.circuit:
             try:
                 # try to apply it to the backend
-                # NOTE we could also handle storing measured vals here
-                cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                val = cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                if val is not None:
```
antalszava, line 363: (Minor suggestion): might be worth adding a small comment e.g. `# if op was a measurement then store the samples returned`

strawberryfields/engine.py
```diff
@@ -277,15 +271,35 @@ def _broadcast_nones(val, dim):
                 p = p.compile(target, **compile_options)
             p.lock()
 
-            self._run_program(p, **kwargs)
-            shots = kwargs.get("shots", 1)
-            self.samples = [_broadcast_nones(p.reg_refs[k].val, shots) for k in sorted(p.reg_refs)]
+            _, values = self._run_program(p, **kwargs)
```
antalszava, line 274: just for my own understanding: was it not necessary to have the `values` be returned because previously we also checked the `RegRef` instances and built up `self.samples` that way?

strawberryfields/engine.py
```diff
@@ -277,15 +271,35 @@ def _broadcast_nones(val, dim):
...
+            if len(values) > 1:
+                sort_order = [
+                    r.ind for c in p.circuit for r in c.reg if "Measure" in c.op.__str__()
+                ]
+
+                # check for duplicate mode-measures
+                if len(sort_order) != len(set(sort_order)):
+                    raise RuntimeError("Modes can only be measured once inside a circuit.")
+
+                self.samples = Result.combine_samples(values, sort_order)
```
antalszava, line 286: could it be worth having `combine_samples` be a method of `Engine`?

strawberryfields/engine.py
antalszava: both are very brief and great solutions! I'd personally suggest simply using `for` loops here so that each object can be well-named and readability could be increased

tests/backend/test_fock_measurement.py
```diff
@@ -123,7 +123,7 @@ def test_fock_measurements(self, setup_backend, cutoff, batch_size, pure, tol):
             ref_result = n[meas_modes]
 
             if batch_size is not None:
-                ref_result = tuple(np.array([i] * batch_size) for i in ref_result)
+                ref_result = np.array([[i] * 2 for i in ref_result]).T.reshape((batch_size, 1, ref_result.shape[0]))
```
antalszava, line 126: could you help me out why ` * 2` is included?

strawberryfields/api/result.py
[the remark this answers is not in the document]
josh146: > if mode_order = [array([1, 3, 4]), array([2, 6])] using 7 modes but only measuring 5. Then, modes = 5 and thus ret[:, m] would cause an out-of-bounds error for m=array([2, 6])

Oh! Good catch. This sounds like a solution for `np.argsort` and `np.split`, as @antalszava suggested, but then you have to also take into account the flattening.

The cost might not be that high, since we know the maximum level of nesting. E.g.,

```python
>>> mode_order = [np.array([0, 5, 2, 9, 6]),  np.array([3, 10])]
>>> shapes = np.cumsum([len(m) for m in mode_order])
array([5, 7])
>>> np.split(np.argsort(np.argsort(np.hstack(mode_order))), shapes)[:-1]
[array([0, 3, 1, 5, 4]), array([2, 6])]
```

strawberryfields/engine.py
thisac: Yes, that's correct. So, previously the samples were only built up from `RegRef` and then `None` was added to all the non-measured modes (with `_broadcast_nones`). No results were actually returned directly from the backends.

strawberryfields/engine.py
thisac: I'm not sure. I feel like the function has more to do with the results than the engine, so I think it might fit well inside `results.py`. @josh146, what do you think?

josh146: Yep, I tend to agree, I feel that it fits in more as a static method of the results class (but it is a very fuzzy line, and could arguably go in both places).

On a tangent, I think it was an oversight that the Results class got moved into the API module, since it is more general than just the API client? 🤔

strawberryfields/engine.py
thisac: Some tests seem to be using it at the moment, so I've left it for now. Perhaps this could be looked at in another PR?

tests/backend/test_fock_measurement.py
thisac: Oh, it should be `batch_size` instead of `2`. It just happens that `batch_size = 2` in these tests, so I think I wrote it in like that instead. The transposition is needed since the batches should be arranged opposite of what the list comprehension gives, and the reshaping since the results should be nested (i.e. adding the `1` inside the shape `(batch_size, 1, ref_result.shape[0])`.

antalszava: sounds reasonable, thanks!

strawberryfields/api/result.py
```diff
@@ -107,6 +102,43 @@ def state(self):
...
+        # create the correct mode order if not all modes are measured
+        if np.max(mode_order_flat) + 1 != modes:
+            shapes = np.cumsum([len(m) for m in mode_order])
+            mode_order = np.split(
+                # argsort twice returns the rank
+                np.argsort(np.argsort(mode_order_flat)),
+                shapes,
+            )[:-1]
+
+        # find the "widest" type in samples_list
```
josh146, line 130: 😆

strawberryfields/api/result.py
```diff
@@ -107,6 +102,43 @@ def state(self):
...
+        type_list = ["c", "f", "i"]
+        while type_list[0] not in [np.array(s).dtype.kind for s in samples_list]:
+            del type_list[0]
+
+        ret = np.empty([shots, modes], dtype=type_list[0] + "8")
```
josh146, line 135: Nice! I wasn't aware of the string representation of NumPy dtypes, learnt something new.

strawberryfields/engine.py
```diff
@@ -277,15 +272,57 @@ def _broadcast_nones(val, dim):
...
+            program (Program): quantum program that is run
+        """
+        # get nested order of measured modes
+        sort_order = [
+            np.array([r.ind for r in c.reg]) for c in program.circuit if "Measure" in c.op.__str__()
+        ]
+
+        # check for duplicate mode-measures
+        if len(np.hstack(sort_order)) != len(set(np.hstack(sort_order))):
+            raise RuntimeError("Modes can only be measured once inside a circuit.")
```
josh146, line 308: Is this true? For example, what about the following case?
```python
with prog.context as q:
    Sgate(0.1) | q[0]
    MeasureX | q[0]  # resets q[0] to the vacuum
    Xgate(q[0].par) | q[0]
    MeasureFock() | q[0]
```

strawberryfields/engine.py
```diff
@@ -277,15 +272,57 @@ def _broadcast_nones(val, dim):
...
+            _, values = self._run_program(p, **kwargs)
             self.run_progs.append(p)
 
+            # check number of measurments and combine samples if needed
+            if len(values) > 1:
+                self._combine_samples(values, p)
+            elif len(values) == 1:
+                self.samples = values[0]
+            else:  # if no measurement was made
+                self.samples = np.array([[]])
```
josh146, line 284: Just out of curiosity, should we be returning `None` here? By returning an empty NumPy array, we are returning something with shape `(1, 0)`, which violates the `(shots, modes)` convention.

On the other hand, `bool(np.array([[]])) == False`, so it would work well in `if` statements (`if results.samples:`). However, this raises a deprecation warning, so this behavior will likely be dropped soon 🤔

strawberryfields/engine.py
```diff
@@ -277,15 +272,57 @@ def _broadcast_nones(val, dim):
...
+        if self.backend_name == "tf":
+            from tensorflow import convert_to_tensor
+
+            if self.backend_options.get("batch_size", 0):  # if batches are used
+                self.samples = []
+                for i in range(self.backend_options.get("batch_size", 0)):
+                    # choose a single batch from values and combine the samples; do for all batches
+                    single_batch_values = [v[i] for v in values]
+                    single_batch_samples = Result.combine_samples(single_batch_values, sort_order)
+                    self.samples.append(single_batch_samples)
```
josh146, line 320: Just to double check what is happening here, we are essentially 'vectorizing' (in the loose sense of the word, since this is a for-loop) the `combine_samples()` function for each batch dimension?

strawberryfields/engine.py
```diff
@@ -341,11 +378,13 @@ def _init_backend(self, init_num_subsystems):
...
     def _run_program(self, prog, **kwargs):
         applied = []
+        values = []
         for cmd in prog.circuit:
             try:
-                # try to apply it to the backend
-                # NOTE we could also handle storing measured vals here
-                cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                # try to apply it to the backend and, if op is a measurement, store it in values
+                val = cmd.op.apply(cmd.reg, self.backend, **kwargs)
```
josh146, line 385: Question related to my comment above re: `RuntimeError("Modes can only be measured once inside a circuit.")`.

Previously, if we ran the program 
```python
with prog.context as q:
    Sgate(0.1) | q[0]
    MeasureX | q[0]  # resets q[0] to the vacuum
    Xgate(q[0].par) | q[0]
    MeasureFock() | q[0]
```

we extracted the returned samples by doing `[r.val for r in regrefs]`. So even though `q[0]` would be measured twice, only the final measured value would be provided (as the measured value of each regref gets overwritten with each measurement).

Now, however, the measurement values are returned directly from the backends, so `values` will have _two_ measured values of `q[0]` in the list, triggering the above exception.

We have two choices here:

1. Keep the existing behaviour, by having `combine_samples` simply take the _last_ value corresponding to modes that appear twice.

2. Allow modes to be measured twice within `combine_samples`, e.g., the above program would give

   ```python
   >>> results.samples
   array([[0.7, 1], [0.1, 0], ...])
   ```
   where column 0 corresponds to the first temporal q[0] measurement, and column 1 corresponding to the second temporal q[1] measurement.

   This changes are `(shots, modes)` convention into `(shots, measurements)`.

---

Side note: I'm realizing now that there will be a second consequence of this PR; samples will now be returned corresponding to the order that the user performed the measurements. E.g., if the user does

```python
MeasureFock | q[1]
MeasureFock | q[0]
```

then the returned samples will be ordered [mode 1, mode 0].

I think this makes sense to some degree, _but_ the more I think about this (and in addition with PR #399), the more I think it makes sense to turn `results.samples` into a class object, so we can provide the additional metainformation.

```pycon
>>> samples.num_modes
4
>>> samples.num_shots
500
>>> samples.num_measurements
6
>>> samples.modes
[0, 2, 1, 3, 1, 0]
>>> samples.mtypes
['Homodyne', 'Fock', ...]
>>> samples.values.shape
[500, 6]
```

strawberryfields/api/result.py
thisac: I just learnt about it as well (with a bit of trial-and-error). `dtype="c"` doesn't really work as a though, so I had to add the `"8"`, also making the `int` and `float` type into 64 bit types, while `dtype="i"` and `dtype="f"` would create 32 bit types. :slightly_smiling_face:

strawberryfields/engine.py
thisac: I added this since I thought it made sense only measuring once per mode per circuit. As you mentioned further down this can, and maybe should, be fixed so that only the last measurement is saved.

strawberryfields/engine.py
antalszava: having a `None` value here sounds reasonable, the if statement could then look as `if results.samples is not None:`

strawberryfields/engine.py
thisac: > 1. Keep the existing behaviour, by having `combine_samples` simply take the _last_ value corresponding to modes that appear twice.

This would make more sense with how things work in this PR.

> 2. Allow modes to be measured twice within `combine_samples`

I'm not sure if this is actually something that's wanted/needed. Perhaps it's nice to have, although it would completely change what is returned to the user.

> Side note: I'm realizing now that there will be a second consequence of this PR; samples will now be returned corresponding to the order that the user performed the measurements. 

This shouldn't be the case since the mode order is taken into account when combining the samples; e.g. `samples = [array([[0, 1]]), array([[2]])]` and `mode_order = [array([0, 3]), array([1])]` would return `array([[0, 2, 1]])` to the user, which would be in ascending mode order (as the docstring for `combine_samples` also states).

> I think this makes sense to some degree, _but_ the more I think about this (and in addition with PR #399), the more I think it makes sense to turn `results.samples` into a class object, so we can provide the additional metainformation.

I think I agree with you, but this maybe should be another PR, right? In that case we could leave this issue to be handled in that PR instead (including `RuntimeError("Modes can only be measured once inside a circuit.")`. Or what does everyone think?

strawberryfields/engine.py
antalszava: Personally I find it confusing to allow multiple measurements on the same mode, even if there is a sensible logic behind it that can be explained further. I'm on board with raising an error to inform the user about it.

The way I usually try to picture these circuits for myself is to think of a circuit diagram. In that sense, I agree with how SF Interactive now deals with measurements: once there was a measurement operator placed on a single wire, no other operations can be placed on that wire anymore. After all, the measurement should result in a "collapse of the wavefunction" on the related wire, right?

Therefore having the logic of resetting to vacuum once a measurement has been done simply would imply to me that we are trying to consider two different circuits.

```python
with prog.context as q:
    Sgate(0.1) | q[0]
    MeasureX | q[0]
```
```python
with prog.context as q:
    Xgate(q[0].par) | q[0]
    MeasureFock() | q[0]
```

strawberryfields/engine.py
thisac: I don't know what would be the best option here. Just changing to `self.samples=None` in the engine would cause issues when running sequential programs since it loops over `self.samples`. This could either be fixed by changing how the results are saved in `results.py`; or by simply not returning anything from the engine, and thus "returning" None; or by adding an `if self.samples is not None:` to line 263 in `engine.py` (above the `for k, v in enumerate(self.samples):`).

> By returning an empty NumPy array, we are returning something with shape (1, 0), which violates the (shots, modes) convention.

An option would be to have `self.samples = np.empty((0, 0))`, which would actually follow the `(shots, measured_modes)` convention. :laughing:

strawberryfields/engine.py
co9olguy: :pray:

strawberryfields/engine.py
co9olguy: If I remember, one of the reasons for allowing a mode to be reused after measurement is so that the user had some way to avoid adding ancillas (which can be costly in the fock space simulations).

Regardless of whether we think this feature makes sense or not, is it truly in-scope for this particular PR to change it?

strawberryfields/engine.py
co9olguy: > > 1. Allow modes to be measured twice within `combine_samples`
> 
> I'm not sure if this is actually something that's wanted/needed. Perhaps it's nice to have, although it would completely change what is returned to the user.
> 

Somewhat related, but I wanted to reiterate that measuring a mode multiple times is a supported feature of SF. It's not a new feature that's being considered. We have to consider the present PR in that context. If we feel the need to change a feature that's been in the library from the start, probably best to do as a separate discussion/PR

strawberryfields/parameters.py
co9olguy, line 358: This feels a bit inelegant/hacky to me. Both the `try/except` and the fact that we need to do the squeezing in this part of the stack. It signals to me that we still don't have a rock-solid "unified" contract between the backends and frontend

tests/backend/test_postselection.py
```diff
@@ -67,6 +67,6 @@ def test_measure_fock(setup_backend, cutoff, batch_size):
         photons_out = sum(res1 + res2)
 
         if batch_size is not None:
-            total_photons = np.tile(total_photons, batch_size)
+            total_photons = np.tile(total_photons, batch_size).sum()
```
co9olguy, line 70: was this a bug before?

tests/conftest.py
```diff
@@ -132,7 +136,7 @@ def backend(monkeypatch):
         m.setattr(dummy_backend, "squeeze", lambda r, modes: None)
         m.setattr(dummy_backend, "rotation", lambda r, modes: None)
         m.setattr(dummy_backend, "beamsplitter", lambda t, r, m1, m2: None)
-        m.setattr(dummy_backend, "measure_homodyne", lambda phi, modes, select, shots: 5)
+        m.setattr(dummy_backend, "measure_homodyne", lambda phi, modes, select, shots: np.array([[5]]))
```
co9olguy, line 139: I know I'm in the minority, but I don't like having to use singleton axes :frowning_face:

tests/frontend/test_engine.py
```diff
@@ -170,6 +170,39 @@ def inspect():
...
+        """Check that an error is raised when measuring the same mode twice"""
+
+        prog = sf.Program(5)
+        with prog.context as q:
+            ops.MeasureX | q[2]
+            ops.MeasureFock() | (q[1], q[3])
+            ops.MeasureFock() | q[2]
+
+        with pytest.raises(
+            RuntimeError, match="Modes can only be measured once inside a circuit."
```
co9olguy, line 203: This behaviour I think is out of scope for this PR, since it is effectively removing a currently supported feature of SF

tests/integration/test_engine_integration.py
```diff
@@ -150,6 +147,27 @@ def test_measured_parameter(self, setup_eng):
 
         eng.run(prog)
 
+    @pytest.mark.backends("tf")
+    @pytest.mark.skipif("BATCHED" not in os.environ, reason="Test for when combining batched samples")
```
co9olguy, line 151: How come we went with os.environ over previous method?

strawberryfields/engine.py
thisac: I agree that this might be a bit out-of-scope for this PR. The change mostly happened by mistake to be honest, since the changes caused other issues when combining the samples when measuring the same mode twice (thus this was the easy fix). Should've checked how it worked before rather than just removing that possibility. :slightly_frowning_face:

tests/conftest.py
thisac: I agree that it looks a bit ugly, but it might be a better long term solution. If e.g. later wanting to add support for measuring more modes and/or shots it would make more sense since they would all have the same shape. 

Maybe an option could be (as @josh146 suggested) having a sample class that could return things to the user in a "nicer" way, while still returning the samples from the backends with shape `(shots, measured_modes)` and storing them in that way inside the new class?

tests/integration/test_engine_integration.py
thisac: The previous way (using the `@property` decorated function) didn't seem to work. It was outside a class, where I believe `@property` isn't supported. When I tried using it, it didn't work as expected, so I think it was broken before. I changed it to check for the environment variable `BATCHED` instead since I couldn't figure out another nicer way (except for perhaps just having an `if batch_size is not None` inside the test.

co9olguy: I seem to remember it used to be that the testing framework itself picked these up from the environment and made them accessible via fixtures (long time ago, so things might have changed since then)

strawberryfields/engine.py
antalszava: That's a fair point, haven't thought of it needing to be iterable! I've found [this comment](https://stackoverflow.com/a/14640048) useful and overall agree with what is described there, that is it's good to return the same datatype. With these points in mind, I think that your last point on having `self.samples = np.empty((0, 0))` does seem pretty reasonable and nice actually! :smile:

tests/backend/test_postselection.py
thisac: Ah, thanks for pointing it out. It's (sort of) a misunderstanding of the test from my side, although I believe this should work as intended as well. 

The issues is a few lines up, with the line `photons_out = sum(res1 + res2)`. Earlier, the returns would have been either lists or tuples (for the `tf` backend) and thus the `+` would act as an append. Now the returns are arrays or tensors, meaning that the `+` would simply add the values at each index. Thus, either removing the `sum(...)` from line 67 or adding a `.sum()` as done here would solve the issue. :slightly_smiling_face:

josh146:
For the moment, I think the best solution is to have the `combine_samples()` method simply use the last measured value if a mode is represented by two measurements. This preserves the existing SF behavior.

strawberryfields/api/result.py
```diff
@@ -114,41 +115,51 @@ def combine_samples(samples_list, mode_order):
...
-        mode_order_flat = np.hstack(mode_order)
 
         shots = len(samples_list[0])
-        modes = len(set(mode_order_flat))
 
-        shapes = np.cumsum([len(m) for m in mode_order])
+        mode_order_flat = np.hstack(mode_order)
+        len_modes = len(set(mode_order_flat))
+
+        # remove duplicate mode-measurements and only return that modes last measurement
```
antalszava, line 124:
```suggestion
        # remove duplicate mode-measurements and only return the modes from the last measurement
```

thisac:
I've now updated the `combine_samples` method to only keep/return the last measured sample if measuring the same mode more than once. This should work for any number of shots/batches and any number of measurements at any time in the circuit.

It could be a bit slow for big circuits with many multiple measurements of the same mode being made, although hopefully the user's will try to avoid measuring the same mode several times (in which case this extra code is simply skipped).

strawberryfields/api/result.py
```diff
@@ -107,6 +103,70 @@ def state(self):
...
+                        samples_shot_list[j][k].remove(s_val)
+
+            # reshape into a list of arrays, ignoring empty ones
+            mode_order = [np.array(m) for m in mode_order_list if m]
+            samples_list = [np.array([s]) for s in zip(*samples_shot_list) if s[0]]
+
+            shapes = np.cumsum([len(m) for m in mode_order])
+
+        # create the correct mode order if not all modes are measured
+        mode_order_flat = np.hstack(mode_order)
```
antalszava, line 151: (seems to be already there on line 121)
```suggestion
```

strawberryfields/api/result.py
```diff
@@ -107,6 +103,70 @@ def state(self):
...
+                same structure as samples_list.
+
+        Returns:
+            array: the samples in ascending mode order with shape ``(shots, measured_modes)``
+        """
+
+        shots = len(samples_list[0])
+
+        mode_order_flat = np.hstack(mode_order)
+        len_modes = len(set(mode_order_flat))
```
antalszava, line 122: (Minor suggestion): `np.unique` seems to perform better 
```suggestion
        len_modes = len(np.unique(mode_order_flat))
```
```python
nums = np.repeat(np.array([1, 2, 3]), 10000)
%timeit len(np.unique(nums))
%timeit len(set(nums))
```
outputs
```
248 µs ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
2.47 ms ± 14.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

strawberryfields/api/result.py
```diff
@@ -107,6 +103,70 @@ def state(self):
             raise AttributeError("The state is undefined for a stateless computation.")
         return self._state
 
+    @staticmethod
+    def combine_samples(samples_list, mode_order):
+        """Combine samples measured at different times into one nested array.
```
antalszava, line 108: Might be worth adding an example here

strawberryfields/api/result.py
```diff
@@ -107,6 +103,70 @@ def state(self):
...
+
+            # reshape into a list of arrays, ignoring empty ones
+            mode_order = [np.array(m) for m in mode_order_list if m]
+            samples_list = [np.array([s]) for s in zip(*samples_shot_list) if s[0]]
+
+            shapes = np.cumsum([len(m) for m in mode_order])
+
+        # create the correct mode order if not all modes are measured
+        mode_order_flat = np.hstack(mode_order)
+        if np.max(mode_order_flat) + 1 != len_modes:
```
antalszava, line 152: (Minor suggestion): could help readability
```suggestion
        not_all_modes_measured = np.max(mode_order_flat) + 1 != len_modes
        if not_all_modes_measured:
```

strawberryfields/api/result.py
```diff
@@ -107,6 +103,70 @@ def state(self):
...
+            array: the samples in ascending mode order with shape ``(shots, measured_modes)``
+        """
+
+        shots = len(samples_list[0])
+
+        mode_order_flat = np.hstack(mode_order)
+        len_modes = len(set(mode_order_flat))
+
+        # remove duplicate mode-measurements and only return the modes from the last measurement
+        if len(mode_order_flat) != len_modes:
```
antalszava, line 125: (Minor suggestion): could help readability
```suggestion
        duplicate_mode_measurements = len(mode_order_flat) != len_modes
        if duplicate_mode_measurements:
```

strawberryfields/api/result.py
```diff
@@ -114,41 +115,52 @@ def combine_samples(samples_list, mode_order):
...
-            samples_list = [np.array(s) for s in np.split(samples_list_trimmed, shapes)[:-1]]
-            mode_order = np.split(mode_order_trimmed, shapes)[:-1]
-
         # find the "widest" type in samples_list
         type_list = ["c", "f", "i"]
         while type_list[0] not in [np.array(s).dtype.kind for s in samples_list]:
             del type_list[0]
 
-        ret = np.empty([shots, modes], dtype=type_list[0] + "8")
+        ret = np.empty([shots, len_modes], dtype=type_list[0] + "8")
```
antalszava, line 163: could there be an overflow for `np.int8` or so?

strawberryfields/api/result.py
thisac: Yes, but `mode_order` has potentially changed (removing duplicates), so it might actually be needed here again. :slightly_smiling_face:

antalszava: oh that's a good one, thanks! :)

strawberryfields/api/result.py
thisac: Nice. I'll change that!

strawberryfields/api/result.py
thisac: I like this. Makes it a bit clearer why this is done. :+1:

strawberryfields/api/result.py
thisac: It shouldn't be a problem since

``` python
np.dtype("i8")
```
gives the output `dtype('int64')`, so it's actually a 64 bit int. :slightly_smiling_face:

strawberryfields/api/result.py
```diff
@@ -107,6 +103,72 @@ def state(self):
...
+                    for j in range(shots):
+                        # remove last sample in shot array so it doesn't disturb the indices
+                        s_val = samples_shot_list[j][k][idx[k][len(intersect) - i - 1]]
+                        samples_shot_list[j][k].remove(s_val)
+
+            # reshape into a list of arrays, ignoring empty ones
+            mode_order = [np.array(m) for m in mode_order_list if m]
+            samples_list = [np.array([s]) for s in zip(*samples_shot_list) if s[0]]
+
+            shapes = np.cumsum([len(m) for m in mode_order])
```
antalszava, line 149: This could perhaps be formulated as follows to break down the intermediate steps a little bit (had a quick chat with Theo on this, curious if any solution would be preferred overall):
```python
# 1. gather the indices of where removal needs to happen
modes_to_remove = []
modes_already = []
for idx, modes in enumerate(reversed(mode_order)):
    removal_indices = []
    for mode in modes:            
        if mode in modes_already:
            removal_indices.append(int(np.argwhere(modes==mode)))
        else:
            modes_already.append(mode)
    modes_to_remove.append((idx, np.array(removal_indices)))

# 2. Remove
for reversed_idx, removal_indices in modes_to_remove:
    idx = len(mode_order) - reversed_idx -1
    if removal_indices.size != 0:
        samples_list[idx] = np.delete(samples_list[idx], removal_indices, axis=1)
        mode_order[idx] = np.delete(mode_order[idx], removal_indices, axis=0)
print(samples_list, mode_order)
```

strawberryfields/api/result.py
antalszava: oh cool, was not aware of this shorthand! nice

strawberryfields/api/result.py
thisac: Thanks @antalszava. I think this is a nice solution as well, although I'm not sure which I prefer. I think it's better if someone could look at this with fresh eyes. @josh146 @co9olguy

josh146: @thisac I wonder if it might make more sense to address this at the _engine_ level rather than here. This would

* avoid this quite complicated logic, and allow it to be reverted to how it was before

* seems cleaner.

For example, in the `eng.run()` method, we have

```python
values = []
for cmd in prog.circuit:
    try:
        # try to apply it to the backend
        # try to apply it to the backend and, if op is a measurement, store it in values
        # NOTE we could also handle storing measured vals here
        val = cmd.op.apply(cmd.reg, self.backend, **kwargs)
        
        if val is not None:
            values.append(val)
```

Perhaps we could make `values` a _dictionary_, indexed by mode number via `cmd.reg`? That way, when a mode is measured again, the values returned overwrite the previous values automatically.

thisac: @josh146 That's actually a nice idea. I feel like that maybe should've been the way to go from the start. :smile: 

I actually started to think about whether it's actually worth having all of this logic in a separate `Results.combine_samples` function. Another solution, which looks a bit more like the earlier one, would be having all of the (much simpler) logic inside `engine.run`. In that case both `Results.combine_samples`, etc. could be removed, since e.g. the code below would return the samples correctly right away.

This seems to pass all the tests; where I've only changed the way the values or stored, as well as added a much simpler `_combining_samples` method:

```python
def _run_program(self, prog, **kwargs):
    applied = []
    values = {}
    batches = self.backend_options.get("batch_size", 0)
    for cmd in prog.circuit:
        try:
            # try to apply it to the backend and, if op is a measurement, store it in values
            val = cmd.op.apply(cmd.reg, self.backend, **kwargs)
            if val is not None:
                for i, r in enumerate(cmd.reg):
                    if batches:
                        values[r.ind] = np.array(val[:, :, i])
                    else:
                        values[r.ind] = val[:, i]

            applied.append(cmd)

        except NotApplicableError:
            # command is not applicable to the current backend type
            raise NotApplicableError(
                "The operation {} cannot be used with {}.".format(cmd.op, self.backend)
            ) from None

        except NotImplementedError:
            # command not directly supported by backend API
            raise NotImplementedError(
                "The operation {} has not been implemented in {} for the arguments {}.".format(
                    cmd.op, self.backend, kwargs
                )
            ) from None

    samples = self._combine_samples(values, kwargs["shots"], batches)

    return applied, samples
```
The `_combine_samples` method above (which can probably be written in a nicer way as well) would be:

```python
def _combine_samples(self, values, shots, batches):
    if values == {}:
        return []

    samples = np.array([[i[j] for _, i in sorted(values.items())] for j in range(shots)])

    if self.backend_name == "tf":
        from tensorflow import convert_to_tensor

        if batches:
            samples = convert_to_tensor(
                np.array(
                    [
                        [[i[b][j] for _, i in sorted(values.items())] for j in range(shots)]
                        for b in range(batches)
                    ]
                )
            )
        else:
            samples = convert_to_tensor(samples)

    return samples
```

This way, there is not that much extra logic trying to filter out and combine the samples afterwards, and I feel like it's a bit clearer to understand what's going on. But then we might as well have just kept the old way of getting the samples from the register instead. :thinking:

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -899,7 +896,10 @@ def measure_homodyne(self, phi, mode, select=None, **kwargs):
 
             self._update_state(new_state)
 
-        return tf.cast(meas_result, dtype=ops.def_type)
+        # `meas_result` will always be a single value since multiple shots is not supported
+        if self.batched:
+            return tf.cast([[[i]] for i in meas_result], dtype=ops.def_type)
```
josh146, line 901: Can the list comprehension be removed here? It feels like something that is achievable with a `tf.reshape`/`tf.expand_dims` and `tf.cast`.

strawberryfields/engine.py
```diff
@@ -277,15 +272,12 @@ def _broadcast_nones(val, dim):
                 p = p.compile(target, **compile_options)
             p.lock()
 
-            self._run_program(p, **kwargs)
-            shots = kwargs.get("shots", 1)
-            self.samples = [_broadcast_nones(p.reg_refs[k].val, shots) for k in sorted(p.reg_refs)]
+            _, self.samples = self._run_program(p, **kwargs)
```
josh146, line 275: This is so much cleaner!

strawberryfields/engine.py
```diff
@@ -341,25 +333,61 @@ def _init_backend(self, init_num_subsystems):
...
-                # NOTE we could also handle storing measured vals here
-                cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                # try to apply it to the backend and, if op is a measurement, store it in values
+                val = cmd.op.apply(cmd.reg, self.backend, **kwargs)
+                if val is not None:
+                    for i, r in enumerate(cmd.reg):
+                        if batches:
+                            samples_dict[r.ind] = val[:, :, i]
+                        else:
+                            samples_dict[r.ind] = val[:, i]
```
josh146, line 347: This is very clear and nice!

strawberryfields/engine.py
```diff
@@ -341,25 +333,61 @@ def _init_backend(self, init_num_subsystems):
...
+
+        samples = self._combine_and_sort_samples(samples_dict)
+
+        return applied, samples
+
+    def _combine_and_sort_samples(self, samples_dict):
+        """Helper function to combine the values in the samples dictionary sorted by its keys."""
+        batches = self.backend_options.get("batch_size", 0)
+
+        if samples_dict == {}:
```
josh146, line 373: Maybe better to have `if not samples_dict:` here?

strawberryfields/engine.py
```diff
@@ -341,25 +333,61 @@ def _init_backend(self, init_num_subsystems):
...
+            return np.empty((0, 0))
+
+        samples = np.transpose([i for _, i in sorted(samples_dict.items())])
```
josh146, line 376: :+1:

strawberryfields/ops.py
```diff
@@ -317,20 +317,13 @@ def apply(self, reg, backend, **kwargs):
...
 
         # store the results in the register reference objects
-        for v, r in zip(values, reg):
+        for v, r in zip(np.transpose(values), reg):
```
josh146, line 322: Nice :)

tests/frontend/test_engine.py
```diff
@@ -170,6 +170,43 @@ def inspect():
...
+            ops.MeasureX | q[2]
+
+        result = eng.run(prog)
+
+        # check that shape is (shots, measured_modes)
+        assert result.samples.shape == (1, 3)
+
+        # check that MesureFock measures `0` while MeasureX does NOT measure `0`.
+        correct_samples = [0, 1, 0]
+        assert [bool(i) for i in result.samples[0]] == correct_samples
```
josh146, line 209: 💯

PR #399, opened 2020-06-01
Counting statistics for post-processing samples

antalszava:
**Context:**
One can include [photon counting measurements](https://strawberryfields.readthedocs.io/en/stable/code/api/strawberryfields.ops.MeasureFock.html?highlight=measurefock#strawberryfields.ops.MeasureFock) within `Program` definitions. Doing so will result in obtaining PNR samples for the mode(s) on which the measurement acts. In general, samples obtained while running a Program on an `Engine` are contained in `Result.samples`. Such samples are in most cases contained in NumPy arrays having a default shape of `(shots, modes)`.

We would like to allow getting counting statistics on samples that were obtained after running a `Program` that solely contained photon counting measurements. This addition is aimed at backends `"fock"` and `"gaussian"`.

* Adds the `number_expectation_pnr` and `number_variance_pnr` functions that take PNR samples
* Adds user input validation checks
* Defines the `number_expectation` and `number_variance` shorthands

**Benefits:**
Counting statistics can be obtained for PNR samples.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

**TODO**
- [X] Further consideration on naming
- [X] Add further unit tests for expval and var
- [X] Adding examples
- [X] Formatting


antalszava:
* Getting a linting issue on `np.argwhere(samples == None)`. Using `samples is None` does not seem to fit here. (might be missing something)
* Open for suggestions on the file organization and/or namespace & import details

strawberryfields/api/post_processing.py
```diff
@@ -0,0 +1,181 @@
...
+    .. code-block:: python
+
+        samples = np.array([[2, 0, None],
+                            [2, 2, None],
+                            [2, 0, None],
+                            [0, 0, None]])
+
+    Getting the variance for these PNR samples:
+
+    >>> number_expectation_pnr(samples)
```
nquesada, line 65: hey @antalszava : Maybe I am missing something, but how does it now which modes to consider? Is it just the ones that do not have `None`s?

antalszava: hi @nquesada, yes, if `modes` was not explicitly defined, then the idea would be considering all the modes where there are no `None` values (the measured modes).

This will actually be simplified by #398, as that change will remove the `None` values from the returned samples completely and only samples for the measured modes will be present. My original thinking was that handling `None` values could be manageable in this PR, but I also found that some parts started to become more complex. Let me know if you'd rather we waited with going forward for this until #398 was merged. It might also make the reviewing process more seamless.

josh146: I think it might make more sense to merge in #398 first, especially if it makes a lot of the validation and the checks here redundant!

antalszava: Sure! In the meantime, potential things I'd be curious to hear thoughts on (if there are any):
* function naming (e.g. having `number_expectation_pnr` as user facing but also having a `number_expectation` as a shorthand)
* import structure (adding the name `number_expectation_pnr` to `api/__init__.py`)
* file structure (having a `post_processing.py` in the `api` folder)

strawberryfields/api/__init__.py
```diff
@@ -23,5 +23,6 @@
 from .connection import Connection, RequestFailedError
 from .job import InvalidJobOperationError, Job, JobStatus
 from .result import Result
+from .post_processing import number_expectation_pnr, number_variance_pnr
 
-__all__ = ["Connection", "Job", "Result"]
+__all__ = ["Connection", "Job", "Result", "number_expectation_pnr", "number_variance_pnr"]
```
josh146, line 28: It feels a bit weird having the post-processing module in the `api` package. These are functions that are relevant to sampled results in general, whether they were generated via the API, local engine, or apps layer.

strawberryfields/api/post_processing.py
josh146: A couple of thoughts:

* For the moment, if both functions have no differences in logic, we should just have a single function.

* I mentioned this above, but I don't think the API module is the right place for sample post-processing. I think (for the moment) the best thing is to turn `sf.utils` into a package, and have `post_processing.py` in there.

antalszava: thanks @josh146! So they are the same function, just allowing further names through module attributes, e.g. at the end of `post_processing.py`:
```python
number_expectation = number_expectation_pnr
```

I've added this because during the preliminary discussion it was suggested that there is some sort of indication to the sample type --- this way we have both a shorthand and a longer version

(Went for your second point with #413)

antalszava:
This should now be ready for review (samples now have no `None` values thanks to #398 :confetti_ball:).

Further changes:
* Added a `quadrature_expectation_homodyne`, `quadrature_variance_homodyne` and an `all_fock_probs_pnr` function
* `post_processing.py` has been moved to the new `utils` package

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module provides functions for post-processing samples.
+"""
+import numpy as np
+
+
```
josh146, line 19: It might be good to move the smaller utility functions first/to the top of the file. This way, it's easier to read the file; by the time you get to the functions that use them, you know what they're doing :)

For example, maybe the following order makes sense?

1. First, general validation functions:

   * `_check_samples`
   * `_check_modes`

2. Then, auxiliary functions used in other post-processing functions:

   * `_product_for_modes`
   * `_sample_variance`
   * `_sample_expectation`

3. Finally, user-facing functions:

   *  `number_expectation_pnr`
   * `number_variance_pnr`
   * `quadrature_variance_homodyne`
   * `all_fock_probs_pnr`

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+        samples = np.array([[2, 0],
+                            [2, 2],
+                            [2, 0],
+                            [0, 0]])
+
+    Getting the expectation value for these PNR samples:
+
+    >>> number_expectation_pnr(samples)
+    1.0
```
josh146, line 35: If possible, I would rewrite this to use a single `>>>` block if possible. Maybe like this?

```pycon
>>> samples = np.array([[2, 0], [2, 2], [2, 0], [0, 0]])
>>> number_expectation_pnr(samples)
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Args:
+        photon_number_samples (ndarray): the photon number samples with a shape
+            of (shots, modes)
```
josh146, line 39:
```suggestion
            of ``(shots, modes)``
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+        modes (Sequence): a flat sequence containing indices of modes to get
+            the expectation value for
```
josh146, line 41:
```suggestion
        modes (Sequence): indices of modes to compute the expectation value over
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module provides functions for post-processing samples.
+"""
+import numpy as np
+
+
+def number_expectation_pnr(photon_number_samples, modes=None):
+    """The expectation value of the number operator from PNR samples.
```
josh146, line 21: It might be good to provide the exact observable that is computed. Perhaps something along the lines of

```rest
The expectation value of the number operator from PNR samples.

If applied to a single mode, this simply corresponds to mean photon number
:math:`\langle n_i\rangle`.

For multiple modes, the expectation value of the tensor of number operators,
:math:`\langle n_{i_0} \otimes n_{i_1} \otimes \cdots \otimes n_{i_m}` is returned.
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+            the expectation value for
+
+    Returns:
+        float: the expectation value from the samples
+    """
+    return _samples_expectation(photon_number_samples, modes)
+
+
+def number_variance_pnr(photon_number_samples, modes=None):
+    """The variance of the number operator from PNR samples.
```
josh146, line 50: Same comments from above apply here as well!

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+        modes (Sequence): a flat sequence containing indices of modes to get
+            the variance for
+
+    Returns:
+        float: the variance from the samples
+    """
+    return _samples_variance(photon_number_samples, modes)
+
+
+def quadrature_expectation_homodyne(homodyne_samples, modes=None):
```
josh146, line 78: As far as I can tell, there is no difference between this function and `number_expectation_pnr` except the docstring?

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+def _samples_expectation(samples, modes):
+    """Utility function for calculating the expectation value of samples from
+    multiple modes.
+
+    Args:
+        homodyne_samples (ndarray): the homodyne samples with a shape of
+            (shots, modes)
+        modes (Sequence): a flat sequence containing indices of modes to get
+            the expectation value for
```
josh146, line 150:
```suggestion
        samples (array): samples with a shape of ``(shots, modes)``
        modes (Sequence): indices of modes to compute the expectation value over
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    product_across_modes = _product_for_modes(samples, modes)
+    variance = product_across_modes.var()
+    return variance
+
+
+def all_fock_probs_pnr(photon_number_samples):
+    r"""The Fock state probabilities for the specified modes.
+
+    Measured modes that are not specified are being traced over. If either all
```
josh146, line 199:
```suggestion
    Measured modes that are not specified are traced over. If either all
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+    probability for this state is correct by indexing into ``probs``:
+
+    >>> probs[(2, 0)]
+    0.5
+
+    We can further check the entire array of probabilities:
+    >>> probs
+    [[0.25 0.   0.  ]
+    [0.   0.   0.  ]
+    [0.5  0.   0.25]]
```
josh146, line 226: really nice!

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Args:
+        photon_number_samples (ndarray): the photon number samples with a shape
+            of (shots, modes)
```
josh146, line 230:
```suggestion
        photon_number_samples (array): the photon number samples with a shape of ``(shots, modes)``
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Returns:
+        array: array of dimension :math:`\underbrace{D\times D\times D\cdots\times D}_{\text{num modes}}`
+            containing the Fock state probabilities, where :math:`D` is the Fock basis cutoff truncation
```
josh146, line 234: How is the cutoff computed? Is the cutoff simply `max_val`? If so, it should probably explain that in the docstring.

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+    """Validation check for the input samples.
+
+    Checks include data types checks and dimension of the input samples.
+
+    Args:
+        samples (ndarray): the photon number samples with a shape of (shots,
+            modes)
+    """
+    if not isinstance(samples, np.ndarray) or samples.ndim != 2:
+        raise Exception("Samples needs to be represented as a two dimensional numpy array.")
```
josh146, line 276:
```suggestion
        raise Exception("Samples needs to be represented as a two dimensional NumPy array.")
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Args:
+        samples (ndarray): the photon number samples with a shape
+            of (shots, modes)
+        modes (Sequence): the input modes to get the expectation value for, a
+            flattened sequence
+    """
+    num_modes = samples.shape[1]
+    flattened_sequence_indices_msg = (
+        "The input modes need to be specified as a flattened sequence of indices!"
```
josh146, line 293: Should it be the following?
```suggestion
        "The input modes need to be specified as a sequence of positive integers."
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+    if modes.ndim != 1 or non_index_modes.size > 0 or len(modes) == 0:
+        raise Exception(flattened_sequence_indices_msg)
+
+    # Checking if valid modes were specified
+    largest_valid_index = num_modes - 1
+    out_of_bounds_modes = modes[modes > largest_valid_index]
+    if out_of_bounds_modes.size > 0:
+        raise Exception(
+            "Cannot specify mode indices {} for a {} mode system!".format(
```
josh146, line 312:
```suggestion
            "Cannot specify mode index {} for a {} mode system.".format(
```

josh146: Note that f-strings are now allowed in SF and PL if you want to use them :)

strawberryfields/utils/post_processing.py
nquesada: I think so.

strawberryfields/utils/post_processing.py
antalszava: Thanks added them! (kept the plural form as a list can be the first argument)

josh146:
> Before falling into the trap of modifying these functions back and forth, I'd just like to double-check if we'd for sure want to abandon that proposal and have only these two user-facing functions for expval and var.

I think your clean implementation in this PR helped solidify that in my head 🙂 Originally, I was under the assumption that each function would have it's own logic. Since that is not the case, I think we can use this to better inform the design.

 > This will have the advantage:
> * will not confuse users into thinking that different functions do different things (a premise that would be otherwise well assumed)
> * less complexity

I agree fully!

> (That is, for example, PNR samples cannot be used to obtain expval/var for position or momentum)

I think solely providing `sample_expectations` and `sample_var` actually makes this _clearer_ from the user perspective. The functions no longer make any assumptions as to the type of measurement, nor the observable that is being measured. It is now up to the user to interpret the output --- the functions will be doing exactly as they say.

strawberryfields/utils/post_processing.py
antalszava: sure, reorganized them!

strawberryfields/utils/post_processing.py
antalszava: thanks added it!

strawberryfields/utils/post_processing.py
antalszava: Thanks! Is this mostly such that no extra space is taken up in the docs?

strawberryfields/utils/post_processing.py
antalszava: Valid point, thanks! Added `The Fock basis cutoff truncation is determined by the maximum Fock state with a non-zero probability.` to the docstring.

antalszava:
Hi @josh146 @nquesada thanks for the comments! I'd hope to have answered them all, this should be ready for a next round

strawberryfields/utils/post_processing.py
josh146: More so that if we decide to use `doctest` in the future :)

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+import numpy as np
+
+
+def _product_for_modes(samples, modes=None):
+    """Getting the product of samples across modes.
+Args:
+        samples (ndarray): the photon number samples with a shape
+            of (shots, modes)
+        modes (Sequence): a flat sequence containing indices of modes to get
+            the expectation value for
```
josh146, line 26: This suggestion was made primarily to fix the indentation, but a couple of other comments:

* The shape should be given using code formatting
* I don't think it's necessary to say 'flat sequence'; this is a given when we say the type is
   ``Sequence``.

```suggestion
    """Getting the product of samples across modes.
        
    Args:
        samples (ndarray): the photon number samples with a shape of ``(shots, modes)``
        modes (Sequence): indices of modes to compute the expectation value for
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def _check_samples(samples):
+    """Validation check for the input samples.
+
+    Checks include data types checks and dimension of the input samples.
+
+    Args:
+        samples (ndarray): the photon number samples with a shape of (shots,
+            modes)
```
josh146, line 43: For consistency, we tend to use ``array`` rather than ``ndarray`` in the rest of the code base.
```suggestion
        samples (array): the photon number samples with a shape of ``(shots, modes)``
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+    """Validation checks for the input modes.
+
+    Checks include data types checks, checks for the dimension of the inputs
+    and the validity of the modes specified.
+
+    Args:
+        samples (ndarray): the photon number samples with a shape
+            of (shots, modes)
+        modes (Sequence): the input modes to get the expectation value for, a
+            flattened sequence
```
josh146, line 59:
```suggestion
        samples (array): the photon number samples with a shape of ``(shots, modes)``
        modes (Sequence): the input modes to get the expectation value for
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+        "The input modes need to be specified as a flattened sequence of non-negative integers!"
+    )
+
+    modes = np.array(modes)
+
+    # Checking if modes is a valid flattened sequence of indices
+    try:
+        non_index_modes = modes[(np.mod(modes, 1) != 0) | (modes < 0)]
+    except Exception:
+        raise Exception(flattened_sequence_indices_msg)
```
josh146, line 72: Interesting approach here 🤔 

What exception is likely to be raised here? Often, it is better to catch a specific exception, rather than all.

josh146: Wondering if it might be better to have separate if statements, as used below?

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+
+    if modes.ndim != 1 or non_index_modes.size > 0 or len(modes) == 0:
+        raise Exception(flattened_sequence_indices_msg)
+
+    # Checking if valid modes were specified
+    largest_valid_index = num_modes - 1
+    out_of_bounds_modes = modes[modes > largest_valid_index]
+    if out_of_bounds_modes.size > 0:
+        raise Exception(
+            f"Cannot specify mode indices {out_of_bounds_modes} for a {num_modes} mode system."
```
josh146, line 82: nice :)

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+    except Exception:
+        raise Exception(flattened_sequence_indices_msg)
+
+    if modes.ndim != 1 or non_index_modes.size > 0 or len(modes) == 0:
+        raise Exception(flattened_sequence_indices_msg)
+
+    # Checking if valid modes were specified
+    largest_valid_index = num_modes - 1
+    out_of_bounds_modes = modes[modes > largest_valid_index]
+    if out_of_bounds_modes.size > 0:
```
josh146, line 80: Maybe clearer to just have the following?
```python
if max(modes) > num_modes - 1:
```

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+        raise Exception(
+            f"Cannot specify mode indices {out_of_bounds_modes} for a {num_modes} mode system."
+        )
+
+
+def samples_expectation(samples, modes=None):
```
josh146, line 86: Really nice docstring here!

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+
+    Args:
+        homodyne_samples (ndarray): the homodyne samples with a shape of
+            (shots, modes)
+        modes (Sequence): a flat sequence containing indices of modes to get
+            the variance for
+
+    Returns:
+        float: the variance from the samples
+    """
```
josh146, line 200: Another really nice docstring :)

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+def all_fock_probs_pnr(photon_number_samples):
+    r"""The Fock state probabilities for the specified modes obtained from PNR
+    samples.
+
+    Measured modes that are not specified are traced over. If either all
+    the modes or no modes were specified, the marginal probabilities are
+    returned.
+
+    The Fock basis cutoff truncation is determined by the maximum Fock state
+    with a non-zero probability.
```
josh146, line 224: 🔥

strawberryfields/utils/post_processing.py
```diff
@@ -0,0 +1,271 @@
...
+    [0.5  0.   0.25]]
+
+    Args:
+        photon_number_samples (array): the photon number samples with a shape
+            of ``(shots, modes)``
+
+    Returns:
+        array: array of dimension :math:`\underbrace{D\times D\times
+            D\cdots\times D}_{\text{num modes}}` containing the Fock state
+            probabilities, where :math:`D` is the Fock basis cutoff truncation
```
josh146, line 252: 🥳

strawberryfields/utils/post_processing.py
antalszava: Thanks, valid point! The reason why this was in a `try-except` block is because it is supposed to handle cases when the array does not contain numerical values but rather other objects (e.g. lists). Added a comment on this and adjusted the `try-except` to except for `TypeError` (this is raised in such cases) and to raise this type of error.

Further changed to raising `ValueError` for other cases.

strawberryfields/utils/post_processing.py
antalszava: thanks, that's a briefer and simpler way! :blush:

antalszava: oh just realized that the original solution also returns the invalid modes, so I'll leave it as is if it's okay

PR #404, opened 2020-06-08
Update README.rst

nquesada:


co9olguy:
We should look into a more reliable coverage service

josh146:
I tried to set up coveralls.io last month, since they now support merging coverage results multiple CI builds. I couldn't even get it to work :/ https://coveralls.io/github/XanaduAI/pennylane

PR #381, opened 2020-04-30
Updates the Fock backend ops.py to use the new Walrus gate signatures

nquesada:
The signature and return values of the `fock_gradients` functions is being modified in this PR https://github.com/XanaduAI/thewalrus/pull/164 . Once PR 164 of The Walrus is merged and a new release is made we will need to update the way these functions are used in SF. This PR does that.

PR #407, opened 2020-06-09
Updates BaseGaussianState to be feature compatible with BaseFockState

josh146:
**Context:** More Fock-basis functionality is available for Gaussian state post-processing thanks to The Walrus. This PR integrates the latest Walrus functionality, elevating the Gaussian state to the same level of functionality as the Fock state.

* Removes the `gaussianbackend.GaussianState` class. It is now sufficient to simply use `BaseGaussianState`, and import needed functions from The Walrus. As a result, several functions in the `gaussianbackend.ops` package are no longer being used and can be removed:

  - `fock_amplitudes_one_mode`

  - `fock_prob`

  - `sm_fidelity`

  - `xmat`

  - `omega`

* `BaseGaussianState` now supports `ket`, `dm`, and `all_fock_probs`. As a result, these are now abstract methods in `BaseState`. The tests have been updated so existing tests for these methods are now run on the Gaussian backend.

* `BaseGaussianState` now uses The Walrus to implement `reduced_dm`, `fidelity`, and `fock_prob`.

* The `states.rst` quickstart has been updated with the new functionality.

**Benefits:** The Gaussian state is now feature complete with the Fock state.

**Possible Drawbacks:** Some of the new functionality is only available in the latest version of The Walrus 0.13.0-dev. Unfortunately, this version of The Walrus breaks Fock gate compatibility. Therefore, this PR branches off #381, as the corresponding Walrus fix has already been implemented there.


strawberryfields/backends/states.py
```diff
@@ -1266,26 +1309,97 @@ def parity_expectation(self, modes):
...
+        cutoff = kwargs.get("cutoff", 10)
+        mu = self._mu
+        cov = self._cov
+
+        if self._pure:
+            return twq.state_vector(
+                mu, cov, hbar=self._hbar, normalize=True, cutoff=cutoff, check_purity=False
+            )
+
+        return None  # pragma: no cover
```
antalszava, line 1322: curious about this one, would this not return `None` implicitly anyways? (I've noticed it is similar to that in `BaseFockState`)

strawberryfields/backends/states.py
```diff
@@ -123,17 +123,46 @@ def mode_indices(self):
         """
         return {v: k for k, v in self._modemap.items()}
 
+    @abc.abstractmethod
+    def ket(self, **kwargs):
+        r"""The numerical state vector for the quantum state.
```
nquesada, line 128:
```suggestion
        r"""The numerical state vector for the quantum state in the Fock basis.
```

strawberryfields/backends/states.py
```diff
@@ -123,17 +123,46 @@ def mode_indices(self):
...
+                states represented in the Fock basis will use their own internal cutoff dimension.
+
+        Returns:
+            array/None: the numerical state vector. Returns None if the state is mixed.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def dm(self, **kwargs):
+        r"""The numerical density matrix for the quantum state.
```
nquesada, line 143:
```suggestion
        r"""The numerical density matrix for the quantum state in the Fock basis.
```

strawberryfields/backends/states.py
```diff
@@ -1241,15 +1282,17 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
             raise ValueError("There can be no duplicates in the modes specified.")
+
         mu = self._mu
         cov = self._cov
+
         if len(modes) == 1:
-            return photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
+            return twq.photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
 
         if len(modes) == 2:
```
nquesada, line 1292: There is now functionality in the Walrus to this for an arbitrary number of modes. I thought @antalszava had added the necessary piping between SF and TW for doing this.

antalszava: that's part of a change I have locally, a missing piece for that too was having the change of TW Fock gate compatibility (the changes from the branch this PR branches off from)

strawberryfields/backends/states.py
```diff
@@ -314,6 +365,7 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
```
antalszava, line 368: Could this be added to `parity_expectation` too?

strawberryfields/backends/states.py
```diff
@@ -1266,26 +1309,97 @@ def parity_expectation(self, modes):
...
 
-    @abc.abstractmethod
+    def ket(self, **kwargs):
+        cutoff = kwargs.get("cutoff", 10)
+        mu = self._mu
+        cov = self._cov
+
+        if self._pure:
+            return twq.state_vector(
+                mu, cov, hbar=self._hbar, normalize=True, cutoff=cutoff, check_purity=False
```
antalszava, line 1319: Just curious for my own understanding: how come `normalize=True` is needed here?

tests/backend/test_states.py
```diff
@@ -177,20 +177,19 @@ def test_rdm(self, setup_backend, tol, cutoff, batch_size):
...
 
-    @pytest.mark.backends("fock", "tf")
     def test_ket(self, setup_backend, pure, cutoff, batch_size, tol):
         """Test that the ket of a displaced state matches analytic result"""
         backend = setup_backend(2)
         backend.displacement(a, 0)
 
         state = backend.state()
-        if not pure:
+        if not pure and backend.short_name != "gaussian":
```
antalszava, line 186: hmm, how come? :thinking: would this not mean that the assertion after the if statement `assert state.is_pure == True` can fail?

strawberryfields/backends/states.py
nquesada: Formally, you are only guaranteed a normalized state if the `cutoff` -> infinity. Typically if the `cutoff` is sufficiently large one is very close to being normalized.

strawberryfields/backends/states.py
josh146: Yep you're right. I copied and pasted this from `BaseFockState.ket()`, so can't remember the original reason.

I suppose it's because we wanted this to be an explicit return of `None`, rather than an implicit return.

strawberryfields/backends/states.py
josh146: Yep! Good catch

tests/backend/test_states.py
josh146: Ah, that's because the `pure` backend argument doesn't have any effect on the Gaussian backend.

* On the Fock backends, `pure=False` forces the backends to use the density matrix representation rather than a ket, even for pure states. It then initializes the state with the attribute `self._pure = False`.

* On the Gaussian backend, however, `pure=False` doesn't change anything --- the state is always represented as a vector of means and covariance matrix. Instead, the value of `self._pure` is set _by the state itself_, by testing the covariance matrix for purity.

This inconsistency then continues between `BaseGaussianState` and `BaseFockState`:

* `BaseFockState.is_pure` returns whether the state is represented by a density matrix or a ket, which is _not_ the same as saying whether it is pure. A pure state can always be represented as a density matrix.

* `BaseGaussianState.is_pure` return whether or not the state is pure or mixed.

These two different approaches explain why
```python
>>> eng = sf.Engine("guassian", backend_options={"pure": False})
>>> results = eng.run(prog)
>>> results.state.is_pure
True
```

There are probably two things that need to change (but out of scope for this PR):

* The `pure` backend option should probably be renamed something else? To specify that (a) it only applies to Fock backends, and (b) it only effects the internal state representation, it doesn't automatically make mix states.

* `BaseFockState.is_pure` should return whether the state is pure or mixed, and be independent of the internal representation.

antalszava: thanks, helped a lot!

strawberryfields/backends/states.py
```diff
@@ -1241,15 +1283,17 @@ def poly_quad_expectation(self, A, d=None, k=0, phi=0, **kwargs):
...
 
         if len(modes) == 2:
-            ni = photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
-            nj = photon_number_mean(mu, cov, modes[1], hbar=self._hbar)
-            return photon_number_covar(mu, cov, modes[1], modes[0], hbar=self._hbar) + ni * nj
+            ni = twq.photon_number_mean(mu, cov, modes[0], hbar=self._hbar)
+            nj = twq.photon_number_mean(mu, cov, modes[1], hbar=self._hbar)
+            return twq.photon_number_covar(mu, cov, modes[1], modes[0], hbar=self._hbar) + ni * nj
 
         raise ValueError(
```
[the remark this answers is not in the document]
josh146: I can update this 🙂 I think @antalszava added this feature prior to the walrus supporting an arbitrary number of modes

.github/CHANGELOG.md
```diff
@@ -2,18 +2,31 @@
...
+
+  - `GaussianState.dm()`
+  - `GaussianState.ket()`
+  - `GaussianState.all_fock_probs()`
+
+  In addition, the existing `GaussianState.reduced_dm()` method now supports
+  multi-mode reduced density matrices.
+
+* Adds new `Xcov` and `Xunitary` compilers for compiling programs into the X
+  architecture. [(#358)](https://github.com/XanaduAI/strawberryfields/pull/358)
```
antalszava, line 19:
```suggestion
* Adds new `Xcov` and `Xunitary` compilers for compiling programs into the X
  architecture.
  [(#358)](https://github.com/XanaduAI/strawberryfields/pull/358)
```

tests/backend/test_states.py
```diff
@@ -344,14 +353,31 @@ def test_number_expectation_displaced_squeezed(self, setup_backend, tol, batch_s
...
 
+        def analytic_var(a, r, phi):
+            magnitude_squared = np.abs(a) ** 2
+            var_ex = - magnitude_squared + magnitude_squared ** 2 + 2 *\
+                magnitude_squared*np.cosh(2*r) - np.exp(-1j*phi) * a ** 2 *\
+                np.cosh(r)*np.sinh(r) - np.exp(1j* phi) * np.conj(a) **2 *\
+                np.cosh(r)*np.sinh(r) + np.sinh(r)**4 - (magnitude_squared +\
+                        np.conj(np.sinh(r))*np.sinh(r)) ** 2 +\
+                np.cosh(r)*np.sinh(r)*np.sinh(2*r)
+            return var_ex
```
josh146, line 364: Thanks @antalszava for calculating these!

PR #409, opened 2020-06-11
[WIP] Fix VGBS

trbromley:
This is a WIP PR to apply some fixes for GBS training in the `apps.train` module.


trbromley:
Closing this PR because we realized that the current functionality works fine. We were getting unexpected numbers compared to a previous script, but that was due to a subtlety in how the problem was set up.

@ixfoduap

PR #410, opened 2020-06-12
[WIP] Add realistic hardware imperfections

thisac:
**Context:**
Currently, the simulations are close to perfect (apart from the dark-counts addition), meaning that they aren't realistically simulating the current hardware. This can be addressed, to an extent, by adding loss to the circuits.

The `Program.comile` method is extended to allow for the keyword argument `allow_imperfections` which automatically applies loss between the circuit gates based on that specific chip's calibration data.

**Benefits:**
Lossy circuits can now be simulated.

**Possible Drawbacks:**

**Related GitHub Issues:**


tests/frontend/test_circuitspecs_X12.py
```diff
@@ -554,11 +554,11 @@ def test_unitary_too_large(self, chip):
...
+    def test_allow_imperfections_warning(self, chip):
         """Test the X12 compile method warns when allowing imperfections"""
         prog = sf.Program(2)
         with prog.context as q:
             ops.MeasureFock() | q
 
         with pytest.warns(UserWarning, match="compile method does not currently support imperfections"):
-            prog.compile("X12", allow_imperfections=True)
\ No newline at end of file
+            prog.compile(chip.short_name, allow_imperfections=True)
```
nquesada, line 564:
```suggestion
            prog.compile(chip.short_name, allow_imperfections=True)

```

strawberryfields/circuitspecs/xcov.py
```diff
@@ -165,4 +165,19 @@ def compile(self, seq, registers):
         for Ui in U2:
             Ui.reg = [registers[r.ind + half_n_modes] for r in Ui.reg]
 
-        return sq_seq + U1 + U2 + meas_seq
+        if allow_imperfections is False:
+            return sq_seq + U1 + U2 + meas_seq
+
+        dark_counts = 3
```
nquesada, line 171:
```suggestion
        dark_counts = 0.03
```

strawberryfields/circuitspecs/xcov.py
```diff
@@ -165,4 +165,19 @@ def compile(self, seq, registers):
...
+        end_to_end_transmission = 0.5
```
nquesada, line 172:
```suggestion
        end_to_end_transmission = 0.2
```

thisac:
This PR should now be getting ready for reviews soon, with most work done by @nquesada. The only question left is how and where to include the parameters for the imperfections. In the future, these would optimally be accessed through an API via which they could be updated regularly to mirror the actual hardware imperfections. This would  probably warrant more in-depth discussions/an ADR, and isn't something that can be implemented yet. For now, these imperfection parameters would be hard-coded into SF.

Options to consider:

1. Setting default values inside the relevant `compile` methods (as it is for the `Xcov` circuit-spec class in this PR at the moment).

2. Setting default values as attributes of the relevant `compile` spec-classes.

3. Setting default values as attributes of the `CircuitSpecs` class. Thus _all_ spec classes will inherit these parameters; even the non-hardware related ones that would not use them.

Which one would make the most sense right for now? Is there another option that might be preferred here? @josh146 @lneuhaus

nquesada:
should we close this PR @thisac ?

thisac:
> should we close this PR @thisac ?

We might as well. It'll be here if we need to come back to it some day, but as far as I know there's no plan on working on this now. If we decide to, it's probably better and easier just to create a new PR then.

Thanks for the reminder @nquesada. Closing this now!

PR #411, opened 2020-06-12
Add a function to qchem for Fock state sampling

soranjh:
This PR adds a function to the `apps.qchem.dynamics` module to generate samples for simulating vibrational quantum dynamics in molecules with a Fock input state.


strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +118,80 @@ def op(t, Ul, w, q):
...
+        Ul (array): normal to local transformation matrix
+        w (array): normal mode frequencies :math:`\omega` in units of :math:`\mbox{cm}^{-1}`
+        n_samples (int): number of samples to be generated
+        loss (float): loss parameter denoting the fraction of generated photons that are lost
+        cutoff (int): cutoff dimension for determining the set of number states for each mode
+
+    Returns:
+        list[list[int]]: a list of samples
+    """
+    if n_samples < 1:
```
ixfoduap, line 157: These are great checks! I think an important one missing is verifying that the length of the input list is equal to the dimension of the unitary

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +118,80 @@ def op(t, Ul, w, q):
...
+    if not 0 <= loss <= 1:
+        raise ValueError("Loss parameter must take a value between zero and one")
+
+    modes = len(Ul)
+    op = evolution(modes)
+    s = []
+
+    eng = sf.Engine("fock", backend_options={"cutoff_dim": cutoff})
+
+    gbs = sf.Program(modes)
```
ixfoduap, line 178: `gbs` is nice, but it may be best to use more conventional notation here. I believe typically we use `prog = sf.Program(modes)`

tests/apps/qchem/test_dynamics.py
```diff
@@ -94,3 +100,91 @@ def test_evolution(time, unitary, frequency, prob):
         p = eng.run(gbs).state.all_fock_probs()
 
     assert np.allclose(p, prob)
+
+
+@pytest.mark.parametrize("d", [d1, d2])
```
ixfoduap, line 105: These tests are fine, but again, I think they are not detecting whether the sampling is working properly. For example, if we apply a different unitary in the code, it won't get caught by the tests. I suggest coding an example where there is no randomness in the sampling. One example could be to use an identity interferometer, or a unitary that permutes the modes. In both cases, the output state is another Fock state that always gives the same result upon a Fock measurement

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +118,80 @@ def op(t, Ul, w, q):
...
+        op(t, Ul, w) | q
+
+        if loss:
+            for _q in q:
+                sf.ops.LossChannel(1 - loss) | _q
+
+        sf.ops.MeasureFock() | q
+
+        for _ in range(n_samples):
+            s.append(eng.run(gbs).samples[0].tolist())
```
trbromley, line 195: I don't think this should be within the `gbs.context` (i.e., we need to lose a tab)

Also, we should be able to just do:
`eng.run(prog, shots=n_samples).samples.tolist()` (see for example `apps.sample.sample`).

tests/apps/qchem/test_dynamics.py
```diff
@@ -94,3 +100,91 @@ def test_evolution(time, unitary, frequency, prob):
...
+
+        def save_hist(*args):
+            call_history.append(args[1])
+            return sf.engine.Result
+
+        call_history = []
+        with monkeypatch.context() as m:
+            m.setattr(sf.engine.Result, "samples", np.array([[0]]))
+            m.setattr(sf.LocalEngine, "run", save_hist)
+            dynamics.sample_fock(*d, loss=0.5)
```
trbromley, line 162: I'm a bit confused what these two tests are doing, they seem similar to `test_loss` and `test_no_loss` that test the `sample.sample` function, but there there are asserts to check what the program looks like.

tests/apps/qchem/test_dynamics.py
trbromley: For me the best test would be one that inspects the program and:
- ensures that the operations are occurring in the expected order
- the unitary and rotation values are as expected

We can then trust SF to sample correctly given that we know we've created the correct program.

So for me, we'd build upon the `test_invalid_loss`/`test_loss` tests and instead have one test to check the order of the operations and another test(s) that looks at the interferometers and rotations.

Regarding the `test_invalid_loss`/`test_loss` tests, you can access the circuit using `call_history[0].circuit`. I then had a quick look at the embedded unitary using `c[3].op.p`. However, I think the `MagicMock` approach used in the `sample.sample` tests should also be a more conventional way of accessing the called arguments.

strawberryfields/apps/qchem/dynamics.py
soranjh: Using `shots=2` gives this error: "The operation MeasureFock has not been implemented in FockBackend for the arguments {'shots': 2}"

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +118,80 @@ def op(t, Ul, w, q):
...
+
+
+def sample_fock(
+    input_state: list,
+    t: float,
+    Ul: np.ndarray,
+    w: np.ndarray,
+    n_samples: int = 1,
+    loss: float = 0.0,
+    cutoff: int = 5,
```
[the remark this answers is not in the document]
soranjh: Some users might not be familiar with the meaning of cutoff here. So I think it is better to keep the default but increase the value to 15.

.github/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 <h3>New features since last release</h3>
 
+* Adds function `sample_fock` to the `apps.qchem.dynamics` module to generate samples for
+  simulating vibrational quantum dynamics in molecules with a Fock input state.
+  [(#411)](https://github.com/XanaduAI/strawberryfields/pull/411)
+
+* Adds function `evolution` to the `apps.qchem.dynamics` module to implement the operations
+  required to simulate vibrational quantum dynamics in molecules.
+  [(#402)](https://github.com/XanaduAI/strawberryfields/pull/402)
```
trbromley, line 11: Awesome! I would have it as a single bullet point:
"Adds the `apps.qchem.dynamics` module for simulating vibrational quantum dynamics in molecules. The `dynamics.evolution()` function provides a custom operation that encodes the input chemical information for use in a Strawberry Fields `Program`. The `sample_fock()` function allows for generation of samples from an input Fock state.
[(#402)](https://github.com/XanaduAI/strawberryfields/pull/402)
[(#411)](https://github.com/XanaduAI/strawberryfields/pull/411)
"
We can then expand as we add more (i.e., the other sample functions).

tests/apps/qchem/test_dynamics.py
soranjh: I have added the asserts. They were missing by mistake.

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +122,92 @@ def op(t, Ul, w, q):
...
+
+
+def sample_fock(
+    input_state: list,
+    t: float,
+    Ul: np.ndarray,
+    w: np.ndarray,
+    n_samples: int = 1,
+    loss: float = 0.0,
+    cutoff: int = 15,
```
ixfoduap, line 134: It's probably not a good idea to have a default cutoff. I think Tom mentioned this before? It's important enough that the user should probably give it some thought instead of relying on default values

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +122,92 @@ def op(t, Ul, w, q):
...
+) -> list:
+    r"""Generate samples for simulating vibrational quantum dynamics with a Fock input state.
```
ixfoduap, line 136:
```suggestion
    r"""Generate samples for simulating vibrational quantum dynamics with an input Fock state.
```

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +122,92 @@ def op(t, Ul, w, q):
...
 
     return op
+
+
+def sample_fock(
+    input_state: list,
+    t: float,
+    Ul: np.ndarray,
+    w: np.ndarray,
+    n_samples: int = 1,
```
ixfoduap, line 132: The number of samples also should probably not have a default value, but always be specified

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +122,92 @@ def op(t, Ul, w, q):
...
+
+    **Example usage:**
+
+    >>> input_state = [0, 2]
+    >>> t = 10.0
+    >>> Ul = np.array([[0.707106781, -0.707106781],
+    >>>                [0.707106781, 0.707106781]])
+    >>> w = np.array([3914.92, 3787.59])
+    >>> n_samples = 5
+    >>> sample_fock(input_state, t, Ul, w, n_samples = n_samples)
```
ixfoduap, line 146: If we remove the default value of `n_samples`, then this should read 
` >>> sample_fock(input_state, t, Ul, w, n_samples )`

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -117,3 +122,92 @@ def op(t, Ul, w, q):
...
+
+    if np.any(np.iscomplex(Ul)):
+        raise ValueError("The normal mode to local mode transformation matrix must be real")
+
+    if not 0 <= loss <= 1:
+        raise ValueError("Loss parameter must take a value between zero and one")
+
+    if not len(input_state) == len(Ul):
+        raise ValueError(
+            "Number of modes in the input state and the normal to local transformation"
```
ixfoduap, line 178:
```suggestion
            "Number of modes in the input state and the normal-to-local transformation"
```

tests/apps/qchem/test_dynamics.py
```diff
@@ -94,3 +103,168 @@ def test_evolution(time, unitary, frequency, prob):
...
+                m.setattr(sf.LocalEngine, "run", mock_eng_run)
+                dynamics.sample_fock(*d)
+                p_func = mock_eng_run.call_args[0][0]
+
+            assert isinstance(p_func.circuit[3].op, -7.374345193888777)
+            assert isinstance(p_func.circuit[4].op, -7.374345193888777)
+
+    def test_interferometer(self, monkeypatch, d):
+        """Test if function correctly uses the interferometer unitaries."""
+        if d == "d1":
```
trbromley, line 244:
```suggestion
        if len(d[0]) == 2:
```

tests/apps/qchem/test_dynamics.py
```diff
@@ -94,3 +103,168 @@ def test_evolution(time, unitary, frequency, prob):
...
+
+            with monkeypatch.context() as m:
+                m.setattr(sf.LocalEngine, "run", mock_eng_run)
+                dynamics.sample_fock(*d)
+                p_func = mock_eng_run.call_args[0][0]
+
+            _, _, U, _, _, _ = d
+
+            assert isinstance(p_func.circuit[2].op, U.T)
+            assert isinstance(p_func.circuit[5].op, U)
```
trbromley, line 255:
```suggestion
            assert np.allclose(p_func.circuit[2].op.p, U.T)
            assert np.allclose(p_func.circuit[5].op.p, U)
```

tests/apps/qchem/test_dynamics.py
trbromley: Tests are great now!

PR #413, opened 2020-06-16
Moving utility functions to a utils package

antalszava:
**Context:**
As further additions would extend the `utils` module (e.g. functions for post-processing of samples) it became desirable to make is a package.

Creates a new `utils` package and imports names from the `utils` module.

**Benefits:**
* The `utils` package can be further extended
* There are no breaking changes for user code that use gates from `utils`

**Possible Drawbacks:**
Importing names from `utils.py` requires managing a long list for `__all__`. The alternative of resorting to an absolute import would, however, this would both introduce discrepancies in the test suite as well as potentially break user code.

**Related GitHub Issues:**
Related PR: #399


strawberryfields/utils/utils.py
```diff
@@ -56,10 +56,28 @@
...
+    "randnc",
+    "random_covariance",
+    "random_symplectic",
+    "random_interferometer",
+    "operation",
+    "is_unitary",
+    "is_channel",
+    "extract_unitary",
+    "extract_channel",
+]
```
josh146, line 80: Now that `utils` is a package, we can start to split these into separate modules 🙂 

Groupings that would make sense include:
* `strawberryfields.utils.states`:

  - `squeezed_cov`
  - `vacuum_state`
  - `coherent_state`
  - `squeezed_state`
  - `displaced_squeezed_state`
  - `fock_state`
  - `cat_state`

* `strawberryfields.utils.random`:

  - `random_covariance`
  - `random_symplectic`
  - `random_interferometer`

The rest.. not really sure yet lol

PR #416, opened 2020-06-17
Move vibronic into new qchem module

trbromley:
This PR moves the existing `vibronic` module into the new `apps.qchem` package, providing a single point to access all quantum chemistry functionality in the apps layer.

Positives:
- Users don't have to combine functions from the `sample` module and the `vibronic` module, they simply use `qchem.vibronic` for both encoding the problem and sampling.
- The new location in the `qchem` module is a natural place for this to go.
- Although a big change from before, it is not a breaking change: users can still interact using the old approach (e.g., `apps.vibronic`), which we achieve using imports. However, the old approach is no longer mentioned in the documentation and tutorials.

Negatives:
- We don't move the `plot.spectrum` function.


tests/apps/qchem/test_vibronic.py
```diff
@@ -0,0 +1,233 @@
...
+
+S3 = [1, 2, 0, 0, 1, 1]
+E3 = 1600.0
+
+
+@pytest.mark.parametrize("sample, sample_energy", [(S1, E1), (S2, E2), (S3, E3)])
+def test_energies(sample, sample_energy):
+    r"""Tests the correctness of the energies generated for GBS samples in
+    ``strawberryfields.apps.qchem.vibronic.energies``."""
+    assert np.allclose(vibronic.energies(sample, w, wp), sample_energy)
```
trbromley, line 111: This is all untouched from before (except for updating naming due to change in location)

tests/apps/qchem/test_vibronic.py
```diff
@@ -0,0 +1,233 @@
...
+    modes, all non-negative integers."""
+    samples = np.array(
+        sf.apps.qchem.vibronic.sample(*p, n_samples=integration_sample_number))
+
+    dims = samples.shape
+
+    assert len(dims) == 2
+    assert dims == (integration_sample_number, len(alpha) * 2)
+    assert samples.dtype == "int"
+    assert (samples >= 0).all()
```
trbromley, line 233: This was taken from the `apps.sample` module.

tests/apps/qchem/test_vibronic.py
```diff
@@ -0,0 +1,233 @@
...
+                p_func = mock_eng_run.call_args[0][0]
+
+            eng = sf.LocalEngine(backend="gaussian")
+
+            state = eng.run(p_func).state
+            cov = state.cov()
+            disp = state.displacement()
+
+            assert np.allclose(cov, 0.5 * state.hbar * np.eye(4 * dim))
+            assert np.allclose(disp, np.zeros(2 * dim))
```
trbromley, line 216: This is the only nontrivial change - the test wasn't actually working before.

tests/apps/qchem/test_vibronic.py
```diff
@@ -0,0 +1,233 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
```
trbromley, line 1: Note that this is mostly copy and paste (see comments below)

strawberryfields/apps/sample.py
```diff
@@ -110,6 +107,7 @@
 import numpy as np
 
 import strawberryfields as sf
+from strawberryfields.apps.qchem.vibronic import sample as vibronic  # pylint: disable=unused-import
```
soranjh, line 110: Do we need this?

tests/apps/qchem/test_vibronic.py
```diff
@@ -0,0 +1,234 @@
...
+
+@pytest.mark.parametrize("p", [p0, p1])
+class TestSample:
+    """Tests for the function ``strawberryfields.apps.qchem.sample``"""
+
+    def test_invalid_n_samples(self, p):
+        """Test if function raises a ``ValueError`` when a number of samples less than one is
+        requested."""
+        with pytest.raises(ValueError, match="Number of samples must be at least one"):
+            sf.apps.qchem.vibronic.sample(*p, -1)
```
soranjh, line 161: Just `vibronic.sample(*p, -1)` is fine since `vibronic` is already imported? Same applies to the tests below.

examples_apps/run_tutorial_vibronic.py
```diff
@@ -5,7 +5,8 @@
 Vibronic spectra
 ================
 
-*Technical details are available in the API documentation:* :doc:`/code/api/strawberryfields.apps.vibronic`
```
josh146, line 8: Instead of making these changes here @trbromley, can you instead make them to the other repo!

strawberryfields/apps/sample.py
trbromley: This is so users can go
```python
from strawberryfields.apps import sample

sample.vibronic()
```
This is how we had it before, and it might be nice to keep it that way since we also show this usage in the paper.

tests/apps/qchem/test_vibronic.py
trbromley: Fixed!

strawberryfields/apps/sample.py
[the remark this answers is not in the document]
ixfoduap: Oh wow, dirty hack! `import sample as vibronic` :rofl:

PR #424, opened 2020-06-22
Fix docs building

trbromley:
A [new version](https://pypi.org/project/sphinx-autodoc-typehints/#history) of sphinx-autodoc-typehints was released yesterday that [drops support](https://github.com/agronholm/sphinx-autodoc-typehints/commit/9a966fe012c7d10baeacd7a55c32cb378d5f5320) for `sphinx < 3.0`. However, we currently version pin `sphinx` to `2.2.2`.

This new release hence caused the documentation to stop building, and a quick fix is to version pin `sphinx-autodoc-typehints` to `1.10.3` as is done in this PR.

Current RTD builds are breaking:
image

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -9,7 +9,7 @@ plotly
 quantum-blackbird
 scipy>=1.0.0
 sklearn
-sphinx-autodoc-typehints
+sphinx-autodoc-typehints==1.10.3
 sphinx-copybutton
 sphinx-automodapi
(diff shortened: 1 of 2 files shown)


trbromley:
Thanks @thisac!

I tried with `sphinx==3.0.4` and `sphinx==3.1.1` and got this error when building:
`TypeError: add_source_parser() takes 2 positional arguments but 3 were given`

Not sure what the problem is :thinking:

PR #428, opened 2020-06-24
Raise error for complex arguments to Coherent, DisplacedSqueezed and Dgate gates

ziofil:
Adds warnings if Dgate, S2gate and BSgate use new default values and errors if the arguments are complex.


strawberryfields/ops.py
```diff
@@ -1730,7 +1732,17 @@ class BSgate(Gate):
...
+        if (np.iscomplex([theta, phi])).any():
+            raise ValueError("The arguments of BSgate(theta, phi) cannot be complex")
+        if phi is None:
+            phi = 0.0
+            # TODO: remove warning in the new release
+            warnings.warn(
+                f"""Warning: since strawberryfields version v0.15.0,
+            BSgate(theta, phi) takes two angles rather than the transmission and
+            reflection coefficients. Using the arguments theta={theta} and phi=0.0"""
+            )
```
antalszava, line 1745: It seems to me that the UI for the BSgate and the S2gate did not change. It's only that after taking the real parameters, internally the complex argument was being used.

This can be checked by having a look at the history (clicking on the line number and choosing to view the git blame).

Link to the BSgate: https://github.com/XanaduAI/strawberryfields/blob/3c7b2a7bec0e05cf2133643bc4f63643db2e5a0e/strawberryfields/ops.py#L1738

strawberryfields/ops.py
```diff
@@ -1806,7 +1818,18 @@ class S2gate(Gate):
...
+            raise ValueError("The arguments of S2gate(r, phi) cannot be complex")
+        if phi is None:
+            phi = 0.0
+            # TODO: remove warning in the new release
+            warnings.warn(
+                f"""Warning: since strawberryfields version v0.15.0,
+            Sgate(r, phi) takes two real arguments which represent
+            the polar decomposition of the complex squeezing parameter.
+            Using the arguments r={r} and phi=0.0"""
+            )
```
antalszava, line 1832: Link for the S2gate:
https://github.com/XanaduAI/strawberryfields/blob/3c7b2a7bec0e05cf2133643bc4f63643db2e5a0e/strawberryfields/ops.py#L1813

strawberryfields/ops.py
ziofil: So, the BSgate now takes 2 angles rather than t,r (one of which could be complex), so there has been some change.

I was convinced I had changed the S2gate, but you're probably right, I've changed so many things lol 😄

For the other operations, for some reason I was convinced I had not modified them in ops.py 🤦‍♂️💆‍♂️
Thank you for the reminder, I'll just shift the errors/warnings there

antalszava: Was that the case really with the BSgate? :thinking: It seems to me that we were deriving t,r using
```python
t = pf.cos(self.p[0])
r = pf.sin(self.p[0]) * pf.exp(1j * self.p[1])
```
before they got passed to
```python
t = par_evaluate(t)
r = evaluate_complex_parameter(r, self.p[1])
backend.beamsplitter(t, r, *reg)
```
Sure thing! :slightly_smiling_face: Yes, it's not so easy to look at the changes after a while :sweat_smile:

josh146: Can verify, the `BSgate` hasn't changed, it has been `BSgate(theta, phi)` from the beginning 🙂 

What _has_ changed is the backend API; the backend call used to be `BaseBackend.beamsplitter(t, r)`, requiring the frontend to convert the input theta, phi into transmission and reflection. But the UI has not changed.

strawberryfields/ops.py
nquesada: I don't think this warning is needed. Only an error must be raised if r is complex (not sure if this is already done).

antalszava:
* The mentioned 3 gates (Coherent, DisplacedSqueezed, Dgate) raise errors if any of the arguments are complex (no warnings)
* Tests were adjusted including checks for raising errors

strawberryfields/ops.py
```diff
@@ -587,8 +587,11 @@ class Coherent(Preparation):
         :math:`\mathbf{V}= \frac{\hbar}{2} I`.
     """
 
-    def __init__(self, r=0.0, p=0.0):
-        super().__init__([r, p])
+    def __init__(self, r=0.0, phi=0.0):
+        if (np.iscomplex([r, phi])).any():
+            raise ValueError("The arguments of Coherent(r, phi) cannot be complex")
+
+        super().__init__([r, phi])
```
nquesada, line 594: what is this `super()` ?

antalszava: `super()` creates an instance of the parent class, `Preparation`. With this syntax we can place a call to `Preparation.__init__`, running the `__init__` method of the parent class.

This was something originally there too, the addition would be the error being raised.

PR #429, opened 2020-06-24
Add connection to device spec API

thisac:
**Context:**
A new device spec API has been created and thus needs to be connected to Strawberry Fields.

A new `DeviceSpec` class has been added, containing device specifications for a certain target, along with a `connection.get_device` method that fetches the device specifications for the target chip and returns a `DeviceSpec` object.

**Benefits:**
Information about the hardware devices can now be refreshed and used in Strawberry Fields without the need to update Strawberry Fields.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


.pylintrc
```diff
@@ -10,7 +10,7 @@ extension-pkg-whitelist=numpy,tensorflow,scipy,networkx,strawberryfields
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis. It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=numpy,tensorflow,scipy,networkx,strawberryfields,strawberryfields.parameters
+ignored-modules=numpy,tensorflow,scipy,networkx,strawberryfields,strawberryfields.parameters,collections.abc
```
josh146, line 13: For some reason, pylint/codefactor was saying 'No Sequence in module collections.abc'

strawberryfields/api/connection.py
```diff
@@ -135,6 +136,29 @@ def use_ssl(self) -> bool:
         """
         return self._use_ssl
 
+    def get_device(self, target: str) -> DeviceSpec:
```
antalszava, line 139: :100:

antalszava, line 139: (Minor suggestion)
```suggestion
    def get_device_spec(self, target: str) -> DeviceSpec:
```

strawberryfields/circuitspecs/circuit_specs.py
```diff
@@ -380,3 +383,9 @@ def __contains__(self, item):
 
     def __repr__(self):
         return ", ".join([str(i) for i in self.ranges])
+
+    def __eq__(self, other):
+        if len(self.ranges) != len(other.ranges):
+            return False
+
+        return all(i == j for i, j in zip(self.ranges, other.ranges))
```
antalszava, line 391: Just a note here: this assumes that the `Ranges` obtained contain `Range` objects in a monotonously increasing fashion. I'd think that that's okay.

```python
from strawberryfields.circuitspecs import Ranges

x = Ranges([0], [0.2, 0.55], [1.0])
y = Ranges([0], [1.0], [0.2, 0.55])

assert x == y
```
```
AssertionError                            Traceback (most recent call last)
<ipython-input-6-cdb229d2c821> in <module>
      4 y = Ranges([0], [1.0], [0.2, 0.55])
      5 
----> 6 assert x == y

AssertionError: 
```

.github/CHANGELOG.md
```diff
@@ -116,6 +116,11 @@
   [(#378)](https://github.com/XanaduAI/strawberryfields/pull/378)
   [(#381)](https://github.com/XanaduAI/strawberryfields/pull/381)
 
+* Strawberry Fields can now access the Xanadu Cloud device specifications API.
+  The ``Connection`` class has a new method ``Connection.get_device(target)``,
```
antalszava, line 120: (Minor suggestion): having `Connection.get_device(target)` might imply that it is a static method as it takes is no `self` argument
```suggestion
  The ``Connection`` class has a new method ``Connection.get_device``,
```

strawberryfields/api/devicespec.py
```diff
@@ -0,0 +1,136 @@
...
+class DeviceSpec:
+    """The specifications for a specific hardware device.
+
+    Args:
+        target (str): name of the target hardware device
+        layout (str): string containing the Blackbird circuit layout
+        modes (int): number of modes supported by the target
+        compiler (list): list of supported compilers
+        gate_parameters (dict): parameters for the circuit gates
+        connection (strawberryfields.api.Connection): connection over which the
```
antalszava, line 35:
```suggestion
        connection (strawberryfields.api.Connection): connection over which
```

strawberryfields/api/devicespec.py
```diff
@@ -0,0 +1,136 @@
...
+    def gate_parameters(self):
+        """dict[str, strawberryfields.circuitspecs.Ranges]: A dictionary of gate parameters
+        and allowed ranges.
+
+        The parameter names correspond to those present in the Blackbird circuit layout.
+
+        **Example**
+
+        >>> spec.gate_parameters
+        {'squeezing_amplitude_0': x=0, x=1, 'phase_0': x=0, 0≤x≤6.283185307179586}
```
antalszava, line 79: (Out of scope): find `'squeezing_amplitude_0': x=0, x=1` slightly confusing, would perhaps prefer `'squeezing_amplitude_0': x ∈ {0, 1}`. For intervals this would get slightly more complicated, having to have unions of sets with `U`. Could just be me.

strawberryfields/api/devicespec.py
```diff
@@ -0,0 +1,136 @@
...
+        correspond to those present in the Blackbird circuit layout. Parameters not
+        present will be assumed to have a value of 0.
+
+        **Example**
+
+        Device specifications can be retrieved from the API by using the
+        :class:`~.Connection` class:
+
+        >>> spec.create_program(squeezing_amplitude_0=0.43)
+        <strawberryfields.program.Program at 0x7fd37e27ff50>
```
antalszava, line 104: Might be missing a `Kwargs` and `Returns` subsection here in the docstring

strawberryfields/api/devicespec.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # evaluate the blackbird template
+        bb = bb(**parameters)
+        prog = sf.io.to_program(bb)
+        return prog
+
+    def refresh(self):
+        """Refreshes the device specifications"""
+        device = self._connection._get_device_dict(self.target)
+        self.__init__(target=self.target, connection=self._connection, **device)
```
antalszava, line 136: (Minor suggestion): feels a bit strange to call on `__init__` again with certain arguments possibly changing. One way could be factoring out setting layout, modes, compiler and gate_parameters into a separate method and calling that in `__init__` and here too.

strawberryfields/api/devicespec.py
thisac: Agreed! We're thinking about updating the `Ranges` class to produce nicer/clearer outputs (it's quite messy right now since the actual output would be `{'squeezing_amplitude_0': squeezing_amplitude_0=0, squeezing_amplitude_0=1, 'phase_0': phase_0=0, 0≤phase_0≤6.283185307179586}` in this case. :confused: 

This would be in a separate PR, though, and it could also solve the issue you mentioned above with the ordering.

strawberryfields/circuitspecs/circuit_specs.py
josh146: Good point @antal!

strawberryfields/api/devicespec.py
thisac: Added! :slightly_smiling_face:

strawberryfields/api/connection.py
josh146: I suppose these two are synonymous in a way?

strawberryfields/api/devicespec.py
josh146: Yep, let's save this for a subsequent PR

josh146:
@thisac, @antalszava: I've gone through and made suggested changes, and merged in master. Might be good if you have one last look through this, and then we can merge it into master and work on finishing #432 🙂

strawberryfields/api/devicespec.py
```diff
@@ -0,0 +1,141 @@
...
+
+        >>> spec.gate_parameters
+        {'squeezing_amplitude_0': x=0, x=1, 'phase_0': x=0, 0≤x≤6.283185307179586}
+        """
+        gate_parameters = dict()
+
+        for gate_name, param_ranges in self._spec["gate_parameters"].items():
+            # convert gate parameter allowed ranges to Range objects
+            range_list = [[i] if not isinstance(i, Sequence) else i for i in param_ranges]
+            gate_parameters[gate_name] = Ranges(*range_list, variable_name=gate_name)
```
thisac, line 86: Maybe we should just use `x` as the ranges variable for now (since that's used in the docstring, and it looks a bit cleaner). Then we can update all of this in a later PR?

```suggestion
            gate_parameters[gate_name] = Ranges(*range_list)
```

strawberryfields/api/connection.py
antalszava: Hmm, generally speaking, I think `get_device` would imply to the reader that by calling this method we obtain some sort of device object. In the context of SF this could be associated as an engine, but for example, it would fit the terminology of PennyLane well as there are `Device` objects there.

As a specific example in SF, a `RemoteEngine` instance stores a `Connection` object that can be accessed with `RemoteEngine.connection`. One could then place a call like `RemoteEngine.connection.get_device` which in itself does not entirely describe that we get details about the device that is being run.

My point is that the reader has to have the background knowledge here that `get_device` returns the specification of the device which I believe could be conveyed by renaming the method.

antalszava:
:100:

thisac:
Thanks @josh146. Looks good! I only had one, very minor, comment on the way `Ranges` is called for the `gate_parameters` property. Other than that, I think it looks fine. :slightly_smiling_face:

strawberryfields/api/devicespec.py
josh146: Good idea!

strawberryfields/api/connection.py
josh146: You've convinced me, I've changed it :)

PR #432, opened 2020-06-26
Device spec API implementation

thisac:
**Context:**
The device spec API need to be implemented so that it is used/called when creating/compiling a program.

Three major changes have been made:

1. `CircuitSpecs` have been renamed to `Compiler`.

2. A new property `eng.spec` has been added to `RemoteEngine` that pings the device spec API and downloads the hardware device specifications and stores it in a `DeviceSpec` object.

3. The `Program.compile` method has been updated to support `DeviceSpec` objects as inputs as well as an optional `compiler` argument.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
@thisac, I've worked a bit on this PR today, finishing the `circuitspec` to `compiler` refactor, and integrating in the device specs. In particular, I've made the following changes:

* Factored out the parameter validation into a separate `DeviceSpec.validate_parameters` method.

* Renamed the `BaseBackend.circuit_spec` property to `BaseBackend.compiler`

* Deleted the TensorFlow circuit spec (since it is identical to the Fock circuit spec).

* Deleted `X8Spec` and `X12Spec`.

* Removed `modes`, `remote`, `parameter_ranges`, and `local` from the `Compiler` class, since they have never been used (and belong in the `DeviceSpec` class instead).

* Previously, `sf.io.to_program` would attempt to both convert a blackbird program to a SF program, _and_ compile it using the `bb.target` attribute. This no longer makes sense, since device targets are no longer local, so implicit compilation has been removed. This probably should not have been occurring at all; a program loader should not implicitly be transforming the program.

* Finally, the engine would skip compilation if `prog.target is not None`. This also does not make sense; having the `target` attribute of a program set does not guarantee that a program has been correctly compiled, and is easily spoofed. For now, every submitted program is recompiled, but we should determine a better approach of marking a program as 'compiled'.

josh146:
@antalszava @thisac I think this is now ready for review! @nquesada, what should the default compiler be for the X chips? `Xcov` or `Xunitary`?

doc/introduction/circuits.rst
```diff
@@ -241,9 +241,23 @@ a particular layout or topology. For example, the ``gbs`` compile target will
...
-.. code-block:: python3
+>>> prog2 = prog.compile('gbs')
+
+Programs can also be compiled for specific hardware devices using the Xanadu Cloud platform.
+After instantiating a remote engine for the target hardware, the device specifications
+can be accessed and used for compilation:
+
+>>> eng = sf.RemoteEngine("X8")
+>>> device = eng.device_spec
+>>> prog2 = prog.compile(device)
```
antalszava, line 252: An empty program (only measurements) seems to raise an error. Might be worth putting a check such that it's checked if `seq[0].op.p[0]` is going to be valid.
```python
import strawberryfields as sf
import strawberryfields.ops as ops

prog = sf.Program(8)

with prog.context as q:
    ops.MeasureFock() | q

eng = sf.RemoteEngine("X8")
device = eng.device_spec
prog2 = prog.compile(device)
```

Output:
```python
---> 11 prog2 = prog.compile(device)

~/xanadu/strawberryfields/strawberryfields/program.py in compile(self, device_or_compiler, force_compiler, **kwargs)
    553             seq = pu.optimize_circuit(seq)
    554 
--> 555         seq = compiler.compile(seq, self.register)
    556 
    557         # create the compiled Program

~/xanadu/strawberryfields/strawberryfields/compilers/xcov.py in compile(self, seq, registers)
     84 
     85         # extract the compiled symplectic matrix
---> 86         S = seq[0].op.p[0]
     87 
     88         if len(used_modes) != n_modes:

IndexError: list index out of range
```

josh146: Good catch @antalszava! It looks like this is an existing bug in Xcov and Xunitary that are present in master, but only now being picked up since they are now the default.

strawberryfields/api/connection.py
```diff
@@ -154,6 +154,7 @@ def _get_device_dict(self, target: str) -> dict:
         response = requests.get(self._url(path), headers=self._headers)
 
         if response.status_code == 200:
+            self.log.info("The device spec %s has been successfully retrieved.", target)
```
antalszava, line 157: :100:

strawberryfields/api/devicespec.py
```diff
@@ -87,6 +97,24 @@ def gate_parameters(self):
...
+        for p, v in parameters.items():
+            if p in self.gate_parameters and v not in self.gate_parameters[p]:
+                # parameter is present in the device specifications
+                # but the user has provided a disallowed value
+                raise ValueError(
+                    f"{p} has invalid value {v}. Only {self.gate_parameters[p]} allowed."
+                )
+
+            if p not in self.gate_parameters:
+                raise ValueError(f"Parameter {p} not a valid parameter for this device")
```
antalszava, line 116: (Minor suggestion): wondering if it was nice to log something if the parameters are valid :thinking: 
`self.log.info("The validation of parameters was completed.")`

Not entirely convinced (doesn't seem super crucial), more of an idea. If decided on, a logger object can be set up for `DeviceSpec` instances similar to how it is managed in `connection.py`.

strawberryfields/compilers/__init__.py
```diff
@@ -31,39 +31,30 @@
...
-specs = (
-    X8Specs,
-    X8_01,
-    X12Specs,
-    X12_01,
-    X12_02,
-    FockSpecs,
-    GaussianSpecs,
-    GBSSpecs,
-    TFSpecs,
```
antalszava, line 59: Is compilation for the TFBackend something that is no longer relevant?

strawberryfields/engine.py
```diff
@@ -44,7 +44,7 @@ class BaseEngine(abc.ABC):
         backend_options (Dict[str, Any]): keyword arguments for the backend
     """
 
-    def __init__(self, backend, backend_options=None):
+    def __init__(self, backend, *, backend_options=None):
```
antalszava, line 47: How come the signature change was needed?

strawberryfields/engine.py
```diff
@@ -317,10 +317,6 @@ class LocalEngine(BaseEngine):
         backend_options (None, Dict[str, Any]): keyword arguments to be passed to the backend
     """
 
-    def __init__(self, backend, *, backend_options=None):
-        backend_options = backend_options or {}
-        super().__init__(backend, backend_options)
-
```
antalszava, line 323: Curious why this could be moved to `BaseEngine`

doc/introduction/circuits.rst
josh146: This has now been fixed as of 4487200 🙂

strawberryfields/compilers/__init__.py
josh146: Yep, the TF backend simply uses the `Fock` compiler. Previously, we required that each backend had its own CircuitSpec, but with the refactor, different backends can now use the same compiler.

strawberryfields/engine.py
josh146: Ah, I  noticed that this `__init__` wasn't doing anything that `BaseEngine.__init__` wasn't doing, and the linter was complaining! Simply deleting this here caused everything to continue working fine.

Note that we plan to merge `BaseEngine` back with `LocalEngine`, since they are essentially the same class, it is the only inheritance, and we don't expect to ever have any other subclasses.

strawberryfields/compilers/__init__.py
antalszava: got it! How does it use the `Fock` compiler? :thinking: Would a `compiler = "fock"` be needed for the TFBackend?

strawberryfields/compilers/xcov.py
```diff
@@ -0,0 +1,165 @@
...
+            raise CircuitError("All modes must be measured.")
+
+        meas_seq = [seq[-1]]
+
+        if len(seq) == 1:
+            # program consists only of measurements
+            sq_seq = [
+                Command(ops.S2gate(0), [registers[i], registers[i + half_n_modes]])
+                for i in range(half_n_modes)
+            ]
```
antalszava, line 82: how come there's squeezing for when only measurements were defined?

strawberryfields/compilers/xunitary.py
```diff
@@ -11,8 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Circuit class specification for the X class of circuits."""
+# pylint: disable=too-many-branches,too-many-statements
+"""General interferometer compiler for the X class of circuits."""
 
+from collections import defaultdict
```
antalszava, line 17: oh, this is so useful! :heart_eyes:

strawberryfields/compilers/xunitary.py
```diff
@@ -22,21 +24,25 @@
...
-from .gbs import GBSSpecs
+from .compiler import Compiler
+from .gbs import GBS
 from .gaussian_unitary import GaussianUnitary
 
 
-class Xunitary(CircuitSpecs):
-    """Circuit specifications for the X class of circuits."""
+def list_duplicates(seq):
+    """Returns a dictionary representing the duplicated values in the sequence
```
antalszava, line 33: (Minor suggestion): a generator object seems to be returned here
```suggestion
    """Returns a generator representing the duplicated values in the sequence
```

strawberryfields/compilers/xcov.py
nquesada: hey Antal, I am not sure I understand your question.

antalszava: Hi Nico, this is a branch where the user only added measurements, hence the program only consists of measurements. I was unsure why commands for `S2gate` are added here, but I would assume it is just because of how the circuit should look like. One thing I missed to parse previously was the squeezing amount being 0.

strawberryfields/program.py
```diff
@@ -436,27 +447,44 @@ def _linked_copy(self):
...
+        >>> device = eng.device_spec
+        >>> prog2 = prog.compile(device, "Xcov")
 
         Args:
-            target (str, ~strawberryfields.circuitspecs.CircuitSpecs): short name of the target
-                circuit specification, or the specification object itself
+            device_or_compiler (str, ~strawberryfields.compilers.Compiler, ~strawberryfields.api.DeviceSpec):
+                compiler name or device specification object to use for program compilation
+            force_compiler (str, ~strawberryfields.compilers.Compiler): Optionally provide a compile strategy.
+                This overrides the compile strategy specified by a hardware :class:`~.DevicSpec`.
```
antalszava, line 487: Having the ability to pass a `~strawberryfields.compilers.Compiler` to both `device_or_compiler` and `force_compiler` seems a little bit odd.

Would having a joint `device_options` dictionary be worth considering here? One that can be passed as a positional argument. It could have two keys: `"device"` and `"compile"`. At least one would have to be in the dictionary, and if both are, then the `compile` would overwrite the compilation strategy (the current `force_compiler` option).

strawberryfields/compilers/xcov.py
nquesada: Understood. Yeah it is to match the chip description and make sure that a correctly formatted blackbird script is returned.

strawberryfields/program.py
```diff
@@ -500,23 +552,27 @@ def compile(self, target, **kwargs):
...
+        run_options = {k: kwargs[k] for k in ALLOWED_RUN_OPTIONS if k in kwargs}
+        compiled.run_options.update(run_options)
 
-        # get run options of compiled program
-        # for the moment, shots is the only supported run option.
-        if "shots" in kwargs:
-            compiled.run_options["shots"] = kwargs["shots"]
+        # set backend options of the program
+        backend_options = {k: kwargs[k] for k in kwargs if k not in ALLOWED_RUN_OPTIONS}
+        compiled.backend_options.update(backend_options)
```
antalszava, line 568: this allows for non allowed backend options too, just checking if that's alright
```python
import strawberryfields as sf
import strawberryfields.ops as ops

prog = sf.Program(8)

with prog.context as q:
    ops.MeasureFock() | q

eng = sf.RemoteEngine("X8")
device = eng.device_spec
prog2 = prog.compile(device, **{'some_kwarg:': 1})
print(prog2.backend_options)
```
```
{'some_kwarg:': 1}
```

tests/frontend/compilers/test_compiler.py
```diff
@@ -0,0 +1,130 @@
...
+        assert circuit[0]['name'] == 'Sgate'
+        assert circuit[1]['name'] == 'Dgate'
+        assert circuit[2]['name'] == 'BSgate'
+        assert circuit[3]['name'] == 'MeasureFock'
+        assert circuit[4]['name'] == 'MeasureFock'
+
+        # check arguments
+        assert circuit[0]['args'] == [0, 0]
+        assert circuit[1]['args'] == [-7.123]
+        assert circuit[2]['args'] == [0]
```
antalszava, line 126: might be missing a bit here: how did the `{sq}` and `{theta}` symbolic parameters get numeric values?

tests/frontend/compilers/test_xcov.py
```diff
@@ -265,40 +309,6 @@ def test_incorrect_s2gate_modes(self, num_pairs):
         ):
             res = prog.compile("Xcov")
 
-    @pytest.mark.parametrize("num_pairs", [4, 5, 6, 7])
```
antalszava, line 268: why could these be deleted?

strawberryfields/program.py
thisac: I agree that this is a bit confusing and odd. Especially since a user could (but shouldn't) set `device_or_compiler` to a compiler _and_ add another `force_compiler` as well, which would simply just not do anything.

tests/frontend/compilers/test_compiler.py
thisac: The line `top = dummy.graph` excecutes:
```python
if bb.is_template():
            params = bb.parameters
            kwargs = {p: 0 for p in params}
```
inside of `compiler.graph` which seems to set all parameters to `0`.

strawberryfields/compilers/__init__.py
josh146: Yes, I think it is added as a class attribute in the backends! Note this is not a new change, the TFBackend previously specified `circuit_spec = "tf"`. But as noted above, this is entirely redundant, since the TF and Fock backends support the same gateset.

tests/frontend/compilers/test_compiler.py
antalszava: could it be a side effect there? :thinking:

strawberryfields/compilers/__init__.py
thisac: So, the `Fock` compiler is set to default (`compiler = "Fock"`) in the `BaseFock` class inside `backends.base.py`. The `TFBackend` inherits from `BaseFock` and doesn't replace the `compiler`. :slightly_smiling_face:

antalszava: thanks guys, these were the missing pieces :)

strawberryfields/program.py
josh146: I agree that it is confusing, and struggled a bit here.

Perhaps one solution is to use the Python `*` syntax, so that positional arguments are not allowed. Then we could do something like

```python
def compile(self, *, device=None, compiler=None):
```

A user could then optionally pass one or the other. This avoids the weird issue with mutually exclusive positional arguments!

strawberryfields/program.py
josh146: Yep, hardware backend options might be unknown by SF, so we can't validate them here! They will be validated by the API instead.

strawberryfields/compilers/__init__.py
josh146: oh that's even nicer!! I did that a couple of weeks ago in this PR, and completely forgot lol

strawberryfields/compilers/xunitary.py
```diff
@@ -103,22 +109,27 @@ def compile(self, seq, registers):
...
+                    r += removed_cmd.op.p[0]
+                    phi_new = removed_cmd.op.p[1]
+
+                    if k > 0 and phi_new != phi:
+                        raise CircuitError("Cannot merge S2gates with different phase values.")
+
+                    phi = phi_new
+
+                i, j = mode
+                B.insert(indices[0], Command(ops.S2gate(r, phi), [registers[i], registers[j]]))
```
josh146, line 132: @nquesada I noticed that your xunitary test file had a commented out test for merging S2 gates. I added the functionality here so that the test would pass

josh146: I only consider the case where all S2gates to be merged have the same value of phi though

tests/frontend/compilers/test_xcov.py
josh146: The compilers are no longer in charge of validating parameter values 🙂 These tests have been deleted and moved to `test_program`

strawberryfields/program.py
antalszava: What would happen if both `device` and `compiler` are left as `None`?

thisac: I think that the suggestion from @josh146 might actually be quite nice. Having both `device` and `compiler` as `None` could perhaps simply just use a default compiler (either `Xcov` or `Xunitary`)?

strawberryfields/program.py
josh146: I would like to revisit how we pass run and backend options around the place, though. It is very 'ad hoc' and should be cleaned up.

thisac:
@josh146 @antalszava I've updated the signature of `Program.compile` to `compile(self, *, device=None, compiler=None)` along with all the places that it was called. I've also added this to the docstring + documentation.

The only thing I think is left now is fixing the pylint issues regarding the tf-backend not overriding inherited methods, but I'm not sure why this is happening now (nor what would be the best way to handle it; disabling the error, etc.)

doc/introduction/circuits.rst
```diff
@@ -241,20 +241,20 @@ a particular layout or topology. For example, the ``gbs`` compile target will
...
+can be accessed and used for compilation using the keywords argument ``device``:
 
 >>> eng = sf.RemoteEngine("X8")
 >>> device = eng.device_spec
->>> prog2 = prog.compile(device)
+>>> prog2 = prog.compile(device=device)
 
 If no compile strategy is supplied, the default compiler from the device
-specification is used. This can be overridden via the ``force_compiler`` argument:
+specification is used. This can be overridden by also defining a compiler:
```
josh146, line 255:
```suggestion
specification is used. This can be overridden by also providing the compiler to be used:
```

doc/introduction/circuits.rst
```diff
@@ -241,20 +241,20 @@ a particular layout or topology. For example, the ``gbs`` compile target will
 compile a circuit consisting of Gaussian operations and Fock measurements
 into canonical Gaussian boson sampling form.
 
->>> prog2 = prog.compile('gbs')
+>>> prog2 = prog.compile(compiler="gbs")
 
 Programs can also be compiled for specific hardware devices using the Xanadu Cloud platform.
 After instantiating a remote engine for the target hardware, the device specifications
-can be accessed and used for compilation:
+can be accessed and used for compilation using the keywords argument ``device``:
```
josh146, line 248:
```suggestion
can be accessed and used for compilation using the ``device`` keyword argument:
```

josh146:
[ch263]

josh146:
[ch263]

PR #433, opened 2020-07-02
Add ability to access all samples

antalszava:
**Context:**
SF allows having multiple measurements on a single mode for memory efficiency. The `Result.samples` attribute stores the samples obtained. This array has a shape of `(shots, modes)` and stores the outcome of the last measurement for each mode.

However, it can be desired to access the outcome of preceding measurements and not just the last.

Adds the instance attribute `_all_samples` to the `LocalEngine` class, a list that stores `(register index, outcome)` tuples. This attribute can be accessed by `eng._all_samples`.

**Benefits:**
All measurement outcomes are accessible.

For example:
```python
import strawberryfields as sf
from strawberryfields.backends.base import BaseBackend


prog = sf.Program(5)

eng = sf.Engine("gaussian", backend_options={"cutoff_dim": 6})

with prog.context as q:
    ops.MeasureFock() | q[2]
    ops.MeasureFock() | (q[1], q[3])
    ops.MeasureX | q[2]

result = eng.run(prog)

# Check that the number of all the samples equals to the number
# of measurements
assert len(eng._all_samples) == 4

print(eng._all_samples)
```
Can output:
```
[(2, 0), (1, 0), (3, 0), (2, -1.6111120837627806)]
```

**Possible Drawbacks:**
There might be solutions which integrate more naturally with the UI for Strawberry Fields. An alternative could be changing from returning samples of `(shots, modes)` to `(shots, measured_modes)`.

Therefore this feature is more for advanced access.

**Related GitHub Issues:**
N/A

### DIFF
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -316,7 +316,7 @@ class LocalEngine(BaseEngine):
     """
 
     def __init__(self, backend, *, backend_options=None):
-        self._all_samples = []
+        self._all_samples = {}
         backend_options = backend_options or {}
         super().__init__(backend, backend_options)
 
@@ -324,7 +324,7 @@ class LocalEngine(BaseEngine):
         return self.__class__.__name__ + "({})".format(self.backend_name)
 
     def reset(self, backend_options=None):
-        self._all_samples = []
+        self._all_samples = {}
         backend_options = backend_options or {}
         super().reset(backend_options)
         self.backend.reset(**self.backend_options)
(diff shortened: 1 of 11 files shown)


tests/frontend/test_engine.py
```diff
@@ -209,6 +209,67 @@ def test_measuring_same_modes(self, eng):
...
+
+        result = eng.run(prog)
+
+        # Check that the number of all the samples equals to the number
+        # of measurements
+        assert len(eng._all_samples) == 4
+
+        correct_samples = [(2, 0), (1, 0), (3, 0), (2, 1)]
+
+        assert [(i,bool(j)) for i,j in eng._all_samples] == correct_samples
```
nquesada, line 249:
```suggestion
        assert [(i, bool(j)) for i, j in eng._all_samples] == correct_samples
```

strawberryfields/engine.py
```diff
@@ -335,11 +337,22 @@ def _run_program(self, prog, **kwargs):
...
+                    # Store the measurement outcomes such that not only the last outcome is retrievable
+                    for ind, reg in enumerate(cmd.reg):
+                        # Retrieve the index and the value
+                        # Values are stored in a nested sequence
+                        tup = (reg.ind, val[0][ind])
+                        self._all_samples.append(tup)
+
                     for i, r in enumerate(cmd.reg):
                         if batches:
                             samples_dict[r.ind] = val[:, :, i]
```
thisac, line 358: This can probably be combined into the same for-loop. Also, this would only work for single-shot, and not at all with batches, right? Since you're only choosing the first index of `val`.

```suggestion
                    for i, r in enumerate(cmd.reg):
                        # Retrieve the index and the value
                        # Values are stored in a nested sequence
                        tup = (reg.ind, val[0][ind])
                        self._all_samples.append(tup)
                        
                        if batches:
                            samples_dict[r.ind] = val[:, :, i]
```

antalszava: Oh, this slipped my mind entirely! :sweat_smile: thank you :heart_eyes: :blush:

antalszava:
@thisac thanks for the comments, they helped a lot! :blush: changed to having a dictionary for bookkeeping (went with having lists as values as we could expect them to be mutated) and added tests on multiple shots and batches.

josh146:
Hi @antalszava, it looks good to me, but I have one suggestion --- can we store all the samples in the created results object? There will likely be use cases in the future where the user has access to the results, but not the engine (and it seems the natural place to store all the samples)

PR #434, opened 2020-07-02
Add duschinsky function to qchem utils

soranjh:
This PR adds the function `duschinsky` to `sf.apps.qchem.utils` for computing Duschinsky rotation matrix and displacement vector.


strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def duschinsky(
+    Li: np.ndarray, Lf: np.ndarray, ri: np.ndarray, rf: np.ndarray, wf: np.ndarray, m: np.ndarray
+) -> Tuple[np.ndarray, np.ndarray]:
+    r"""Generate the Duschinsky rotation matrix and displacement vector.
+
+    **Example usage:**
+
+    >>> Li = np.array([[-0.08727946], [ 0.00000000], [ 0.00000000],
```
ixfoduap, line 31: Does the function not work if the inputs are lists instead of numpy arrays? :thinking: Seems annoying to use `np.array` in front of every input

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    >>> ri = np.array([-0.0236542994, 0.0000000000, 0.0000000000,
+    >>>                 1.2236542994, 0.0000000000, 0.0000000000])
+    >>> rf = np.array([ 0.0000000000, 0.0000000000, 0.0000000000,
+    >>>                 1.4397000000, 0.0000000000, 0.0000000000])
+    >>> wf = np.array([1363.210])
+    >>> m = np.array([11.00931] * 3 + [1.00782] * 3)
+    >>> Ud, delta = duschinsky(Li, Lf, ri, rf, wf, m)
+    (array([[0.99999546]]), array([-1.1755024]))
+
+    Args:
```
ixfoduap, line 44: I think it should be `(np.array)` for the inputs

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Tuple
+
+import numpy as np
+from scipy.constants import c, h, m_u, pi
+
+
+def duschinsky(
+    Li: np.ndarray, Lf: np.ndarray, ri: np.ndarray, rf: np.ndarray, wf: np.ndarray, m: np.ndarray
+) -> Tuple[np.ndarray, np.ndarray]:
+    r"""Generate the Duschinsky rotation matrix and displacement vector.
```
ixfoduap, line 27: Maybe worth adding a bit more information here about how they are computed and what the inputs are

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    (array([[0.99999546]]), array([-1.1755024]))
+
+    Args:
+        Li (array): normal modes of the initial electronic state
+        Lf (array): normal modes of the final electronic state
+        ri (array): equilibrium molecular geometry of the initial electronic state
+        rf (array): equilibrium molecular geometry of the final electronic state
+        wf (array): normal mode frequencies of the final electronic state in units of
+        :math:`\mbox{cm}^{-1}`
+        m (array): atomic masses in units of unified atomic mass unit
```
ixfoduap, line 50:
```suggestion
        m (array): atomic masses in unified atomic mass units
```

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Returns:
+        tuple[array, array]: Duschinsky rotation matrix :math:`Ud`, Duschinsky displacement vector
+        :math:`\delta`
+    """
+    Ud = (Lf.T * m ** 0.5) @ (Li.T * m ** 0.5).T
+
+    d = (ri - rf) @ (Lf.T * m).T
+
+    l0 = np.diag((h / (wf * 100.0 * c)) ** 0.5 / 2.0 / pi) * 1.0e10 / m_u ** 0.5
```
ixfoduap, line 60: what is `m_u`? Mass units?

strawberryfields/apps/qchem/utils.py
soranjh: `np.array` is consistent with what we have in other functions such as `gbs_params` in `qchem.vibronic`. I f we switch to lists, we should probably do it everywhere.

strawberryfields/apps/qchem/utils.py
soranjh: We typically use `(array)` in the rest of the `apps` functions.

strawberryfields/apps/qchem/utils.py
soranjh: It is an imported `scipy` constant for converting atomic mass unit to kg.

strawberryfields/apps/qchem/utils.py
soranjh: Done.

strawberryfields/apps/qchem/__init__.py
```diff
@@ -20,6 +20,8 @@
 
     dynamics
     vibronic
+    utils
 """
 import strawberryfields.apps.qchem.dynamics
 import strawberryfields.apps.qchem.vibronic
+import strawberryfields.apps.qchem.utils
```
trbromley, line 27: Do we definitely want to make the `utils` module visible to the user?
The alternative being to do:
```suggestion
"""
import strawberryfields.apps.qchem.dynamics
import strawberryfields.apps.qchem.vibronic
from strawberryfields.apps.qchem.utils import duschinsky
```
This would allow users to go:
```python
from strawberryfields.apps import qchem

qchem.duschinsky(...)
```

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    Args:
+        Li (array): normal modes of the initial electronic state
+        Lf (array): normal modes of the final electronic state
+        ri (array): equilibrium molecular geometry of the initial electronic state
+        rf (array): equilibrium molecular geometry of the final electronic state
+        wf (array): normal mode frequencies of the final electronic state in units of :math:`\mbox{cm}^{-1}`
+        m (array): atomic masses in unified atomic mass units
+
+    Returns:
+        tuple[array, array]: Duschinsky rotation matrix :math:`U_d`, Duschinsky displacement vector
```
agran2018, line 83: The `U_d` is a bit confusing. Above I thought it was a typo but now I see it stands for Duschinsky, I guess. However, above we are also talking about a displacement `d`...,  I would suggest to use `U_D` or just remove the `D`.

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        :math:`\delta`
+    """
+    Ud = (Lf.T * m ** 0.5) @ (Li.T * m ** 0.5).T
+
+    d = (ri - rf) @ (Lf.T * m).T
```
agran2018, line 88: same comment here regarding the `d`

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+    l0 = np.diag((h / (wf * 100.0 * c)) ** 0.5 / 2.0 / pi) * 1.0e10 / m_u ** 0.5
```
agran2018, line 90: why do you need the factor `100`?, to transform units of `c` ?. do the imported constant stick to the MKS system?.

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+    delta = np.array(d @ np.linalg.inv(l0))
```
agran2018, line 92: Just a comment here: matrix inversion for large matrices may introduce numerical errors. I don't know if this is a problem for diagonal matrices, good to check. One possibility to avoid this is to define `l0_inv` above directly above so no need to perform a matrix inversion.

tests/apps/qchem/test_utils_qchem.py
```diff
@@ -0,0 +1,102 @@
...
+class TestDuschinsky:
+    """Tests for the function ``strawberryfields.apps.qchem.vibronic.duschinsky``"""
+
+    def test_duschinsky_parameters(self, p):
+        """Test if function outputs the correct duschinsky parameters."""
+
+        Li, Lf, ri, rf, wf, m, Ud, delta = p
+
+        assert np.allclose(utils.duschinsky(Li, Lf, ri, rf, wf, m)[0], Ud)
+        assert np.allclose(utils.duschinsky(Li, Lf, ri, rf, wf, m)[1], delta)
```
agran2018, line 102: nice!

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+    where :math:`U_d` is a matrix known as the Duschinsky rotation matrix and the displacement parameter
+    :math:`d` is related to the structural changes of the molecule during a vibronic transition.
+    The normal coordinates of a molecule can be represented in terms of atomic displacements as:
+
+    .. math:: q = L^T \sqrt{m} (r -r_e),
+
+    where :math:`r_e` represents the equilibrium geometry of the molecule, :math:`m` represents
+    atomic masses and :math:`L` is a matrix containing the eigenvectors of the mass-weighted
+    Hessian. The Duschinsky parameters :math:`U_d` and :math:`d` can be obtained as:
```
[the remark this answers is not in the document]
soranjh: I convert `U` to `U_d` everywhere.

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    >>> wf = np.array([1363.210])
+    >>> m = np.array([11.00931] * 3 + [1.00782] * 3)
+    >>> Ud, delta = duschinsky(Li, Lf, ri, rf, wf, m)
+    (array([[0.99999546]]), array([-1.1755024]))
+
+    Args:
+        Li (array): normal modes of the initial electronic state
+        Lf (array): normal modes of the final electronic state
+        ri (array): equilibrium molecular geometry of the initial electronic state
+        rf (array): equilibrium molecular geometry of the final electronic state
```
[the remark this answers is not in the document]
soranjh: I have tried to use lower-case as much as possible and at the same time stay consistent with literature notations.

strawberryfields/apps/qchem/utils.py
soranjh: 100 converts cm to m as wave numbers are reported in cm^-1.

soranjh:
> Thank you @soranjh. I have left some comments.
> 
> Question: is there a utility function to extract the input parameters from the output of the frequency calculations for the initial and final states?.

We are thinking of adding it.

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,95 @@
...
+        wf (array): normal mode frequencies of the final electronic state in units of :math:`\mbox{cm}^{-1}`
+        m (array): atomic masses in unified atomic mass units
+
+    Returns:
+        tuple[array, array]: Duschinsky rotation matrix :math:`U`, Duschinsky displacement vector
+        :math:`\delta`
+    """
+    U = (Lf.T * m ** 0.5) @ (Li.T * m ** 0.5).T
+
+    d = (ri - rf) @ (Lf.T * m).T
```
trbromley, line 89: Also not clear why this equation is different to the function docstring?
image

- Why is the order different?
- Why is the `m` term not square rooted?

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+    .. math:: q = L^T \sqrt{m} (r -r_e),
+
+    where :math:`r_e` represents the equilibrium geometry of the molecule, :math:`m` represents
+    atomic masses and :math:`L` is a matrix containing the eigenvectors of the mass-weighted
+    Hessian. The Duschinsky parameters :math:`U` and :math:`d` can be obtained as:
+
+    .. math:: U = L_f^T L_i,
+
+    .. math:: d = L_f^T \sqrt{m} (r_e^i-r_e^f).
```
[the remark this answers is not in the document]
soranjh: They are defined in line 29 as "The Duschinsky transformation relates the normal coordinates of the initial and final states in a vibronic transition, :math:`q_i` and :math:`q_f` respectively".

soranjh:
Thanks Tom!

> * Some of the equations in the docstring don't seem to be matching what is going on in the function, in particular `m` cropping up in unexpected places in the function. It would be good if both could match to make future maintainability easier.

The code and docstring equations match now. (they were looking different because quantum chemistry packages typically print normal modes in mass-unweighted format and I thought we can make it easier for users to just regenerate mass-weighted ones in the code. But it can be confusing for people that are not familiar with outputs of quantum chemistry packages so I reorganised things and mentioned it clearly in the docstring Args.) 

> * The input and output dimensions of the function surprised me, I'm curious why we expect 1 dimensional output?

Please note that the vibrational normal modes for a molecule containing N atoms are (3N) x (3N - 6) matrices for nonlinear molecules, and (3N) x (3N - 5) matrices for linear molecules. The 6 and 5 here refer to the number of translational and rotational degrees of freedom that should be subtracted from the total degrees of freedom. So for a diatomic molecule, which is necessarily linear, we have only 1 (3 * 2 - 5) vibrational normal mode that is defined by 6 Cartesian displacement components (3 for each atom). In this case the Duschinsky rotation matrix is simply a 1 x 1 matrix and the displacement vector also has one component. But for a nonlinear molecule with 4 atoms provided in the ` test_utils_qchem`, we have 6 vibrational modes and the Duschinsky rotation matrix is a 6 x 6 matrix and the displacement vector also has 6 components. I could use a larger example in the "Example usage" section but as you can see the amount of data is huge and does not render nicely.

I have added a paragraph to the docstring to explain these details. Please also note that we are thinking of adding a function that reads the output of quantum chemistry data and extracts the normal modes. We can explain the fundamentals a bit more in that function or maybe in a tutorial.

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    The Duschinsky transformation relates the normal coordinates of the initial and
+    final states in a vibronic transition, :math:`q_i` and :math:`q_f` respectively, as:
+
+    .. math:: q_f = U_d q_i + d.
+
+    where :math:`U_d` is a matrix known as the Duschinsky rotation matrix and the displacement parameter
+    :math:`d` is related to the structural changes of the molecule during a vibronic transition.
+    The normal coordinates of a molecule can be represented in terms of atomic displacements as:
+
+    .. math:: q = L^T \sqrt{m} (r -r_e),
```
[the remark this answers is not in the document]
ixfoduap: Python conventions are typically more important than scientific conventions. Variable and input names should be lowercase in python, so I think it's better to use `r`. 

This raises the question of using `li`, `lf` instead of `Li`, `Lf`, but since these are matrices, I think it's clearer to use uppercase.

strawberryfields/apps/qchem/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    where :math:`l` is a diagonal matrix containing the vibrational frequencies :math:`\omega` of
+    the final state:
+
+    .. math:: l_{kk} = \left ( \frac{\hbar }{2 \pi \omega_k c} \right )^{1/2},
+
+    where :math:`\hbar` is the reduced Planck constant and :math:`c` is the speed of light.
+
+    The vibrational normal mode matrix for a molecule with :math:`M` vibrational modes and
+    :math:`N` atoms is a :math:`3N * M` matrix where :math:`M = 3N - 6` for nonlinear molecules and
+    :math:`M = 3N - 5` for linear molecules. The Duschinsky rotation matrix for this molecule is a
```
ixfoduap, line 63:
```suggestion
    :math:`M = 3N - 5` for linear molecules. The Duschinsky rotation matrix of a molecule is an
```

tests/apps/qchem/test_utils_qchem.py
```diff
@@ -0,0 +1,100 @@
+# Copyright 2019 Xanadu Quantum Technologies Inc.
```
trbromley, line 1:
```suggestion
# Copyright 2020 Xanadu Quantum Technologies Inc.
```

PR #436, opened 2020-07-03
Update analytic expression in variance test

antalszava:
**Context:**
A test case for a displaced squeezed circuit was added for calculating variances. However, the analytic expression provided was incorrect.

Updates the analytic expression used for the displaced squeezed variance test.

**Benefits:**
The tolerance value used can be decreased.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


tests/backend/test_states.py
```diff
@@ -354,31 +354,32 @@ def test_number_expectation_displaced_squeezed(self, setup_backend, tol, batch_s
...
-            return var_ex
-
-        var_tol = 0.01
-        v0 = analytic_var(a0, r0, phi0)
-        v1 = analytic_var(a1, r1, phi1)
+            return squared_term
 
+        # Increasing the tolerance when using
+        # a cutoff
+        var_tol = tol if isinstance(backend, backends.BaseGaussian) else 7 * tol
```
antalszava, line 368: @josh146 was looking at possibilities for specifying the cutoff for `setup_backend` that is being used here (going from the default of 6 to 8 would mean the original tolerance value could be used). Would there be some way to specify a custom value for the test case?

josh146: @antalszava yes this is possible, you can simply do this in the test:

```python
backend = setup_backend(3)
# increase the cutoff to 8 for accuracy
backend.reset(cutoff_dim=8, pure=pure)
```

antalszava:
* Was revisiting analytic expressions for https://github.com/XanaduAI/pennylane-sf/pull/42 and this turned out to be amiss. :slightly_smiling_face: 
* Yes, the increased tolerance made it pass. :confused: If the parameters for displacement and squeezing are high and the Fock backend is used (with lower cutoff values), then the difference between the obtained result for the variance and the analytic value can be greater than the standard tolerance used here. However, it turned out this was not the case.

PR #438, opened 2020-07-10
Adds an Xstrict compiler

josh146:
Adds a strict compiler for the X class of circuits.

Ensure that the compiled program exactly matches the devices topology. As a result, this compiler only accepts :`S2gate`, `MZgate`, and `MeasureFock()` operations.

This compiler _must_ be used with an X series device specification.

```python
>>> eng = sf.RemoteEngine("X8")
>>> spec = eng.device_spec
>>> prog.compile(device=spec, compiler="Xstrict")
```


josh146:
[ch269]

heltluke:
I note that the tests involve adding dummy `Rgate` parameters or so-called "final phases". That is, if I understand correctly, for the Xstrict compiler to work on e.g. X8_01, the supplied BB program must contain 4 `S2gate`s, 12 `MZgate`s, _and_ 8 `Rgate`s.

Is this correct? If so, I suppose the docstring for the `Xstrict` class should be updated to mention the necessity of the inconsequential `Rgate`s. However, I also wonder whether these gates should be supplied by users of the Xstrict compiler, or inserted by SF...

josh146:
> However, I also wonder whether these gates should be supplied by users of the Xstrict compiler, or inserted by SF...

@heltluke this is a good point. Is requiring the user add in the final phase gates _too_ strict? What we could do is have the `Xstrict` compiler add in final phase gates with `phi=0` if they don't exist, so that the user doesn't need to specify them

heltluke:
>Is requiring the user add in the final phase gates _too_ strict? What we could do is have the `Xstrict` compiler add in final phase gates with `phi=0` if they don't exist, so that the user doesn't need to specify them.

That sounds incredibly convenient to me, but I guess we should run this by a number of others as well. @tymomboe @DavidSPhillips @lneuhaus @xanadu-zvernon

strawberryfields/program.py
```diff
@@ -572,7 +572,15 @@ def _get_compiler(compiler_or_name):
...
             bb_device = bb.loads(device.layout)
             bb_compiled = sf.io.to_blackbird(compiled)
-            user_parameters = match_template(bb_device, bb_compiled)
+
+            try:
+                user_parameters = match_template(bb_device, bb_compiled)
+            except bb.utils.TemplateError:
+                raise CircuitError(
+                    "Program cannot be used with the compiler '{}' "
+                    "due to incompatible topology.".format(compiler.short_name)
```
antalszava, line 581: Just for my own understanding: would it be sure that the reason is always `incompatible topology`? ([Seems like there could be other cases for `TemplateError` too](https://github.com/XanaduAI/blackbird/blob/60e1ae43cd97e8759496a5ecc177b16f9b0b8299/blackbird_python/blackbird/utils.py#L187), though I'd imagine that these are unlikely to arise here? :thinking: )

josh146: Oh, good catch @antalszava. Although, I would say from looking at the source code that the first three are unlikely to occur (possible even impossible to occur), since, by construction,

* Both arguments will always be a template
* This function uses the same version of BB to compute both arguments

josh146:
> That sounds incredibly convenient to me, but I guess we should run this by a number of others as well.

@heltluke I've just gone through the logic again, and unfortunately this is somewhat non-trivial just by construction. The way this compiler is written, it _avoids_ compilation and just performs a simple equivalence check. I might merge this in now as-is, and we can make a new PR to add in this behaviour.

PR #439, opened 2020-07-13
Check for complex arguments for symbolic parameters

antalszava:
**Context:**
The parameters of the `Dgate`, `Coherent` and `Displacedsqueezed` gates are checked such that no complex values are passed as gate parameters. This check is performed on the arguments when creating the gates. At this point, however, parameters can be symbolic. In such a case the check does not apply to the parameter value as it is substituted later (at runtime).

An example of a gate with a symbolic parameter:
```python
prog = sf.Program(1)

with prog.context as q:
    ops.MeasureHD | q[0]
    ops.Dgate(q[0].par) | q

print(q[0].val)
```
Outputs:
```
None
```
Moves the checking of arguments to the `_apply` method. Checks are applied to the output of the `par_evaluate` method such that numeric values are substituted to gate parameters.

With this change, checks will happen when the gates are applied (for example calling the `engine.run` method) rather than upon the creation of the gates.

**Benefits:**
The values used for symbolic parameters of the `Dgate`, `Coherent` and `Displacedsqueezed` gates are checked now too.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


strawberryfields/ops.py
```diff
@@ -588,14 +588,17 @@ class Coherent(Preparation):
...
-        if (np.iscomplex([r, phi])).any():
-            raise ValueError("The arguments of Coherent(r, phi) cannot be complex")
-
         super().__init__([r, phi])
 
     def _apply(self, reg, backend, **kwargs):
         r = par_evaluate(self.p[0])
         phi = par_evaluate(self.p[1])
+
+        tf_complex = any(hasattr(arg, "numpy") and np.iscomplex(arg.numpy()) for arg in [r, phi])
```
antalszava, line 597: @josh146 thank you so much for your point regarding TF! :heart_eyes:

This logic now assumes that the TF data structure (e.g. `tf.Variable`) would have a `numpy` attribute. Could there be cases which this would not cover? :thinking:

josh146: That should be fine! Good to always check if the gradient can still be computed (sometimes minor TF modifications can break the gradient tape):

```python
prog = sf.Program(1)
alpha = prog.params("alpha)

with prog.context as q:
    ops.Dgate(alpha) | q[0]

eng = sf.Engine("tf")
a = tf.Variable(0.5)

with tf.GradientTape() as tape:
    res = eng.run(prog, args={"alpha": a})
    loss = state.mean_photon([0])

grad = tape.gradient(loss, a)
print(grad)
```

Another option would be to perform the validation _only_ if the argument is a `numbers.Number` or NumPy array.

antalszava: That's a good idea, thank you! Perhaps this test case would cover that, right? https://github.com/XanaduAI/strawberryfields/blob/54490931368be926996b86a68c2d1083c1779d47/tests/integration/test_tf_integration.py#L337

Apparently calling `mean_photon` with a non-integer value (e.g. `state.mean_photon([0])`) seems to result in an error, although it would read perfectly fine (a problem with reshaping comes up). :thinking: This got me wondering, would it be would worth placing a check there?

antalszava: > Another option would be to perform the validation only if the argument is a numbers.Number or NumPy array.

How so? Would that attempt to substitute the `(np.iscomplex([r, phi])).any()` check? :thinking:

josh146: > Perhaps this test case would cover that, right? 

Looks like it does! If you set `a = tf.Variable(0.5 + 0.2j)`, is the correct exception still raised?

> Apparently calling mean_photon with a non-integer value (e.g. state.mean_photon([0])) seems to result in an error, 

I wouldn't worry too much about that case, since it is less likely to occur due to user error (the docstring specifies list of ints)

josh146: > How so? Would that attempt to substitute the (np.iscomplex([r, phi])).any() check? 🤔

Yes, I was thinking of
```python
if isinstance(r, (numbers.Number, np.ndarray)) and isinstance(phi, (numbers.Number, np.ndarray)):
    # put (np.iscomplex([r, phi])).any() check here
```

But this was only in case having r,phi be TF tensors caused issues!

antalszava: > If you set a = tf.Variable(0.5 + 0.2j), is the correct exception still `raised?`

Yep! Think this would align with the new `test_complex_symbolic_tf` test case.

> Yes, I was thinking of
> ```python
> if isinstance(r, (numbers.Number, np.ndarray)) and isinstance(phi, (numbers.Number, np.ndarray)):
>     # put (np.iscomplex([r, phi])).any() check here
> ```
> 
> But this was only in case having r,phi be TF tensors caused issues!

Oh, I see! With the current version, the TF Variable and Tensor cases should be fine:

**TF Variable**
```python
In [17]: r = 0                                                                                                                                    

In [18]: phi = tf.Variable(0.3+9j)                                                                                                                

In [19]: tf_complex = any(hasattr(arg, "numpy") and np.iscomplex(arg.numpy()) for arg in [r, phi])                                                

In [20]: if_statement = (np.iscomplex([r, phi])).any() or tf_complex                                                                              

In [21]: if_statement                                                                                                                             
Out[21]: True
```

**TF Tensor**

(Assuming that the `tf.Tensor` instance contains exactly one element)
```python
In [25]: phi = tf.constant([-3.25 + 5.75j])                                                                                                       

In [26]: tf_complex = any(hasattr(arg, "numpy") and np.iscomplex(arg.numpy()) for arg in [r, phi])                                                

In [27]: if_statement = (np.iscomplex([r, phi])).any() or tf_complex                                                                              

In [28]: if_statement                                                                                                                             
Out[28]: True

In [29]: r                                                                                                                                        
Out[29]: 0

In [30]: phi                                                                                                                                      
Out[30]: <tf.Tensor: id=43, shape=(1,), dtype=complex128, numpy=array([-3.25+5.75j])>
```

josh146: Perfect! Feel free to merge this in :)

PR #440, opened 2020-07-31
Time domain program class

nquesada:
**Context:**
Implements the time domain program class

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


tests/frontend/test_tdmprogram.py
```diff
@@ -0,0 +1,72 @@
...
+	    ops.BSgate(p[0]) | (q[2], q[1])
+	    ops.Rgate(p[1]) | q[2]
+	    ops.MeasureHomodyne(p[2]) | q[0]
+	eng = sf.Engine('gaussian')
+	result = eng.run(prog)
+	samples = reshape_samples(result.all_samples, c=1)
+	assert samples.shape == (1, len(R)*c)
+
+def test_epr():
+	sq_r = 1.0
```
nquesada, line 55: @fab1an-q : This is the test that I was trying to implement. It does not quite work (you can heck this by pulling this branch and running this function).
Once properly implemented the check should be something relatively easy. If we call `var` the variance of the squeezed collective quadrature then one could do something like `assert np.allclose(np.allclose(var, np.exp(-2*r), atol=atol)`

fab1an-q: I found two errors:
1) The gate parameters are incorrect for producing EPR states. It should be:
```
alpha = [0, np.pi/4]*c
phi = [np.pi/2, 0]*c
```

instead of:
```
alpha = [0, np.pi/2]*c
phi = [np.pi/4, 0]*c
```
I gave you the wrong parameters here, sorry.

2) I think the order in the `with` statement was mixed up:
```
with prog.context(alpha, phi, M, copies=copies) as (p, q):
	ops.Sgate(sq_r, 0) | q[2]
	ops.Rgate(p[0]) | q[2] # the Rgate's argument should be phi, i.e. p[1]
	ops.BSgate(p[1]) | (q[2], q[1]) # the BSgate's argument should be alpha, i.e. p[0]
	ops.MeasureHomodyne(p[2]) | q[0]
```

In the `prog.context` constructor, the order is `alpha, phi`. However, in the `with` statement, `p[0]` corresponds to the `Rgate` (whose argument should be `phi`) and `p[1]` corresponds to the `BSgate` (whose argument should be alpha).

The correct way would be:
```
with prog.context(alpha, phi, M, copies=copies) as (p, q):
    ops.Sgate(sq_r, 0) | q[2]
    ops.BSgate(p[0]) | (q[2], q[1])
    ops.Rgate(p[1]) | q[2]
    ops.MeasureHomodyne(p[2]) | q[0]
```

fab1an-q: I committed a corrected version including some code to plot nearest-neighbour correlations. This plots now show the characteristic EPR correlations:
EPR_test

strawberryfields/tdm/tdmprogram2.py
```diff
@@ -0,0 +1,225 @@
...
+        slice(0, 6, None)   for spatial mode A
+        slice(6, 10, None)  for spatial mode B
+        slice(10, 19, None) for spatial mode C
+        slice(19, 20, None) for spatial mode D.
+
+        This helps to define:
+        q[sm[0]] as concurrent modes of spatial mode A
+        q[sm[1]] as concurrent modes of spatial mode B
+        q[sm[2]] as concurrent modes of spatial mode C
+        q[sm[3]] as concurrent modes of spatial mode D.
```
josh146, line 185: This is really nice!

nquesada:
@josh146 There are a couple of things that pylint is complaining about and that I am not sure are intentional. Wondering if you could have a look? If they are intentional you can let me know and I can add the relevant `pylint disables`

nquesada:
Finally, just to confirm: Do we want eliminate any and all reference to `shift=after`? 
Two options here:
1. A hard removal where we can simply remove the `shift` option altogether and it is always assumed that it amount to `shift=end`
2. Leave the keyword shift in there with the default value? I the user uses any other value for `shift` we raise a warning. If in the future we decide to add more options for `shift` it will be simpler to add.

What do @josh146 and @fab1an-q think?

fab1an-q:
We could give the users the option to pass an integer as shift parameter, defining the step size of the qumode shift in each time bin. Might be interesting for some to experiment with it.

However, I think we should change the string "end". The name was useful to distinguish from shift="after", but we decided to remove this option (I believe?). So maybe something like shift="auto" (as compared to a user-controlled shift) makes more sense?

fab1an-q:
Just realized that the option for integers as shift parameters is currently implemented in the `TDMProgram`:

```
```

I think we could leave it this way. But if our automated shift-handling works as fine as we hope, I acknowledge that only few people will _not_ want to use it. So I am also ok with removing `shift` entirely from the arguments of the context manager.

josh146:
> However, I think we should change the string "end"

:+1:

> I think we could leave it this way. But if our automated shift-handling works as fine as we hope, I acknowledge that only few people will not want to use it. So I am also ok with removing shift entirely from the arguments of the context manager.

If it is less effort to leave it in/as-is, I'm happy to leave it as is!

nquesada:
Alright @josh146 , @fab1an-q and @josh146 . I think we are close to the end. 
I have lightly reformatted the great functions @fab1an-q wrote into proper tests by adding `assert` statements and calculating the theoretical values expected.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+                q = tuple(q_aux)
+
+            elif isinstance(self.shift, int):
+                q = shift_by(q, self.shift)  # shift at end of each time bin
+
+    def apply_op(self, cmd, q, t):
+        """Apply a particular operation on register q at timestep t"""
+        params = cmd.op.p.copy()
+
+        if par_is_symbolic(params[0]):
```
nquesada, line 220: @josh146 : I believe this `if` block should be removed. I believe one should never have symbolic parameters for a sampling circuit. If you think this is the case feel free to delete the block.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+        if isinstance(cmd.op, ops.Measurement):
+            modes = [r.ind for r in cmd.reg]
+            measurement_reg.extend(modes)
+            spatial_modes += len(modes)
+
+    if not spatial_modes:
+        raise ValueError("Must be at least one measurement.")
+
+    if spatial_modes is not len(N):
+        raise ValueError("Number of measurements operators must match number of spatial_modes.")
```
fab1an-q, line 65: `raise ValueError("Number of measurement operators must match number of spatial modes.")`

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+            self.N = N
+        self.concurr_modes = sum(self.N)
+
+        self.total_timebins = 0
+        self.spatial_modes = 0
+        self.measured_modes = []
+        super().__init__(num_subsystems=self.concurr_modes, name=name)
+
+    # pylint: disable=arguments-differ, invalid-overridden-method
+    def context(self, *args, copies=1, shift="end"):
```
fab1an-q, line 142: The string `"end"` only made sense as long as we had `shift="after"` as an alternative. How about `"auto"` or `"standard"` or similar to distinguish from a user-defined shift (which takes place at the end as well)?

nquesada: `default`?

tests/frontend/test_tdmprogram.py
```diff
@@ -0,0 +1,425 @@
...
+    phi = [np.pi / 2] * n
+    theta = [0, np.pi / 2] * (n // 2)  # Note that we measure x for mode i and the p for mode i+1.
+    reshaped_samples = np.array(singleloop(sq_r, alpha, phi, theta, copies)).reshape(copies, n)
+    nullifier = lambda x: np.array([-x[i - 2] + x[i - 1] - x[i] for i in range(2, len(x) - 2, 2)])[1:]
+    nullifier_samples = np.array([nullifier(y) for y in reshaped_samples])
+    delta = np.var(nullifier_samples, axis=0)
+    assert np.allclose(delta, 3 * np.exp(-2 * sq_r), rtol=5 / np.sqrt(copies))
+
+
+def test_two_by_n_cluster():
```
fab1an-q, line 238: The name of the function sounds like a 2D cluster state. Isn't the experiment about a 1D cluster state?

tests/frontend/test_tdmprogram.py
```diff
@@ -0,0 +1,425 @@
...
+    delayC = 5
+    delayD = 0
+
+    # concurrent modes in each spatial mode
+    concurrA = 1 + delayA
+    concurrB = 1 + delayB
+    concurrC = 1 + delayC
+    concurrD = 1 + delayD
+
+    N = [concurrA, concurrB, concurrC, concurrD]
```
fab1an-q, line 354: Btw, even though some of this code looks redundant, in my opinion this is the best way to define and understand how many concurrent modes we have in each spatial mode and, thereby, how `N` is defined.

tests/frontend/test_tdmprogram.py
```diff
@@ -0,0 +1,425 @@
...
+
+    nX1var = np.var(nX1)
+    nX2var = np.var(nX2)
+    nP1var = np.var(nP1)
+    nP2var = np.var(nP2)
+
+    assert np.allclose(nX1var, 4 * np.exp(-2 * sq_r), rtol=5 / np.sqrt(ntot))
+    assert np.allclose(nX2var, 4 * np.exp(-2 * sq_r), rtol=5 / np.sqrt(ntot))
+    assert np.allclose(nP1var, 4 * np.exp(-2 * sq_r), rtol=5 / np.sqrt(ntot))
+    assert np.allclose(nP2var, 4 * np.exp(-2 * sq_r), rtol=5 / np.sqrt(ntot))
```
fab1an-q, line 425: Thanks for formalizing all these tests, it looks great! One day, you'll have to teach me about `assert` and `allclose`.

fab1an-q (approved):

tests/frontend/test_tdmprogram.py
nquesada: when running `pytest tests/frontend/test_tdmprogram.py -x --tb=short --cov=strawberryfields/tdm --cov-report term-missing` I get
image

tests/frontend/test_tdmprogram.py
nquesada: changed it to `test_one_dimensional_cluster_tokyo`

strawberryfields/tdm/tdmprogram2.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # if shift == "after" and list(set(measurement_reg)) != [0]:
+    #     raise ValueError("Measurements allowed only on q[0].")
+
+    #####################################################################################
+    ##### THIS CHANGE ACCOUNTS FOR NEW PROPOSAL: ONE BELT PER SPATIAL MODE ##############
+    #####################################################################################
+    # N is now a list with number of qumodes for each spatial mode
+    if spatial_modes is not len(N):
+        raise ValueError("Numer of measurements operators must match number of spatial_modes.")
```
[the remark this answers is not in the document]
fab1an-q: `"Number of measurement operators"` instead of `"Number of measurements operators"`. (my typo)

tests/frontend/test_tdmprogram.py
```diff
@@ -0,0 +1,425 @@
...
+    sq_r = 5
+    alpha = [np.arccos(np.sqrt(1 / (n - i + 1))) if i != n + 1 else 0 for i in range(n + vac_modes)]
+    alpha[0] = 0.0
+    phi = [0] * (n + vac_modes)
+    phi[0] = np.pi / 2
+
+    # Measuring X nullifier
+    theta = [0] * (n + vac_modes)
+    samples_X = singleloop(sq_r, alpha, phi, theta, copies)
+    reshaped_samples_X = np.array(samples_X).reshape([copies, n + vac_modes])
```
fab1an-q, line 200: What does `reshape([copies, n + vac_modes])` do here?

nquesada: it turns the vector of length `copies * (n+vac_modes)` into a matrix of shape `[copies, n+vac_modes]`

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+        if isinstance(cmd.op, ops.Measurement):
+            modes = [r.ind for r in cmd.reg]
+            measurement_reg.extend(modes)
+            spatial_modes += len(modes)
+
+    if not spatial_modes:
+        raise ValueError("Must be at least one measurement.")
+
+    if spatial_modes is not len(N):
+        raise ValueError("Number of measurements operators must match number of spatial modes.")
```
fab1an-q, line 65: Sorry if I make this comment too often but the typo is still there. I believe it should say "measurement operators" instead of "measurement**S** operators".

tests/frontend/test_tdmprogram.py
fab1an-q: Love it! Does this work for >1 spatial modes too?

strawberryfields/tdm/tdmprogram.py
nquesada: Corrected now :)

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+
+def validate_measurements(circuit, N):
+    """Validate the TDM program measurements are correct"""
+    spatial_modes = 0
+    measurement_reg = []
+
+    for cmd in circuit:
+        if isinstance(cmd.op, ops.Measurement):
+            modes = [r.ind for r in cmd.reg]
+            measurement_reg.extend(modes)
```
josh146, line 58: Is `measurement_reg` used here? If not, is it redundant and can be deleted?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+    modes = reg_getter(q)
+
+    if not isinstance(modes, tuple):
+        modes = (modes,)
+
+    return modes
+
+
+def validate_measurements(circuit, N):
+    """Validate the TDM program measurements are correct"""
```
josh146, line 51: Might be good to extend this docstring with Args and Returns. Otherwise it's not so clear from looking at it what `N` should be!

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+        raise ValueError("Must be at least one measurement.")
+
+    if spatial_modes is not len(N):
+        raise ValueError("Number of measurement operators must match number of spatial modes.")
+
+    return spatial_modes
+
+
+def input_check(args, copies):
+    """Checks the input arguments have consistent dimensions"""
```
josh146, line 71: We should also extend the docstring here

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+    if len(set(param_lengths)) != 1:
+        raise ValueError("Gate-parameter lists must be of equal length.")
+
+    # check copies
+    if not isinstance(copies, int) or copies < 1:
+        raise TypeError("Number of copies must be a positive integer.")
+
+
+def _get_mode_order(num_of_values, N):
+    """Get the order by which the modes where measured"""
```
josh146, line 83: and here

josh146:
```suggestion
    """Get the order by which the modes were measured"""
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+    N = N.copy()
```
josh146, line 84: How come a copy of N is made here? It doesn't look like N is ever modified

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+        all_modes.append(ra * ceil(max(N) / len(ra)))
+
+    mode_order = [i for j in zip(*all_modes) for i in j]
+    mode_order *= ceil(num_of_values / len(mode_order))
+
+    return mode_order[:num_of_values]
+
+
+def reshape_samples(all_samples, modes, N):
+    """Reshapes the samples dict so that they have the expected correct shape
```
josh146, line 98:
```suggestion
    """Reshapes the samples dict so that they have the expected correct shape.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Reshapes the samples dict so that they the measured modes are the ones
```
josh146, line 100:
```suggestion
    Corrects the :attr:`~.Results.all_samples` dictionary so that the measured modes are the ones
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+    defined to be measured in the circuit instead of being spread over a larger
+    number of modes due to the mode-shifting occuring in TDMProgram.
```
josh146, line 102:
```suggestion
    defined to be measured in the circuit, instead of being spread over a larger
    number of modes due to the mode-shifting occurring in :class:`~.TDMProgram`.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Args:
+        all_samples (dict[int, list]): the raw measured samples
+        modes (Sequence[int]): the modes that are measured in the circuit
+        N (Sequence[int]): the number of used modes per belt/spatial modes
```
josh146, line 107: What does `used modes` mean in this context?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+    for i, m in enumerate(mode_order):
+        idx = modes[i % len(N)]
+        if idx not in new_samples:
+            new_samples[idx] = []
+        new_samples[idx].append(all_samples[m].pop(0))
+    return new_samples
+
+
+class TDMProgram(sf.Program):
+    """Represents a photonic quantum circuit in the time domain encoding"""
```
josh146, line 126: We should extend this docstring to include details, arguments, and minimal examples. See the [`Program`](https://strawberryfields.readthedocs.io/en/stable/code/api/strawberryfields.Program.html) docstring (source code [here](https://strawberryfields.readthedocs.io/en/stable/_modules/strawberryfields/program.html#Program)) for a good example.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,232 @@
...
+                self.measured_modes.append(cmd.reg[0].ind)
+
+        for i in range(self.total_timebins):
+            for cmd in cmds:
+                self.apply_op(cmd, q, i)
+
+            if self.shift == "default":
+                # shift each spatial mode SEPARATELY by one step
+                q_aux = list(q)
+                # Here the index used to be i in changed to j, check this is fine.
```
josh146, line 208: I think this comment can now be removed
```suggestion
```

strawberryfields/tdm/tdmprogram.py
josh146: Spoke offline, but for context this is required for the TDMProgram logic. What happens internally is:

* Inside the with context, all gate arguments are replaced with sympy variables. This is used to track which parameter is used in which gate.

* After the with context exits, the TDM loop is constructed, replacing the parameters with the numeric values given by the user.

tests/frontend/test_tdmprogram.py
```diff
@@ -0,0 +1,422 @@
...
+    x = singleloop(sq_r, alpha, phi, theta, copies)
+    np.random.seed(42)
+    y = singleloop(sq_r, alpha, phi, theta, copies, shift=1)
+    assert np.allclose(x, y)
+
+
+def test_str_tdm_method():
+    """Testing the string method"""
+    prog = tdmprogram.TDMProgram(N=1)
+    assert prog.__str__() == "<TDMProgram: concurrent modes=1, time bins=0, spatial modes=0>"
```
josh146, line 132: 😆

tests/frontend/test_tdmprogram.py
josh146: The tests look great @fab1an-q and @nquesada!

josh146 (commented):
This is looking really really nice, thanks for all the hard work (especially writing the tests and fixing the shifting!)

I think the only thing we have to do is fix up the documentation and then it is ready to merge. The following things need to be done:

* Work out what the imports look like. Do we want the user to do `sf.TDMProgram`, or `sf.tdm.TDMProgram`?

* Add the TDM package to the documentation (I can help do this)

* Flesh out the TDMProgram docstring

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,327 @@
...
+    for i, m in enumerate(mode_order):
+        idx = modes[i % len(N)]
+        if idx not in new_samples:
+            new_samples[idx] = []
+        new_samples[idx].append(all_samples[m].pop(0))
+    return new_samples
+
+
+class TDMProgram(sf.Program):
+    r"""Represents a photonic quantum circuit in the time domain encoding.
```
josh146, line 126: @fab1an-q I had a go beginning to fill out the TDMProgram docstring, could you take a look and add/edit it as needed? There are a few parts I didn't know fully how to describe :)

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,327 @@
...
+      If the ``p`` variable is not used, the gate argument is assumed to be **constant**
+      across all time-bins.
+
+    * ``copies=1`` *(int)*: the number of times to repeat the time-domain program.
+      For example, if a sequence of 10 gate arguments is provided, and ``copies==15``,
+      then there will be a total of :math:`10\times 15` time bins in the time domain
+      program. The sequence corresponding to the gate arguments is repeated ``copies``
+      number of times.
+
+    * ``shift="default"`` *(str or int)*:
```
josh146, line 217: @fab1an-q this is one option I wasn't sure how best to describe

josh146:
Sorry everyone, `black` seems to have changed its mind over what formatting it prefers - a whole heap of unrelated files have been added to this PR! Please ignore them :)

strawberryfields/tdm/tdmprogram.py
josh146: The build docs are visible here: https://strawberryfields--440.org.readthedocs.build/en/440/code/api/strawberryfields.tdm.TDMProgram.html

They will be automatically re-built after every commit

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,327 @@
...
+      If the ``p`` variable is not used, the gate argument is assumed to be **constant**
+      across all time-bins.
+
+    * ``copies=1`` *(int)*: the number of times to repeat the time-domain program.
+      For example, if a sequence of 10 gate arguments is provided, and ``copies==15``,
+      then there will be a total of :math:`10\times 15` time bins in the time domain
+      program. The sequence corresponding to the gate arguments is repeated ``copies``
+      number of times.
+
+    * ``shift="default"`` *(str or int)*: defines in which way the qumode register is shifted at the end of each time bin. If set to "default", the qumode register is shifted such that each measured qumode reappears as a fresh mode at the beginning of the subsequent time bin. This is equivalent to a measured qumode being removed from the representation (by measurement) and a new one being added (by a light source). If set to an integer value, the register will shift by a step size of this integer at the end of each time bin.
```
fab1an-q, line 217: Feel free to modify as you wish. Question: Always wondered how you set the word wrap column in these strings?

josh146: I have an extension in my text editor that automatically wraps to 100 characters :)

josh146:
> I expected ops.MeasureHomodyne(p[1]) here...?

@fab1an-q oops, yes, fixed. Good catch!

strawberryfields/tdm/tdmprogram.py
fab1an-q: One thing that we haven't explained yet is how to assign the proper qumode indices to the individual concurrent modes: 

- For each spatial mode the number of concurrent modes should equal: 1 + number of time bins in a potential delay line in this spatial mode.
- Divide the total number of concurrent modes into the individual spatial modes with the indices filling up spatial mode A before going to spatial mode B and so on. For instance, if a setup has
2 concurrent modes in spatial mode A,
1 concurrent mode in spatial mode B,
3 concurrent modes in spatial mode C,
then
`q[0]`, `q[1]` will be assigned to A,
`q[2]` will be assigned to B,
`q[3]`, `q[4]`, `q[5]` will be assigned to C
- In each spatial mode, the lowest index corresponds to the oldest mode (the one closest to the detector), whereas the highest index corresponds to the youngest mode (the one being produced at the light source).

Not sure if these instructions should be part of the docstring or rather occur somewhere else? Also, I think a simple example drawing could be really helpful for the user.

nquesada: Hi @fab1an-q . Not sure if this should go in the docstring but it is super useful for the tutorial that should come along with this.

fab1an-q: Users could benefit from some of our examples in `test_tdmprogram.py` as well. And we also have examples of how we can plot quadrature correlations and so on. This is obviously not relevant for the docstring, but maybe there is another place these examples could fit into?

nquesada: Absolutely! They are the basis for the tutorial.  I will continue working on it this week and once it is in reasonable shape will share it with everyone.

fab1an-q: Oh ok. I kind of thought the docstring already were the tutorial. Thanks @nquesada!

josh146: > One thing that we haven't explained yet is how to assign the proper qumode indices to the individual concurrent modes:

It depends how 'core' this is to understand. What we can always do is add more information to the bottom of the docstring, in a

```rest
.. UsageDetails::
```
indented section. It will look like this: https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.layers.BasicEntanglerLayers.html (and be minimized by default).

> This is obviously not relevant for the docstring, but maybe there is another place these examples could fit into?

Could these go in the tutorials?

josh146: @fab1an-q think of the docstring as a quick reference page; e.g., condensed reference information showing everything you need to know about the function.

The tutorial is a separate page, that is more 'telling a story'. That is, showing off the functionality via a set of fully worked through examples (including any theory!)

fab1an-q: Thanks @josh146! I admit I have misinterpreted the docstring as tutorial. I now understand that it's not the whole story. Yes, all my suggestions above seem to fit well into the tutorial then.

tests/frontend/test_tdmprogram.py
nquesada: Is there a better way to test this?

strawberryfields/tdm/tdmprogram.py
thisac: Oh, that was quite unclear from my side. It simply means the number of modes per rotating belt (with one measurement per belt). So, e.g. the 3 first modes might be rotating between 0, 1 and 3, while the following 2 modes might be rotating between 4 and 5, thus `N = [3, 2]` and there should be a measurement on a mode each belt. Could perhaps simply remove the word `used`?

Is this correct @fab1an-q @nquesada?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,353 @@
...
+    time domain program using the Python-embedded Blackbird syntax.
+
+    Once created, time domain programs can be executed on Strawberry Fields'
+    suite of built-in simulators in an efficient manner, or submitted
+    to be executed on compatible hardware.
+
+    Args:
+        N (int or Sequence[int]): If an integer, the number of concurrent (or 'alive')
+            modes in each time bin. Alternatively, a sequence of integers
+            may be provided, corresponding to
```
thisac, line 162: This sentence ends quite abruptly. I'm guessing that something is missing here... :thinking:

strawberryfields/tdm/tdmprogram.py
nquesada: Concurrent?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,353 @@
...
+    MeasureHomodyne(3) | (q[0])
+    Sgate(0.32, 0) | (q[0])
+    BSgate(2, 0) | (q[1], q[0])
+    MeasureHomodyne(4) | (q[1])
+    Sgate(0.32, 0) | (q[1])
+    BSgate(1, 0) | (q[0], q[1])
+    MeasureHomodyne(3) | (q[0])
+    Sgate(0.32, 0) | (q[0])
+    BSgate(2, 0) | (q[1], q[0])
+    MeasureHomodyne(4) | (q[1])
```
thisac, line 203: Do we want to update the print as well to print something that makes more sense to the user? It feels like the `TDMProgram` is trying to copy the hardware, and thus "hide" the fact that it actually shifts the modes around. This could make users a bit confused if expecting the circuit to look like the one they've defined, as well as having `all_samples` being measured on fewer/different modes than those that are printed using `prog.print()`. :thinking:

strawberryfields/tdm/tdmprogram.py
fab1an-q: Agree with @nquesada. Best to replace "used" by "concurrent".

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,353 @@
...
+        ra = list(range(sum(N[:i]), sum(N[: i + 1])))
+        all_modes.append(ra * ceil(max(N) / len(ra)))
+
+    mode_order = [i for j in zip(*all_modes) for i in j]
+    mode_order *= ceil(num_of_values / len(mode_order))
+
+    return mode_order[:num_of_values]
+
+
+def reshape_samples(all_samples, modes, N):
```
thisac, line 119: A note on the `reshape_samples` function. Currently it changes `all_samples`, leaving it empty, and returns a new `new_samples` that then replaces the, now empty, `all_samples` (inside `engine.py`). Thinking about it, this might not be the best solution. Either a deep copy could be made of `all_samples` in the beginning, but this could be computationally heavy if having many samples, or `new_samples` can replace `all_samples` instead of returning it, which might just make more sense than what's happening now. :thinking: 

What do you think @nquesada, @josh146?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,351 @@
...
+    """Reshapes the samples dict so that they have the expected correct shape.
+
+    Corrects the :attr:`~.Results.all_samples` dictionary so that the measured modes are the ones
+    defined to be measured in the circuit, instead of being spread over a larger
+    number of modes due to the mode-shifting occurring in :class:`~.TDMProgram`.
+
+    Args:
+        all_samples (dict[int, list]): the raw measured samples
+        modes (Sequence[int]): the modes that are measured in the circuit
+        N (Sequence[int]): the number of used modes per belt/spatial modes
```
nquesada, line 127:
```suggestion
        N (Sequence[int]): the number of concurrent modes per belt/spatial modes
```

strawberryfields/tdm/tdmprogram.py
nquesada: I kind of like this print...

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,351 @@
...
+    time domain program using the Python-embedded Blackbird syntax.
+
+    Once created, time domain programs can be executed on Strawberry Fields'
+    suite of built-in simulators in an efficient manner, or submitted
+    to be executed on compatible hardware.
+
+    Args:
+        N (int or Sequence[int]): If an integer, the number of concurrent (or 'alive')
+            modes in each time bin. Alternatively, a sequence of integers
+            may be provided, corresponding to
```
nquesada, line 160:
```suggestion
            may be provided, corresponding to the number of concurrent modes in the possibly multiple bands in the circuit.
```

strawberryfields/tdm/tdmprogram.py
nquesada: I'd prefer the one that minimizes computational costs given that the number of samples involved here can be massive.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,351 @@
...
+    Return:
+        spatial_modes (int): number of spatial modes
+    """
+    spatial_modes = 0
+    measurement_reg = []
+
+    for cmd in circuit:
+        if isinstance(cmd.op, ops.Measurement):
+            modes = [r.ind for r in cmd.reg]
+            measurement_reg.extend(modes)
```
nquesada, line 74:
```suggestion
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -0,0 +1,351 @@
...
+
+    Args:
+        circuit (list): a list containing commands or measurements specifying a circuit
+        N (list): list giving the number of concurrent modes per band
+
+    Return:
+        spatial_modes (int): number of spatial modes
+    """
+    spatial_modes = 0
+    measurement_reg = []
```
nquesada, line 69:
```suggestion
```

nquesada (commented):
removes `measuremen_reg` from `validate_measurements`

josh146 (approved):
Great work everyone!!

strawberryfields/tdm/tdmprogram.py
josh146: > or new_samples can replace all_samples instead of returning it

The only way I can see this working is if `reshape_samples` is either a method of the engine, _or_ it accepts and modifies an engine object directly? In either case though, I'm happy with this function for now 🙂 I agree with @nquesada - I would favour computational efficiency with some weird behaviour over more elegant behaviour with an added cost

strawberryfields/tdm/tdmprogram.py
josh146: I can see how it might be confusing though! Perhaps `prog.print()` could receive a keyword argument switching the display mode. E.g., `prog.print()` displays it without the shift, `prog.print(simulation_mode=True)` displays it with the shift?

fab1an-q: No objections against additional print options but, personally, I like the way the circuit is printed in the above example. In my opinion, it does not hide the fact that the modes are shifted but, rather, just reflects that qumodes are being indexed completely differently in a TDMProgram (indices corresponding to the "location" in the TDM setup instead of sticking to their qumode throughout the interferometer). Once the users have accepted this paradigm, then I think the above representation will seem intuitive to them. @thisac, do you think an explicit mention of the shift in the print would help the user to see what's going on? Like for instance

```
>>> prog.print()
Sgate(0.32, 0) | (q[1])
BSgate(1, 0) | (q[0], q[1])
MeasureHomodyne(3) | (q[0])
Shift | (q[0],q[1])
Sgate(0.32, 0) | (q[0])
BSgate(2, 0) | (q[1], q[0])
MeasureHomodyne(4) | (q[1])
Shift | (q[0],q[1])
Sgate(0.32, 0) | (q[1])
BSgate(1, 0) | (q[0], q[1])
MeasureHomodyne(3) | (q[0])
Shift | (q[0],q[1])
Sgate(0.32, 0) | (q[0])
...
...
```

fab1an-q: qumode conveyor belt (1)

I know it's a bit stupid, but this is how I picture a TDMProgram. The gates and the qumode indices are defined by the static upper disc. The actual light modes are written on the lower disc, rotating from one window to another. Not too relevant for this discussion, but found it cute and wanted to share. :)

thisac: > Once the users have accepted this paradigm, then I think the above representation will seem intuitive to them.

I might be missing part of the idea here, but I sort of thought that part of the reason for having a `TDMProgram` was to not make the user having to know about the inner workings of it at all. :thinking: 

> @thisac, do you think an explicit mention of the shift in the print would help the user to see what's going on? Like for instance

That could help a bit perhaps, but I don't think it's necessary. I kind of like the suggestion from @josh146; having both prints available. But I don't really consider this an important issue, so feel free leave it as it is. :slightly_smiling_face:

josh146:
@nquesada it looks fine to me?

image

Issue #443, opened 2020-08-24
IndexError during program compliation


heltluke:
#### Issue description
An `IndexError` during program compilation occurs for squeezing amplitudes of 0 and identity Interferometers.

* *Expected behavior:* (What you expect to happen)
Program compiles cleanly.

* *Actual behavior:* (What actually happens)
The `IndexError` appears.

* *Reproduces how often:* (What percentage of the time does it reproduce?)
100% of the time.

* *System information:* (post the output of `import strawberryfields as sf; sf.about()`)
```
Python version:            3.7.6
Platform info:             Linux-5.4.0-42-generic-x86_64-with-debian-buster-sid
Installation path:         /home/heltluke/anaconda3/envs/sf_latest/lib/python3.7/site-packages/strawberryfields
Strawberry Fields version: 0.15.0
Numpy version:             1.18.1
Scipy version:             1.4.1
SymPy version:             1.5.1
NetworkX version:          2.4
The Walrus version:        0.13.0
Blackbird version:         0.2.3
TensorFlow version:        None
```

#### Source code and tracebacks
 Here is an example that will reproduce the error:
```python
n_modes = 4
squeezing_amplitudes = [0] * n_modes
unitary = np.identity(n_modes)
prog = sf.Program(n_modes * 2)
with prog.context as q:
    for i in range(n_modes):
        sf.ops.S2gate(squeezing_amplitudes[i]) | (q[i], q[i + n_modes])
    for qumodes in (q[:n_modes], q[n_modes:]):
        sf.ops.Interferometer(unitary) | qumodes
    sf.ops.MeasureFock() | q
prog.compile(compiler="Xcov")
```

The error itself is
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-34-0ca84a06d898> in <module>
----> 1 prog.compile(compiler="Xcov")
~/anaconda3/envs/sf_latest/lib/python3.7/site-packages/strawberryfields/program.py in compile(self, device, compiler, **kwargs)
    554             seq = pu.optimize_circuit(seq)
    555 
--> 556         seq = compiler.compile(seq, self.register)
    557 
    558         # create the compiled Program
~/anaconda3/envs/sf_latest/lib/python3.7/site-packages/strawberryfields/compilers/xcov.py in compile(self, seq, registers)
    124 
    125             # extract the compiled symplectic matrix
--> 126             S = seq[0].op.p[0]
    127 
[...]

PR #444, opened 2020-08-24
Compiling bug

nquesada:
Addresses https://github.com/XanaduAI/strawberryfields/issues/443.
The issue arises because the `gaussian_unitary` compiler returns an empty list if the commands of your program amounts to doing nothing, but in the `xcov` compiler it is always assumed that something is returned.


strawberryfields/compilers/xcov.py
```diff
@@ -92,8 +92,18 @@ def compile(self, seq, registers):
...
+            # Since this compiler does not allow for displacements
+            # when its parameter are passed to the GaussianUnitary compiler
+            # the latter either returns a GaussianTransform + MeasureFock
+            # or just MeasureFock. This is because the GaussianUnitary checks
+            # if the symplectic matrix is just the identity it simply elides it
+
+
             # extract the compiled symplectic matrix
-            S = seq[0].op.p[0]
+            try:
```
nquesada, line 103: A much better way to do this would be to check that `seq[0].op` is a `MeasureFock` command, but I am not sure how to do this.

nquesada: `isinstance` is my friend :)

strawberryfields/compilers/xcov.py
```diff
@@ -92,8 +92,16 @@ def compile(self, seq, registers):
             # determine the modes that are acted on by the symplectic transformation
             used_modes = [x.ind for x in seq[0].reg]
 
+            # Since this compiler does not allow for displacements
+            # when its parameter are passed to the GaussianUnitary compiler
+            # the latter either returns a GaussianTransform + MeasureFock
+            # or just MeasureFock. This is because the GaussianUnitary checks
+            # if the symplectic matrix is just the identity it simply elides it
             # extract the compiled symplectic matrix
```
thisac, line 130: Maybe there should still be an empty line here, even if Black complains. :thinking: 

```suggestion
            # if the symplectic matrix is just the identity it simply elides it
            
            # extract the compiled symplectic matrix
```

strawberryfields/compilers/xcov.py
```diff
@@ -122,8 +122,17 @@ def compile(self, seq, registers):
             # determine the modes that are acted on by the symplectic transformation
             used_modes = [x.ind for x in seq[0].reg]
 
+            # Since this compiler does not allow for displacements
+            # when its parameter are passed to the GaussianUnitary compiler
```
josh146, line 126:
```suggestion
            # when its parameters are passed to the GaussianUnitary compiler,
```

strawberryfields/compilers/xcov.py
```diff
@@ -122,8 +122,17 @@ def compile(self, seq, registers):
...
+            # the latter either returns a GaussianTransform + MeasureFock
+            # or just MeasureFock. This is because the GaussianUnitary checks
+            # if the symplectic matrix is just the identity it simply elides it
```
josh146, line 129:
```suggestion
            # or just MeasureFock. This is because the GaussianUnitary checks
            # if the symplectic matrix is just the identity; if so, it simply elides it
```

josh146: Also, I didn't know of the word 'elides'! I learnt something

tests/frontend/compilers/test_xcov.py
```diff
@@ -529,3 +529,18 @@ def test_identity_program(self, tol):
         # extract the Gaussian symplectic matrix
         res = res.compile(compiler="gaussian_unitary")
         assert not res.circuit
+
+
+    def test_nothing_happens_and_nothing_crashes(self):
```
josh146, line 534: 😆

tests/frontend/compilers/test_xcov.py
```diff
@@ -529,3 +529,18 @@ def test_identity_program(self, tol):
...
+        squeezing_amplitudes = [0] * n_modes
+        unitary = np.identity(n_modes)
+        prog = sf.Program(n_modes * 2)
+        with prog.context as q:
+            for i in range(n_modes):
+                ops.S2gate(squeezing_amplitudes[i]) | (q[i], q[i + n_modes])
+            for qumodes in (q[:n_modes], q[n_modes:]):
+                ops.Interferometer(unitary) | qumodes
+            ops.MeasureFock() | q
+        prog.compile(compiler="Xcov")
```
josh146, line 546: I suppose there should be an assert statement here, just to make sure the compiled program is valid/compilation is doing what it should.

josh146 (approved):
Thanks @nquesada! Before merging this in, could you add a item to the bug fix sectionj of the changelog

PR #445, opened 2020-08-26
Changes the default compiler to be Xunitary

josh146:
**Context:** The current default compiler for devices that don't specify a preferred compiler is `Xcov`. This compiler is the most general, but can lead to permutations of the unitary.

**Description of the Change:** Changes the default compiler to `Xunitary`. This compiler only compiles the unitary, not the initial squeezers, however avoids any unintentional permutations.

**Benefits:** Avoids permutations of the qumodes

**Possible Drawbacks:** Slightly more strict; initial `S2gate` operations must match the chip layout.

### DIFF
--- a/strawberryfields/api/devicespec.py
+++ b/strawberryfields/api/devicespec.py
@@ -74,7 +74,7 @@ class DeviceSpec:
 
         # For now, use Xcov compiler by default for devices
         # if the default compiler is not specified.
-        return "Xcov"
+        return "Xunitary"
 
     @property
     def gate_parameters(self):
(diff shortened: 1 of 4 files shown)


strawberryfields/api/devicespec.py
```diff
@@ -74,7 +74,7 @@ def default_compiler(self):
 
         # For now, use Xcov compiler by default for devices
```
heltluke, line 75:
```suggestion
        # For now, use Xunitary compiler by default for devices
```

PR #447, opened 2020-08-26
[Compilation] Adding compiled_with Program attribute

antalszava:
**Context:**
When using the `RemoteEngine` with Strawberry Fields, users do not necessarily have to provide a compiler or compile their programs. Compilation happens implicitly and SF helps them with this step.

However, there are cases when this is undesirable as a pre-compiled program gets recompiled. If a program is compiled with a specific compiler (e.g. `Xstrict`) and is passed to `RemoteEngine`, then it will be recompiled using the default compiler of `RemoteEngine` (`Xcov`). This behaviour is unwanted, as compilation takes place twice and recompilation might also result in changed gates within the program.

We would like to have the following behaviour:
1. If an uncompiled program is provided to the engine, the engine implicitly compiles it
2. If the user compiles a program, we tag it as compiled (storing the compiled device and compiler) via a private attribute. The engine then simply (a) checks the program was compiled for the engine device, and (b) validates it against Xstrict. If either fail, an error is raised.

Adds the `Program._compiled_with` private attribute that is set when a `Program` is compiled or when a `DeviceSpec` is used to create a program.

**Benefits:**
Users can manually compile programs without SF recompiling it with a different compiler.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


strawberryfields/api/devicespec.py
```diff
@@ -151,6 +151,7 @@ def create_program(self, **parameters):
         # evaluate the blackbird template
         bb = bb(**parameters)
         prog = sf.io.to_program(bb)
+        prog._compiled_with = self.default_compiler
```
josh146, line 154: I would potentially have a tuple here:

```python
prog._compile_info = (self, self.default_compiler)
```

e.g., the program stores both the device spec used to compile it, as well as the compiler used.

strawberryfields/engine.py
```diff
@@ -614,10 +615,19 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
-        program = program.compile(device=device, **compile_options)
+        if program.compiled_with is None:
+            msg = f"Compiling program for device {device.target} using compiler {compiler_name}."
+            self.log.info(msg)
+
+            program = program.compile(device=device, **compile_options)
+
+        elif program.compiled_with != compiler_name:
+            msg = f"No compilation, program already compiled for {program.compiled_with}. Set recompile=true to compile with {compiler_name}."
+            self.log.info(msg)
```
josh146, line 627: By storing both the device target _and_ the compiler for compiled programs, we can make the check a bit more fine grained here. For example,

```python
if program.compile_info is None:
    # program is uncompiled
    msg = f"Compiling program for device {device.target} using compiler {compiler_name}."
    self.log.info(msg)
    program = program.compile(device=device, **compile_options)

elif program.compile_info[0] != device:
    # program was compiled for a different device
    if not recompile:
        raise ValueError(f"Cannot use program compiled for {program._compile_info[0].target} for target {self.target}")
    program = program.compile(device=device, **compile_options)

elif program.compile_info[1] != compiler_name:
    # program was compiled for the device, but using a different compiler
    if compile_options is None:
        # perhaps we put a optional validation check here? depends how much we trust the compilation
        program.compile(device=device, compiler="Xstrict")
    else:
        # user has explicitly provided compile_options to the engine, so best to apply these.
        program = program.compile(device=device, **compile_options)
```

strawberryfields/program.py
```diff
@@ -655,6 +658,18 @@ def target(self):
         """
         return self._target
 
+    @property
+    def compiled_with(self):
```
josh146, line 662: Nice!

josh146 (commented):
Looking good @antalszava!

strawberryfields/engine.py
```diff
@@ -609,15 +610,63 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
         """
         # get the specific chip to submit the program to
         compile_options = compile_options or {}
+        recompile = compile_options.pop("recompile", False)
```
josh146, line 613: should `recompile` be part of compile options? It feels like it should be a separate keyword argument (since it is not passed to `Program.compile`, as the docstring says)

strawberryfields/engine.py
```diff
@@ -609,15 +610,63 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
+            self.log.info(msg)
+            program = program.compile(device=device, **compile_options)
+
+        elif program.compile_info[1] != compiler_name:
+            # program was compiled for the device, but using a different compiler
+
+            if compile_options is None or not compile_options:
+                msg = (
+                    f"No compile options specified, compiling program for "
+                    f"device {device.target} using compiler Xstrict."
```
josh146, line 650: Perhaps we can make this a bit clearer?
```suggestion
                    f"Program previously compiled for {device.target} using {program.compile_info[1]}. "
                    f"Validating program against the Xstrict compiler."
```

strawberryfields/engine.py
```diff
@@ -609,15 +610,63 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
+                )
+                self.log.info(msg)
+
+                program = program.compile(device=device, compiler="Xstrict")
+            else:
+                # User has explicitly provided compile_options to the engine
+                msg = (
+                    f"Compiling program for device {device.target}"
+                    f"using the specified compiler options."
```
josh146, line 659:
```suggestion
                    f"using the specified compiler options: {compile_options}."
```

strawberryfields/engine.py
```diff
@@ -609,15 +610,63 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
+                )
+
+            msg = f"Recompiling program for device {device.target} using compiler {compiler_name}."
+            self.log.info(msg)
+            program = program.compile(device=device, **compile_options)
+
+        elif program.compile_info[1] != compiler_name:
+            # program was compiled for the device, but using a different compiler
+
+            if compile_options is None or not compile_options:
```
josh146, line 647: Should this take into account `recompile`?
```python
            if (compile_options is None or not compile_options) and recompile:
```

strawberryfields/engine.py
```diff
@@ -609,15 +610,63 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
+                )
+                self.log.info(msg)
+
+                program = program.compile(device=device, **compile_options)
+        else:
+            msg = (
+                f"No compilation, compiled program compatible with device "
+                f"{device.target} and compiler {compiler_name}."
+            )
+            self.log.info(msg)
```
josh146, line 669: Does this `else` correspond to 'Program was compiled for the device using the default device compiler'? Perhaps this should be combined with the previous `elif`, so that

* The program is still validated against `Xstrict`
* The program is recompiled if `recompile=True` and `compile_options` are provided (as far as I can tell, this won't happen here; a program already compiled against the default compiler won't be recompiled even if the user provides `compile_options`).

josh146 (commented):
Thanks Antal, this is looking good! I left some questions regarding the logical flow in `eng.run_async` 🙂

tests/api/test_remote_engine.py
```diff
@@ -207,3 +208,164 @@ def test_default_compiler(self, prog, monkeypatch, caplog):
...
+        prog._compile_info = (X8_spec, "dummy_compiler")
+
+        engine = sf.RemoteEngine("X8")
+        with pytest.raises(
+            ValueError, match="Cannot use program compiled for"
+        ):
+            program = engine.run_async(prog, shots=10)
+
+    def test_recompile_different_device(self, prog, monkeypatch, caplog):
+        """Tests that no error is raised if the program was compiled for
```
heltluke, line 259: What exactly does this mean? It's not immediately obvious to me how a program compiled for one device could be recompiled for another.

heltluke:
This looks awesome! Thanks @antalszava!

strawberryfields/engine.py
thisac: An addition to this: wouldn't it make sense to _always_ recompile if `recompile=True`? In that case this could be outside/above of most of this logic, and simply jump over it if `recompile=True`.

thisac (commented):
Looks great @antalszava! I just had a few comments. :rocket:

strawberryfields/engine.py
```diff
@@ -609,15 +610,63 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
-        program = program.compile(device=device, **compile_options)
+        if program.compile_info is None:
+            # program is uncompiled
+            msg = f"Compiling program for device {device.target} using compiler {compiler_name}."
+            self.log.info(msg)
+            program = program.compile(device=device, **compile_options)
+
+        elif (
+            program.compile_info[0].target != device.target
+            or program.compile_info[0]._spec != device._spec
```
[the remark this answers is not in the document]
antalszava: What if a `DeviceSpec` is explicitly created so? The following snippet creates a `DeviceSpec` by specifying `X8` as target (instead of `X8_01`).

```python
import strawberryfields as sf
import strawberryfields.ops as ops
U = random_interferometer(4)

from strawberryfields.api import DeviceSpec

import textwrap

test_spec = {
    "layout": textwrap.dedent(
        """\
        name template_4x2_X8
        version 1.0
        target X8_01 (shots=1)

        # for n spatial degrees, first n signal modes, then n idler modes, all phases zero
        S2gate({squeezing_amplitude_0}, 0.0) | [0, 4]
        S2gate({squeezing_amplitude_1}, 0.0) | [1, 5]
        S2gate({squeezing_amplitude_2}, 0.0) | [2, 6]
        S2gate({squeezing_amplitude_3}, 0.0) | [3, 7]

        # standard 4x4 interferometer for the signal modes (the lower ones in frequency)
        # even phase indices correspond to internal Mach-Zehnder interferometer phases
        # odd phase indices correspond to external Mach-Zehnder interferometer phases
        MZgate({phase_0}, {phase_1}) | [0, 1]
        MZgate({phase_2}, {phase_3}) | [2, 3]
        MZgate({phase_4}, {phase_5}) | [1, 2]
        MZgate({phase_6}, {phase_7}) | [0, 1]
        MZgate({phase_8}, {phase_9}) | [2, 3]
        MZgate({phase_10}, {phase_11}) | [1, 2]

        # duplicate the interferometer for the idler modes (the higher ones in frequency)
        MZgate({phase_0}, {phase_1}) | [4, 5]
        MZgate({phase_2}, {phase_3}) | [6, 7]
        MZgate({phase_4}, {phase_5}) | [5, 6]
        MZgate({phase_6}, {phase_7}) | [4, 5]
        MZgate({phase_8}, {phase_9}) | [6, 7]
        MZgate({phase_10}, {phase_11}) | [5, 6]

        # add final dummy phases to allow mapping any unitary to this template (these do not
        # affect the photon number measurement)
        Rgate({final_phase_0}) | [0]
        Rgate({final_phase_1}) | [1]
        Rgate({final_phase_2}) | [2]
        Rgate({final_phase_3}) | [3]
        Rgate({final_phase_4}) | [4]
        Rgate({final_phase_5}) | [5]
        Rgate({final_phase_6}) | [6]
        Rgate({final_phase_7}) | [7]

        # measurement in Fock basis
        MeasureFock() | [0, 1, 2, 3, 4, 5, 6, 7]
    """
    ),
    "modes": 8,
    "compiler": [],
    "gate_parameters": {
        "squeezing_amplitude_0": [0, 1],
        "squeezing_amplitude_1": [0, 1],
        "squeezing_amplitude_2": [0, 1],
        "squeezing_amplitude_3": [0, 1],
        "phase_0": [0, [0, 6.283185307179586]],
        "phase_1": [0, [0, 6.283185307179586]],
        "phase_2": [0, [0, 6.283185307179586]],
        "phase_3": [0, [0, 6.283185307179586]],
        "phase_4": [0, [0, 6.283185307179586]],
        "phase_5": [0, [0, 6.283185307179586]],
        "phase_6": [0, [0, 6.283185307179586]],
        "phase_7": [0, [0, 6.283185307179586]],
        "phase_8": [0, [0, 6.283185307179586]],
        "phase_9": [0, [0, 6.283185307179586]],
        "phase_10": [0, [0, 6.283185307179586]],
        "phase_11": [0, [0, 6.283185307179586]],
        "final_phase_0": [0, [0, 6.283185307179586]],
        "final_phase_1": [0, [0, 6.283185307179586]],
        "final_phase_2": [0, [0, 6.283185307179586]],
        "final_phase_3": [0, [0, 6.283185307179586]],
        "final_phase_4": [0, [0, 6.283185307179586]],
        "final_phase_5": [0, [0, 6.283185307179586]],
        "final_phase_6": [0, [0, 6.283185307179586]],
        "final_phase_7": [0, [0, 6.283185307179586]],
    },
}

def generate_X8_params(r, p):
    return {
        "squeezing_amplitude_0": r,
        "squeezing_amplitude_1": r,
        "squeezing_amplitude_2": r,
        "squeezing_amplitude_3": r,
        "phase_0": p,
        "phase_1": p,
        "phase_2": p,
        "phase_3": p,
        "phase_4": p,
        "phase_5": p,
        "phase_6": p,
        "phase_7": p,
        "phase_8": p,
        "phase_9": p,
        "phase_10": p,
        "phase_11": p,
        "final_phase_0": 1.24,
        "final_phase_1": 0.54,
        "final_phase_2": 4.12,
        "final_phase_3": 0,
        "final_phase_4": 1.24,
        "final_phase_5": 0.54,
        "final_phase_6": 4.12,
        "final_phase_7": 0,
    }

X8_spec = DeviceSpec(target="X8", connection=None, spec=test_spec) # <--- specifying X8 as target instead of X8_01

params = generate_X8_params(1, 0.3)

prog = X8_spec.create_program(**params)

eng = sf.RemoteEngine("X8_01")
device = eng.device_spec

res = eng.run(prog, shots=10)
```

strawberryfields/engine.py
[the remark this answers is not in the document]
antalszava: fair point, thanks!

antalszava: makes sense, thanks!

strawberryfields/engine.py
antalszava: Thanks for the comments! @thisac that's a fair point, one reason to keep it is as is to be able to log messages that are more specific to the case

strawberryfields/engine.py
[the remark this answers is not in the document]
thisac: :thinking: Good point @antalszava. That might actually be a case when these can differ, although I'm not sure if that's actually something that should be supported. Is there a reason _not_ to raise an error in the `DeviceSpec` class if someone attempts to create a device spec with a different target than the one defined in the layout?

I think that every time the engine is initiated, the `self.target` will always be a specific target (e.g. `X8_01`) rather than (e.g. `X8`). So, when calling the device spec API and creating the `DeviceSpec` object, there will never be a case where they differ.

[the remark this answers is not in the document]
josh146: Thanks @antalszava. The only way this seems to be possible is if a user manually creates a device spec such that the target attribute does not match the created engine... This feels like something that should be validated (_if_ it is ever likely to happen) at a much lower level?

tests/api/test_remote_engine.py
antalszava: What is meant to be tested here, is that even if the program was compiled for a device that is not compatible with the default device specification, since `recompile=True`, recompilation would take place anyways.

Reworded the docstring, let me know if it would be better to further add details here.

strawberryfields/engine.py
antalszava: Reiterated on this point @thisac, it made me rethink how these branches of the if statement could align together. Thanks again :)

tests/api/test_remote_engine.py
heltluke: Got it, thanks!

strawberryfields/engine.py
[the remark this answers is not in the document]
antalszava: > Is there a reason not to raise an error in the DeviceSpec class if someone attempts to create a device spec with a different target than the one defined in the layout?

That's a good question, not entirely sure myself either.

> This feels like something that should be validated (if it is ever likely to happen) at a much lower level?

What overall comes to mind is to remove the `target` argument from the `DeviceSpec` class and define `DeviceSpec._target` based on the `layout` that is being passed (by using the blackbird io function `loads` and extracting the target). This would ensure that `DeviceSpec._target` cannot differ from the target specified in `DeviceSpec._spec`. Could that potentially cause any major breaks? :thinking:

Any alternative validation methods would need to be included in `RemoteEngine.run`, as that is where `Program.target` and `RemoteEngine.target` can be compared.

antalszava:
Thanks for the comments @josh146 @thisac @heltluke! :slightly_smiling_face: Updated the PR and addressed them

strawberryfields/engine.py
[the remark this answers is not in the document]
josh146: >  This would ensure that DeviceSpec._target cannot differ from the target specified in DeviceSpec._spec. Could that potentially cause any major breaks? 🤔

Hmm, it could potentially cause some breaks, so let's mark this to investigate in a separate PR

strawberryfields/engine.py
```diff
@@ -603,21 +604,75 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
+                # program was compiled for a different device
+                raise ValueError(
+                    "Cannot use program compiled with "
+                    f"{program._compile_info[0].target} for target {self.target}. "
+                    'Pass the "recompile=True" keyword argument '
+                    f"to compile with {compiler_name}."
+                )
+
+            if program.compile_info[1] != compiler_name:
+                # program was compiled for the device, but using a different compiler
```
josh146, line 642: I think this is rather an important point, but `compiler_name`, from looking at the source code, is just the **default** compiler specified for the device. For example, the default compiler for `X8` is `Xunitary`. However, this doesn't mean that a program compiled using `Xcov` can't be executed on X8; this program should execute fine, even though it used a different compiler!

If a program _was_ compiled for the device, but using a different compiler, we shouldn't reject the job (via an exception), nor should we force recompilation. Instead, we should simply verify that the program is valid, by running it through `Xstrict`.

For example, assuming recompile=False and `RemoteEngine("X8")`:

* Provided program compiled for X8 using `Xcov != (compiler_name == Xunitary)`: should still run, simply validate

* Provided program compiled for X12 (a _different_ device): reject

josh146: So I think we should simply remove this if statement here, and we can simply rely on the validation happening on line 668

josh146 (changes requested):
Hey @antalszava, this is looking good - I only have concerns over one of the if statements in the logic. Once that is resolved though, I think this is close to merging!

strawberryfields/engine.py
antalszava: Thanks so much for catching this @josh146! :100: Modified it accordingly

antalszava:
Modified based on the comment from Josh and also tidied up the tests so that they reflect the logic now.

strawberryfields/engine.py
```diff
@@ -603,21 +604,64 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
...
+        program_is_compiled = program.compile_info is not None
+
+        if program_is_compiled and not recompile:
+            # error handling for program compilation:
+            # program was compiled but recompilation was not allowed by the
+            # user
+
+            if (
+                program.compile_info[0].target != device.target
+                or program.compile_info[0]._spec != device._spec
```
thisac, line 631: Did we reach a conclusion regarding this? I guess it doesn't harm to keep both checks, although I still feel like just checking `device._spec` should be enough, and then update `DeviceSpec` by either removing `target` or adding a warning/error when creating using a different target than the one in the layout.

tests/api/test_remote_engine.py
```diff
@@ -207,3 +208,135 @@ def test_default_compiler(self, prog, monkeypatch, caplog):
...
+        monkeypatch.setattr(Program, "compile", lambda *args, **kwargs: self.MockProgram())
+
+        # Leaving compile_info as None
+        assert prog.compile_info == None
+
+        engine = RemoteEngine("X8")
+        program = engine.run_async(prog, shots=10)
+
+        assert isinstance(program, self.MockProgram)
+        assert caplog.records[-1].message == "Compiling program for device X8_01 using compiler Xunitary."
```
thisac, line 256: Nice! :fire:

tests/api/test_remote_engine.py
```diff
@@ -207,3 +208,135 @@ def test_default_compiler(self, prog, monkeypatch, caplog):
...
+        compiler = "Xunitary"
+
+        caplog.set_level(logging.INFO)
+        test_device_dict = mock_device_dict.copy()
+        test_device_dict["compiler"] = compiler
+
+        monkeypatch.setattr(Connection, "create_job", lambda self, target, program, run_options: program)
+        monkeypatch.setattr(Connection, "_get_device_dict", lambda *args: test_device_dict)
+
+        # This will be popped
```
thisac, line 270: Is this comment a remnant from when `recompile` was part of `compile_options`? :thinking:

thisac (approved):
Looks great @antalszava. Just a few minor comments, but I think everything looks good. :100:

tests/api/test_remote_engine.py
antalszava: Yes, thanks for catching it!

strawberryfields/engine.py
antalszava: Thanks for raising the point again! Created https://github.com/XanaduAI/strawberryfields/issues/449 as per @josh146's suggestion to mark it.

From my perspective, if there's no reason against having both checks, I'd rather keep them for now. The reason is just super safety in exchange for minor overhead.

strawberryfields/engine.py
```diff
@@ -603,21 +604,64 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
         Keyword Args:
             shots (Optional[int]): The number of shots for which to run the job. If this
                 argument is not provided, the shots are derived from the given ``program``.
+            recompile (bool): specifies if ``program`` should be recompiled
+                using the default compilation options
```
josh146, line 608:
```suggestion
            recompile (bool): Specifies if ``program`` should be recompiled
                using ``compile_options``, or if not provided, the default compilation options.
```

strawberryfields/engine.py
```diff
@@ -603,21 +604,64 @@ def run_async(self, program: Program, *, compile_options=None, **kwargs) -> Job:
         Keyword Args:
             shots (Optional[int]): The number of shots for which to run the job. If this
                 argument is not provided, the shots are derived from the given ``program``.
+            recompile (bool): specifies if ``program`` should be recompiled
```
josh146, line 607: Just want to double check with either @heltluke or @lneuhaus: does adding SF specific `eng.run()` keyword arguments here have any implications for HW-specific keyword arguments on your end?

strawberryfields/engine.py
josh146: Yep, I don't think there is a reason for now against having both checks. It just looks a bit surprising, and indicates potential inconsistency earlier in the stack that should be double checked!

josh146 (approved):
Looks wonderful, thanks @antalszava 💪

strawberryfields/engine.py
heltluke: It is something we'll need to prepare for, as we presently do some validation of our own, but I don't think it's much of an issue.

Could we coordinate the release of SF that will include this change with an appropriate update to gbsdriver?

josh146: :+1: Perhaps what we should do on our end is 'remove' all SF-specific keywords from `kwargs` before blackbird serialization

heltluke: That could also work. :)

PR #450, opened 2020-09-13
Increment version number to v0.15.1

antalszava:
**Summary:**
* Adjusts the signature of the `RemoteEngine.run` method to be able to accept the `recompile` argument added in #447 to the `RemoteEngine.run_async` method.
* Increments the version number to `0.15.1`

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,6 +1,4 @@
-# Release 0.16.0 (development release)
-
-<h3>New features since last release</h3>
+# Release 0.15.1 (current release)
 
 <h3>Improvements</h3>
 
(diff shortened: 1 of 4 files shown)


strawberryfields/engine.py
```diff
@@ -548,7 +548,9 @@ def device_spec(self):
             self._spec = self._connection.get_device_spec(self.target)
         return self._spec
 
-    def run(self, program: Program, *, compile_options=None, **kwargs) -> Optional[Result]:
+    def run(
+        self, program: Program, *, compile_options=None, recompile=False, **kwargs
```
antalszava, line 552: Explicitly adding `recompile=False` here was needed as othwerwise `recompile` was an element in `kwargs`. In `run_async` the `recompile` parameter was separated from `kwargs`, hence `run` needs to be adjusted too in its signature and the way it calls `run_async`.

.github/CHANGELOG.md
```diff
@@ -23,13 +21,16 @@
...
 
+* Updates the `README.rst` file and hardware access links.
+  [(#448)](https://github.com/XanaduAI/strawberryfields/pull/448)
+
 <h3>Contributors</h3>
 
 This release contains contributions from (in alphabetical order):
 
-Josh Izaac, Nicolás Quesada, Antal Száva
+Theodor Isacsson, Josh Izaac, Nathan Killoran, Nicolás Quesada, Antal Száva
```
josh146, line 31: 🎉

josh146 (approved):
🍾

PR #451, opened 2020-09-15
Add qchem vibronic transition operation

soranjh:
This PR adds the function `VibronicTransition` to the `apps.qchem.vibronic` module. This function generates a custom ``sf`` operation for applying the Doktorov operator on a given state.


strawberryfields/apps/qchem/vibronic.py
```diff
@@ -48,7 +48,9 @@
 displacement vector and Duschinsky matrix, to the required GBS parameters. Additionally, this
 function computes two-mode squeezing parameters :math:`t`, from the molecule's temperature, which
 are required by the GBS algorithm to compute vibronic spectra for molecules at finite temperature.
-The :func:`sample` function then takes the computed GBS parameters and generates samples.
+The :func:`sample` function then takes the computed GBS parameters and generates samples. The
```
trbromley, line 51: Maybe we could add a few words to flow from the previous sentence? E.g.,
```rst
Alternatively, the :func:`~.VibronicTransition` function returns a custom ``sf`` operation for application of the
Doktorov operator within the conventional :class:`~.Program` interface, allowing for greater freedom on the 
choice of input state and any subsequent quantum operations.
```

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +140,55 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
     return [np.dot(s[: len(s) // 2], wp) - np.dot(s[len(s) // 2 :], w) for s in samples]
 
 
+def VibronicTransition(modes: int):
```
trbromley, line 143: Will we also be changing the `evolution()` function to caps?

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -207,15 +258,9 @@ def sample(
             for i in range(n_modes):
                 sf.ops.S2gate(t[i]) | (q[i], q[i + n_modes])
 
-        sf.ops.Interferometer(U1) | q[:n_modes]
-
-        for i in range(n_modes):
-            sf.ops.Sgate(r[i]) | q[i]
+        op = VibronicTransition(len(q))
```
trbromley, line 261: I think it's ok (since also tests pass), but I was wondering if it's bad practice to do this within the context statement.

trbromley (approved):
Thanks @soranjh, looks great!

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +140,55 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
...
+    - U2 (array): unitary matrix for the second interferometer
+    - alpha (array): displacement parameters
+
+    **Example usage:**
+
+    >>> modes = 2
+    >>> transform =  VibronicTransition(modes)
+    >>> p = sf.Program(modes)
+    >>> with p.context as q:
+    >>>     transform(U1, r, U2, alpha) | q
```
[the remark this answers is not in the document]
ixfoduap: I'm not currently sure how exactly to make this happen :thinking:

tests/apps/qchem/test_vibronic.py
```diff
@@ -232,3 +232,47 @@ def test_sample_integration(p, integration_sample_number):
...
+            [2.26762512e-06, 2.68454082e-05, 1.52080070e-04, 5.56423985e-04],
+            [2.80701727e-06, 2.52203270e-05, 1.14702166e-04, 3.51760396e-04],
+            [1.14518212e-07, 1.37859685e-06, 7.96673530e-06, 2.98259165e-05],
+        ]
+    )
+
+    p1 = [U1, r, U2, alpha, prob]
+
+    @pytest.mark.parametrize("params", [p1])
+    def test_operation(self, params):
```
ixfoduap, line 260: This is just one test and I'm not sure it's enough. Could there be some potential errors we are not catching here? :thinking:

strawberryfields/apps/qchem/vibronic.py
soranjh: Yes, there will be a PR for adding `TimeEvolution` function. We can discuss this further in the next PR.

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +140,55 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
     return [np.dot(s[: len(s) // 2], wp) - np.dot(s[len(s) // 2 :], w) for s in samples]
 
 
+def VibronicTransition(modes: int):
+    r"""Generates a custom ``sf`` operation for applying the Doktorov operator
```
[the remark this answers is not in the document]
soranjh: We initially used "custom ``sf`` operation" to emphasise that the function uses the format in `sf.utils.operation`.

strawberryfields/apps/qchem/vibronic.py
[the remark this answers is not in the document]
soranjh: Not sure either.

strawberryfields/apps/qchem/vibronic.py
soranjh: I brought `op = VibronicTransition(n_modes)` out of the context statement.

tests/apps/qchem/test_vibronic.py
soranjh: We can test the length of the input parameters. In terms of the accuracy of the operations, I think the current test is pretty strong.

strawberryfields/apps/qchem/vibronic.py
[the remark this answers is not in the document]
trbromley: Nice idea!

I think this may work:
```python
def VibronicTransition(U1, r, U2, alpha):
    
    modes = len(U1)
    
    @operation(modes)
    def op(q):

        n_modes = len(alpha)

        sf.ops.Interferometer(U1) | q

        for i in range(n_modes):
            sf.ops.Sgate(r[i]) | q[i]

        sf.ops.Interferometer(U2) | q

        for i in range(n_modes):
            sf.ops.Dgate(np.abs(alpha[i]), np.angle(alpha[i])) | q[i]
    
    return op()
```
Alternatively, we could inherit from `Operation` or maybe `Decomposition` and define the `VibronicTransition` class like other SF gates, although the above looks easier.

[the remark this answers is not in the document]
soranjh: Yes, it works. Thanks Tom.

tests/apps/qchem/test_vibronic.py
agran2018: A very naive question.
If you swap the matrices U1 and U2 entering the interferometer operations, would the resulting probabilities be different?

agran2018 (commented):
Thanks @soranjh. Just left one question on the unit test.
Another question, why not to call this function ``Doktorov``?. I noticed in the test that there are two "vibronic" prefixes `` vibronic.VibronicTransition(U1, r, U2, alpha) | q``.

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -49,6 +49,9 @@
 function computes two-mode squeezing parameters :math:`t`, from the molecule's temperature, which
 are required by the GBS algorithm to compute vibronic spectra for molecules at finite temperature.
 The :func:`sample` function then takes the computed GBS parameters and generates samples.
+The :func:`~.VibronicTransition` function returns a custom ``sf`` operation for application of the
+Doktorov operator within the conventional :class:`~.Program` interface, allowing for greater freedom
+on the choice of input state and any subsequent quantum operations.
```
trbromley, line 54:
```suggestion
The :func:`~.VibronicTransition` function is an ``sf`` operation that can be used within the
conventional :class:`~.Program` interface for application of the Doktorov operator, allowing for
greater freedom on the choice of input state and any subsequent quantum operations.
```
Sorry for the second revision :laughing:

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +141,51 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
     return [np.dot(s[: len(s) // 2], wp) - np.dot(s[len(s) // 2 :], w) for s in samples]
 
 
+def VibronicTransition(U1: np.ndarray, r: np.ndarray, U2: np.ndarray, alpha: np.ndarray):
+    r"""Generates a custom ``sf`` operation for applying the Doktorov operator
```
trbromley, line 145:
```suggestion
    r"""An ``sf`` operation for applying the Doktorov operator
```

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +141,51 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
...
+    :math:`\hat{U}_{\text{Dok}} = \hat{D}({\alpha}) \hat{R}(U_2) \hat{S}({r}) \hat{R}(U_1)` on a
+    given state.
+
+    The custom operation returned by this function can be used as part of a
+    Strawberry Fields :class:`~.Program` just like any other operation from the :mod:`~.ops` module.
+
```
trbromley, line 151:
```suggestion

```

tests/apps/qchem/test_vibronic.py
```diff
@@ -232,3 +232,45 @@ def test_sample_integration(p, integration_sample_number):
...
+
+        U1, r, U2, alpha, prob = params
+
+        eng = sf.LocalEngine(backend="gaussian")
+
+        gbs = sf.Program(len(U1))
+
+        with gbs.context as q:
+
+            vibronic.VibronicTransition(U1, r, U2, alpha) | q
```
trbromley, line 272: Any negatives of this approach? I can't think of any.

soranjh: Seems to be working fine.

tests/apps/qchem/test_vibronic.py
soranjh: Yes. The order of the operations is important.

soranjh:
> Thanks @soranjh. Just left one question on the unit test.
> Another question, why not to call this function `Doktorov`?. I noticed in the test that there are two "vibronic" prefixes ` vibronic.VibronicTransition(U1, r, U2, alpha) | q`.

We also thought of `Doktorov` and it is still an option.  The idea is to name the operators in a similar way to the existing [sf.ops](https://strawberryfields.readthedocs.io/en/stable/code/sf_ops.html) operators, e.g., MeasureFock. We will make all the operations available as `qchem.OperatorName` in the future.

tests/apps/qchem/test_vibronic.py
ixfoduap: I think we may want to add a test that checks that the operations are added in the right order, like we did before with other functions. And together with this test, it should be enough

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +141,42 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
     return [np.dot(s[: len(s) // 2], wp) - np.dot(s[len(s) // 2 :], w) for s in samples]
 
 
+def VibronicTransition(U1: np.ndarray, r: np.ndarray, U2: np.ndarray, alpha: np.ndarray):
+    r"""An ``sf`` operation for applying the Doktorov operator
```
ixfoduap, line 145: I am not a big fan of saying " an `sf` operation", it just doesn't read well. Can we just say "An operation for..." ?

strawberryfields/apps/qchem/vibronic.py
```diff
@@ -138,6 +141,42 @@ def energies(samples: list, w: np.ndarray, wp: np.ndarray) -> Union[list, float]
...
+    :math:`\hat{U}_{\text{Dok}} = \hat{D}({\alpha}) \hat{R}(U_2) \hat{S}({r}) \hat{R}(U_1)` on a
+    given state.
```
ixfoduap, line 147: Maybe worth adding a sentence explaining the connection between Doktorov and VibronicTransition, e.g. "The Doktorov operator describes the transformation of the vibrational modes of a molecule when it undergoes a transition from an initial to a final electronic state"

tests/apps/qchem/test_vibronic.py
ixfoduap: I really think this is best. In practice, I would prefer to use 
``` pythn3
from strawberryfields.qchem.vibronic import VibtronicTransition
```
but we don't need to do that in a test

ixfoduap (approved):
Should be good to go! I just left a couple of suggestions regarding the docstring

PR #455, opened 2020-09-16
Add qchem time evolution operation

soranjh:
This PR adds the function `TimeEvolution` to the `apps.qchem.dynamics` module. This function generates a custom ``sf`` operation for applying a time evolution operator on a given state.


trbromley (approved):
Approved once my suggestions are addressed.

Not sure what the pylint problem is, I don't get that locally (actually I get a `no-value-for-parameter` on line 133). Looks like codecov is just being awkward because I made coverage locally and it was fine.

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -92,51 +92,40 @@
 from strawberryfields.utils import operation
 
 
-def evolution(modes: int):
-    r"""Generates a custom ``sf`` operation for performing the transformation
-    :math:`U(t) = U_l e^{-i\hat{H}t/\hbar} U_l^\dagger` on a given state.
+def TimeEvolution(t: float, w: np.ndarray):
+    r"""An ``sf`` operation for performing the transformation
```
ixfoduap, line 96: Same comment as in #451 : I'm not a big fan of using "An `sf` operation". I prefer just "an operation"

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -92,51 +92,40 @@
 from strawberryfields.utils import operation
 
 
-def evolution(modes: int):
-    r"""Generates a custom ``sf`` operation for performing the transformation
-    :math:`U(t) = U_l e^{-i\hat{H}t/\hbar} U_l^\dagger` on a given state.
+def TimeEvolution(t: float, w: np.ndarray):
```
ixfoduap, line 95: I think I prefer to reverse the order of the arguments so that the function is
```python3
TimeEvolution(w, t)
```

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -206,7 +195,11 @@ def sample_fock(
         for i in range(modes):
             sf.ops.Fock(input_state[i]) | q[i]
 
-        op(t, Ul, w) | q
+        sf.ops.Interferometer(Ul.T) | q
+
+        TimeEvolution(t, w) | q
```
ixfoduap, line 200: Should change this too if we change order to `(w, t)`

tests/apps/qchem/test_dynamics.py
```diff
@@ -114,26 +114,47 @@
...
 
         sf.ops.Fock(2) | q[0]
 
-        op(time, unitary, frequency) | q
+        sf.ops.Interferometer(unitary.T) | q
 
-        p = eng.run(gbs).state.all_fock_probs()
+        dynamics.TimeEvolution(time, frequency) | q
+
+        sf.ops.Interferometer(unitary) | q
```
ixfoduap, line 133: I believe it is best to not include the interferometer in the test; it's better to just test the action of `TimeEvolution`

ixfoduap (approved):
Looks great! I'm suggesting a change in order for the arguments (t,w) --> (w, t) and removing the unitary from the main test

tests/apps/qchem/test_dynamics.py
soranjh: In this case the `all_fock_probs()` will simply reflect the input state. For example:
`array([[0., 0., 0., 0.],
           [0., 0., 0., 0.],
           [1., 0., 0., 0.],
           [0., 0., 0., 0.]])`
If this is fine, then I can remove the interferometers.

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -64,8 +64,8 @@
 
 This module contains functions for implementing this algorithm.
 
-- The function :func:`~.evolution` returns a custom ``sf`` operation that contains the required
-  unitary and rotation operations explained in steps 2-4 of the algorithm.
+- The function :func:`~.TimeEvolution` is an operation that contains the required
+  rotation operations explained in step 3 of the algorithm.
 
 - The function :func:`~.sample_fock` generates samples for simulating vibrational quantum dynamics
   in molecules with a Fock input state.
```
agran2018, line 71: maybe:
"from a initial state represented in the Fock space"    ?

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -92,51 +92,40 @@
...
 
     Args:
-        modes (int): number of modes
-
-    Returns:
-        an ``sf`` operation for enacting the dynamics transformation
-    Return type:
-        op
+        w (array): normal mode frequencies :math:`\omega` in units of :math:`\mbox{cm}^{-1}` that
+            compose the Hamiltonian :math:`\hat{H} = \sum_i \hbar \omega_i a_i^\dagger a_i`
```
agran2018, line 114: Maybe:

"
normal mode frequencies :math:`\omega` in units of :math:`\mbox{cm}^{-1}` entering 
 the Hamiltonian :math:`\hat{H} = \sum_i^N \hbar \omega_i a_i^\dagger a_i` for :math:`N` independent quantum Harmonic oscillators

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -193,7 +182,7 @@ def sample_fock(
         raise ValueError("Number of photons in each input state mode must be smaller than cutoff")
```
agran2018, line 182:
```suggestion
        raise ValueError("Number of photons in each mode must be smaller than cutoff")
```

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -92,51 +92,40 @@
 from strawberryfields.utils import operation
 
 
-def evolution(modes: int):
-    r"""Generates a custom ``sf`` operation for performing the transformation
-    :math:`U(t) = U_l e^{-i\hat{H}t/\hbar} U_l^\dagger` on a given state.
+def TimeEvolution(w: np.ndarray, t: float):
+    r"""An operation for performing the transformation
+    :math:`e^{-i\hat{H}t/\hbar}` on a given state.
```
agran2018, line 97: Is this statement general?. Can you use this function to simulate the the real-time dynamics for any Hamiltonian or is it constrained to the case of independent quantum harmonic oscillators?.

strawberryfields/apps/qchem/dynamics.py
```diff
@@ -268,9 +261,9 @@ def sample_tmsv(
 
     This function generates samples from a GBS device with two-mode squeezed vacuum input states.
     Given :math:`N` squeezing parameters and an :math:`N`-dimensional normal-to-local transformation
-    matrix, a GBS device with :math:`2N` modes is simulated. The evolution operator acts on only
-    the first :math:`N` modes in the device. Samples are generated by measuring the number of photons
-    in each of the :math:`2N` modes.
+    matrix, a GBS device with :math:`2N` modes is simulated. The :func:`~.TimeEvolution` operator
+    acts on only the first :math:`N` modes in the device. Samples are generated by measuring the
```
agran2018, line 265:
```suggestion
    acts only on the first :math:`N` modes in the device. Samples are generated by measuring the
```

agran2018 (commented):
Thanks @soranjh, just left few comments.

strawberryfields/apps/qchem/dynamics.py
ixfoduap: This is a great point! This applies only to evolution under the free Hamiltonian of independent quantum harmonic oscillators. So we can add something like "where :math:`H = \sum_k \hbar \omega_w (\hat{n_k} + 1/2)` is the free Hamiltonian of the system"

tests/apps/qchem/test_dynamics.py
ixfoduap: It's fine to keep as is, but you could test evolution for a different initial state for which the evolution is not trivial. The only issue is that this test can fail if there is an issue with Interferometer, even if TimeEvolution works fine

PR #456, opened 2020-09-18
Add the cancel_pending JobStatus

antalszava:
**Context:**
Adds the `cancel_pending` `JobStatus`.

### DIFF
--- a/tests/api/test_connection.py
+++ b/tests/api/test_connection.py
@@ -15,7 +15,6 @@
 Unit tests for strawberryfields.api.connection
 """
 from datetime import datetime
-from time import sleep
 import io
 import os
 
--- a/tests/api/test_connection.py
+++ b/tests/api/test_connection.py
@@ -280,30 +280,6 @@ class TestConnection:
         # A successful cancellation does not raise an exception
         connection.cancel_job("123")
 
-    def test_cancel_pending_job(self, connection, monkeypatch):
-        """Tests a job cancellation first is pending and then is successful by
-        using a mock server."""
-        # A custom `mock_return` that checks for expected arguments
-        def _mock_return(return_value):
-            def function(*args, **kwargs):
-                assert kwargs.get("json") == {"status": "cancelled"}
-                return return_value
-
-            return function
-
-        server = MockCancelerServer()
-        monkeypatch.setattr(requests, "patch", server.patch)
-
-        # A successful cancellation does not raise an exception
-        connection.cancel_job("123")
-
-        # Job cancellation is first pending
-        server.request_dict["123"] == JobStatus.CANCEL_PENDING.value
-        sleep(0.2)
-
-        # Job cancellation is successful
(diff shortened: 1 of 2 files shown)


tests/api/test_connection.py
```diff
@@ -260,6 +280,30 @@ def function(*args, **kwargs):
...
+    def test_cancel_pending_job(self, connection, monkeypatch):
+        """Tests a job cancellation first is pending and then is successful by
+        using a mock server."""
+        # A custom `mock_return` that checks for expected arguments
+        def _mock_return(return_value):
+            def function(*args, **kwargs):
+                assert kwargs.get("json") == {"status": "cancelled"}
+                return return_value
+
+            return function
```
pauljxtan, line 292: Correct me if I'm wrong, but this mock doesn't seem to be used anywhere :thinking: And it seems that the same logic is covered in `MockCancelerServer.patch`, so maybe you meant to delete this?

tests/api/test_connection.py
```diff
@@ -260,6 +280,30 @@ def function(*args, **kwargs):
...
+        # A successful cancellation does not raise an exception
+        connection.cancel_job("123")
+
+        # Job cancellation is first pending
+        server.request_dict["123"] == JobStatus.CANCEL_PENDING.value
+        sleep(0.2)
+
+        # Job cancellation is successful
+        server.request_dict["123"] == JobStatus.CANCELLED.value
+
```
pauljxtan, line 306: Also, I wonder if we're really testing anything meaningful in this particular case: since we manually set the value, then check the value directly, it feels like we are validating the behaviour of `time.sleep` more than anything else :smiley: 

I think the transition from `cancel_pending` to `cancelled` might be better tested at the integration level between SF and the platform API, which would help to simplify things here a bit. But let me know if I'm missing something important.

tests/api/test_connection.py
thisac, line 292: This `_mock_return` doesn't seem to be used anywhere. :thinking:

tests/api/test_connection.py
```diff
@@ -260,6 +280,30 @@ def function(*args, **kwargs):
...
+        # A successful cancellation does not raise an exception
+        connection.cancel_job("123")
+
+        # Job cancellation is first pending
+        server.request_dict["123"] == JobStatus.CANCEL_PENDING.value
+        sleep(0.2)
+
+        # Job cancellation is successful
+        server.request_dict["123"] == JobStatus.CANCELLED.value
```
thisac, line 305: Shouldn't these be `assert`s?

pauljxtan (commented):
Looks great @antalszava - just have some small comments.

antalszava:
Thanks for the comments @pauljxtan @thisac! :100: Agreed, might simply just be best to add it to the list and have it that way :)

pauljxtan (approved):
:100:

PR #460, opened 2020-09-24
Add qchem function to read gamess output

soranjh:
This PR adds the function `read_gamess` to the `qchem` module to extracts the atomic coordinates, atomic masses, vibrational frequencies, and normal modes of a molecule from the output file of a vibrational frequency calculation performed with the GAMESS quantum chemistry package.


strawberryfields/apps/qchem/utils.py
```diff
@@ -96,3 +96,59 @@ def duschinsky(
...
+def read_gamess(file) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    r"""Reads molecular data from the output file generated by the GAMESS quantum chemistry package.
+
+    This function extracts the atomic coordinates, atomic masses, vibrational frequencies,
+    and normal modes of a molecule from the output file of a vibrational frequency calculation
+    performed with the GAMESS quantum chemistry package.
+
+    **Example usage:**
+
+    >>> r, m, w, l = read_gamess('../BH_frq.out')
```
ixfoduap, line 110: Maybe nice to print each of these four with a short comment explaining what they stand for? Would also be good to give the `.out` file a clearer name. For example
```
>>> r, m, w, l = read_gamess('[molecule_name]_data.out'
>>> r  # atomic coordinates
array([0.000000, 0.000000, 0.000000, 1.253604, 0.000000, 0.000000]
```

strawberryfields/apps/qchem/utils.py
```diff
@@ -96,3 +96,59 @@ def duschinsky(
...
+    >>> r
+    array([0.000000, 0.000000, 0.000000, 1.253604, 0.000000, 0.000000]
+
+    Args:
+        file: name and the path of the GAMESS output file
```
ixfoduap, line 115: Maybe just say "path to the GAMESS output file" since the path includes the name?

strawberryfields/apps/qchem/utils.py
```diff
@@ -96,3 +96,59 @@ def duschinsky(
...
+    with open(file, "r") as f:
+
+        r = []
+        m = []
+        w = []
+        l = []
+
+        for line in f:
+
+            if "INPUT CARD> $data" in line or "INPUT CARD> $DATA" in line:
```
ixfoduap, line 130: So GAMESS sometimes uses uppercase and other times it does not? :thinking:

strawberryfields/apps/qchem/utils.py
```diff
@@ -96,3 +96,59 @@ def duschinsky(
...
+    array([0.000000, 0.000000, 0.000000, 1.253604, 0.000000, 0.000000]
+
+    Args:
+        file: name and the path of the GAMESS output file
+
+    Returns:
+        tuple[array, array, array, array]: atomic coordinates, atomic masses, normal mode
+        frequencies, normal modes
+    """
+    with open(file, "r") as f:
```
ixfoduap, line 121: One general comment: How confident are we that GAMESS files always "look the same"? This function is relying heavily on these files having a very specific, reliable structure. Is this a good assumption?

ixfoduap: For example, I wondered if we should raise any errors. But raising errors would mean there is a problem with the file, right? 
Maybe we should at least test that the input is a GAMESS file

ixfoduap (commented):
Such a useful function! Just left a comment regarding checking that the file is correct

strawberryfields/apps/qchem/utils.py
soranjh: This part is actually defined by the user. Some users prefer lowercase and some use uppercase.

strawberryfields/apps/qchem/utils.py
soranjh: To my knowledge the format is pretty standard for vibrational frequency calculations. I tested the function on two outputs generated independently by myself and Alain for different molecules. I will add one or few conditions for raising errors.

strawberryfields/apps/qchem/utils.py
agran2018: You may add a keyword argument so the user can indicate the path to his/her file like we do in the function ``read_structure`` of PL-QChem. https://github.com/PennyLaneAI/pennylane/blob/936a5cf626b2f933dbb6a766bb1baa78a739cfcc/qchem/pennylane_qchem/qchem/structure.py#L149

strawberryfields/apps/qchem/utils.py
agran2018: The keywords used in the input files of GAMESS are not case sensitive.

strawberryfields/apps/qchem/utils.py
agran2018: @soranjh, probably to explore: GAMESS outputs a file with extension ".dat" which I think it is precisely intended for post-processing purposes. That's why the format of this file is purely numeric. I think it contains the information you are looking for (after the molecular orbitals printing group). Perhaps, it might make the function simpler(not 100% sure). Just attached a sample of this file below. (just hacked the extension of the file to be ".log" as github does not support ".dat"
[dft_hessian_gs_b3lyp-ccpVTZ.log](https://github.com/XanaduAI/strawberryfields/files/5285160/dft_hessian_gs_b3lyp-ccpVTZ.log)

strawberryfields/apps/qchem/utils.py
```diff
@@ -96,3 +96,59 @@ def duschinsky(
...
+
+        for line in f:
+
+            if "INPUT CARD> $data" in line or "INPUT CARD> $DATA" in line:
+                line = [next(f) for _ in range(3)][-1]
+                while "end" not in line and "END" not in line:
+                    r.append(np.array(line.rstrip().split()[-3:], float))
+                    line = next(f).rstrip()
+
+            if "ATOMIC WEIGHTS" in line:
```
agran2018, line 136: Probably it is convenient to raise an error if the output group is not found and print an error message. For example, what happens if the output file does not correspond to a frequency calculation?.

agran2018 (commented):
Thank you @soranjh. This is a useful functionality.
I have left some comments. The most important one is whether is more convenient or not using the data file ".dat" punched by GAMESS which is more intended for post-processing.

trbromley (commented):
Thanks @soranjh! This looks like a useful function, but I'm worried that it's susceptible to be error-prone. Are we sure that:

- There is no other python package for interfacing with GAMESS that offers a similar functionality?
- Is converting GAMESS files a crucial part of how a user would interact with the SF qchem package? I was thinking this was a little outside scope for SF, but I am not knowledgeable on the chemistry side.

agran2018:
> Thanks @soranjh! This looks like a useful function, but I'm worried that it's susceptible to be error-prone. Are we sure that:
> 
> * There is no other python package for interfacing with GAMESS that offers a similar functionality?

Now that Tom mention this. The ezSpectrum developed by the group of Ana Krylov comes with the python utility ``make_xml.py`` that extracts the information coming coming out of electronic structure calculations for the ground and all excited (electronic) states to be included in the vibronic spectrum calculation (See here http://iopenshell.usc.edu/downloads/manual.pdf). One possibility is to post-process this file. One advantage of this approach is that ``make_xml.py`` supports the output of several quantum chemistry packages. In this sense we are not forcing the user to perform his/her calculation with GAMESS.

> * Is converting GAMESS files a crucial part of how a user would interact with the SF qchem package? I was thinking this was a little outside scope for SF, but I am not knowledgeable on the chemistry side.
@trbromley, if you meant processing the GAMESS file, yes, getting all these numbers from electronic structure calculations is not straightforward. For example, in principle, you have to gather all the information for all excited states.

strawberryfields/apps/qchem/utils.py
```diff
@@ -96,3 +96,59 @@ def duschinsky(
...
+
+        r = []
+        m = []
+        w = []
+        l = []
+
+        for line in f:
+
+            if "INPUT CARD> $data" in line or "INPUT CARD> $DATA" in line:
+                line = [next(f) for _ in range(3)][-1]
```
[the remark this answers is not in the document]
soranjh: Yes, there is a line dedicated to title.

agran2018 (approved):

ixfoduap (approved):

PR #471, opened 2020-10-19
Fix reduced dm

### LINKED ISSUE #469: reduced_dm not correct for subsystems of more than 1 mode
#### Issue description

* *Expected behavior:* Calling `reduced_dm` with a list of modes with length longer than one should give an output with the correct shape (e.g., for a system with three modes, using `modes=[0,1]` should return a density matrix with shape `(D, D, D, D)`, where `D` is the cutoff dimension)

* *Actual behavior:* The output tensors do not have the correct shape, they are missing required axes, or become scalars. I have diagnosed the bug to the following lines in `tfbackend/states.py`:
```
  for m in modes:
            reduced = reduced_density_matrix(reduced, m, False, batched=self.batched)
```
The function `reduced_density_matrix`, according to its docstring, traces out all modes except `m`. Thus, the loop above always produces a density matrix of shape `(D, D)` after its first iteration. This is not how one produces reduced density matrices when more than two subsystems are kept

* *Reproduces how often:* occurs when using the TF backend and more than 2 modes, but not the Fock backend

* *System information:* 
```
sf.about()
Python 3.8.2 (default, May  7 2020, 20:00:49) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.18.1 -- An enhanced Interactive Python. Type '?' for help.
PyDev console: using IPython 7.18.1
Python version:            3.8.2
Platform info:             Linux-5.4.0-48-generic-x86_64-with-glibc2.10
Installation path:         /home/nathan/dev/strawberryfields/strawberryfields
Strawberry Fields version: 0.15.0-dev0
Numpy version:             1.18.4
Scipy version:             1.5.2
SymPy version:             1.6.2
NetworkX version:          2.5
The Walrus version:        0.13.0
Blackbird version:         0.2.4
TensorFlow version:        2.3.1
```

Minimal (non)working example: 
```
import strawberryfields as sf

modes = 3
fock_cutoff = 3
prog = sf.Program(modes)
eng = sf.Engine("tf", backend_options={"cutoff_dim": fock_cutoff})
result = eng.run(prog)
state = result.state
d01 = state.reduced_dm([0,1])
d02 = state.reduced_dm([0,2])
print(d01.shape)
print(d02.shape)
>>> ()  # should be (3,3,3,3)
>>> ()  # should be (3,3,3,3)
```

thisac:
**Context:**
`reduced_dm` in the tf backend doesn't work with several modes, since `sf.backends.tf.reduced_density_matrix` doesn't either.

`reduced_density_matrix` is updated to support several modes and `reduce_dm` is patched to allow for this as well.

**Benefits:**
The reduced density matrix is now calculated correctly when tracing out over several modes (in the tf backend).

**Possible Drawbacks:**
None

**Related GitHub Issues:**
Fixes #469


strawberryfields/backends/tfbackend/ops.py
```diff
@@ -1077,11 +1077,14 @@ def partial_trace(system, mode, state_is_pure, batched=False):
     return reduced_state
 
 
-def reduced_density_matrix(system, mode, state_is_pure, batched=False):
+def reduced_density_matrix(system, modes, state_is_pure, batched=False):
     """
-    Trace out all subsystems except 'mode' from 'system'.
+    Trace out all subsystems except the ones in the 'modes' from 'system'. 'modes' can be either an int or a list.
```
josh146, line 1082:
```suggestion
    Trace out all subsystems except those specified in ``modes`` from ``system``. ``modes`` can be either an int or a list.
```

strawberryfields/backends/tfbackend/states.py
```diff
@@ -327,8 +327,7 @@ def reduced_dm(self, modes, **kwargs):
             )
 
         reduced = self.dm(**kwargs)
-        for m in modes:
-            reduced = reduced_density_matrix(reduced, m, False, batched=self.batched)
+        reduced = reduced_density_matrix(reduced, modes, False, batched=self.batched)
```
josh146, line 330: :+1:

josh146 (approved):
Looks good to me @thisac! Does this bring the behaviour of `state.reduced_dm` consistent between the Fock and TF backends?

By the way, don't forget to update the changelog to include this bug fix.

thisac:
Thanks @josh146!

> Does this bring the behaviour of state.reduced_dm consistent between the Fock and TF backends?

Yes, it should work the same now, since this issue wasn't present in the Fock backend.

.github/CHANGELOG.md
```diff
@@ -32,6 +32,9 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where `reduced_dm` gives the wrong output when passing it several modes.
```
co9olguy, line 35: be good to clarify where this function lives

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,26 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
...
+        when tracing out more than one mode."""
+        zero_photon_state = np.zeros([cutoff])
+        zero_photon_state[0] = 1.
+        one_photon_state = np.zeros([cutoff])
+        one_photon_state[1] = 1.
+
+        # create a single-photon state in the second mode
+        state = np.outer(zero_photon_state, one_photon_state)
+        state = np.multiply.outer(state, zero_photon_state)
+        state_dm = mixed(state)
```
co9olguy, line 67: Since this is really just taking an outer product, might be better to not rely on it, and make the test more "unit"al

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,26 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
...
+        state_dm = tf.constant(state_dm)
+
+        reduced_dm = reduced_density_matrix(state_dm, modes, state_is_pure=False)
+
+        alpha_str = np.array(["ab", "cd", "ef"])
+        expected = np.einsum(f"abcdef->{''.join(alpha_str[modes])}", state_dm)
```
co9olguy, line 73: blew my mind how you put a string inside a string here :grinning:

co9olguy: Would be good to use `tf.einsum` here since these are tensorflow objects

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,26 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
         expected = tf.constant(expected)
 
         assert np.allclose(reduced_dm, expected, atol=tol, rtol=0)
+
+    @pytest.mark.parametrize("modes", [[0], [1], [2], [0, 1], [0, 2], [1, 2], [0, 1, 2]])
+    def test_reduced_density_matrix_multiple_modes(self, setup_backend, cutoff, modes, tol):
+        """Test the reduced_density_matrix returns the correct reduced density matrices
+        when tracing out more than one mode."""
```
co9olguy, line 58:
```suggestion
        """Test the reduced_density_matrix returns the correct reduced density matrices."""
```

co9olguy: since you also test the case where you trace out one or none

tests/backend/test_tf_ops.py
co9olguy: Is this `einsum` correct? For a reduced density matrix on, say, mode 0, I would expect the formula to be
```
abccdd->ab
```
i.e., we sum the diagonals explicitly. My understanding of einsum is that when you use `->` any index missing from the left of `->` is summed over, but your example would sum over both the row and column indices of the traced out subsystems independently, not the diagonal, as is needed for partial trace. Please can you confirm? (likely this specific example masks it because there is only one nonzero entry)

co9olguy (changes requested):
Thanks @thisac, I think your test case usage of `einsum` may be incorrect (only passing by coincidence)?

tests/backend/test_tf_ops.py
thisac: I rewrote the test completely, causing `tf.einsum` not to work since it doesn't support the same axis appearing several times in the einstring (unless it's a trace).

thisac: Oh, just saw your second reply. Hadn't reloaded the page. :laughing: 

You're completely right @co9olguy. It was a quite bad mistake there. Thanks for spotting it! :1st_place_medal: I rewrote the test, creating instead GHZ state and using that for the calculations. I think I got a bit confused with what was happening earlier as well.

co9olguy: That's a good point. They still don't support that yet? If not, we'll have to indeed use `np.einsum` then

tests/backend/test_tf_ops.py
```diff
@@ -16,11 +16,11 @@
 import pytest
 
 import numpy as np
-tf = pytest.importorskip("tensorflow", minversion="2.0")
+import tensorflow as tf
 
 from strawberryfields.backends.tfbackend.ops import reduced_density_matrix
 
-@pytest.mark.backends("tf")
+@pytest.mark.backends("tf", "fock")
```
co9olguy, line 23: looks like a merge overwrote changes that are already set in the base branch

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,25 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
         expected = tf.constant(expected)
 
         assert np.allclose(reduced_dm, expected, atol=tol, rtol=0)
+
+    @pytest.mark.parametrize("modes", [[0], [1], [2], [0, 1], [0, 2], [1, 2], [0, 1, 2]])
+    def test_reduced_density_matrix_multiple_modes(self, setup_backend, cutoff, modes, tol):
+        """Test the reduced_density_matrix returns the correct reduced density matrices."""
```
co9olguy, line 57:
```suggestion
        """Test that reduced_density_matrix returns the correct reduced density matrices."""
```

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,25 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
...
+        state = np.zeros([cutoff, cutoff, cutoff])
+        state[0, 0, 0] = 1 / np.sqrt(2)
+        state[1, 1, 1] = 1 / np.sqrt(2)
+        state = np.multiply.outer(state, state.conj())
+
+        # transpose the state due to different conventions
```
co9olguy, line 63: What are the different conventions? We should only need to use SF convention here :thinking:

tests/backend/test_tf_ops.py
thisac, line 23:
```suggestion
@pytest.mark.backends("tf")
```

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,25 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
...
+        state = state.transpose([0, 3, 1, 4, 2, 5])
+        state = tf.constant(state)
+
+        reduced_dm = reduced_density_matrix(state, modes, state_is_pure=False)
+
+        alpha_str = np.array(["ab", "cd", "ef"])
+        betha_str = np.array(["aa", "bb", "cc"])
```
co9olguy, line 70: `beta`?

tests/backend/test_tf_ops.py
```diff
@@ -16,7 +16,7 @@
 import pytest
 
 import numpy as np
-tf = pytest.importorskip("tensorflow", minversion="2.0")
+import tensorflow as tf
```
co9olguy, line 19: pls revert this merge, it is overwritting a step @josh146 asked for in code review

co9olguy:
Now that I think of it, the fact that tf didn't support full `einsum` semantics was exactly the reason why the reduced density matrix and partial trace functions looked the way they did

tests/backend/test_tf_ops.py
```diff
@@ -51,3 +51,25 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
...
+        # transpose the state due to different conventions
+        state = state.transpose([0, 3, 1, 4, 2, 5])
+        state = tf.constant(state)
+
+        reduced_dm = reduced_density_matrix(state, modes, state_is_pure=False)
+
+        alpha_str = np.array(["ab", "cd", "ef"])
+        betha_str = np.array(["aa", "bb", "cc"])
+        betha_str[modes] = alpha_str[modes]
+        expected = np.einsum(f"{''.join(betha_str)}->{''.join(alpha_str[modes])}", state)
```
co9olguy, line 72: Starting to get a bit hard to parse. Perhaps instead of trying to get too fancy, we can just parametrize in the decorator over the einsum strings instead?

tests/backend/test_tf_ops.py
thisac: Oh, my bad. Is it OK to just change this line back to `tf = pytest.importorskip("tensorflow", minversion="2.0")`, or are you suggesting somehow reverting the merge? Not sure exactly how that's done.

tests/backend/test_tf_ops.py
thisac: Yes, should be beta. :laughing:

tests/backend/test_tf_ops.py
thisac: True, that would work as well. Would you prefer doing that rather than splitting up the einsum and adding a few extra lines and comments?

tests/backend/test_tf_ops.py
thisac: I think it's the way [`np.multiply.outer`](https://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.ufunc.outer.html) works, by ordering the indices, is different from what's done in SF. I could be wrong though, but I remember this being the case in TW at least. :thinking:

tests/backend/test_tf_ops.py
co9olguy: all I'm suggesting is undoing this change. How you achieve it is up to you :laughing:

tests/backend/test_tf_ops.py
co9olguy: ahh yes, got it thanks :smiley:

tests/backend/test_tf_ops.py
co9olguy: yeah I think that might make it easier for future readers of this code :smiley:

thisac:
Thanks @co9olguy for the comments (and sorry for the bad merge). I looked into the conventions again, and [SF](https://strawberryfields.readthedocs.io/en/stable/code/api/strawberryfields.ops.DensityMatrix.html) differs from [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.outer.html#numpy.ufunc.outer), thus the transpose. I hope the test is easier to comprehend now.

tests/backend/test_tf_ops.py
```diff
@@ -52,8 +52,16 @@ def test_reduced_density_matrix_fock_states(self, num_modes, setup_backend, cuto
...
-    def test_reduced_density_matrix_multiple_modes(self, setup_backend, cutoff, modes, tol):
+    @pytest.mark.parametrize("modes,einstr", [
+        ([0], "abccee->ab"),
+        ([1], "aacdee->cd"),
+        ([2], "aaccef->ef"),
+        ([0, 1], "abcdee->abcd"),
+        ([0, 2], "abccef->abef"),
+        ([1, 2], "aacdef->cdef"),
+        ([0, 1, 2], "abcdef->abcdef")
+    ])
```
co9olguy, line 63: Yeah, this is definitely easier to parse!

co9olguy (approved):

PR #475, opened 2020-10-22
Rename io module to io_utils

antalszava:
**Context:**
The `strawberryfields/io.py` file implements loading and saving functions.

The name of the module can at times clash e.g., with the standard `io` package and result in unexpected behaviour.

Running `pip` in the `strawberryfields` subdirectory:
```
Fatal Python error: init_sys_streams: can't initialize sys standard streams
  File "/strawberryfields/strawberryfields/io.py", line 21, in <module>
ModuleNotFoundError: No module named 'blackbird'
Aborted (core dumped)
```

Renames `strawberryfields/io.py` to `strawberryfields/io_utils.py`

**Benefits:**
Command-line tools like `pip` and `vim` can be used from the `strawberryfields` subdirectory and no unexpected behaviour would further arise due to the name clash.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


josh146:
Hey @antalszava, just copying over my comment from https://github.com/PennyLaneAI/pennylane/pull/866#issuecomment-714912421:

> Hey @antalszava, I think this is expected behaviour. Having a module named `io.py` or `collections.py` would be troublesome if they were top-level modules, however these two modules are namespaced within the `pennylane` package, so should never be treated as top-level modules and should never clash.
>
> By running the Python interpreter from _inside_ the PennyLane package, what seems to be happening is that your Python interpreter is adding the current path to `sys.path`, and is thus attempting to load the modules as top-level. The solution here is not to run Python from inside the package directory, thus avoiding issues with the Python path.
>
> In fact, namespacing is a common solution so as to avoid issues with clashing module names. For example, the following packages also include a (namespaced) `collections` module:
>
> * [`matplotlib.collections`](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/collections.py)
> * [`sqlalchemy.orm.collections`](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/collections.py)
>
> *Side note:* I am not able to reproduce this issue.
>
> ```python
> >>> !pwd
> /home/josh/xanadu/pennylane/pennylane
> >>> import collections
> >>> collections.__file__
> '/home/josh/miniconda3/envs/37/lib/python3.7/collections/__init__.py'
> ```

antalszava:
Thank you! :) Closing this one too

PR #476, opened 2020-10-23
TDM hardware connection

thisac:
**Context:**
This PR contains changes that will allow
* TDM device specs to be retrieved from the API
* `TDMPrograms` to be submitted to the hardware
* `TDMPrograms` to be converted into Blackbird programs

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/tdm/tdmprogram.py
```diff
@@ -360,6 +363,14 @@ def apply_op(self, cmd, q, t):
 
         self.append(cmd.op.__class__(*params), get_modes(cmd, q))
 
+    def assert_number_of_modes(self, device):
+        """Check that the number of modes in the program is valid for the given device."""
+        if device.timebins > device.modes["max"]["temporal"]:
```
antalszava, line 368: Might become `device.modes["temporal"]["max"]`

requirements.txt
```diff
@@ -5,7 +5,7 @@ sympy>=1.5
 tensorflow>=2.0
 tensorboard>=2.0
 networkx>=2.0
-quantum-blackbird==0.2.3
+git+git://github.com/XanaduAI/blackbird@master#egg=blackbird
```
thisac, line 8: This is causing the readthedocs to fail, and since a new release of Blackbird is around the corner I'll probably remove this and pin it to the latest release instead.

tests/frontend/test_io.py
```diff
@@ -528,6 +551,52 @@ def test_gate_multimode(self):
...
+        version 1.0
+        type tdm (temporal_modes=2, copies=3)
+
+        int array p0 =
+            1, 2
+        int array p1 =
+            3, 4
+
+        Sgate(0.7, 0) | 1
+        BSgate(p0, 0.0) | [0, 1]
```
antalszava, line 568: As a sanity check, would be cool to have `Rgate` too here imho

strawberryfields/io.py
```diff
@@ -81,7 +83,29 @@ def to_blackbird(prog, version="1.0"):
...
+                    if str(p) == str(v):
+                        op["kwargs"][k] = p.name
+
         bb._operations.append(op)
+    # add the specific "tdm" metadata to the Blackbird program
+    if prog.type == "tdm":
+        bb._type["name"] = "tdm"
+        bb._type["options"].update(
+            {
+                "temporal_modes": prog.timebins,
```
antalszava, line 102: Might be worth ironing out this one across the components (either going with `temporal modes` in SF too or having `timebins` elsewhere). Any thoughts?

tests/frontend/test_io.py
```diff
@@ -335,6 +337,27 @@ def test_free_par_str(self):
         assert bb.operations[0] == {"op": "Sgate", "modes": [0], "args": ['{r}', 0.0], "kwargs": {}}
         assert bb.operations[1] == {"op": "Zgate", "modes": [1], "args": ['3*log(-{alpha})'], "kwargs": {}}
 
+    def test_tdm_program(self):
+        prog = TDMProgram(2)
+
+        with prog.context([1, 2], [3, 4], copies=3) as (p, q):
+            ops.Sgate(0.7, 0) | q[1]
+            ops.BSgate(p[0]) | (q[0], q[1])
```
antalszava, line 345: Similar thought on having `Rgate` here

tests/frontend/test_tdmprogram.py
```diff
@@ -18,11 +18,13 @@
 from strawberryfields import ops
 from strawberryfields.tdm import tdmprogram
 
+pytestmark = pytest.mark.frontend
+
 # make test deterministic
 np.random.seed(42)
 
 
-def singleloop(r, alpha, phi, theta, copies, shift="default", hbar=2):
```
antalszava, line 25: Not entirely sure about the reason here (recall hearing about the change, but can't recall the exact reason :thinking: )

tests/frontend/test_tdmprogram.py
```diff
@@ -416,7 +402,53 @@ def test_two_dimensional_cluster_tokyo():
...
+        (200, 2, 1, "contains 200 temporal modes"),
+        (50, 42, 1, "contains 42 concurrent modes"),
+        (50, 2, 2, "contains 2 spatial modes"),
+    ]
+)
+def test_assert_number_of_modes(temporal_modes, concurrent_modes, spatial_modes, match):
+    """Test that an exception is raised if the compiler
+    is called with a device spec with an incorrect number of modes"""
+
+    class DummyCircuit(sf.compilers.compiler.Compiler):
```
antalszava, line 421:
```suggestion
    class DummyCompiler(sf.compilers.compiler.Compiler):
```

tests/frontend/test_tdmprogram.py
```diff
@@ -416,7 +402,53 @@ def test_two_dimensional_cluster_tokyo():
...
+        """A circuit with 2 modes"""
```
antalszava, line 422:
```suggestion
        """A compiler with 2 gates"""
```

tests/frontend/test_tdmprogram.py
```diff
@@ -416,7 +402,53 @@ def test_two_dimensional_cluster_tokyo():
...
+
+    params = np.ones(temporal_modes)
+    with prog.context(params, params, copies=3) as (p, q):
+        ops.Sgate(0.7, 0) | q[1]
+        ops.BSgate(p[0]) | (q[0], q[1])
+        for i in range(spatial_modes):
+            ops.MeasureHomodyne(p[1]) | q[i]
+
+    with pytest.raises(sf.program_utils.CircuitError, match=match):
+        new_prog = prog.compile(device=spec, compiler=DummyCircuit())
```
antalszava, line 454:
```suggestion
        new_prog = prog.compile(device=spec, compiler=DummyCompiler())
```

tests/frontend/test_tdmprogram.py
```diff
@@ -416,7 +402,53 @@ def test_two_dimensional_cluster_tokyo():
...
+@pytest.mark.parametrize(
+    "temporal_modes,concurrent_modes,spatial_modes,match", [
+        (200, 2, 1, "contains 200 temporal modes"),
+        (50, 42, 1, "contains 42 concurrent modes"),
+        (50, 2, 2, "contains 2 spatial modes"),
+    ]
+)
+def test_assert_number_of_modes(temporal_modes, concurrent_modes, spatial_modes, match):
+    """Test that an exception is raised if the compiler
+    is called with a device spec with an incorrect number of modes"""
```
antalszava, line 419: Awesome test!

antalszava (commented):
Hi @thisac, this looks really great! :confetti_ball: Left a couple of comments, questions. Everything seems nice overall, would just revisit for another review round to double-check your thoughts

strawberryfields/tdm/tdmprogram.py
```diff
@@ -350,6 +382,10 @@ def construct_circuit(self):
             elif isinstance(self.shift, int):
                 q = shift_by(q, self.shift)  # shift at end of each time bin
 
+        self.unrolled_circuit = self.circuit.copy()
```
antalszava, line 385: Took some time for me to understand how the `copy` calls and setting the refs worked between `self.circuit`, `self.rolled_circuit` and `self.unrolled_circuit`. Maybe worth placing a couple of comments? (Similar to this suggestion or something more descriptive)

Also: could there be a case where `self.circuit` does not equal neither `self.rolled_circuit` nor `self.unrolled_circuit`?
```suggestion
        # Unrolling the circuit for the first time: storing a copy of the unrolled circuit
        self.unrolled_circuit = self.circuit.copy()
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -291,15 +312,26 @@ def __exit__(self, ex_type, ex_value, ex_tb):
 
         if ex_type is None:
             self.spatial_modes = validate_measurements(self.circuit, self.N)
-            self.construct_circuit()
+            self.timebins = len(self.tdm_params[0])
+            self.total_timebins = self.timebins * self.copies
+            self.rolled_circuit = self.circuit.copy()
```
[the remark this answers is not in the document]
josh146: Yep! That's right

strawberryfields/tdm/tdmprogram.py
```diff
@@ -291,15 +312,26 @@ def __exit__(self, ex_type, ex_value, ex_tb):
...
+        self.circuit = self.rolled_circuit
+        return self
+
+    def unroll(self):
         """Construct program with the register shift"""
-        cmds = self.circuit.copy()
-        self.circuit = []
+        if self.unrolled_circuit is not None:
+            self.circuit = self.unrolled_circuit
+            return self
```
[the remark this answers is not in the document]
josh146: > We modify the `TDMProgram`

Is this a possible step? The assumption with the current workflow is that it is not possible to modify the program (at least, in a user-facing manner) after creation. If it turns out it is possible to modify the program after creation, a simply solution is to just `lock()` the program when it exits the `with` context.

strawberryfields/tdm/tdmprogram.py
josh146: Good suggestion @antalszava.

> Also: could there be a case where self.circuit does not equal neither self.rolled_circuit nor self.unrolled_circuit?

There shouldn't be, no. If there is, then there is a bug in the logic.

nquesada:
Thanks for the suggestions @antalszava .

strawberryfields/api/devicespec.py
```diff
@@ -97,17 +98,27 @@ def gate_parameters(self):
 
         return gate_parameters
 
+    def fill_template(self, program):
+        """Fill template with parameter values from a program"""
+        if program.type == "tdm" and program.timebins:
+            self._spec["layout"] = self._spec["layout"].format(
+                target=program.target, tm=program.timebins
+            )
```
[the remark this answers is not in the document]
thisac: Good point Antal! 

This should probably be changed to use either `device.target` or `self.target` (i.e. `RemoteEngine.target`). These two should always be the same, since `device.target` is created by creating a `DeviceSpec` with `target=self.target` in `RemoteEngine`. `program.target` should be the same _after_ compilation, but will be `None` before compilation (I think), meaning that the above code would only work for already compiled programs.

I'll update this!

strawberryfields/io.py
thisac: Thought about this as well. Not sure what the preference would be. I started using `temporal modes` and then later noticed that `timebins` was used inside `TDMProgram`. I don't necessarily mind using both, but I agree it might be a bit confusing having two different names for the same thing. I guess it depends on how commonly used these terms are outside of our software libraries. Any inputs @nquesada?

strawberryfields/io.py
```diff
@@ -146,6 +173,63 @@ def to_program(bb):
...
+def _convert_tdm(bb):
+    # pylint: disable=import-outside-toplevel
+    from strawberryfields.tdm.tdmprogram import TDMProgram
+
+    prog = TDMProgram(max(bb.modes) + 1, name=bb.name)
+
+    args = [bb._var[f"p{i}"].flatten() for i in range(max(bb.modes) + 1)]
+    # append the quantum operations
+    with prog.context(*args, copies=bb.programtype["options"]["copies"]) as (p, q):
+        for op in bb.operations:
```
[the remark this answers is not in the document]
thisac: Thought about this as well, and decided to go for duplication instead for two reasons:

1. It made it clearer to understand what was happening, avoiding the use of a lot of `if-tdmprogram` branches.

2. Things might change with TDMProgram in the future, and this might make it easier to maintain both program conversions, allowing for more program-specific changes in the future. It's difficult to say what might change, but since they use completely different contexts and also differ in the use of symbolic parameters (including the use of rolling/unrolling in `TDMProgram`) there could be significant updates here that would make it easier to maintain if separating the two conversions.

I'm happy to make this more compact though if others think that's better.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -175,27 +176,42 @@ class TDMProgram(sf.Program):
...
+    simulated registers. This is performed automatically by the local engine, however, we can
+    visualize the unrolling manually via the
+    :meth:`~.unroll` method.
+
+    >>> prog.unroll().print()
+    Sgate(0.7, 0) | (q[1])
     BSgate(1, 0) | (q[0], q[1])
     MeasureHomodyne(3) | (q[0])
-    Sgate(0.32, 0) | (q[0])
+    Sgate(0.7, 0) | (q[0])
```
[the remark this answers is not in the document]
thisac: Yes, I'm guessing it's simply has to do with the value for `Sgate` used in the context above, right @nquesada?

tests/frontend/test_io.py
thisac: Thanks Antal. This made me discover a small bug in the code. :tada:

tests/frontend/test_io.py
```diff
@@ -528,6 +551,52 @@ def test_gate_multimode(self):
...
+        assert prog.circuit[2].reg[0].ind == 0
+
+        assert prog.concurr_modes == 2
+        assert prog.timebins == 2
+        assert prog.copies == 3
+        assert prog.spatial_modes == 1
+        assert prog.free_params == {'p0': FreeParameter("p0"), 'p1': FreeParameter("p1")}
+        assert all(prog.tdm_params[0] == np.array([1, 2]))
+        assert all(prog.tdm_params[1] == np.array([3, 4]))
+
```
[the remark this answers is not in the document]
thisac: There's no `BlackbirdSyntaxError` being raised in `io.py`. That error should tested in another place. Although I added tests for the `NameError` and `ValueError` instead. :slightly_smiling_face:

strawberryfields/engine.py
```diff
@@ -632,6 +637,8 @@ def run_async(
         kwargs.update(self._backend_options)
 
         device = self.device_spec
+        if program.type == "tdm":
+            device.fill_template(program, device.target)
```
antalszava, line 641: @thisac seems like an adjustment is needed here, correct? `target` is derived from the `device` later
```suggestion
            device.fill_template(program)
```

thisac: Thanks @antalszava. :1st_place_medal:

strawberryfields/api/devicespec.py
```diff
@@ -97,17 +98,27 @@ def gate_parameters(self):
 
         return gate_parameters
 
+    def fill_template(self, program):
+        """Fill template with parameter values from a program"""
+        if program.type == "tdm" and program.timebins:
```
josh146, line 103: Is it ever possible for `program.type == "tdm"` and not `progam.timebins`?

strawberryfields/io.py
josh146: Also @fab1an-q!

strawberryfields/io.py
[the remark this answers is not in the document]
josh146: I would keep this separate for now as well; it is still too early, and we might make changes in the near term that cause the logic to deviate

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
josh146: This is a result of having multiple PRs merged into this one 😆 This was me, at one point I noticed that the print statement was incorrect (`prog`, defined above, uses 0.7)

tests/api/test_devicespec.py
```diff
@@ -49,6 +49,45 @@
...
+        {{r_array}}
+    float array p2[1, {tm}] =
+        {{m_array}}
+
+    Sgate(1, 0) | 1
+    BSgate(p0) | (1, 0)
+    Rgate(p1) | 1
+    MeasureHomodyne(p2) | 0
+    """
+)
```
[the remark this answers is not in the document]
josh146: I suppose this will now need to be updated!

strawberryfields/io.py
fab1an-q: It might be subtle but the difference between timebins and temporal modes occurs when we have more than one spatial mode where within one timebin we may deal with many temporal qumodes (think of the Tokyo and DTU experiments). This is why I prefer timebins, personally.

strawberryfields/api/devicespec.py
thisac: No it isn't. I think I added this because before, `self.timebins` wasn't set in the `__init__` method, but it is now, so this can be removed here.

strawberryfields/program.py
```diff
@@ -529,17 +544,7 @@ def _get_compiler(compiler_or_name):
...
-                # Program subsystems may be created and destroyed during execution. The length
-                # of the program registers represents the total number of modes that has ever existed.
-                modes_total = len(self.reg_refs)
-
-                if modes_total > device.modes:
-                    raise CircuitError(
-                        f"This program contains {modes_total} modes, but the device '{target}' "
-                        f"only supports a {device.modes}-mode program."
-                    )
+                self.assert_number_of_modes(device)
```
thisac, line 547: @josh146 @antalszava Just realized that the recently merged [No modes check for simulon PR](https://github.com/XanaduAI/strawberryfields/pull/482) clashes with this one, since that PR added a check for `isinstance(device.modes, int)` rather than simply checking for `device.modes != "simulon"` (as was first proposed). That doesn't work with the changes here, so we'd have to update it by e.g. moving it to inside the `assert_number_of_modes` method, or changing the check (since not only integers should eval `True` here.

antalszava: Hmm, if I understand correctly, the clash is related to a `TDMProgram` inheriting this method, correct? Would it be solved by having `TDMProgram.compile` be redefined in #484? :thinking: We also have #483 open

thisac: Oh, that's true. We should add a call to `assert_number_of_modes` in `TDMProgram.compile` as well then, and perhaps just leave the `isinstance(device.modes, int)` here. I feel like this is becoming a bit confusing though, but we're planning on adding checks for other mode entries (e.g. dicts) as well, right? In that case perhaps it would be best just to expand the `assert_number_of_modes` method to check for all possibilities.

antalszava: Yes, I think you're right. The way I understand how a potential way could be:
1. Wait for #484 to be merged in here *or* create a separate condition to resolve conflict in `program.py` and allow the tests to pass e.g.,
```python
            if device.modes is not None and (isinstance(device.modes, int) or device.target == "tdm"):
                self.assert_number_of_modes(device)
```
2. Once #484 was merged, if the new condition in 1. was added, remove `or device.target == "tdm"`
3. When resolving #483, expand `assert_number_of_modes` to check for `dict` types

strawberryfields/api/devicespec.py
```diff
@@ -97,17 +99,35 @@ def gate_parameters(self):
...
+        if self.layout_is_formatted():
+            return
+
+        if program.type == "tdm" and program.timebins:
+            self._spec["layout"] = self._spec["layout"].format(
+                target=self.target, tm=program.timebins
+            )
+        else:
+            # TODO: update when `self._spec["layout"]` is returned as an unformatted string
+            raise NotImplementedError("Formatting not required or supported for non-TDM programs.")
```
josh146, line 118: Please add this to the clubhouse backlog so that we don't forget!

strawberryfields/io.py
```diff
@@ -40,6 +41,7 @@ def to_blackbird(prog, version="1.0"):
         blackbird.BlackbirdProgram:
     """
     bb = blackbird.BlackbirdProgram(name=prog.name, version=version)
+    bb._modes = list(prog.reg_refs.keys())
```
josh146, line 44: Was this an oversight that this wasn't being set before?

strawberryfields/io.py
```diff
@@ -146,6 +176,70 @@ def to_program(bb):
...
+    # pylint: disable=import-outside-toplevel
+    from strawberryfields.tdm.tdmprogram import TDMProgram
+
+    prog = TDMProgram(max(bb.modes) + 1, name=bb.name)
+
+    # retrieve all the free parameters in the Blackbird program (e.g. "p0", "p1"
+    # etc.) and add their corresponding values to args
+    args = []
+    for k in bb._var.keys():
+        if k[0] == "p" and k[1:].isdigit():
```
josh146, line 189: Nice, I didn't know about `isdigit`!

strawberryfields/io.py
```diff
@@ -146,6 +176,70 @@ def to_program(bb):
...
+                for k, v in kwargs.items():
+                    if isinstance(v, str) and v[0] == "p" and v[1:].isdigit():
+                        kwargs[k] = sfpar.FreeParameter(v)
+
+                # Convert symbolic expressions in args/kwargs containing measured and free parameters to
+                # symbolic expressions containing the corresponding MeasuredParameter and FreeParameter instances.
+                args = sfpar.par_convert(args, prog)
+                vals = sfpar.par_convert(kwargs.values(), prog)
+                kwargs = dict(zip(kwargs.keys(), vals))
+                gate(*args, **kwargs) | regrefs  # pylint:disable=expression-not-assigned
```
josh146, line 227: :+1:

strawberryfields/tdm/tdmprogram.py
```diff
@@ -175,27 +176,42 @@ class TDMProgram(sf.Program):
...
+    we must use the ``prog.parameters`` attribute:
+
+    >>> prog.parameters
+    {'p0': [1, 2], 'p1': [3, 4]}
+
+    When we simulate a time-domain program, it is first unrolled by the engine; unrolling involves
+    explicitly repeating the single time-bin sequence constructed above, and *shifting* the
+    simulated registers. This is performed automatically by the local engine, however, we can
+    visualize the unrolling manually via the
+    :meth:`~.unroll` method.
```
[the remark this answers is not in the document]
josh146:
```suggestion
    simulated registers. This 'unrolling' procedure is performed automatically by the engine, however, we can
    visualize the unrolled program by calling the :meth:`~.unroll` method.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -282,6 +310,121 @@ def context(self, *args, copies=1, shift="default"):
...
+        """Compile the time-domain program given a Strawberry Fields photonic hardware device specification.
+        At this stage the compilation is simply a check that the program matches the device.
+
+        Args:
+            device (~strawberryfields.api.DeviceSpec): device specification object to use for
+                program compilation
+            compiler (str, ~strawberryfields.compilers.Compiler): Compiler name or compile strategy
+                to use. If a device is specified, this overrides the compile strategy specified by
+                the hardware :class:`~.DeviceSpec`. For :class:`~.TDMProgram` this should be set to
+                None.
```
josh146, line 323: This needs to be updated 🙂 Since we can now pass `gaussian`, `tdm_one_loop`, etc.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -282,6 +310,121 @@ def context(self, *args, copies=1, shift="default"):
...
+        Returns:
+            Program: compiled program
+        """
+        if compiler == "gaussian":
+            return super().compile(device=device, compiler=compiler)
+
+        if compiler == "tdm_one_loop" and device == "tdm_one_loop_device":
+            # Do something clever
+            compiler = "TD2"  # Devious but it will work, Josh said
+            # Do not add a return statement
```
josh146, line 334: 🙈

josh146: I'm struggling to recall the point of this if statement @nquesada 

* `device` cannot be a string, yet it is being compared to a string. So this if statement will never be activated.

* Even if the `if` statement _is_ activated, the internal logic doesn't seem to do anything.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -360,6 +396,24 @@ def apply_op(self, cmd, q, t):
 
         self.append(cmd.op.__class__(*params), get_modes(cmd, q))
 
+    def assert_number_of_modes(self, device):
+        """Check that the number of modes in the program is valid for the given device."""
+        if self.timebins > device.modes["temporal"]["max"]:
+            raise CircuitError(
+                f"This program contains {self.timebins} temporal modes, but the device '{device.target}' "
+                f"only supports up to {device.modes['temporal']['max']} modes."
+            )
```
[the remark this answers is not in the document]
josh146: I think this suggestion is worth doing

tests/frontend/test_io.py
```diff
@@ -528,6 +554,96 @@ def test_gate_multimode(self):
...
+        int array p42 =
+            1, 3, 5
+
+        np | [0, 1]
+        """)
+
+        bb = blackbird.loads(bb_script)
+
+        with pytest.raises(NameError, match="operation np not defined"):
+            io.to_program(bb)
```
josh146, line 646: Very thorough testing!

tests/frontend/test_tdmprogram.py
```diff
@@ -406,17 +432,219 @@ def test_two_dimensional_cluster_tokyo():
...
+        device1 = DeviceSpec("x", device_spec1, connection=None)
+        c = 1
+        copies = 1
+        sq_r = 0.5643
+        alpha = [0.5, 0] * c
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] * c
+        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        with pytest.raises(sf.program_utils.CircuitError, match="spatial modes, but the device"):
+            prog.compile(device=device1, compiler="TD2")
```
josh146, line 650: also some amazing testing 💯

josh146 (commented):
This is looking great! Very thorough and almost merge ready.

A couple of general comments:

* Don't forget to update the changelog!

* I can see that the `io` and `compilation` logic is being tested, but I didn't see any new (unit) tests regarding the new engine logic (the rolling/unrolling)?

strawberryfields/tdm/tdmprogram.py
nquesada: Hey @josh146 --- this is partly my fault. I still need to add the logic there to some "clever" transformations. I left those comments there just a reminder for myself. They can be removed and I can the later add a branch with the actual logic.

josh146: no worries @nquesada, I was worried that I was forgetting my own devious reasoning here 😆

strawberryfields/io.py
thisac: Yes, it wasn't there before. :grimacing:

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
thisac: @antalszava @josh146 It's actually possible to change the `TDMProgram` after creation by appending more operations to the program. They are, however, _not_ represented in the unrolled circuit if the program has already been unrolled earlier, and the new gates will "disappear" if unrolling again (since the unrolling only used `unrolled_circuit` to unroll it if it's already been done before).

The program gets locked when the engine is run, so after that it's no longer possible.

I guess this could be fixed either by forcing an unrolling each time an operation is appended, or simply by locking the circuit early (I think the first option seems nice, and seems fairly simple to implement). Perhaps we could add this as an issue and fix it in a separate PR later?

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
thisac: I'm not sure if that's actually a better solution, looking closer at it. I think it's fine having three checks, and separate error raises, since they differ in either having to not be `!=` or `>` the temporal/concurrent/spatial modes value in the device spec.

strawberryfields/api/devicespec.py
thisac: [Backlog ticket added!](https://app.clubhouse.io/xanaduai/story/2813/update-fill-template-to-support-non-tdm-layouts)

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
 
 <h3>New features since last release</h3>
 
+* `TDMProgram`s can now be compiled and submitted via the API.
+  [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
```
josh146, line 6:
```suggestion
* `TDMProgram` objects can now be compiled and submitted via the API.
  [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
```
(ReST doesn't like characters occurring after the backticks)

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
 <h3>Improvements</h3>
 
+* `copies` are renamed `shots` inside of `TDMProgram` and is now passed via
+  the `eng.run()` method instead of via the program context.
+  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```
josh146, line 12:
```suggestion
* The `copies` option when constructing a `TDMProgram` have been removed. Instead, the number of
  copies of a TDM algorithm can now be set by passing the `shots` keyword argument to 
  the `eng.run()` method.
  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
+  ```pycon
+    >>> with prog.context([1, 2], [3, 4]) as (p, q):
+    ...     ops.Sgate(0.7, 0) | q[1]
+    ...     ops.BSgate(p[0]) | (q[0], q[1])
+    ...     ops.MeasureHomodyne(p[1]) | q[0]
+
+    >>> eng = sf.Engine("gaussian")
+    >>> results = eng.run(prog, shots=3)
+  ```
```
josh146, line 22:
```suggestion
  ```pycon
  >>> with prog.context([1, 2], [3, 4]) as (p, q):
  ...     ops.Sgate(0.7, 0) | q[1]
  ...     ops.BSgate(p[0]) | (q[0], q[1])
  ...     ops.MeasureHomodyne(p[1]) | q[0]
  >>> results = eng.run(prog, shots=3)
  ```
```

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
+  Furthermore, `TDMProgram.unrolled_circuit` only contains the single-shot
+  unrolled circuit, while the unrolling including shots is returned by the
+  `TDMProgram.unroll()` method, and is strictly contained to this method
+  (and thus not an attribute of the `TDMProgram` class any longer).
```
josh146, line 27:
```suggestion
  Furthermore, the `TDMProgram.unrolled_circuit` attribute now only contains the single-shot
  unrolled circuit. Unrolling with multiple shots can still be specified via the `unroll` method:
  `TDMProgram.unroll(shots=60)`.
```

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
+* The `reshape_samples` function is updated to return a dictionary where
+  each key correspond to a spatial mode and the values have shape ``(shots,
+  timebins)``.
+  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```
josh146, line 32: Since this is an internal implementation detail, perhaps it's not worth including?

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
+* The `Result.samples` is updated for TDM programs to return samples of
+  shape `(shots, spatial modes, timebins)` instead of `(shots, spatial
+  modes * timebins)`.
+  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```
josh146, line 37:
```suggestion
* The `Result.samples` returned by TDM programs has been updated to return
  samples of shape `(shots, spatial modes, timebins)` instead of `(shots, spatial
  modes * timebins)`.
  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
+* A sample post-processing function is added that allows user's to move
+  vacuum mode measurements from the first shots to the last shots, and
+  potentially crop out the final shots containing these measurements.
+  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```
josh146, line 42:
```suggestion
* A sample post-processing function is added that allows users to move
  vacuum mode measurements from the first shots to the last shots, and
  potentially crop out the final shots containing these measurements.
  [(#489)](https://github.com/XanaduAI/strawberryfields/pull/489)
```

Potentially also not to needed to be included, since it is more of an internal implementation detail?

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+
+* `TDMProgram`s can be converted into Blackbird programs and vice versa.
+  [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
```
josh146, line 45:
```suggestion
* `TDMProgram` objects can now be serialized into Blackbird scripts, and vice versa.
  [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
```

strawberryfields/api/devicespec.py
```diff
@@ -97,17 +99,35 @@ def gate_parameters(self):
 
         return gate_parameters
 
+    def layout_is_formatted(self):
+        """bool: Whether the device layout is formatted or not."""
+        p = re.compile(r"{{\w*}}")
+        return not bool(p.search(self.layout))
```
josh146, line 105: Minor: perhaps this could be done with less overhead by simply returning `"{{" in self.layout`?

.github/CHANGELOG.md
josh146, line 37: This is another reason why I suggest not using a holding branch for future projects --- so many changelog entries refer to the same (large) PR 😆

.github/CHANGELOG.md
```diff
@@ -2,18 +2,68 @@
...
+  `io.to_blackbird` function.
+  [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
+
 <h3>Documentation</h3>
 
 <h3>Contributors</h3>
 
 This release contains contributions from (in alphabetical order):
 
+Theodor Isacsson, Fabian Laudenbach, Nicolas Quesada
```
josh146, line 65:
```suggestion
Theodor Isacsson, Josh Izaac, Fabian Laudenbach, Nicolas Quesada, Antal Száva.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
     param_lengths = [len(param) for param in args]
     if len(set(param_lengths)) != 1:
         raise ValueError("Gate-parameter lists must be of equal length.")
 
-    # check copies
-    if not isinstance(copies, int) or copies < 1:
-        raise TypeError("Number of copies must be a positive integer.")
 
+def _get_mode_order(num_of_values, modes, N, timebins):
+    """Get the order by which the modes were measured
```
josh146, line 101:
```suggestion
    """Get the order in which the modes were measured.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
+
+    The mode order is determined by the circuit and the mode-shifting occurring in
+    :class:`~.TDMProgram`. For the following circuit, the mode order returned by this
+    function would be [0, 2, 0, 1], duplicated shots number of times:
```
josh146, line 105:
```suggestion
    function would be ``[0, 2, 0, 1]``, duplicated shots number of times:
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
+    which would then be reshaped, and returned, as follows:
+
+    .. code-block:: pycon
+
+        {
+            0: np.array([[(sample 0), (sample 2)],
+                         [(sample 4), (sample 6)]]),
+            2: np.array([[(sample 1), (sample 3)],
+                         [(sample 5), (sample 7)]]),
+        }
```
josh146, line 180: 👌

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
-    defined to be measured in the circuit, instead of being spread over a larger
+    Corrects the :attr:`~.Results.all_samples` dictionary so that the measured modes are
+    the ones defined to be measured in the circuit, instead of being spread over a larger
     number of modes due to the mode-shifting occurring in :class:`~.TDMProgram`.
 
+    The function iterates through samples obtained from the unrolled circuit to populate
+    and return a new samples dictionary with the shape ``{spatial mode: (shots,
+    timebins)}``. E.g., this unrolled circuit:
+
+    .. code-block:: pycon
```
josh146, line 146: This codeblock (and the two below) can probably be changed to `python`, since there is no python console syntax present

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
+        if last_mode_in_timebin:
+            timebin_idx = (timebin_idx + 1) % timebins
+
+    # transpose each value so that it has shape `(shots, timebins)`
+    return {k: np.array(v).T for k, v in new_samples.items()}
+
+
+def move_vac_modes(samples, N, crop=False):
+    """Post-processing function for TDM samples. Moves all measured vacuum modes
+    from the first shot of the samples array to the end of the last shot.
```
josh146, line 221: "Post-processing function for TDM samples" is a bit superfluous, we can probably remove it:
```suggestion
    """Moves all measured vacuum modes from the first shot of the
    returned TDM samples array to the end of the last shot.
```

josh146: I have to admit, I still don't quite follow how come this function was needed. It would be good to get some context here!

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
+            the possibly multiple bands in the circuit.
+
+    Keyword args:
+        crop (bool): whether to remove all the shots containing measured vacuum
+            modes at the end
+
+    Returns:
+        ndarray[float]: the post-processed samples
+
+    """
```
josh146, line 238:
```suggestion
    Returns:
        array[float]: the post-processed samples
    """
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
+    # transpose each value so that it has shape `(shots, timebins)`
+    return {k: np.array(v).T for k, v in new_samples.items()}
+
+
+def move_vac_modes(samples, N, crop=False):
+    """Post-processing function for TDM samples. Moves all measured vacuum modes
+    from the first shot of the samples array to the end of the last shot.
+
+    Args:
+        samples (ndarray[float]): samples as received from ``TDMProgram``, with the
```
josh146, line 224:
```suggestion
        samples (array[float]): samples as received from ``TDMProgram``, with the
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,176 @@ def validate_measurements(circuit, N):
...
+        samples = samples[:num_of_shots_with_vac_modes]
+
+    return samples
 
 
 class TDMProgram(sf.Program):
     r"""Represents a photonic quantum circuit in the time domain encoding.
 
-    The ``TDMProgam`` class provides a context manager for easily defining
+    The ``TDMProgram`` class provides a context manager for easily defining
```
josh146, line 257: Good catch!

strawberryfields/tdm/tdmprogram.py
```diff
@@ -268,20 +387,139 @@ def __init__(self, N, name=None):
...
+    def compile(self, *, device=None, compiler=None):
+        """Compile the time-domain program given a Strawberry Fields photonic hardware device specification.
+        At this stage the compilation is simply a check that the program matches the device.
+
+        Args:
+            device (~strawberryfields.api.DeviceSpec): device specification object to use for
+                program compilation
+            compiler (str, ~strawberryfields.compilers.Compiler): Compiler name or compile strategy
+                to use. If a device is specified, this overrides the compile strategy specified by
+                the hardware :class:`~.DeviceSpec`.
```
josh146, line 424: It might be worth including a list of supported compiler names here (currently, `"gaussian"` and `"TD2"`.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -268,20 +387,139 @@ def __init__(self, N, name=None):
...
+        input_check(args)
         self.tdm_params = args
         self.shift = shift
         self.loop_vars = self.params(*[f"p{i}" for i in range(len(args))])
         return self
 
+    # pylint: disable=too-many-branches
+    def compile(self, *, device=None, compiler=None):
+        """Compile the time-domain program given a Strawberry Fields photonic hardware device specification.
+        At this stage the compilation is simply a check that the program matches the device.
```
josh146, line 417:
```suggestion
        """Compile the time-domain program given a Strawberry Fields photonic hardware device specification.
        
        Currently, the compilation is simply a check that the program matches the device.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -268,20 +387,139 @@ def __init__(self, N, name=None):
...
+                to use. If a device is specified, this overrides the compile strategy specified by
+                the hardware :class:`~.DeviceSpec`.
+
+        Returns:
+            Program: compiled program
+        """
+        if compiler == "gaussian":
+            return super().compile(device=device, compiler=compiler)
+
+        if compiler == "TD2" and device is not None:
```
josh146, line 432: Should this simply be?
```suggestion
        if device is not None:
```

This way, we allow the behaviour `prog.compile(device=spec)`.

Or is there a reason for requiring `prog.compile(device=spec, compiler="TD2")` to be explicit?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -268,20 +387,139 @@ def __init__(self, N, name=None):
...
+                        # If it is a numerical value check directly
+                        if not program_param in param_range:
+                            raise CircuitError(
+                                "Program cannot be used with the device '{}' "
+                                "due to incompatible parameter. Parameter has value '{}'"
+                                "while its valid range is '{}'".format(
+                                    device.target, program_param, param_range
+                                )
+                            )
+        return self
```
josh146, line 521: Should the logic be like the following?

```python
if compiler == "gaussian":
    return ...

if compiler == "TD2" and device is not None:
    ...
    return self

raise CircuitError("Unknown compiler etc.")
```

This avoids the case where `prog.compile()` or `prog.compile(compiler="TD3")` simply returns the existing program, which might not be intuitive.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -268,20 +387,139 @@ def __init__(self, N, name=None):
...
+
+        Returns:
+            Program: compiled program
+        """
+        if compiler == "gaussian":
+            return super().compile(device=device, compiler=compiler)
+
+        if compiler == "TD2" and device is not None:
+
+            if device.modes is not None:
```
josh146, line 434: Should we also be checking that the device spec provided corresponds to a TDM hardware device? I imagine there might be issues if a user tries to compile a TDM program using a GBS hardware device spec! This allows us to catch that and provide a nice error message.

josh146 (changes requested):
Monstrous work by everyone 💪 I left some thoughts, comments and suggestions, but getting very very close.

strawberryfields/tdm/tdmprogram.py
thisac: I think this sounds reasonable. This would mean that by not specifying a compiler, it would always just check the circuit "strictly", right? And then we could add other compilers later if we want to have more flexibility.

strawberryfields/tdm/tdmprogram.py
thisac: That makes sense also! :+1:

strawberryfields/tdm/tdmprogram.py
josh146: Yep, which is what I expect intuitively as a user if I were to call `prog.compile(device=spec)`.

strawberryfields/tdm/tdmprogram.py
thisac: I like this! It makes sense. Although, if we decide to change `if compiler == "TD2" and device is not None:` to `if device is not None:`, there would need to be another error message.

strawberryfields/io.py
```diff
@@ -146,6 +176,71 @@ def to_program(bb):
...
+            if op["op"] in ops.__all__:
+                # get the quantum operation from the sf.ops module
+                gate = getattr(ops, op["op"])
+            else:
+                raise NameError("Quantum operation {} not defined!".format(op["op"]))
+
+            # create the list of regrefs
+            regrefs = [q[i] for i in op["modes"]]
+
+            if "args" in op:
```
antalszava, line 211: Perhaps `"kwargs" in op` would anyways be assumed `if "args" in op`, yet think it's good to keep it explicit (and potentially think if there would be an edge case where it could happen that exactly one is defined).
```suggestion
            if "args" in op and "kwargs" in op:
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -413,27 +413,34 @@ def context(self, *args, shift="default"):
...
             return super().compile(device=device, compiler=compiler)
 
-        if compiler == "TD2" and device is not None:
+        if device is not None:
+            device_layout = bb.loads(device.layout)
+
+            if device_layout.programtype["name"] != "tdm":
+                raise TypeError(
+                    'TDM compiler only supports "tdm" type device specification layouts. '
+                    "Received {} type.".format(device_layout.programtype["name"])
```
josh146, line 439: 👍

strawberryfields/tdm/tdmprogram.py
```diff
@@ -413,27 +413,34 @@ def context(self, *args, shift="default"):
...
         if compiler == "gaussian":
             return super().compile(device=device, compiler=compiler)
 
-        if compiler == "TD2" and device is not None:
+        if device is not None:
+            device_layout = bb.loads(device.layout)
+
+            if device_layout.programtype["name"] != "tdm":
+                raise TypeError(
+                    'TDM compiler only supports "tdm" type device specification layouts. '
```
josh146, line 438:
```suggestion
                    "TDM compiler only supports "tdm" type device specifications. "
```
(perhaps needs to be Blacked?)

strawberryfields/tdm/tdmprogram.py
```diff
@@ -518,7 +525,9 @@ def compile(self, *, device=None, compiler=None):
                                     device.target, program_param, param_range
                                 )
                             )
-        return self
+            return self
+
+        raise CircuitError("TDM programs cannot be compiled without a valid device specification.")
```
josh146, line 530: :+1:

josh146 (approved):
💯 🥇

strawberryfields/io.py
```diff
@@ -146,6 +176,71 @@ def to_program(bb):
...
+    # pylint: disable=import-outside-toplevel
+    from strawberryfields.tdm.tdmprogram import TDMProgram
+
+    prog = TDMProgram(max(bb.modes) + 1, name=bb.name)
+
+    # retrieve all the free parameters in the Blackbird program (e.g. "p0", "p1"
+    # etc.) and add their corresponding values to args
+    args = []
+    for k in bb._var.keys():
+        if k[0] == "p" and k[1:].isdigit():
```
antalszava, line 190: Could create something like `is_bb_free_param = lambda param: isinstance(param, str) and param[0] == "p" and param[1:].isdigit()` to help with readability and usage here and at later occurences.

strawberryfields/tdm/tdmprogram.py
thisac: Nope, there's a `"tdm"` inside of the string, so this is correct. Although it might be better to just write `tdm` instead. :thinking:

strawberryfields/program.py
```diff
@@ -449,6 +451,19 @@ def _linked_copy(self):
             p.source = self.source
         return p
 
+    def assert_number_of_modes(self, device):
+        """Check that the number of modes in the program is valid for the given device."""
+
+        # Program subsystems may be created and destroyed during execution. The length
+        # of the program registers represents the total number of modes that has ever existed.
```
antalszava, line 458:
```suggestion
        # of the program registers represents the total number of modes that ever existed.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -81,69 +85,175 @@ def validate_measurements(circuit, N):
...
+
+    would return the dictionary
+
+    .. code-block:: python
+
+        {
+            0: np.array([(sample 0), (sample 2), (sample 4), (sample 6)]),
+            2: np.array([(sample 1), (sample 5)]),
+            1: np.array([(sample 3), (sample 7)]),
+        }
```
antalszava, line 169: Thanks for adding this! :)

strawberryfields/tdm/tdmprogram.py
```diff
@@ -268,20 +386,149 @@ def __init__(self, N, name=None):
...
+
+                    # make sure that hardcoded parameters in the device layout are correct
+                    if not isinstance(param_name, str):
+                        if not program_param == param_name:
+                            raise CircuitError(
+                                "Program cannot be used with the device '{}' "
+                                "due to incompatible parameter. Parameter has value '{}'"
+                                "while its valid value is '{}'".format(
+                                    device.target, program_param, param_name
+                                )
```
antalszava, line 497: Could these with the next two messages be factored out? There seems to be a minor difference

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
antalszava: The original suggestion aimed at only factoring out the string for the message, but keeping the three checks separated.

Inside of the string, the only difference seems to be the type of the mode (e.g., `This program contains {self.timebins} temporal modes`, `This program contains {self.concurr_modes} concurrent modes`, etc.). Here we could go with `This program contains {num_modes} {mode_type} modes`. It just aimed at transitioning to have a single string that we'd have to maintain. Although it involves moving from using `f` strings to using `format`, think that could be fine.

tests/api/test_devicespec.py
```diff
@@ -115,3 +156,38 @@ def test_refresh(self, connection, monkeypatch):
         spec.refresh()
         assert spec.modes != device_dict["modes"]
         assert spec.modes == 42
+
+    def test_fill_template(self, connection, monkeypatch):
+        """Test that the fill_template methods works"""
```
antalszava, line 161:
```suggestion
        """Test that the fill_template method works"""
```

tests/frontend/test_tdmprogram.py
```diff
@@ -383,40 +417,276 @@ def test_two_dimensional_cluster_tokyo():
...
+    def test_tdm_wrong_layout(self):
+        """Test the correct error is raised when the tdm circuit gates don't match the device spec"""
+        sq_r = 0.5643
+        c = 2
+        alpha = [np.pi / 4, 0] * c
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] + [np.pi / 2, np.pi / 2]
+        prog = tdmprogram.TDMProgram(N=2)
+        with prog.context(alpha, phi, theta) as (p, q):
+            ops.Dgate(sq_r) | q[1]  # Here one should have an Sgate
```
antalszava, line 529: Thanks for these comments!

antalszava (approved):
Looks good to me! :100: :1st_place_medal: Great one! :blush: 

Left a couple of thoughts on code quality, mostly minor. One thing that stayed in my mind was to make sure that `"kwargs"` is always defined in a bb template when `"args"` is (might be trivial but could necessitate adding one or two small tests, not quite sure).

Other than that I think it's also good to keep the room open here for modifications after some functional testing (that might uncover things that we might have missed).

requirements.txt
```diff
@@ -5,7 +5,7 @@ sympy>=1.5
 tensorflow>=2.0
 tensorboard>=2.0
 networkx>=2.0
-quantum-blackbird==0.2.3
+quantum-blackbird>=0.3.0
```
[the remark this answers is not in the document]
thisac: Not really, that I know of. Might be best to just pin it to `==0.3` instead, allowing non-breaking releases and bugfixes, right?

strawberryfields/engine.py
```diff
@@ -469,12 +478,14 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
             program, args=args, compile_options=compile_options, **eng_run_options
         )
 
-        if isinstance(program, TDMProgram):
+        if not isinstance(program, collections.abc.Sequence) and program.type == "tdm":
```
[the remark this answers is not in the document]
thisac: This was mainly because importing `TDMProgram` caused a circular import in other places and thus broke things. Here, actually, both should work equally well, but `program.type == "tdm"` is used elsewhere, so it was just to be a bit more consistent.

requirements.txt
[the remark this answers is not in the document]
antalszava: Good question :thinking: Not quite sure actually. The advantage of `quantum-blackbird>=0.3.0` would be that we might not need to update `requirements.txt` if there's a new bb version and that new version is not necessarily required by Strawberry Fields.

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
thisac: I see your point (and I almost agree with it). But there's the slight extra difference that one of the messages has the additional `up to` in it. If you're fine with it, I might just keep this as it is. I think it works fine, and is the simplest solution without having to re-word it in a slightly different way.

requirements.txt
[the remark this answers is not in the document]
thisac: I'll keep it like this for now then, and we can pin it later if needed. There probably won't be any major Blackbird changes for a little while.

tests/frontend/test_tdmprogram.py
```diff
@@ -12,101 +12,97 @@
...
+
+pytestmark = pytest.mark.frontend
 
 # make test deterministic
 np.random.seed(42)
 
 
-def singleloop(r, alpha, phi, theta, copies, shift="default", hbar=2):
+def singleloop(r, alpha, phi, theta, shots, shift="default"):
     """Single delay loop with program.
```
nquesada, line 32: Single loop program?

tests/frontend/test_tdmprogram.py
```diff
@@ -12,101 +12,97 @@
...
     Args:
         r (float): squeezing parameter
         alpha (Sequence[float]): beamsplitter angles
         phi (Sequence[float]): rotation angles
         theta (Sequence[float]): homodyne measurement angles
-        hbar (float): value in appearing in the commutation relation
+        shots (int): number of shots
+        shift (string): type of shift used in the program
     Returns:
         (list): homodyne samples from the single loop simulation
```
nquesada, line 42: I think this is now an array, no?

nquesada (approved):
left some very minor suggestions. Really happy to see this one going forward. Thank you all for your work on this!

tests/frontend/test_tdmprogram.py
thisac: Fixed, thanks!

tests/frontend/test_tdmprogram.py
thisac: Changed to `Single-loop program.`

PR #478, opened 2020-10-23
renames construct to unroll and does unrolling only in engine

nquesada:


strawberryfields/tdm/tdmprogram.py
```diff
@@ -294,16 +296,22 @@ def __exit__(self, ex_type, ex_value, ex_tb):
...
+            self.total_timebins = self.timebins * self.copies
+            # self.construct_circuit()
+            self.rolled_circuit = self.circuit.copy()
 
-    def construct_circuit(self):
+    def roll(self):
+        """Represent the in a compressed way using without rolling the for loops"""
+        self.circuit = self.rolled_circuit
+
+    def unroll(self):
```
thisac, line 308: Should there be an `self.unrolled_circuit = self.circuit.copy()` at the end of this method? Would `self.unrolled_circuit` be used at all otherwise; will it ever be set?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -335,12 +343,12 @@ def construct_circuit(self):
...
 
-        for cmd in cmds:
+        for cmd in self.rolled_circuit:
             if isinstance(cmd.op, ops.Measurement):
                 self.measured_modes.append(cmd.reg[0].ind)
 
         for i in range(self.total_timebins):
-            for cmd in cmds:
+            for cmd in self.rolled_circuit:
                 self.apply_op(cmd, q, i)
```
thisac, line 352: Would `self.rolled_circuit` ever be changed here? I mean via the `cmd`s somehow? I don't think so, but it might be good to double check since we're not using copies here.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -294,16 +296,22 @@ def __exit__(self, ex_type, ex_value, ex_tb):
 
         if ex_type is None:
             self.spatial_modes = validate_measurements(self.circuit, self.N)
-            self.construct_circuit()
+            self.timebins = len(self.tdm_params[0])
+            self.total_timebins = self.timebins * self.copies
+            # self.construct_circuit()
```
thisac, line 301:
```suggestion
```

thisac (approved):
Looks good Nico. :+1: I believe this would work well. I like the `roll` - `unroll` solution. :slightly_smiling_face:

strawberryfields/engine.py
```diff
@@ -418,6 +418,9 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
         # pylint: disable=import-outside-toplevel
         from strawberryfields.tdm.tdmprogram import TDMProgram, reshape_samples
 
+        if isinstance(program, TDMProgram):
+            program.unroll()
+
```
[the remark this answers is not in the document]
nquesada: yup that sounds cool!

strawberryfields/tdm/tdmprogram.py
nquesada: I guess if you never ever pass it to an engine to be run, it is kind of pointless to unroll it.

josh146: oh, nice catch Theo!

josh146:
@nquesada I pushed a small change to update the docstring, https://github.com/XanaduAI/strawberryfields/pull/478/commits/2c2ed25956983adc9c14ddfd802ff50a96e313e2, feel free to make any modifications

josh146 (approved):
Thanks @nquesada and @thisac! I think this is now ready to be merged, however we need to update the changelog before merging (this change _is_ userfacing if a user prints out their TDM program, but can probably just go in the 'improvements' section)

tests/frontend/test_tdmprogram.py
```diff
@@ -250,7 +253,7 @@ def test_one_dimensional_cluster_tokyo():
     # first half of cluster state measured in X, second half in P
     theta1 = [0] * int(n / 2) + [np.pi / 2] * int(n / 2)  # measurement angles for detector A
     theta2 = theta1  # measurement angles for detector B
-
+    sf.hbar = 2.0
```
josh146, line 256: Hey Nico, inside the tests, it's best to do it like this:

```python
def test_one_dimensional_cluster_tokyo(monkeypatch):
    monkeypatch.setattr(sf, "hbar", 2.0)
```
`monkeypatch` is a built in pytest fixture that ensures that global attributes are only modified inside the test. This avoids the setting of `sf.hbar` from affecting other tests.

josh146: All the tests below will also need to be updated to use monkeypatching

tests/frontend/test_tdmprogram.py
```diff
@@ -34,15 +37,15 @@ def singleloop(r, alpha, phi, theta, copies, shift="default", hbar=2):
     Returns:
         (list): homodyne samples from the single loop simulation
     """
-
+    sf.hbar = 2.0
```
josh146, line 40: This should be removed, and instead all tests that call it should monkeypatch the value of `hbar` (to avoid this global variable change from affecting other files)

PR #479, opened 2020-10-26
Update ops.py

nquesada:
Adds some special cases of this interferometer, which can help theorists trying to figure what it does.

### DIFF
--- a/strawberryfields/ops.py
+++ b/strawberryfields/ops.py
@@ -1765,7 +1765,7 @@ class MZgate(Gate):
         phi_ex (float): external phase
 
     .. details::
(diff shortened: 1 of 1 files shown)


strawberryfields/ops.py
```diff
@@ -1763,6 +1763,10 @@ class MZgate(Gate):
     Args:
         phi_in (float): internal phase
         phi_ex (float): external phase
+
+
+    This gate becomes the identity for phi_in = np.pi and phi_ex=0 and permutes the modes for phi_in = 0 and phi_ex=0.
```
josh146, line 1768:
```suggestion
    This gate becomes the identity for ``phi_in=np.pi`` and ``phi_ex=0``, and permutes the modes
    for ``phi_in=0`` and ``phi_ex=0``.
```

josh146 (approved):

PR #480, opened 2020-10-26
Add pytest-randomly

thisac:
**Context:**
Some tests in `test_tdmprogram.py` seems to be failing stochastically. They seem to pass locally when running a single test file, but fail in the PR, causing some confusion.

[pytest-randomly](https://pypi.org/project/pytest-randomly/) is added to the SF tests and automatically set to a seed of 42, which is the same that was used earlier in many cases. The makefile and travis are updated to use this.

**Benefits:**
Tests can now be run deterministically.

**Possible Drawbacks:**
Optimally, we would remove the `randomly-seed=42` from the tests, causing the seed to be set randomly each time, helping us discover potential bugs due to the stochastic nature of some tests. By resetting the seed before each test, we may miss failures that might be present. This might cause some headaches, though.

**Related GitHub Issues:**
None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,11 @@ matrix:
     - env: BATCHED=1 OPTIONS="tf and mixed"
 install:
   - pip install -r requirements.txt
-  - pip install wheel codecov pytest pytest-cov pytest-mock --upgrade
+  - pip install wheel codecov pytest pytest-cov pytest-randomly pytest-mock --upgrade
   - python3 setup.py bdist_wheel
   - pip install dist/StrawberryFields*.whl
 script:
-  - python3 -m pytest tests --cov=strawberryfields -p no:warnings --tb=native -m "$OPTIONS"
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -17,11 +17,11 @@ matrix:
     - env: BATCHED=1 OPTIONS="tf and mixed"
 install:
   - pip install -r requirements.txt
-  - pip install wheel codecov pytest pytest-cov pytest-mock --upgrade
+  - pip install wheel codecov pytest pytest-cov pytest-randomly pytest-mock --upgrade
   - python3 setup.py bdist_wheel
   - pip install dist/StrawberryFields*.whl
 script:
-  - python3 -m pytest tests --cov=strawberryfields -p no:warnings --tb=native -m "$OPTIONS"
+  - python3 -m pytest tests --cov=strawberryfields -p no:warnings --randomly-seed=42 --tb=native -m "$OPTIONS"
```
josh146, line 24:
```suggestion
  - python3 -m pytest tests --cov=strawberryfields -p no:warnings --randomly-seed=43 --tb=native -m "$OPTIONS"
```

josh146 (approved):

PR #482, opened 2020-11-02
No modes check for simulon

antalszava:


thisac (approved):
:+1: @antalszava.

strawberryfields/program.py
```diff
@@ -528,7 +528,7 @@ def _get_compiler(compiler_or_name):
             else:
                 compiler = _get_compiler(compiler)
 
-            if device.modes is not None:
+            if device.modes is not None and target != "simulon":
```
josh146, line 531: 🤔 
I'm quite wary of hardcoding in a target name here, especially if the target name changes in future.

Do we have a way of specifying in the device API that a target accepts an arbitrary number of modes, or a range of modes? If so, we should alter the logic below to take this into account.

E.g.,

```python
if modes_total in device.modes:
```
(or something like that)

antalszava: > I'm quite wary of hardcoding in a target name here, especially if the target name changes in future.

Fair point! I've changed to checking for `int` objects.

> Do we have a way of specifying in the device API that a target accepts an arbitrary number of modes, or a range of modes? 

Hmm, I'm not quite sure I understand :thinking: Certain devices might accept an arbitrary number of modes, or a range of modes. One thing to note is that here we would only like to make this check "stricter" so that it works for our previous devices (and incorporate the assumption that `device.modes` is an integer).

josh146 (approved):

strawberryfields/program.py
```diff
@@ -528,7 +528,8 @@ def _get_compiler(compiler_or_name):
             else:
                 compiler = _get_compiler(compiler)
 
-            if device.modes is not None:
+            # TODO: add validation for modes defined with dictionaries
```
josh146, line 531:
```suggestion
            # TODO: add validation for device specs that provide a dictionary for `device.modes`.
```

PR #484, opened 2020-11-03
Tdm compiler

nquesada:
Adds a basic compiler method for the tdm program. As currently written it requires a modification of the way a given device is specified in as a `DeviceSpec` instance.


tests/frontend/test_tdmprogram.py
```diff
@@ -451,4 +499,158 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
...
+############################################################################
+# For the test below to work, the BB cript had to be changed
+# so that the squezing parameter is a dummy symbolic variable called p0
+# and now the names insides the BB script match the names in the
+# dictionary giving the allowed ranges for the variables
+# in the DeviceSpec object
+############################################################################
+
+tm = 4
+device_spec = {
```
nquesada, line 519: Note that the device spec will have to be updated to something similar to this.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+            )
+        # Second check: the gates act in the correct modes
+        program_modes = [[r.ind for r in cmd.reg] for cmd in self.rolled_circuit]
+        device_modes = [op["modes"] for op in device_layout.operations]
+        if program_modes != device_modes:
+            raise CircuitError(
+                "Program cannot be used with the device '{}' "
+                "due to incompatible mode ordering.".format(device.target)
+            )
+        # Third check: the parameters of the gates are valid
```
nquesada, line 344: The next bit of code is  a bit clunky, wondering if there are better ways to the checks of the parameters themselves.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+        Returns:
+            Program: compiled program
+        """
+        device_layout = bb.loads(device.layout)
+        # First check: the gates are in the correct order
+        program_gates = [cmd.op.__class__.__name__ for cmd in self.rolled_circuit]
+        device_gates = [op["op"] for op in device_layout.operations]
+        if device_gates != program_gates:
+            raise CircuitError(
+                "Program cannot be used with the device '{}' "
```
nquesada, line 333: Also happy to get suggestions in terms of more meaningful error messages

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+            len_params_program = len(self.rolled_circuit[i].op.p)
+            len_params_device = len(param_names)
+            # The next if is to make sure we do not flag incorrectly things like Sgate(r,0) beign different Sgate(r)
+            if len_params_device < len_params_program:
+                for j in range(1, len_params_program):
+                    if self.rolled_circuit[i].op.p[j] != 0:
+                        raise CircuitError(
+                            "Program cannot be used with the device '{}' "
+                            "due to incompatible parameter.".format(device.target)
+                        )
```
thisac, line 359: This assumes that a potential second parameter, that's not defined in the layout,_always_ is 0, right?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+            # Now we will check explicitly if the parameters in the program match
+            counter = 0  # counts the number of symbolic variables, which are labeled consecutively by the context method
+            for k, param_name in enumerate(param_names):
+                # Obtain the relelvant parameter range from the device
```
thisac, line 363:
```suggestion
                # Obtain the relevant parameter range from the device
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+                param_range = device.gate_parameters[param_name]
+                # Obtain the value of the corresponding parameter in the program
+                program_param = self.rolled_circuit[i].op.p[k]
+                if sf.parameters.par_is_symbolic(program_param):
+                    # If it is a symbolic value go and lookup its corresponding list in self.tdm_params
+                    local_p_vals = self.tdm_params[counter]
+                    print(len(local_p_vals), device.modes["temporal"]["max"])
```
thisac, line 370:
```suggestion
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+                    if len(local_p_vals) > device.modes["temporal"]["max"]:
+                        raise CircuitError(
+                            "Program cannot be used with the device '{}' "
+                            "due to not having enough temporal modes.".format(device.target)
+                        )
```
thisac, line 375: This should already be checked in the `assert_number_of_modes` method. Is it really needed here?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+
+                    for x in local_p_vals:
+                        if not x in param_range:
+                            raise CircuitError(
+                                "Program cannot be used with the device '{}' "
+                                "due to incompatible parameter.".format(device.target)
+                            )
```
thisac, line 382: I guess this could include the parameter range and the value as well. :slightly_smiling_face:

tests/frontend/test_tdmprogram.py
```diff
@@ -451,4 +499,158 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
             ops.MeasureHomodyne(p[1]) | q[i]
 
     with pytest.raises(sf.program_utils.CircuitError, match=match):
-        new_prog = prog.compile(device=spec, compiler=DummyCompiler())
+        new_prog = prog.compile(device=spec, compiler=DummyCircuit())
```
thisac, line 502: This should probably be left as `DummyCompiler`. A bit confusing perhaps when things are being changed in different PRs back and forth. :smile:

thisac (commented):
Nice additions @nquesada. :+1: Thanks for working on this! Just a few minor comments, along with a suggestion to add some spaces in-between code parts. The compile method felt a bit compacted.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+    def compile(self, *, device=None, compiler=None, **kwargs):
+        """Compile the program given a Strawberry Fields photonic hardware device specification.
+        At this stage the compilation is simply a check that the program matches the device.
+
+        Args:
+            device (~strawberryfields.api.DeviceSpec): device specification object to use for
+                program compilation
+            compiler (str, ~strawberryfields.compilers.Compiler): Compiler name or compile strategy
+                to use. If a device is specified, this overrides the compile strategy specified by
+                the hardware :class:`~.DevicSpec`.
```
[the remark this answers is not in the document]
josh146:
```suggestion
            compiler (str, ~strawberryfields.compilers.Compiler): Compiler name or compile strategy
                to use. If a device is specified, this overrides the compile strategy specified by
                the hardware :class:`~.DeviceSpec`.
```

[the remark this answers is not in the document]
josh146: We could simply raise an error if `compiler` is anything except `None`, which might be a bit cleaner.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
...
+
+        Keyword Args:
+            optimize (bool): If True, try to optimize the program by merging and canceling gates.
+                The default is False.
+            warn_connected (bool): If True, the user is warned if the quantum circuit is not weakly
+                connected. The default is True.
```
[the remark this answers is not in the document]
josh146: Yep, I would remove them from here. They are not needed to keep the same signature as `Program.compile`, like the Args are.

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
josh146: Further, I would update the docstring here to specify that TDM compilation does not use a compiler, and that `compiler` should always be `None`.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -410,8 +495,8 @@ def assert_number_of_modes(self, device):
             )
         if self.spatial_modes > device.modes["spatial"]:
             raise CircuitError(
-                f"This program contains {self.spatial_modes} spatial modes, but the device '{device.target}' "
-                f"only supports {device.modes['spatial']} modes."
+                f"This program contains {device.timebins} temporal modes, but the device '{device.target}' "
+                f"only supports up to {device.modes['max']['temporal']} modes."
```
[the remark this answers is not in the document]
josh146: Perhaps this came from a merge conflict at some point?

josh146 (commented):
A bit of a fly-by review, so I haven't gone through the logic in proper detail!

strawberryfields/tdm/tdmprogram.py
nquesada: yup

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,91 @@ def context(self, *args, copies=1, shift="default"):
         self.loop_vars = self.params(*[f"p{i}" for i in range(len(args))])
         return self
 
+    def compile(self, *, device=None, compiler=None, **kwargs):
+        """Compile the program given a Strawberry Fields photonic hardware device specification.
+        At this stage the compilation is simply a check that the program matches the device.
```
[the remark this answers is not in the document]
nquesada: How would you update it?

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
nquesada: why should it be changed?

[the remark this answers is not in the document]
nquesada: I did not modify those lines. Feel free to change them to whatever it should be.

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
nquesada: Removed.

strawberryfields/tdm/tdmprogram.py
nquesada: Thanks, had not notices that method!

strawberryfields/tdm/tdmprogram.py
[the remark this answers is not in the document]
nquesada: I changed it a little bit.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -410,8 +494,8 @@ def assert_number_of_modes(self, device):
             )
         if self.spatial_modes > device.modes["spatial"]:
             raise CircuitError(
-                f"This program contains {self.spatial_modes} spatial modes, but the device '{device.target}' "
-                f"only supports {device.modes['spatial']} modes."
+                f"This program contains {device.timebins} temporal modes, but the device '{device.target}' "
+                f"only supports up to {device.modes['max']['temporal']} modes."
```
thisac, line 505:
```suggestion
                f"This program contains {self.spatial_modes} spatial modes, but the device '{device.target}' "
                f"only supports {device.modes['spatial']} modes."
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
...
+            compiler (str, ~strawberryfields.compilers.Compiler): Compiler name or compile strategy
+                to use. If a device is specified, this overrides the compile strategy specified by
+                the hardware :class:`~.DeviceSpec`. For :class:`~.TDMProgram` this should be set to
+                None.
+
+        Returns:
+            Program: compiled program
+        """
+        if compiler is not None:
+            raise CircuitError("Program can only be compiled with default compiler")
```
antalszava, line 323: Just curious: would this be applicable should there be multiple tdm compilers? That is, find it a bit strange that a `compiler` can be explicitly passed atm but an error is raised if that does happen

strawberryfields/tdm/tdmprogram.py
antalszava: Maybe something like `Program does not match the physical architecture of the device`? As per the wording from here https://strawberryfields.ai/photonics/demos/tutorial_X8.html#hardware-compilation

antalszava: Just saw the next check: maybe worth referring here to the gates and their expected order in this message. Even if we manage to pass this first check, the topology/architecture can still be different if gates act on different modes as expected (second check).

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
...
+        # First check: the gates are in the correct order
+        program_gates = [cmd.op.__class__.__name__ for cmd in self.rolled_circuit]
+        device_gates = [op["op"] for op in device_layout.operations]
+        if device_gates != program_gates:
+            raise CircuitError(
+                "Program cannot be used with the device '{}' "
+                "due to incompatible topology.".format(device.target)
+            )
+
+        # Second check: the gates act in the correct modes
```
antalszava, line 336:
```suggestion
        # Second check: the gates act on the correct modes
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
...
+        # Third check: the parameters of the gates are valid
+        gate_params_ranges = device.gate_parameters
+        # We will loop over the different operations in the device specification
+
+        for i, operation in enumerate(device_layout.operations):
+            # We obtain the name of the parameter(s)
+            param_names = operation["args"]
+            len_params_program = len(self.rolled_circuit[i].op.p)
+            len_params_device = len(param_names)
+            # The next if is to make sure we do not flag incorrectly things like Sgate(r,0) beign different Sgate(r)
```
antalszava, line 354:
```suggestion
            # The next if is to make sure we do not flag incorrectly things like Sgate(r,0) being different Sgate(r)
```

strawberryfields/tdm/tdmprogram.py
antalszava: maybe worth adding this information as a comment

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
...
+
+        # Third check: the parameters of the gates are valid
+        gate_params_ranges = device.gate_parameters
+        # We will loop over the different operations in the device specification
+
+        for i, operation in enumerate(device_layout.operations):
+            # We obtain the name of the parameter(s)
+            param_names = operation["args"]
+            len_params_program = len(self.rolled_circuit[i].op.p)
+            len_params_device = len(param_names)
```
antalszava, line 353: (Minor suggestion): it can be slightly more difficult to read names where only the last word is different (`len_params_program` & `len_params_device`). How about `program_params_len` & `device_params_len`?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
...
+
+            if len_params_device < len_params_program:
+                for j in range(1, len_params_program):
+                    if self.rolled_circuit[i].op.p[j] != 0:
+                        raise CircuitError(
+                            "Program cannot be used with the device '{}' "
+                            "due to incompatible parameter.".format(device.target)
+                        )
+            # Now we will check explicitly if the parameters in the program match
+            counter = 0  # counts the number of symbolic variables, which are labeled consecutively by the context method
```
antalszava, line 364: (Minor suggestion): how about `num_symbolic_param`/`num_symbolic` as a name?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -382,7 +475,6 @@ def unroll(self):
             elif isinstance(self.shift, int):
                 q = shift_by(q, self.shift)  # shift at end of each time bin
 
-        # Unrolling the circuit for the first time: storing a copy of the unrolled circuit
```
antalszava, line 385: How come this comment was removed?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
...
+            counter = 0  # counts the number of symbolic variables, which are labeled consecutively by the context method
+
+            for k, param_name in enumerate(param_names):
+                # Obtain the relevant parameter range from the device
+                param_range = device.gate_parameters[param_name]
+                # Obtain the value of the corresponding parameter in the program
+                program_param = self.rolled_circuit[i].op.p[k]
+
+                if sf.parameters.par_is_symbolic(program_param):
+                    # If it is a symbolic value go and lookup its corresponding list in self.tdm_params
```
antalszava, line 373: Nice comments!

tests/frontend/test_tdmprogram.py
```diff
@@ -452,3 +500,157 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
 
     with pytest.raises(sf.program_utils.CircuitError, match=match):
         new_prog = prog.compile(device=spec, compiler=DummyCompiler())
+
+
+## Test for the compilation
+
+import blackbird as bb
```
antalszava, line 507: Maybe worth moving the import to the top level of the file?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,98 @@ def context(self, *args, copies=1, shift="default"):
         self.loop_vars = self.params(*[f"p{i}" for i in range(len(args))])
         return self
 
+    def compile(self, *, device=None, compiler=None):
+        """Compile the time-domain program given a Strawberry Fields photonic hardware device specification.
+        At this stage the compilation is simply a check that the program matches the device.
```
antalszava, line 309: Just curious: for general CV circuits we'd have separate compilers. Was the motivation behind adding the logic here for simplicity (only have one type of compilation for TDM & it's just a check)?

tests/frontend/test_tdmprogram.py
```diff
@@ -452,3 +500,157 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
...
+    },
+}
+device_spec["layout"] = device_spec["layout"].format(tm=tm)
+
+from strawberryfields.api.devicespec import DeviceSpec
+
+device = DeviceSpec("tdm", device_spec, connection=None)
+
+
+def test_tdm_wrong_layout():
```
antalszava, line 537: Just curious: how come no test classes were defined in this file?

tests/frontend/test_tdmprogram.py
```diff
@@ -452,3 +500,157 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
...
+    "gate_parameters": {
+        "p0": [0.5643],
+        "p1": [0, [0, 6.283185307179586]],
+        "p2": [0, [0, 3.141592653589793], 3.141592653589793],
+        "p3": [0, [0, 6.283185307179586]],
+    },
+}
+device_spec["layout"] = device_spec["layout"].format(tm=tm)
+
+from strawberryfields.api.devicespec import DeviceSpec
```
antalszava, line 532: Similarly wondering about the import here

tests/frontend/test_tdmprogram.py
```diff
@@ -452,3 +500,157 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
...
+    "modes": {"concurrent": 2, "spatial": 1, "temporal": {"max": 100}},
+    "compiler": ["tdm"],
+    "gate_parameters": {
+        "p0": [0.5643],
+        "p1": [0, [0, 6.283185307179586]],
+        "p2": [0, [0, 3.141592653589793], 3.141592653589793],
+        "p3": [0, [0, 6.283185307179586]],
+    },
+}
+device_spec["layout"] = device_spec["layout"].format(tm=tm)
```
antalszava, line 530: Could we maybe have `{target}` inside of the layout and substitute that too explicitly `device_spec["layout"] = device_spec["layout"].format(target=target, tm=tm)`? Would conform with the original layout and it would not be missed when looking here.
```suggestion

target = "tdm"
tm = 4
device_spec = {
    "layout": "name template_tdm\nversion 1.0\ntarget {target} (shots=1)\ntype tdm (temporal_modes={tm}, copies=1)\nfloat array p0[1, {tm}] =\n    {{rs_array}}\nfloat array p1[1, {tm}] =\n    {{bs_array}}\nfloat array p2[1, {tm}] =\n    {{r_array}}\nfloat array p3[1, {tm}] =\n    {{m_array}}\n\nSgate(p0) | 1\nBSgate(p1) | (1, 0)\nRgate(p2) | 1\nMeasureHomodyne(p3) | 0\n",
    "modes": {"concurrent": 2, "spatial": 1, "temporal": {"max": 100}},
    "compiler": ["tdm"],
    "gate_parameters": {
        "p0": [0.5643],
        "p1": [0, [0, 6.283185307179586]],
        "p2": [0, [0, 3.141592653589793], 3.141592653589793],
        "p3": [0, [0, 6.283185307179586]],
    },
}
device_spec["layout"] = device_spec["layout"].format(target=target, tm=tm)
```

tests/frontend/test_tdmprogram.py
```diff
@@ -452,3 +500,157 @@ class DummyCompiler(sf.compilers.compiler.Compiler):
...
+    c = 2
+    copies = 10
+    alpha = [np.pi / 4, 0] * c
+    phi = [0, np.pi / 2] * c
+    theta = [0, 0] + [np.pi / 2, np.pi / 2]
+    shift = "default"
+    prog = tdmprogram.TDMProgram(N=2)
+    with prog.context(alpha, phi, theta, copies=copies, shift=shift) as (p, q):
+        ops.Sgate(sq_r) | q[1]
+        ops.BSgate(p[0]) | (q[0], q[1])
```
antalszava, line 569: Maybe worth noting that here's the diff?

antalszava (approved):
Looks good to me! :100: Had a couple of questions/minor suggestions, but nothing major

strawberryfields/tdm/tdmprogram.py
nquesada: @josh146 might know better what is the logic behind this.

strawberryfields/tdm/tdmprogram.py
nquesada: The motivation, as far as I can tell, is that for TDM programs the only compilation that one can hope to do is to check that the program matches exactly the device. There (at least as of now) clever tricks to figure out if one can map it or not.

strawberryfields/tdm/tdmprogram.py
josh146: Not so much logic, more so I was attempting to stick to the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) 😆 I was worried that changing the signature of an inherited method would result in user error elsewhere.

But happy to remove it from the signature if we think that will be less confusing.

antalszava: Fair point, thanks for the clarification!

tests/frontend/test_tdmprogram.py
nquesada: Because I wrote them :P

strawberryfields/tdm/tdmprogram.py
nquesada: Makes complete sense. Thanks!

strawberryfields/tdm/tdmprogram.py
nquesada: Added.

strawberryfields/tdm/tdmprogram.py
nquesada: Thanks, yes, it is much more descriptive.

tests/frontend/test_tdmprogram.py
nquesada: Done.

tests/frontend/test_tdmprogram.py
nquesada: Done.

tests/frontend/test_tdmprogram.py
nquesada: Done.

tests/frontend/test_tdmprogram.py
nquesada: Now they follow the tenets of the OOP church :)

strawberryfields/tdm/tdmprogram.py
nquesada: Thanks!

strawberryfields/tdm/tdmprogram.py
thisac: Another idea (which might be completely unreasonable) would be to let the user define any compiler, and then simply call `super().compile(device=device, compiler=compiler)` and hope for the best. One thing that won't pass is the `bb.match_template` function, but that could either be fixed or skipped for `TDMPrograms`. Not sure if there are other things that this will break though. :thinking:

strawberryfields/engine.py
```diff
@@ -661,33 +661,33 @@ def run_async(
...
 
-        if program.type != "tdm":
-            if not program_is_compiled:
-                # program is not compiled
-                msg = (
-                    f"Compiling program for device {device.target} using compiler {compiler_name}."
-                )
-                self.log.info(msg)
-                program = program.compile(device=device, **compile_options)
+        #if program.type != "tdm":
```
nquesada, line 664:
```suggestion
```

strawberryfields/engine.py
```diff
@@ -661,33 +661,30 @@ def run_async(
                     f"to compile with {compiler_name}."
                 )
 
-        if program.type != "tdm":
```
antalszava, line 664: Hmm, I'm getting the following error for a remote job, if this `if` statement is removed :thinking::
```python
/strawberryfields/strawberryfields/engine.py in run(self, program, compile_options, recompile, **kwargs)
    589         """
    590         job = self.run_async(
--> 591             program, compile_options=compile_options, recompile=recompile, **kwargs
    592         )
    593         try:

/strawberryfields/strawberryfields/engine.py in run_async(self, program, compile_options, recompile, **kwargs)
    689         # update the run options if provided
    690         run_options = {}
--> 691         run_options.update(program.run_options)
    692         run_options.update(kwargs or {})
    693 

AttributeError: 'NoneType' object has no attribute 'run_options'
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,106 @@ def context(self, *args, copies=1, shift="default"):
...
+                    "due to incompatible mode ordering.".format(device.target)
+                )
+
+            # Third check: the parameters of the gates are valid
+            gate_params_ranges = device.gate_parameters
+            # We will loop over the different operations in the device specification
+
+            for i, operation in enumerate(device_layout.operations):
+                # We obtain the name of the parameter(s)
+                param_names = operation["args"]
```
thisac, line 359: `operation["args"]` will only be a list of names/strings if the parameters are named `p0`, `p1`, etc. What if we want to add a fixed gate to the device spec layout that has hard-coded parameters. Could this ever be the case? 

If so, it might be worth adding a check that this is a string and, if not, simply check if `self.rolled_circuit[i].op.p[k] == operation["args"][k]`. Otherwise, the line `param_range = device.gate_parameters[param_name]` below will raise a `KeyError`.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,106 @@ def context(self, *args, copies=1, shift="default"):
...
+
+                for k, param_name in enumerate(param_names):
+                    # Obtain the relevant parameter range from the device
+                    param_range = device.gate_parameters[param_name]
+                    # Obtain the value of the corresponding parameter in the program
+                    program_param = self.rolled_circuit[i].op.p[k]
+
+                    if sf.parameters.par_is_symbolic(program_param):
+                        # If it is a symbolic value go and lookup its corresponding list in self.tdm_params
+                        self.assert_number_of_modes(device)
```
thisac, line 382: Why is this only called if it's a symbolic parameter? Would this be best called simply at the top of this method, right under `if compiler == "td2" and device is not None:`, since we'd always like to check whether the correct number of modes are used? :thinking:

strawberryfields/tdm/tdmprogram.py
```diff
@@ -325,10 +325,10 @@ def compile(self, *, device=None, compiler=None):
         # The next two if statements need to be updated, in particular with the name of the programs
         if compiler == "tdm_one_loop" and device == "tdm_one_loop_device":
             # Do something clever
-            compiler = "tdm"  # Devious but it will work, Josh said
+            compiler = "td2"  # Devious but it will work, Josh said
```
antalszava, line 328:
```suggestion
            compiler = "TD2"  # Devious but it will work, Josh said
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -325,10 +325,10 @@ def compile(self, *, device=None, compiler=None):
...
             # Do not add a return statement
 
-        if compiler == "tdm" and device is not None:
+        if compiler == "td2" and device is not None:
```
antalszava, line 331:
```suggestion
        if compiler == "TD2" and device is not None:
```

tests/api/test_devicespec.py
```diff
@@ -79,12 +79,12 @@
             "max": 100
         }
     },
-    "compiler": ["tdm"],
+    "compiler": ["td2"],
```
antalszava, line 82:
```suggestion
    "compiler": ["TD2"],
```

tests/frontend/test_tdmprogram.py
```diff
@@ -489,12 +489,12 @@ def singleloop_program(r, alpha, phi, theta, copies):
     return prog
 
 
-target = "tdm"
+target = "td2"
```
antalszava, line 492:
```suggestion
target = "TD2"
```

antalszava (commented):
Two things to note:
* Not quite sure if the compilation specific logic is good to be performed for a time-domain device, I got an error there
* Would have to go for changing the target and compiler name to `TD2` (note the upper case letters). Started leaving suggestions, but better to just use `sed` or in file search & replace.

Other than this, looks good to me! :100:

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +304,106 @@ def context(self, *args, copies=1, shift="default"):
...
+            for i, operation in enumerate(device_layout.operations):
+                # We obtain the name of the parameter(s)
+                param_names = operation["args"]
+                program_params_len = len(self.rolled_circuit[i].op.p)
+                device_params_len = len(param_names)
+                # The next if is to make sure we do not flag incorrectly things like Sgate(r,0) being different Sgate(r)
+                # This assumes that a potential second parameter, that is not explicitly defined in the layout, is always 0.
+                if device_params_len < program_params_len:
+                    for j in range(1, program_params_len):
+                        if self.rolled_circuit[i].op.p[j] != 0:
```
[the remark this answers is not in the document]
nquesada: Yup that was the intent. Maybe I am assuming too much about the future development of SF, but I kind of like the rule implied by it.

strawberryfields/tdm/tdmprogram.py
nquesada: I see. I did not know about this feature of BB. Sounds like a good addition!

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+    "compiler": ["td2"],
+    "gate_parameters": {
+        "p0": [0.5643],
+        "p1": [0, [0, 6.283185307179586]],
+        "p2": [0, [0, 3.141592653589793], 3.141592653589793],
+        "p3": [0, [0, 6.283185307179586]],
+    },
+}
+device_spec["layout"] = device_spec["layout"].format(target=target, tm=tm)
+device = DeviceSpec("td2", device_spec, connection=None)
```
thisac, line 506:
```suggestion
device = DeviceSpec("TD2", device_spec, connection=None)
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+            ops.Dgate(sq_r) | q[1]  # Here one should have an Sgate
+            ops.BSgate(p[0]) | (q[1], q[0])
+            ops.Rgate(p[1]) | q[1]
+            ops.MeasureHomodyne(p[2]) | q[0]
+        eng = sf.Engine("gaussian")
+        with pytest.raises(
+            sf.program_utils.CircuitError,
+            match="The gates or the order of gates used in the Program",
+        ):
+            prog.compile(device=device, compiler="td2")
```
thisac, line 529:
```suggestion
            prog.compile(device=device, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        with prog.context(alpha, phi, theta, copies=copies) as (p, q):
+            ops.Sgate(sq_r) | q[1]
+            ops.BSgate(p[0]) | (q[0], q[1])  # The order should be (q[1], q[0])
+            ops.Rgate(p[1]) | q[1]
+            ops.MeasureHomodyne(p[2]) | q[0]
+        eng = sf.Engine("gaussian")
+        with pytest.raises(
+            sf.program_utils.CircuitError, match="due to incompatible mode ordering."
+        ):
+            prog.compile(device=device, compiler="td2")
```
thisac, line 549:
```suggestion
            prog.compile(device=device, compiler="TD2")
```

strawberryfields/tdm/tdmprogram.py
nquesada: You are completely right. There was direct logic testing for the number of temporal modes there. After you told about `assert_number_of_modes` I simply switched it, but it makes more sense to do the check right at the beginning. We  might have to change the test a bit since some warning will be triggered before other but that's OK.

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        """Test the correct error is raised when the tdm circuit explicit parameters are not within the allowed ranges"""
+        sq_r = 2  # This squeezing is not in the allowed range of squeezing parameters
+        c = 2
+        copies = 10
+        alpha = [np.pi / 4, 0] * c
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] + [np.pi / 2, np.pi / 2]
+        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        with pytest.raises(sf.program_utils.CircuitError, match="due to incompatible parameter."):
+            prog.compile(device=device, compiler="td2")
```
thisac, line 561:
```suggestion
            prog.compile(device=device, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        with prog.context(alpha, phi, theta, copies=copies) as (p, q):
+            ops.Sgate(sq_r, 0.4) | q[
+                1
+            ]  # Note that the Sgate has a second parameter that is non-zero
+            ops.BSgate(p[0]) | (q[1], q[0])
+            ops.Rgate(p[1]) | q[1]
+            ops.MeasureHomodyne(p[2]) | q[0]
+        eng = sf.Engine("gaussian")
+        with pytest.raises(sf.program_utils.CircuitError, match="due to incompatible parameter."):
+            prog.compile(device=device, compiler="td2")
```
thisac, line 581:
```suggestion
            prog.compile(device=device, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        """Test the correct error is raised when the tdm circuit symbolic parameters are not within the allowed ranges"""
+        sq_r = 0.5643
+        c = 2
+        copies = 200
+        alpha = [137, 0] * c  # Note that alpha is outside the allowed range
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] + [np.pi / 2, np.pi / 2]
+        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        with pytest.raises(sf.program_utils.CircuitError, match="due to incompatible parameter."):
+            prog.compile(device=device, compiler="td2")
```
thisac, line 593:
```suggestion
            prog.compile(device=device, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        """Test the correct error is raised when the tdm circuit has way too many temporal modes"""
+        sq_r = 0.5643
+        c = 100  # Note that we are requesting more temporal modes (2*c = 200) than what is allowed.
+        copies = 1
+        alpha = [0.5, 0] * c
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] * c
+        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        with pytest.raises(sf.program_utils.CircuitError, match="temporal modes, but the device"):
+            prog.compile(device=device, compiler="td2")
```
thisac, line 605:
```suggestion
            prog.compile(device=device, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        device1 = DeviceSpec("x", device_spec1, connection=None)
+        c = 1
+        copies = 1
+        sq_r = 0.5643
+        alpha = [0.5, 0] * c
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] * c
+        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        with pytest.raises(sf.program_utils.CircuitError, match="concurrent modes, but the device"):
+            prog.compile(device=device1, compiler="td2")
```
thisac, line 621:
```suggestion
            prog.compile(device=device1, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        device1 = DeviceSpec("x", device_spec1, connection=None)
+        c = 1
+        copies = 1
+        sq_r = 0.5643
+        alpha = [0.5, 0] * c
+        phi = [0, np.pi / 2] * c
+        theta = [0, 0] * c
+        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        with pytest.raises(sf.program_utils.CircuitError, match="spatial modes, but the device"):
+            prog.compile(device=device1, compiler="td2")
```
thisac, line 637:
```suggestion
            prog.compile(device=device1, compiler="TD2")
```

tests/frontend/test_tdmprogram.py
```diff
@@ -407,48 +467,171 @@ def test_two_dimensional_cluster_tokyo():
...
+        ops.MeasureHomodyne(p[2]) | q[0]
+    return prog
+
+
+target = "td2"
+tm = 4
+device_spec = {
+    "layout": "name template_tdm\nversion 1.0\ntarget {target} (shots=1)\ntype tdm (temporal_modes={tm}, copies=1)\nfloat array p0[1, {tm}] =\n    {{rs_array}}\nfloat array p1[1, {tm}] =\n    {{bs_array}}\nfloat array p2[1, {tm}] =\n    {{r_array}}\nfloat array p3[1, {tm}] =\n    {{m_array}}\n\nSgate(p0) | 1\nBSgate(p1) | (1, 0)\nRgate(p2) | 1\nMeasureHomodyne(p3) | 0\n",
+    "modes": {"concurrent": 2, "spatial": 1, "temporal": {"max": 100}},
+    "compiler": ["td2"],
```
thisac, line 497:
```suggestion
    "compiler": ["TD2"],
```

strawberryfields/tdm/tdmprogram.py
thisac: It's more that the `operation["args"]` have always strictly returned values before. It's just the TDM addition that added the ability to let them be parameters (named `p0`, `p1`, etc.), and thus letting them be names/strings instead.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -360,7 +359,7 @@ def compile(self, *, device=None, compiler=None):
                 program_params_len = len(self.rolled_circuit[i].op.p)
                 device_params_len = len(param_names)
                 # The next if is to make sure we do not flag incorrectly things like Sgate(r,0) being different Sgate(r)
-                # This assumes that a potential second parameter, that is not explicitly defined in the layout, is always 0.
+                # This assumes that parameters other than the first one are zero if not explicitly stated.
```
thisac, line 362: :100:

strawberryfields/engine.py
thisac: This was two-fold: `TDMProgram.compile` wasn't returning the compiled program (which it doesn't really need to since it changes nothing), but for compatibility I set it to return `self`; and also there was no `TDMprogram.run_options` dict, which I also added since it actually is needed both here and for the Blackbird to TDMProgram conversion.

thisac (approved):
Thanks @nquesada! :100:

strawberryfields/tdm/tdmprogram.py
```diff
@@ -303,6 +310,121 @@ def context(self, *args, copies=1, shift="default"):
...
+                    if not isinstance(param_name, str):
+                        if not program_param == param_name:
+                            raise CircuitError(
+                                "Program cannot be used with the device '{}' "
+                                "due to incompatible parameter. Parameter has value '{}'"
+                                "while its valid value is '{}'".format(
+                                    device.target, program_param, param_name
+                                )
+                            )
+                        continue
```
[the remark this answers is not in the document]
nquesada: Amazing! Thanks much!

PR #487, opened 2020-11-11
Rearrange qchem utility functions

soranjh:
This PR moves the `prob` and `marginals` functions from `apps.qchem.dynamics` to `apps.qchem.utils`. These functions were initially created as utility functions for simulating vibrational dynamics of molecules. However, they can also be used in other applications such as simulating vibrational excitations and therefore should be moved to the `utils` module which is created for hosting general-purpose utility functions. The arrangement of the `qchem` functions in this PR is shown in the image below. 

qchem


josh146:
@soranjh does this require any changes to existing SF demos?

trbromley (approved):
Thanks @soranjh! I agree we should check if there are any corresponding tutorials, but I couldn't see any.

soranjh:
> @soranjh does this require any changes to existing SF demos?

Hi @Josh. No, the change does not affect any demos.

ixfoduap (approved):

PR #489, opened 2020-11-13
Remove copies from TDMProgram and use shots

thisac:
**Context:**
Copies are removed from `TDMProgram` and shots are inserted in the unrolling via `engine.run` instead.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/engine.py
```diff
@@ -420,7 +420,10 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
 
         if not isinstance(program, collections.abc.Sequence) and program.type == "tdm":
             # At this time we do not support lists of tdm programs
-            program.unroll()
+            shots = kwargs.get("shots", 1)
+            program.unroll(shots=shots)
+            # set shots to 1 so that the gaussian backend works
```
nquesada, line 425: Great that you added this comment!

josh146: Good catch! I have to admit, even with this comment I was still confused for a second 😆 I suggest adding a few more details:

```suggestion
            # Shots >1 for a TDM program simply corresponds to creating
            # multiple copies of the program, and appending them to run sequentially.
            # As a result, we set the backend shots to 1 for the Gaussian backend.
```

strawberryfields/engine.py
```diff
@@ -476,9 +479,9 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
         if not isinstance(program, collections.abc.Sequence) and program.type == "tdm":
             # At this time we do not support lists of tdm programs
             result._all_samples = reshape_samples(
-                result.all_samples, program.measured_modes, program.N
+                result.all_samples, program.measured_modes, program.N, program.timebins
             )
-            result._samples = np.array(list(result.all_samples.values()))
+            result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
```
josh146, line 484: How come this transposition is needed? Can it simply be absorbed into the `reshape_samples` function?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -125,21 +121,32 @@ def reshape_samples(all_samples, modes, N):
         all_samples (dict[int, list]): the raw measured samples
         modes (Sequence[int]): the modes that are measured in the circuit
         N (Sequence[int]): the number of concurrent modes per belt/spatial modes
+        timebins (int): the number of timebins/temporal modes in the program per shot
 
     Returns:
-        new_samples (dict[int, list]): the re-shaped samples
+        new_samples (dict[int, list]): the re-shaped samples, where each key correspond to a spatial
+            mode and the values have shape `(shots, timebins)`
```
josh146, line 128:
```suggestion
            mode and the values have shape ``(shots, timebins)``
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -244,12 +251,6 @@ class TDMProgram(sf.Program):
       If the ``p`` variable is not used, the gate argument is assumed to be **constant**
       across all time-bins.
 
-    * ``copies=1`` *(int)*: the number of times to repeat the time-domain program.
-      For example, if a sequence of 10 gate arguments is provided, and ``copies==15``,
-      then there will be a total of :math:`10\times 15` time bins in the time domain
-      program. The sequence corresponding to the gate arguments is repeated ``copies``
-      number of times.
```
josh146, line 251: :+1: 🗑️

strawberryfields/tdm/tdmprogram.py
```diff
@@ -447,10 +445,10 @@ def roll(self):
         self.circuit = self.rolled_circuit
         return self
 
-    def unroll(self):
+    def unroll(self, shots=1):
```
josh146, line 448: Out of curiosity, how come a default value was added here for shots? Does it make more sense to have `unroll(self, shots)`, so that shots always have to be provided?

tests/frontend/test_tdmprogram.py
```diff
@@ -27,93 +27,81 @@
...
     prog = tdmprogram.TDMProgram(N=2)
-    with prog.context(alpha, phi, theta, copies=copies, shift=shift) as (p, q):
+    with prog.context(alpha, phi, theta, shift=shift) as (p, q):
         ops.Sgate(r, 0) | q[1]
         ops.BSgate(p[0]) | (q[0], q[1])
         ops.Rgate(p[1]) | q[1]
         ops.MeasureHomodyne(p[2]) | q[0]
     eng = sf.Engine("gaussian")
-    result = eng.run(prog)
+    result = eng.run(prog, shots=shots)
```
josh146, line 50: Awesome!

josh146 (approved):
Left some minor comments, but happy to approve! We shouldn't forget to update the tutorial over the photonics repo, however.

P.S. changelog 😉

josh146 (changes requested):
Actually, I just realised that Travis is not being run - I'm not sure I am comfortable approving without confirmation that all tests pass. We should port the tests to github actions ASAP.

nquesada:
@josh146 : this anyway cannot be merged yet since I still need to modify the tests so that they take into account the new shape of the samples.

nquesada:
Also, once this is merged we will need to update the tutorial.

strawberryfields/tdm/tdmprogram.py
thisac: I think I was just a bit quick adding this here. It makes sense letting it always having to be provided. I'll fix it!

strawberryfields/engine.py
thisac: It's because of the shape change we decided to have for both `all_samples`, `{spatial mode: (shots, timebins)}`, and `samples`, `(shots, spatial modes, timebins)`, along with that `reshape_samples` only actually reshapes and returns `all_samples`. `samples` is created here, outside of the `reshape_samples` function, although I guess this could be move inside of the `reshape_samples` instead, if we want, so that it returns the tuple `(all_samples, samples)` instead.

josh146 (approved):
Looks good from my end! Small question though - should there be a unittest for `move_vac_modes`? I couldn't spot one.

strawberryfields/engine.py
```diff
@@ -420,7 +420,12 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
 
         if not isinstance(program, collections.abc.Sequence) and program.type == "tdm":
             # At this time we do not support lists of tdm programs
-            program.unroll()
+            shots = kwargs.get("shots", 1)
+            program.unroll(shots=shots)
+            # Shots >1 for a TDM program simply corresponds to creating
+            # multiple copies of the program, and appending them to run sequentially.
+            # As a result, we set the backend shots to 1 for the Gaussian backend.
+            kwargs["shots"] = 1
```
antalszava, line 428: :100:

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,136 @@ def validate_measurements(circuit, N):
...
+    dictionary with the shape ``{spatial mode: (shots, timebins)}``. E.g., this unrolled circuit:
+
+    .. code-block:: pycon
+
+        ...
+        MeasureHomodyne(0) | (q[0])  # shot 0, timebin 0, spatial mode 0  (sample 0)
+        MeasureHomodyne(0) | (q[2])  # shot 0, timebin 0, spatial mode 2  (sample 1)
+        ...
+        MeasureHomodyne(0) | (q[0])  # shot 0, timebin 1, spatial mode 0  (sample 2)
+        MeasureHomodyne(0) | (q[1])  # shot 0, timebin 1, spatial mode 2  (sample 3)
```
antalszava, line 130: One question here (think I'm personally just missing something): should the two target modes here be swapped for the measurements? :thinking: 

I'd think that we're shifting with one mode in timebin 1:
1. `MeasureHomodyne(0) | (q[0])` -> `MeasureHomodyne(0) | (q[1])` (spatial mode 0) 
2. `MeasureHomodyne(0) | (q[2])` -> `MeasureHomodyne(0) | (q[0])` (spatial mode 2)

And would have expected:
```
        MeasureHomodyne(0) | (q[1])  # shot 0, timebin 1, spatial mode 0  (sample 2)
        MeasureHomodyne(0) | (q[0])  # shot 0, timebin 1, spatial mode 2  (sample 3)
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,136 @@ def validate_measurements(circuit, N):
...
     if len(set(param_lengths)) != 1:
         raise ValueError("Gate-parameter lists must be of equal length.")
 
-    # check copies
-    if not isinstance(copies, int) or copies < 1:
-        raise TypeError("Number of copies must be a positive integer.")
-
 
-def _get_mode_order(num_of_values, N):
+def _get_mode_order(num_of_values, N, timebins):
```
antalszava, line 99: Would we eventually go with the `timebins` naming in SF (opposed to `temporal_modes`)?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,136 @@ def validate_measurements(circuit, N):
...
     """Get the order by which the modes were measured"""
```
antalszava, line 100: (Out of scope:) I find that it would be nice to have an example here and add the arguments, return values to the docstring.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,136 @@ def validate_measurements(circuit, N):
...
 
-def reshape_samples(all_samples, modes, N):
+def reshape_samples(all_samples, modes, N, timebins):
     """Reshapes the samples dict so that they have the expected correct shape.
 
     Corrects the :attr:`~.Results.all_samples` dictionary so that the measured modes are the ones
     defined to be measured in the circuit, instead of being spread over a larger
     number of modes due to the mode-shifting occurring in :class:`~.TDMProgram`.
 
+    The function iterates through the unrolled circuit and populates, and returns, a new samples
```
antalszava, line 120:
```suggestion
    The function iterates through samples obtained from the unrolled circuit to populate and return a new samples
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,136 @@ def validate_measurements(circuit, N):
...
+        MeasureHomodyne(0) | (q[0])  # shot 0, timebin 1, spatial mode 0  (sample 2)
+        MeasureHomodyne(0) | (q[1])  # shot 0, timebin 1, spatial mode 2  (sample 3)
+        ...
+        MeasureHomodyne(0) | (q[0])  # shot 1, timebin 0, spatial mode 0  (sample 4)
+        MeasureHomodyne(0) | (q[2])  # shot 1, timebin 0, spatial mode 2  (sample 5)
+        ...
+        MeasureHomodyne(0) | (q[0])  # shot 1, timebin 1, spatial mode 0  (sample 6)
+        MeasureHomodyne(0) | (q[1])  # shot 1, timebin 1, spatial mode 2  (sample 7)
+
+    would return the dictionary
```
antalszava, line 138: How about including the equivalent `all_samples` that is produced by the above circuit?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,136 @@ def validate_measurements(circuit, N):
...
     """
+    # calculate the total number of samples and the order in which they were measured
     num_of_values = len([i for j in all_samples.values() for i in j])
-    mode_order = _get_mode_order(num_of_values, N)
+    mode_order = _get_mode_order(num_of_values, N, timebins)
 
-    # go backwards through all_samples and add them into the correct mode
+    # iterate backwards through all_samples and add them into the correct mode
     new_samples = dict()
+    t = 0
```
antalszava, line 165: This might make it more descriptive
```suggestion
    timebin_idx = 0
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
         if idx not in new_samples:
-            new_samples[idx] = []
-        new_samples[idx].append(all_samples[m].pop(0))
-    return new_samples
+            new_samples[idx] = [[] for _ in range(timebins)]
+        new_samples[idx][t].append(all_samples[m].pop(0)[0])
+
+        # populate each spatial mode in one timebin before moving on to the next timebin
+        # when each timebin has been filled, move to the next shot
+        if (i + 1) % len(N) == 0:
```
antalszava, line 174: (Minor suggestion): naming the condition could make the operation more descriptive
```suggestion
        # when each timebin has been filled, move to the next shot
        last_mode_in_timebin = (i + 1) % len(N) == 0
        if last_mode_in_timebin:
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
+    # iterate backwards through all_samples and add them into the correct mode
     new_samples = dict()
+    t = 0
     for i, m in enumerate(mode_order):
         idx = modes[i % len(N)]
         if idx not in new_samples:
-            new_samples[idx] = []
-        new_samples[idx].append(all_samples[m].pop(0))
-    return new_samples
+            new_samples[idx] = [[] for _ in range(timebins)]
```
antalszava, line 169:
```suggestion
            # create an entry for the new mode with a nested list for each timebin
            new_samples[idx] = [[] for _ in range(timebins)]
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
     num_of_values = len([i for j in all_samples.values() for i in j])
-    mode_order = _get_mode_order(num_of_values, N)
+    mode_order = _get_mode_order(num_of_values, N, timebins)
 
-    # go backwards through all_samples and add them into the correct mode
+    # iterate backwards through all_samples and add them into the correct mode
     new_samples = dict()
+    t = 0
     for i, m in enumerate(mode_order):
         idx = modes[i % len(N)]
```
antalszava, line 167: I think it would be nice to have this be called `mode_idx`, would make the rest of the function more readable imho.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
+    # calculate the total number of samples and the order in which they were measured
     num_of_values = len([i for j in all_samples.values() for i in j])
-    mode_order = _get_mode_order(num_of_values, N)
+    mode_order = _get_mode_order(num_of_values, N, timebins)
 
-    # go backwards through all_samples and add them into the correct mode
+    # iterate backwards through all_samples and add them into the correct mode
     new_samples = dict()
+    t = 0
     for i, m in enumerate(mode_order):
```
antalszava, line 166: (Minor suggestion)
```suggestion
    for i, mode in enumerate(mode_order):
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
-    return new_samples
+            new_samples[idx] = [[] for _ in range(timebins)]
+        new_samples[idx][t].append(all_samples[m].pop(0)[0])
+
+        # populate each spatial mode in one timebin before moving on to the next timebin
+        # when each timebin has been filled, move to the next shot
+        if (i + 1) % len(N) == 0:
+            t = (t + 1) % timebins
+
+    # transpose each value so that it has shape `(shots, timebins)`
```
antalszava, line 177: Could the mode order creation be adjusted so that transposition would not be needed here? (Such that `new_samples` is populated `(shots, timebins)` by default, might not be a simple change though so feel free to disregard.)

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
+    return {k: np.array(v).T for k, v in new_samples.items()}
+
+
+def move_vac_modes(samples, N, crop=False):
+    """Post-processing function for TDM samples
+
+    Moves all measured vacuum modes from the first shot of the samples
+    array to the end of the last shot.
```
antalszava, line 185:
```suggestion
    """Post-processing function for TDM samples, moves all measured vacuum modes from the first shot of the samples
    array to the end of the last shot.
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
+
+    Args:
+        samples (ndarray[float]): samples as received from TDMProgram, with the
```
antalszava, line 188:
```suggestion
        samples (ndarray[float]): samples as received from ``TDMProgram``, with the
```

strawberryfields/tdm/tdmprogram.py
thisac: So, in this case this would correspond to `N = [1, 2]`, i.e. one timebin in the first band and 2 timebins in the second. Since the first measurement is in the first band (which only has timebin 0), it stays there:

`MeasureHomodyne(0) | (q[0]) -> MeasureHomodyne(0) | (q[0])`

The second measurement would start on band 2 (which has timebins 1 and 2), and would loop around to timebin 1:

`MeasureHomodyne(0) | (q[2]) -> MeasureHomodyne(0) | (q[1])`

You can check it by running the following:

```pycon
>>> with prog.context([1,2],[4,5]) as (p, q): 
...     BSgate(p[0]) | (q[0], q[1]) 
...     Rgate(p[1]) | q[1] 
...     MeasureHomodyne(p[0]) | q[0] 
...     MeasureHomodyne(p[1]) | q[2] 

>>> shots = 2 
>>> result = eng.run(prog, shots=shots)                                    

>>> for c in prog.unrolled_circuit: print(c)
                            
BSgate(1, 0) | (q[0], q[1])
Rgate(4) | (q[1])
MeasureHomodyne(1) | (q[0])
MeasureHomodyne(4) | (q[2])
BSgate(2, 0) | (q[0], q[2])
Rgate(5) | (q[2])
MeasureHomodyne(2) | (q[0])
MeasureHomodyne(5) | (q[1])
```

But this is just one example of the program, and this docstring example doesn't actually tell which circuit we're running (since it's only concerned about the reshaping). :slightly_smiling_face: 

The `...`s would be the other gates in the circuit, that also aren't important for the reshaping.

strawberryfields/tdm/tdmprogram.py
thisac: Don't think we decided upon anything. I feel like, at least, timebins is what is best used inside of `tdmprogram.py` since it's been using that all along, while temporal modes have been used for Blackbird and thus `sf.io` and the device spec also. I think it's fine keeping it like this for, but I guess this is still up for discussion since it might be a bit confusing for users.

strawberryfields/tdm/tdmprogram.py
thisac: What do you mean? The dictionary in the example would be the `all_samples` that's returned by the above circuit (only with `(sample X)` instead of the actual sample value. Or do you mean the earlier `all_samples` that is passed to the `reshape_samples` function? :thinking:

strawberryfields/tdm/tdmprogram.py
thisac: It potentially could, although it wasn't as obvious to me how to do this. I think this is fine, since it would probably take quite a bit of fixing to get it to work the other way (if that's possible without extra overhead anyway at all that is).

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
+    Returns:
+        ndarray[float]: the post-processed samples
+
+    """
+    num_of_vac_modes = np.max(N) - 1
+    shape = samples.shape
+
+    flat_samples = np.ravel(samples)
+    samples = np.append(flat_samples[num_of_vac_modes:], [0] * num_of_vac_modes)
+    samples = samples.reshape(shape)
```
antalszava, line 209: Wondering if this could be done without having to ravel/reshape the entire samples array :thinking: Maybe using two indices could work? (There wasn't a good alternative that came to mind though. :confused: )

strawberryfields/tdm/tdmprogram.py
```diff
@@ -83,63 +84,134 @@ def validate_measurements(circuit, N):
...
+
+    if crop:
+        samples = samples[: -num_of_vac_modes // (np.prod(shape[1:]) + 1)]
```
antalszava, line 212: Would be nice to break this down into steps (define new variables with descriptive names).

tests/frontend/test_tdmprogram.py
```diff
@@ -146,23 +135,29 @@ def test_epr(self):
...
+        sq_r = 5.0
         c = 2
-        copies = 200
+        shots = 100
 
         # This will generate c EPRstates per copy. I chose c = 4 because it allows us to make 4 EPR pairs per copy that can each be measured in different basis permutations.
-        alpha = [np.pi / 4, 0] * c
-        phi = [0, np.pi / 2] * c
+        alpha = [0, np.pi / 4] * c
+        phi = [np.pi / 2, 0] * c
```
antalszava, line 145: How come these were changed?

strawberryfields/tdm/tdmprogram.py
thisac: I don't think that this is such a big issue though since (correct me if I'm wrong) I believe `np.ravel` and `.reshape` only changes references and doesn't make any copies if the argument is already a NumPy array. So there's quite little overhead even for large sample arrays. Not sure about the appending though.

tests/frontend/test_tdmprogram.py
thisac: @nquesada :question:

strawberryfields/engine.py
antalszava: Worth leaving a comment on this.

antalszava (commented):
This looks like a nice change! :100:  Had some comments/questions. Mostly think that further comments and adjustments can be made in the quality of the code, but overall looked good to me.

How would `blackbird` be affected? `copies` is an optional field if I remember correctly, so it would be a "nice to have" adjustment to remove it there?

strawberryfields/tdm/tdmprogram.py
thisac: I would say it's fine leaving this as it is, since the mode order wouldn't be something that visible to the user. It's simply a hidden helper function collecting which modes that were measured in the simulator (nothing to do with the actual modes being returned).

thisac:
> How would blackbird be affected? copies is an optional field if I remember correctly, so it would be a "nice to have" adjustment to remove it there?

@antalszava There's nothing to be changed in Blackbird at all. It doesn't know about `copies`; it simply accepts any keywords. The only changes would be to remove `copies` from Blackbird scripts/templates that have it (but it doesn't harm leaving them either).

strawberryfields/tdm/tdmprogram.py
antalszava: Oh got ya, thank you!

When `shots=2`, got the following error though:
```
~/xanadu/strawberryfields/strawberryfields/engine.py in _run_program(self, prog, **kwargs)
    371                         cmd.op, self.backend, kwargs
    372                     )
--> 373                 ) from None
    374 
    375         samples = self._combine_and_sort_samples(samples_dict)

NotImplementedError: The operation MeasureHomodyne(1) has not been implemented in GaussianBackend for the arguments {'shots': 2}.
```

strawberryfields/tdm/tdmprogram.py
antalszava: Think this will be confusing in the future in the code base.

strawberryfields/tdm/tdmprogram.py
antalszava: Right. Although docstrings are not written entirely for users, I personally found it pretty difficult to parse this function.

strawberryfields/tdm/tdmprogram.py
thisac: Are you running it on the correct branch? That error shouldn't be happening (since, for a TDMProgram, it should set shots to 1 before moving on after unrolling the circuit in `eng.run`).

strawberryfields/tdm/tdmprogram.py
antalszava: > do you mean the earlier all_samples that is passed to the reshape_samples function? thinking

Yes, this one.

strawberryfields/tdm/tdmprogram.py
antalszava: Hmm, I think you might be right! Checked on an array and there doesn't seem to be considerable overhead added by the ravel/reshape part :+1:

strawberryfields/tdm/tdmprogram.py
antalszava: You're right, was my bad! Got confused with branch names

tests/frontend/test_tdmprogram.py
```diff
@@ -640,11 +646,42 @@ def test_tdm_inconsistent_spatial_modes(self):
...
-        prog = singleloop_program(sq_r, alpha, phi, theta, copies)
+        prog = singleloop_program(sq_r, alpha, phi, theta)
         with pytest.raises(sf.program_utils.CircuitError, match="spatial modes, but the device"):
             prog.compile(device=device1, compiler="TD2")
+
+class TestTDMProgramFunctions:
+    """Test functions in the tdmprogram module"""
+
+    @pytest.mark.parametrize("N, crop, expected", [
+        (
```
antalszava, line 661: Worth adding an `N=1`, `crop=True` case to cover the edge case found there.

antalszava (approved):
Looks good to me! Curious about the parameter changes in the tests, but could be addressed in the TDM HW connection PR too.

PR #493, opened 2020-11-26
Fix result representation for TDM programs

thisac:
**Context:**
Currently the representation of the `Result` object doesn't work for TDM programs due to the samples having dimension 3 (shots, spatial modes, timebins) instead of 2.

Quick fix that assumes that it's a TDM program if the samples dim is 3, and thus prints the correct representation.

**Benefits:**
Just writing `eng.run(tdm_prog)` or `res = eng.run(tdm_prog); print(res)` works, and actually prints something useful.

**Possible Drawbacks:**
Might not be the nicest solution, but since `Result` doesn't know what type of program the samples are from, I think this works well for now. Might need to be revisited later with some further changes to the `Result` class (incorporating e.g. where the results are from).

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/api/result.py
+++ b/strawberryfields/api/result.py
@@ -86,7 +86,7 @@ class Result:
 
         Returns:
             dict[int, list]: mode index associated with the list of
-                measurement outcomes
+            measurement outcomes
         """
         return self._all_samples
 
@@ -116,15 +116,18 @@ class Result:
 
     def __repr__(self):
         """String representation."""
-        try:
+        if self.samples.ndim == 2:
+            # if samples has dim 2, assume they're GBS
             shots, modes = self.samples.shape
-        except ValueError:
-            # if samples has dim 3, then we assume they're from a TDMProgram
+            return "<Result: shots={}, num_modes={}, contains state={}>".format(
+                shots, modes, self._is_stateful
+            )
+
+        if self.samples.ndim == 3:
+            # if samples has dim 3, assume they're TDM
             shots, modes, timebins = self.samples.shape
-            return "<Result: shots={}, spatial_modes={}, timebins={} contains state={}>".format(
+            return "<Result: shots={}, spatial_modes={}, timebins={}, contains state={}>".format(
                 shots, modes, timebins, self._is_stateful
             )
 
-        return "<Result: shots={}, num_modes={}, contains state={}>".format(
-            shots, modes, self._is_stateful
-        )
+        return "<Result: contains state={}>".format(self._is_stateful)
--- a/strawberryfields/api/result.py
+++ b/strawberryfields/api/result.py
@@ -119,7 +119,7 @@ class Result:
         try:
             shots, modes = self.samples.shape
         except ValueError:
-            # if the samples has dim 3, then they're from a TDMProgram
+            # if samples has dim 3, then we assume they're from a TDMProgram
             shots, modes, timebins = self.samples.shape
             return "<Result: shots={}, spatial_modes={}, timebins={} contains state={}>".format(
                 shots, modes, timebins, self._is_stateful
--- a/strawberryfields/api/result.py
+++ b/strawberryfields/api/result.py
@@ -44,9 +44,7 @@ class Result:
     >>> eng = sf.Engine("gaussian")
     >>> results = eng.run(prog)
     >>> print(results)
-    Result: 3 subsystems
-        state: <GaussianState: num_modes=3, pure=True, hbar=2>
-        samples: [[0, 0, 0]]
+    <Result: shots=1, num_modes=3, contains state=True>
     >>> results.samples
     np.array([[0, 0, 0]])
     >>> results.state.is_pure()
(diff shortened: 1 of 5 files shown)


strawberryfields/api/result.py
```diff
@@ -118,7 +118,15 @@ def state(self):
 
     def __repr__(self):
         """String representation."""
-        shots, modes = self.samples.shape
+        try:
+            shots, modes = self.samples.shape
+        except ValueError:
+            # if the samples has dim 3, then they're from a TDMProgram
```
antalszava, line 124:
```suggestion
            # if samples has dim 3, then we assume they're from a TDMProgram
```

strawberryfields/api/result.py
```diff
@@ -118,7 +118,15 @@ def state(self):
...
+            shots, modes, timebins = self.samples.shape
+            return "<Result: spatial_modes={}, shots={}, timebins={} contains state={}>".format(
+                modes, shots, timebins, self._is_stateful
```
antalszava, line 127: Think this would be good to keep the order from the shape. Let me know if the change in order was deliberate though!
```suggestion
            return "<Result: shots={}, spatial_modes={}, timebins={} contains state={}>".format(
                shots, modes, timebins, self._is_stateful
```

strawberryfields/api/result.py
```diff
@@ -118,7 +118,15 @@ def state(self):
 
     def __repr__(self):
         """String representation."""
-        shots, modes = self.samples.shape
+        try:
+            shots, modes = self.samples.shape
+        except ValueError:
```
antalszava, line 123: Think this would work fine for now actually! :100: Would be great to add a test though.

antalszava (commented):
Hi @thisac the fix is looking great! A test might be nice and then I think it should be good to go.

strawberryfields/api/result.py
thisac: Thanks @antalszava. I've added a test now! You were a bit too quick (and I maybe tagged you for a review a bit too quickly). :laughing:

strawberryfields/api/result.py
thisac: I also think this looks nicer to be honest. I just kept the same form as the current repr (for GBS samples) which starts with modes, and then shots. I might just update it in both places if you also think that makes more sense.

tests/api/test_result.py
```diff
@@ -54,3 +54,14 @@ def test_state_print(self, capfd):
...
         assert "shots=3" in out
         assert "contains state=True" in out
+
+    def test_tdm_print(self, capfd):
+        """Test that printing a result object with TDM samples provides the correct output."""
+        samples = np.ones((2, 3, 4))
+        result = Result(samples)
+        print(result)
+        out, err = capfd.readouterr()
+        assert "spatial_modes=3" in out
```
antalszava, line 64: :100: Might be cool to check the order (not too important though).

strawberryfields/api/result.py
antalszava: Hmm, think it's good to keep the shape order. As for changing for GBS, not quite sure if that would cause any troubles somewhere (don't think so) :thinking:.

antalszava (approved):
Looks good to me! :100:

tests/api/test_result.py
thisac: I think this is fine (since the order isn't actually that important I would say).

josh146 (changes requested):
Thanks @thisac - just a small comment regarding the existing try-except logic

strawberryfields/api/result.py
```diff
@@ -44,9 +44,7 @@ class Result:
     >>> eng = sf.Engine("gaussian")
     >>> results = eng.run(prog)
     >>> print(results)
-    Result: 3 subsystems
-        state: <GaussianState: num_modes=3, pure=True, hbar=2>
-        samples: [[0, 0, 0]]
+    <Result: shots=1, num_modes=3, contains state=True>
```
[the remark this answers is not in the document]
thisac: Yes it was, not sure why.

strawberryfields/api/result.py
```diff
@@ -118,7 +116,15 @@ def state(self):
 
     def __repr__(self):
         """String representation."""
-        shots, modes = self.samples.shape
-        return "<Result: num_modes={}, shots={}, contains state={}>".format(
-            modes, shots, self._is_stateful
+        try:
+            shots, modes = self.samples.shape
+        except ValueError:
```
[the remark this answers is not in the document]
thisac: Thanks @josh146. I think this is a good solution as well. Updated it to follow your logic above and added a test.

strawberryfields/engine.py
```diff
@@ -598,7 +598,7 @@ def run(
 
         Returns:
             [strawberryfields.api.Result, None]: the job result if successful, and
-            ``None`` otherwise
+                ``None`` otherwise
```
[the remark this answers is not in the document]
thisac: Oh, I didn't actually know. Fixed now!

strawberryfields/api/result.py
```diff
@@ -118,7 +116,18 @@ def state(self):
 
     def __repr__(self):
         """String representation."""
-        shots, modes = self.samples.shape
-        return "<Result: num_modes={}, shots={}, contains state={}>".format(
-            modes, shots, self._is_stateful
-        )
+        if self.samples.ndim == 2:
+            # if samples has dim 2, assume they're GBS
```
josh146, line 120:
```suggestion
            # if samples has dim 2, assume they're from a standard Program
```
Might not be too accurate to say 'GBS' here, since this applies to _any_ executed `sf.Program` object.

tests/api/test_result.py
```diff
@@ -54,3 +54,29 @@ def test_state_print(self, capfd):
...
+        assert "shots=2" in out
+        assert "timebins=4" in out
+        assert f"contains state={stateful}" in out
+
+    @pytest.mark.parametrize("stateful", [True, False])
+    def test_unkown_shape_print(self, stateful, capfd):
+        """Test that printing a result object with samples with an unknown shape
+        provides the correct output."""
```
josh146, line 73: Nice!

josh146 (approved):
Looks great! 💯

PR #496, opened 2020-12-02
Add program code generator function

thisac:
**Context:**
The convert-to-python-code functionality in SF interactive should be moved to SF.

Added a `generate_code` function to `io.py` that generates SF code from a `Program` or a `TDMProgram`, returning the code as a string. If an engine is passed, the engine-related code is also serialized.

**Benefits:**
It's possible to convert a program/engine into code directly via SF.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -255,24 +255,7 @@ def generate_code(prog, eng=None):
     Returns:
         str: the Strawberry Fields code, for constructing the program, as a string
     """
-    code_seq = ["import strawberryfields as sf"]
-
-    operations = []
-    op_imports = []
-    for cmd in prog.circuit:
-        op_name = cmd.op.__class__.__name__
-        if op_name not in op_imports:
-            op_imports.append(op_name)
-
-        if prog.type == "tdm":
-            format_dict = {k: f"p[{k[1:]}]" for k in prog.parameters.keys()}
-            op = cmd.__str__().format(**format_dict)
-        else:
-            op = cmd.__str__()
-
-        operations.append(op)
-
-    code_seq.append("from strawberryfields.ops import " + ", ".join(op_imports) + "\n")
+    code_seq = ["import strawberryfields as sf", "from strawberryfields import ops\n"]
 
     if prog.type == "tdm":
         code_seq.append(f"prog = sf.TDMProgram(N={prog.N})")
@@ -282,25 +265,41 @@ def generate_code(prog, eng=None):
     if eng:
         eng_type = eng.__class__.__name__
         if eng_type == "RemoteEngine":
-            code_seq.append(f"eng = sf.RemoteEngine({eng.target})")
+            code_seq.append(f'eng = sf.RemoteEngine("{eng.target}")')
         else:
             if eng.backend_options:
                 code_seq.append(
-                    f"eng = sf.Engine({eng.backend_name}, backend_options={eng.backend_options})\n"
+                    f'eng = sf.Engine("{eng.backend_name}", backend_options={eng.backend_options})'
                 )
             else:
-                code_seq.append(f"eng = sf.Engine({eng.backend_name})\n")
+                code_seq.append(f'eng = sf.Engine("{eng.backend_name}")')
 
     if prog.type == "tdm":
         tdm_params = [f"{par}" for par in prog.tdm_params]
-        code_seq.append("with prog.context(" + ", ".join(tdm_params) + ") as (p, q):")
+        code_seq.append("\nwith prog.context(" + ", ".join(tdm_params) + ") as (p, q):")
     else:
-        code_seq.append("with prog.context as q:")
+        code_seq.append("\nwith prog.context as q:")
 
-    code_seq.append("    " + "\n    ".join(operations) + "\n")
+    for cmd in prog.circuit:
+        if prog.type == "tdm":
+            format_dict = {k: f"p[{k[1:]}]" for k in prog.parameters.keys()}
+            name = cmd.op.__class__.__name__
+            params_str = str(cmd.op.p)[1:-1].format(**format_dict)
+        else:
+            name = cmd.op.__class__.__name__
+            params_str = str(cmd.op.p)[1:-1]
+
+        modes = [f"q[{r.ind}]" for r in cmd.reg]
+        if len(modes) == 1:
+            modes_str = ", ".join(modes)
+        else:
+            modes_str = "(" + ", ".join(modes) + ")"
+        op = f"    ops.{name}({params_str}) | {modes_str}"
+
+        code_seq.append(op)
 
     if eng:
-        code_seq.append("results = eng.run(prog, shots=1)")
+        code_seq.append("\nresults = eng.run(prog)")
 
     return "\n".join(code_seq)
 
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -300,7 +300,7 @@ def serialize_program(prog, eng=None):
     code_seq.append("    " + "\n    ".join(operations) + "\n")
 
     if eng:
-        code_seq.append(f"results = eng.run()")
+        code_seq.append("results = eng.run()")
 
     return "\n".join(code_seq)
 
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -332,7 +332,7 @@ def _factor_out_pi(num_list, precision=12):
             a.append(str(p))
             continue
 
-        if np.isclose(p % (), [0, factor]).any() and p != 0:
+        if np.isclose(p % factor, [0, factor]).any() and p != 0:
             gcd = np.gcd(int(p / factor), precision)
             if gcd == precision:
                 if int(p/np.pi) == 1:
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -316,7 +316,7 @@ def generate_code(prog, eng=None):
 
     # check if program is of TDM type and format the context as appropriate
     if prog.type == "tdm":
-        # if the context arrays contain pi-values, rewrite them as pi-expressions
+        # if the context arrays contain pi-values, factor out multiples of np.pi
         tdm_params = [f"[{_factor_out_pi(par)}]" for par in prog.tdm_params]
         code_seq.append("\nwith prog.context(" + ", ".join(tdm_params) + ") as (p, q):")
     else:
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -272,12 +272,10 @@ def generate_code(prog, eng=None):
         else:
             if eng.backend_options:
                 formatting_str = (
-                    f'"{eng.backend_name}", backend_options=' +
-                    f'{{"cutoff_dim": {eng.backend_options["cutoff_dim"]}}}'
+                    f'"{eng.backend_name}", backend_options='
+                    + f'{{"cutoff_dim": {eng.backend_options["cutoff_dim"]}}}'
                 )
-                code_seq.append(
-                    f'eng = sf.Engine({formatting_str})'
-                )
+                code_seq.append(f"eng = sf.Engine({formatting_str})")
             else:
                 code_seq.append(f'eng = sf.Engine("{eng.backend_name}")')
 
@@ -335,7 +333,7 @@ def _factor_out_pi(num_list, precision=12):
         if np.isclose(p % factor, [0, factor]).any() and p != 0:
             gcd = np.gcd(int(p / factor), precision)
             if gcd == precision:
-                if int(p/np.pi) == 1:
+                if int(p / np.pi) == 1:
                     a.append("np.pi")
                 else:
                     a.append(f"{int(p/np.pi)}*np.pi")
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -263,6 +263,8 @@ def generate_code(prog, eng=None):
     else:
         code_seq.append(f"prog = sf.Program({prog.num_subsystems})")
 
+    # check if an engine is supplied; if so, format and add backend/target
+    # along with backend options
     if eng:
         eng_type = eng.__class__.__name__
         if eng_type == "RemoteEngine":
@@ -279,12 +281,14 @@ def generate_code(prog, eng=None):
             else:
                 code_seq.append(f'eng = sf.Engine("{eng.backend_name}")')
 
+    # check if program is of TDM type and format the context as appropriate
     if prog.type == "tdm":
         tdm_params = [f"[{_factor_out_pi(par)}]" for par in prog.tdm_params]
         code_seq.append("\nwith prog.context(" + ", ".join(tdm_params) + ") as (p, q):")
     else:
         code_seq.append("\nwith prog.context as q:")
 
+    # add the operations, and replace any free parameters with e.g. `p[0]`, `p[1]`
     for cmd in prog.circuit:
         name = cmd.op.__class__.__name__
         if prog.type == "tdm":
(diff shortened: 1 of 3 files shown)


strawberryfields/io.py
```diff
@@ -244,6 +245,66 @@ def is_free_param(param):
     return prog
 
 
+def serialize_program(prog, eng=None):
+    """Converts a Strawberry Fields program into valid Strawberry Fields code.
```
josh146, line 249: Maybe this requires a better function name.

I'm not such a fan of `serialize_program`, for several reasons:

* It doesn't really emphasize the end result (i.e., produces a string containing valid Python code)

* It could potentially be confused with `io.save`, which is ideally what we want users to use when they are thinking 'what is the best way to serialize my SF program'.

* While it _is_ a form of serialization, it's not a very *good* form of serialization. Firstsly, you can't really de-serialize it easily (I'm not sure running it through the Python interpreter counts 😆). Secondly, it makes a few assumptions when serializing to 'fill in the gaps'.

Maybe `generate_python_code`, `generate_python`, or some other permutation thereof?

strawberryfields/io.py
```diff
@@ -244,6 +245,66 @@ def is_free_param(param):
...
+
+    Args:
+        prog (Program): the Strawberry Fields program
+        eng (Engine): the Strawberryfields engine
```
josh146, line 253: Oh, this is a good idea to have the engine as input! I can't remember if the server version was already like this haha

strawberryfields/io.py
```diff
@@ -244,6 +245,66 @@ def is_free_param(param):
...
+    operations = []
+    op_imports = []
+    for cmd in prog.circuit:
+        op_name = cmd.op.__class__.__name__
+        if op_name not in op_imports:
+            op_imports.append(op_name)
+
+        if prog.type == "tdm":
+            format_dict = {k: f"p[{k[1:]}]" for k in prog.parameters.keys()}
+            op = cmd.__str__().format(**format_dict)
```
josh146, line 269: 💯

strawberryfields/io.py
```diff
@@ -244,6 +245,66 @@ def is_free_param(param):
...
+        else:
+            op = cmd.__str__()
+
+        operations.append(op)
+
+    code_seq.append("from strawberryfields.ops import " + ", ".join(op_imports) + "\n")
```
josh146, line 275: Very minor: we could alternatively just do `from strawberryfields import ops`, and then prepend all operations with `ops.`. Not sure if this has any advantages over this approach though.

strawberryfields/io.py
```diff
@@ -244,6 +245,66 @@ def is_free_param(param):
...
+    if prog.type == "tdm":
+        tdm_params = [f"{par}" for par in prog.tdm_params]
+        code_seq.append("with prog.context(" + ", ".join(tdm_params) + ") as (p, q):")
+    else:
+        code_seq.append("with prog.context as q:")
+
+    code_seq.append("    " + "\n    ".join(operations) + "\n")
+
+    if eng:
+        code_seq.append("results = eng.run()")
```
josh146, line 303: `eng.run` is missing the program input, as well as other runtime options (shots, etc.)

strawberryfields/io.py
```diff
@@ -244,6 +245,66 @@ def is_free_param(param):
...
+        code_seq.append(f"prog = sf.Program({prog.num_subsystems})")
+
+    if eng:
+        eng_type = eng.__class__.__name__
+        if eng_type == "RemoteEngine":
+            code_seq.append(f"eng = sf.RemoteEngine({eng.target})")
+        else:
+            if eng.backend_options:
+                code_seq.append(
+                    f"eng = sf.Engine({eng.backend_name}, backend_options={eng.backend_options})\n"
```
josh146, line 289: 💪

josh146 (commented):
Looking good @thisac! I like that it accepts the engine as an argument, will make it very flexible.

Since @antalszava recently worked on this on the server side, would be good to also get his thoughts (Edit: I see he is already tagged 💪 )

strawberryfields/io.py
thisac: Thanks for the comments @josh146. What you're saying makes sense, and I'll update it to something better. :slightly_smiling_face:  :+1:

strawberryfields/io.py
thisac: Nope, it didn't. I actually wrote this one from scratch since I thought that was easier. I like being able to pass an engine as well. It makes it possible to generate code using different engines and programs (and if skipping the engine, it'll just return the circuit code).

strawberryfields/io.py
thisac: Not sure what we'd prefer actually. I don't mind either way; I just chose this since I got the feeling that this would be the preferred way. The server actually has `from strawberryfields.ops import *`, but I'm not super-fond of that. I think either like this or, as you suggested, `from strawberryfields import ops`. I'm happy either way. :smile:

strawberryfields/io.py
thisac: :scream:

thisac: I missed adding the `prog`, but `shots` and `modes` are a bit difficult since they're actually never stored in neither the engine nor the program. They're only passed when running the engine; not while creating it. 

Either we can just leave it at `eng.run(prog)`, 

or add e.g. `shots=1` always, `eng.run(prog, shots=1)`, so that this option is clear and easy to change,

or have it as another argument in the function `generate_code(prog, eng=None, shots=1)`,

or maybe even as `generate_code(prog, eng=None, **kwargs)`, where kwargs is the same as the `engine.run` kwargs.

@josh146 Do you have any other ideas? I think I'm leaning towards the second option. :thinking:

josh146: Oh right 🤔 

Do shots not get serialized when converting from a blackbird script to a SF program? I thought they did.

But I guess if you are starting directly from a SF program, no info will be there...

strawberryfields/io.py
josh146: Same, I actually have no preference! Feel free to choose which ever one you prefer.

The _only_ downside that I can think of: if the number of operations is large, the number of imports here will cause a very long line.

strawberryfields/io.py
thisac: Just had a look, and `shots` can either be part of `prog.run_options` or passed directly via `eng.run`. If passed via `eng.run` `prog.run_options` is _not_ updated. That would be easy to do, but I'm not sure if it's a good thing that running the engine would change the `shots` defined earlier in `run_options`. :thinking:

strawberryfields/io.py
thisac: I went with `from strawberryfields import ops`. :fireworks:

strawberryfields/io.py
antalszava: More on the visual side at a later stage, maybe this could also be a drop-down option? That way we wouldn't have to worry about input validation (e.g., making sure that the user did indeed input the correct engine).

antalszava (commented):
Looking good so far! :100:

strawberryfields/io.py
josh146: @antalszava how do you mean exactly?

strawberryfields/io.py
```diff
@@ -244,6 +246,108 @@ def is_free_param(param):
     return prog
 
 
+def generate_code(prog, eng=None):
+    """Converts a Strawberry Fields program into valid Strawberry Fields code.
+
+    Args:
+        prog (Program): the Strawberry Fields program
+        eng (Engine): the Strawberryfields engine
```
josh146, line 254: Might be a good idea to add here what happens if `eng` is not supplied.

strawberryfields/io.py
```diff
@@ -244,6 +246,108 @@ def is_free_param(param):
...
+
+    Returns:
+        str: the Strawberry Fields code, for constructing the program, as a string
+    """
```
josh146, line 258: I would say worth adding an example here as well

strawberryfields/io.py
```diff
@@ -244,6 +246,108 @@ def is_free_param(param):
...
+    # along with backend options
+    if eng:
+        eng_type = eng.__class__.__name__
+        if eng_type == "RemoteEngine":
+            code_seq.append(f'eng = sf.RemoteEngine("{eng.target}")')
+        else:
+            if eng.backend_options:
+                formatting_str = (
+                    f'"{eng.backend_name}", backend_options='
+                    + f'{{"cutoff_dim": {eng.backend_options["cutoff_dim"]}}}'
```
josh146, line 276: I suppose this line doesn't distinguish between the Fock backend and the Gaussian backend? As far as I remember, the Gaussian backend does not permit `cutoff_dim` as a backend option

josh146: Maybe we could have an if statement `if "cutoff_dim" in eng.backend_options:`

strawberryfields/io.py
```diff
@@ -244,6 +246,108 @@ def is_free_param(param):
...
+                )
+                code_seq.append(f"eng = sf.Engine({formatting_str})")
+            else:
+                code_seq.append(f'eng = sf.Engine("{eng.backend_name}")')
+
+    # check if program is of TDM type and format the context as appropriate
+    if prog.type == "tdm":
+        tdm_params = [f"[{_factor_out_pi(par)}]" for par in prog.tdm_params]
```
josh146, line 284: How come pi is factored out here?

strawberryfields/io.py
```diff
@@ -244,6 +246,108 @@ def is_free_param(param):
...
+                if int(p / np.pi) == 1:
+                    a.append("np.pi")
+                else:
+                    a.append(f"{int(p/np.pi)}*np.pi")
+            else:
+                coeff = int(p / factor / gcd)
+                if coeff == 1:
+                    a.append(f"np.pi/{int(precision/gcd)}")
+                else:
+                    a.append(f"{coeff}*np.pi/{int(precision / gcd)}")
```
josh146, line 345: Oh, I got it! This makes sense.

strawberryfields/io.py
josh146: Ah! I read the function below and now get it. Might make sense to add a one line comment here just for clarity

tests/frontend/test_io.py
```diff
@@ -644,6 +644,151 @@ def test_gate_not_defined_tdm(self):
...
+            assert row == expected[i]
+
+    @pytest.mark.parametrize(
+        "value", [
+            "np.pi", "np.pi/2", "np.pi/12", "2*np.pi", "2*np.pi/3",
+            "0", "42", "9.87", "-0.2", "no_number", "{p3}"
+        ],
+    )
+    def test_factor_out_pi(self, value):
+        """Test for the factor_out_pi function"""
```
josh146, line 781: Might be worth making this more descriptive

tests/frontend/test_io.py
```diff
@@ -644,6 +644,151 @@ def test_gate_not_defined_tdm(self):
...
+            ops.BSgate(2*np.pi, 0.62) | (q[0], q[1])
+            ops.MeasureFock() | q[0]
+
+        results = eng.run(prog)
+
+        code = io.generate_code(prog, eng)
+
+        code_list = code.split("\n")
+        formatting_str = f"\"{engine_kwargs['backend']}\""
+        '"fock", backend_options={"cutoff_dim": 5}'
```
josh146, line 740: This string looks like it has no effect?

tests/frontend/test_io.py
```diff
@@ -644,6 +644,151 @@ def test_gate_not_defined_tdm(self):
...
+        if "backend_options" in engine_kwargs:
+            formatting_str += (
+                ", backend_options=" +
+                f'{{"cutoff_dim": {engine_kwargs["backend_options"]["cutoff_dim"]}}}'
```
josh146, line 744: Inside parenthesis, the string concatenation is implicit (so the `+` is not needed)

strawberryfields/io.py
josh146: Oh, I see what is happening. The current logic assumes that the existence of `backend_options` implies that the backend is a Fock backend, with a cutoff argument. However this might not be the case in general.

josh146 (commented):
Looking really good @thisac! Really nice tests, and it's awesome that it also supports TDM programs - will make it very easy to incorporate TDM programs into SF interactive in the future.

Only one major comment, regarding the printing of `backend_options`.

strawberryfields/io.py
thisac: You're right! I missed that, just thinking that `cutoff_dim` would be the only backend option, but it's true that this might not always be the case. Thanks for spotting it!

strawberryfields/io.py
```diff
@@ -281,6 +316,7 @@ def generate_code(prog, eng=None):
 
     # check if program is of TDM type and format the context as appropriate
     if prog.type == "tdm":
+        # if the context arrays contain pi-values, rewrite them as pi-expressions
```
josh146, line 319:
```suggestion
        # if the context arrays contain pi-values, factor out multiples of np.pi
```

tests/frontend/test_io.py
```diff
@@ -778,7 +777,12 @@ def test_generate_code_tdm(self):
         ],
     )
     def test_factor_out_pi(self, value):
-        """Test for the factor_out_pi function"""
+        """Test that the factor_out_pi function is able to convert floats
+        that are equal to a pi expression, to strings containing a pi
+        expression.
+
+        For example, the float 6.28318530718 should be converted to the string "2*np.pi"
+        """
```
josh146: This is a pretty cool utility function. Sounds like it was fun to write.

josh146 (approved):
Looks great @thisac!

strawberryfields/io.py
antalszava: The `eng` argument that is passed to `generate_code` is given as user input. This in itself would then require input validation, so that there is some sort of an error raised if the wrong value is inputted.

My thinking previously was that that can be okay when embedded into a webpage because we could then create a drop-down menu where the possible options are pre-defined (hence there's no need for input validation). For other use cases maybe it's worth including some sort of validation or defining the accepted type of engines here in advance.

strawberryfields/io.py
```diff
@@ -244,6 +246,145 @@ def is_free_param(param):
...
+
+        prog = sf.Program(3)
+        eng = sf.Engine("fock", backend_options={"cutoff_dim": 5})
+
+        with prog.context as q:
+            ops.Sgate(2*np.pi/3, 0.0) | q[1]
+            ops.BSgate(0.6, 0.1) | (q[2], q[0])
+            ops.MeasureFock() | (q[0], q[1], q[2])
+
+        results = eng.run(prog)
```
antalszava, line 284: :partying_face:

strawberryfields/io.py
```diff
@@ -244,6 +246,145 @@ def is_free_param(param):
...
+                    f'"{eng.backend_name}", backend_options='
+                    + f'{{"cutoff_dim": {eng.backend_options["cutoff_dim"]}}}'
+                )
+                code_seq.append(f"eng = sf.Engine({formatting_str})")
+            else:
+                code_seq.append(f'eng = sf.Engine("{eng.backend_name}")')
+
+    # check if program is of TDM type and format the context as appropriate
+    if prog.type == "tdm":
+        # if the context arrays contain pi-values, factor out multiples of np.pi
```
antalszava, line 319: Hmm, I'm still not clear on the motivation of factoring out pi :thinking: Might just be me missing something.

strawberryfields/io.py
```diff
@@ -244,6 +246,145 @@ def is_free_param(param):
...
+
+        code_seq.append(op)
+
+    if eng:
+        code_seq.append("\nresults = eng.run(prog)")
+
+    return "\n".join(code_seq)
+
+
+def _factor_out_pi(num_list, precision=12):
```
antalszava, line 349: Wondering if it would be worth adjusting `precision` to the value that can be set by [`np.set_printoptions`](https://numpy.org/doc/stable/reference/generated/numpy.set_printoptions.html). Right now we'd always use `precision=12` internally, correct?

antalszava (approved):
Looks good to me overall! :100: :blush: Curious about the usage of `factor_out_pi` (think it's mainly for precision?), though it's not something major.

strawberryfields/io.py
thisac: It simply allows a program that contains e.g. `Sgate(3*np.pi/2)` to actually return a string code representation that looks the same, instead of just returning a float value instead, like `Sgate(4.71238898038)`.

thisac: Maybe `factor out` is a bit confusing (although that's what we're doing, when it's possible). Not sure if there's a better name. :thinking:

strawberryfields/io.py
thisac: This is a bit different though, and again, this might be a misleading argument name. Here `precision` simply means how large denominators it should use for the factors (e.g. `precision=12` means it tries to factor out `np.pi/12`, while `precision=2` means that `np.pi/2` works, but `np.pi/3` will return `1.0471975512`).

antalszava: Got it! Could maybe go with the name `denominator` or `denom`?

antalszava:
Thanks for the clarification @thisac! The main piece missing for me was that internally we'd already have the numeric values for gate parameters (which actually makes a lot of sense now) and we'd like to convert those numeric values to expressions that contain `np.pi`.

antalszava:
[ch1778]

PR #497, opened 2020-12-03
Test a possible bug in the backend.

nquesada:


elib20 (approved):

josh146:
@nquesada what bug does this fix? It would be good to include a test for it, just for (a) verification is fixes it, and (b) making sure we catch this bug in the future.

Also don't forget to update the changelog!

co9olguy:
How come this bug needed to be tested via PR? :thinking:

nquesada:
@co9olguy so that all the tests are run?

co9olguy:
How come you didn't want to run the tests locally?

PR #500, opened 2020-12-04
Fix run_options not being used

thisac:
**Context:**
Setting the number of shots in `TDMProgram.run_options` doesn't change the number of shots used when calling `eng.run()`, without passing a shots value.

Shots now checked and retrieved in the following order: shots passed via `eng.run(prog, shots=shots)` > `prog.run_options` > default value, which is 1. The left-most available being the one that is used.

**Benefits:**
If `TDMProgram.run_options` is set, that value is used rather than falling back on the default value. This can be the case, for example, when converting from a Blackbird script. Without this fix the shots values set in the Blackbird script would simply be ignored when running the engine.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -424,8 +424,9 @@ class LocalEngine(BaseEngine):
             not isinstance(program, collections.abc.Sequence) and program.type == "tdm"
         )
         if valid_tdm_program:
+            # priority order for the shots value should be kwargs > run_options > 1
+            shots = kwargs.get("shots", program.run_options.get("shots", 1))
 
-            shots = kwargs.get("shots", 1)
             program.unroll(shots=shots)
             # Shots >1 for a TDM program simply corresponds to creating
             # multiple copies of the program, and appending them to run sequentially.
--- a/tests/frontend/test_engine.py
+++ b/tests/frontend/test_engine.py
@@ -69,6 +69,46 @@ class TestEngine:
 class TestEngineProgramInteraction:
     """Test the Engine class and its interaction with Program instances."""
 
+    def test_shots_default(self):
+        """Test that default shots (1) is used"""
+        prog = sf.Program(1)
+        eng = sf.Engine("gaussian")
+
+        with prog.context as q:
+            ops.Sgate(0.5) | q[0]
+            ops.MeasureFock() | q
+
+        results = eng.run(prog)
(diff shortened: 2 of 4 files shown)


strawberryfields/engine.py
```diff
@@ -424,8 +424,9 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
             not isinstance(program, collections.abc.Sequence) and program.type == "tdm"
         )
         if valid_tdm_program:
+            # priority order for the shots value should be kwargs > run_options > 1
+            shots = kwargs.get("shots", program.run_options.get("shots", 1))
```
josh146, line 428: I like how this use of nested `get` makes it easy to see the priority

josh146 (approved):
Nice 💪

antalszava (commented):
Looking good! 💯 Maybe a test case could be helpful to cover the previous non-working case and verify that shots are extracted well now.

tests/frontend/test_engine.py
```diff
@@ -69,6 +69,46 @@ def test_bad_backend(self):
...
+        eng = sf.Engine("gaussian")
+
+        with prog.context as q:
+            ops.Sgate(0.5) | q[0]
+            ops.MeasureFock() | q
+
+        prog.run_options = {"shots": 5}
+        results = eng.run(prog, shots=2)
+        assert results.samples.shape[0] == 2
+        assert prog.run_options["shots"] == 5
```
antalszava, line 110: These tests look awesome! :1st_place_medal: :rocket:

tests/frontend/test_tdmprogram.py
```diff
@@ -706,3 +706,46 @@ def test_move_vac_modes(self, N, crop, expected):
...
+        eng = sf.Engine("gaussian")
+
+        with prog.context([1,2], [3,4]) as (p, q):
+            ops.Sgate(p[0]) | q[0]
+            ops.MeasureHomodyne(p[1]) | q[0]
+
+        prog.run_options = {"shots": 5}
+        results = eng.run(prog, shots=2)
+        assert results.samples.shape[0] == 2
+        assert prog.run_options["shots"] == 5
```
antalszava, line 751: :100:

antalszava (approved):
Looks good to me! :100:

PR #507, opened 2020-12-11
Fix bug in Dgate, Coherent, and DisplacedSqueezed

trbromley:
This PR fixes a bug in Dgate, Coherent, and DisplacedSqueezed where TensorFlow tensors raise an error if fed with an extra dimension due to batching.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -99,7 +99,8 @@
 
 This release contains contributions from (in alphabetical order):
 
-Tom Bromley, Jack Brown, Theodor Isacsson, Josh Izaac, Fabian Laudenbach, Nicolas Quesada, Antal Száva.
+Tom Bromley, Jack Brown, Theodor Isacsson, Josh Izaac, Fabian Laudenbach, Nicolas Quesada,
+Antal Száva.
 
 # Release 0.16.0 (current release)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -48,6 +48,7 @@
 
 * Fixes a bug where the `Dgate` does not support TensorFlow tensors if the tensor has an added
   dimension due to the existence of batching.
+  [(#507)](https://github.com/XanaduAI/strawberryfields/pull/507)
 
 * Fixed issue with `reshape_samples` where the samples were sometimes
   reshaped in the wrong way.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -46,6 +46,9 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the `Dgate` does not support TensorFlow tensors if the tensor has an added
+  dimension due to the existence of batching.
+
 * Fixed issue with `reshape_samples` where the samples were sometimes
   reshaped in the wrong way.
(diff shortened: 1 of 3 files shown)


co9olguy (approved):
Are any other gates guilty of this?

josh146 (approved):
Thanks @trbromley!

@co9olguy: this will affect any op that previously accepted a complex argument. Only other ones that come to mind are `Coherent` and `Catstate`.

co9olguy:
Should we test/fix those ops in this PR then?

trbromley:
Thanks @co9olguy, I have added checks for the other gates @josh146 mentioned: daef1c59941946a8d2d23733b66b5e1ae2b04248. I also can't see any other gates where this issue may occur.

.github/CHANGELOG.md
```diff
@@ -69,6 +69,10 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where `Dgate`, `Coherent` and `DisplacedSqueezed` do not support TensorFlow tensors
```
josh146, line 72:
```suggestion
* Fixes a bug where `Dgate`, `Coherent`, and `DisplacedSqueezed` do not support TensorFlow tensors
```

tests/frontend/test_ops_gate.py
```diff
@@ -220,3 +227,21 @@ def test_merge_measured_pars():
     # gates that have different p[1] parameters
     with pytest.raises(MergeFailure, match="Don't know how to merge these gates."):
         assert D.merge(G)
+
+
+@pytest.mark.parametrize("gate", [ops.Dgate, ops.Coherent, ops.DisplacedSqueezed])
+@pytest.mark.skipif(not tf_available, reason="Test only works if TF is installed")
+def test_tf_batch_in_gates_previously_supporting_complex(gate):
```
josh146, line 234: Minor, but could we not just use `pytest.importorskip` inside the test function? This would avoid the if statement at the top of the file as well

josh146 (approved):

strawberryfields/ops.py
```diff
@@ -594,9 +594,10 @@ def _apply(self, reg, backend, **kwargs):
         r = par_evaluate(self.p[0])
         phi = par_evaluate(self.p[1])
 
-        tf_complex = any(hasattr(arg, "numpy") and np.iscomplex(arg.numpy()) for arg in [r, phi])
+        np_args = [arg.numpy() if hasattr(arg, "numpy") else arg for arg in [r, phi]]
+        is_complex = any([np.iscomplexobj(np.real_if_close(arg)) for arg in np_args])
```
trbromley, line 598: Note that we need the `np.real_if_close()`, otherwise the `integration/test_algorithms.py::test_gaussian_gate_teleportation` test will fail. The test fails on the line:
```python
Xgate(s * q[0].par) | q[2]
```
Here, `q[0].par` is a real-valued object with a complex dtype.

trbromley:
The approach has been updated due to some issues with failing tests in Python 3.6 :thinking: Should be good to merge now.

strawberryfields/ops.py
josh146: Thanks for catching this Tom; so this only failed on Python 3.6?

PR #510, opened 2020-12-15
Fock state and quadrature plotting 

antalszava:
**Context:**
The functionality to plot the Fock state probability distribution and quadrature probability distributions currently available with SF interactive should be moved to SF.

Adds two plotting functions:
* `plot_fock`: plots the Fock state probabilities given a state object, uses the `generate_fock_chart` auxiliary function;
* `plot_quad`:  given a state object and arrays of x and p discretized quadrature values, uses the `generate_quad_chart` auxiliary function.

The `generate_fock_chart` and `generate_quad_chart` auxiliary functions create `plotly` compatible json charts.

**Benefits:**
Fock state probabilities and x and p-quadrature probabilities can be plotted.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -31,6 +31,9 @@
 
 <h3>Improvements</h3>
 
+* `Connection` objects now send versioned requests to the platform API.
+  [(#512)](https://github.com/XanaduAI/strawberryfields/pull/512)
+
 * The `copies` option when constructing a `TDMProgram` have been removed. Instead, the number of
   copies of a TDM algorithm can now be set by passing the `shots` keyword argument to
   the `eng.run()` method.
@@ -92,7 +95,7 @@
 
 This release contains contributions from (in alphabetical order):
 
-Theodor Isacsson, Josh Izaac, Fabian Laudenbach, Nicolas Quesada, Antal Száva.
+Jack Brown, Theodor Isacsson, Josh Izaac, Fabian Laudenbach, Nicolas Quesada, Antal Száva.
 
 # Release 0.16.0 (current release)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -25,8 +25,8 @@
   sf.plot_wigner(state, mode, xvec, pvec, renderer="browser")
   ```
 
-* Fock state probabilities can be plotted directly via Strawberry Fields using
-  Plot.ly.
+* Fock state marginal probabilities can be plotted directly via Strawberry
+  Fields using Plot.ly.
   [(#510)](https://github.com/XanaduAI/strawberryfields/pull/510)
 
   ```python
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -5,6 +5,10 @@
 * `TDMProgram` objects can now be compiled and submitted via the API.
   [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
 
+* Strawberry Fields code can be generated from a program (and an engine) by
+  calling `sf.io.generate_code(program, eng=engine)`.
+  [(#496)](https://github.com/XanaduAI/strawberryfields/pull/496)
+
 <h3>Improvements</h3>
 
 * The `copies` option when constructing a `TDMProgram` have been removed. Instead, the number of
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -5,6 +5,26 @@
 * `TDMProgram` objects can now be compiled and submitted via the API.
   [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
 
+* Wigner functions can be plotted directly via Strawberry Fields using Plot.ly.
+  [(#495)](https://github.com/XanaduAI/strawberryfields/pull/495)
+
+  ```python
+  prog = sf.Program(1)
+  eng = sf.Engine('fock', backend_options={"cutoff_dim": 10})
+
+  with prog.context as q:
+    gamma = 2
+    Vgate(gamma) | q[0]
+
+  state = eng.run(prog).state
+
+  xvec = np.arange(-4, 4, 0.01)
+  pvec = np.arange(-4, 4, 0.01)
+  mode = 0
+
+  sf.plot_wigner(state, mode, xvec, pvec, renderer="browser")
+  ```
+
 * Strawberry Fields code can be generated from a program (and an engine) by
   calling `sf.io.generate_code(program, eng=engine)`.
   [(#496)](https://github.com/XanaduAI/strawberryfields/pull/496)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -25,6 +25,45 @@
   sf.plot_wigner(state, mode, xvec, pvec, renderer="browser")
   ```
 
+* Fock state probabilities can be plotted directly via Strawberry Fields using
+  Plot.ly.
+  [(#510)](https://github.com/XanaduAI/strawberryfields/pull/510)
+
+  ```python
+  prog = sf.Program(1)
+  eng = sf.Engine('fock', backend_options={"cutoff_dim":5})
+
+  with prog.context as q:
+      Sgate(0.5) | q[0]
+
+  state = eng.run(prog).state
+  state.all_fock_probs()
+
+  modes = [0]
+
+  sf.plot_fock(state, modes, cutoff=5, renderer="browser")
+  ```
+
+* Position and momentum quadrature probabilities can be plotted directly via
+  Strawberry Fields using Plot.ly.
+  [(#510)](https://github.com/XanaduAI/strawberryfields/pull/510)
+
+  ```python
+  prog = sf.Program(1)
+  eng = sf.Engine('fock', backend_options={"cutoff_dim":5})
+
+  with prog.context as q:
+      Sgate(0.5) | q[0]
+
+  state = eng.run(prog).state
+
+  modes = [0]
+  xvec = np.arange(-4, 4, 0.1)
+  pvec = np.arange(-4, 4, 0.1)
+
+  sf.plot_quad(state, modes, xvec, pvec, renderer="browser")
+  ```
+
 * Strawberry Fields code can be generated from a program (and an engine) by
   calling `sf.io.generate_code(program, eng=engine)`.
   [(#496)](https://github.com/XanaduAI/strawberryfields/pull/496)
--- a/doc/code/sf_plot.rst
+++ b/doc/code/sf_plot.rst
@@ -1,0 +1,13 @@
+sf.plot
+=======
+
+.. currentmodule:: strawberryfields.plot
+
+.. warning::
+
+    Unless you are a Strawberry Fields developer, you likely do not need
+    to use these classes directly.
+
+.. automodapi:: strawberryfields.plot
+    :no-heading:
+    :skip: copy, deepcopy
--- a/strawberryfields/plot.py
+++ b/strawberryfields/plot.py
@@ -201,7 +201,7 @@ def plot_fock(state, modes, cutoff=None, renderer="browser"):
     n = np.arange(cutoff)
     mean = [np.sum(n * probs).real for probs in photon_dists]
 
-    xlabels = ["|{}>".format(i) for i in range(0, cutoff, 1)]
+    xlabels = [fr"$|{i}\rangle$" for i in range(0, cutoff, 1)]
 
     new_chart = generate_fock_chart(modes, photon_dists, mean, xlabels)
     pio.show(new_chart)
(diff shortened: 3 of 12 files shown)


tests/frontend/test_sf_plot.py
```diff
@@ -54,3 +62,271 @@ def test_no_errors(self, mode, renderer, contours, prog, monkeypatch):
...
+    config = {
+        "modeBarButtonsToRemove": ["lasso2d", "select2d", "toggleSpikelines"],
+        "displaylogo": False,
+    }
+    quad_chart = {"data": data, "layout": layout, "config": config}
+    return quad_chart
+
+
+class TestFockProbPlotting:
+    """Test the Fock state probabilities plotting function"""
```
antalszava, line 219:
```suggestion
    """Test the Fock state probabilities plotting function."""
```

nquesada (approved):

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+
+def plot_fock(state, modes, cutoff=None, renderer="browser"):
+    """Plot the Fock state probabilities with Plot.ly.
+
+    Args:
+        state (:class:`.BaseState`): the state used for plotting
+        modes (list): list of modes to generate output for
+        cutoff (int): the cutoff value determining the maximum Fock state to
+            get probabilities for, required if state is a
+            :class:`~.BaseGaussianState`
```
josh146, line 182:
```suggestion
        cutoff (int): The cutoff value determining the maximum Fock state to
            get probabilities for. Only required if state is a
            :class:`~.BaseGaussianState`
```

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+    },
+}
+
+
+def plot_fock(state, modes, cutoff=None, renderer="browser"):
+    """Plot the Fock state probabilities with Plot.ly.
+
+    Args:
+        state (:class:`.BaseState`): the state used for plotting
+        modes (list): list of modes to generate output for
```
josh146, line 179:
```suggestion
        modes (list): List of modes to generate output for. If more than one mode
            is provided, then the reduced state is computed for each mode, and marginal
            Fock probabilities plotted per mode.
```

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+        cutoff = state.cutoff_dim
+
+    pio = _get_plotly()
+    pio.renderers.default = renderer
+
+    num_modes = len(modes)
+
+    # Reduced density matrices
+    rho = [state.reduced_dm(n, cutoff=cutoff) for n in range(num_modes)]
+    photon_dists = np.array([np.real(np.diagonal(p)) for p in rho])
```
josh146, line 199: @nquesada, is this the fastest way of doing it?

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+        "displaylogo": False,
+    },
+}
+
+
+def plot_quad(state, modes, xvec, pvec, renderer="browser"):
+    """Plot the Wigner function with Plot.ly.
+
+    Args:
+        state (:class:`.BaseState`): the state used for plotting
```
josh146, line 298:
```suggestion
        state (.BaseState): the state used for plotting
```

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+    n = np.arange(cutoff)
+    mean = [np.sum(n * probs).real for probs in photon_dists]
+
+    xlabels = [fr"$|{i}\rangle$" for i in range(0, cutoff, 1)]
+
+    new_chart = generate_fock_chart(modes, photon_dists, mean, xlabels)
+    pio.show(new_chart)
+
+
+def generate_fock_chart(modes, photon_dists, mean, xlabels):
```
josh146, line 210: @antalszava I think we should pass the state to _this_ function, and do the photon dist computation here, instead.

I just suddenly realised that if running SF on a server (e.g., simulon), there is no function that can be called, given a state, to compute the JSON plot data.

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+    },
+    "config": {
+        "modeBarButtonsToRemove": ["lasso2d", "select2d", "toggleSpikelines"],
+        "displaylogo": False,
+    },
+}
+
+
+def plot_quad(state, modes, xvec, pvec, renderer="browser"):
+    """Plot the Wigner function with Plot.ly.
```
josh146, line 295: Docstring needs an update :)

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+        xvec (array): array of discretized :math:`x` quadrature values
+        pvec (array): array of discretized :math:`p` quadrature values
+        renderer (string): the renderer for plotting with Plot.ly
+    """
+    pio = _get_plotly()
+    pio.renderers.default = renderer
+
+    for mode in modes:
+        p_probs = state.p_quad_values(mode, xvec, pvec).tolist()
+        x_probs = state.x_quad_values(mode, xvec, pvec).tolist()
```
josh146, line 309: Perhaps these should also be moved into `generate_quad_chart`?

josh146 (commented):
Looking good @antalszava! I only have one major question - should we have the 'generate plot' functions take the state as an argument as well?

strawberryfields/plot.py
```diff
@@ -120,3 +123,236 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+    },
+    "config": {
+        "modeBarButtonsToRemove": ["zoom2d", "lasso2d", "select2d", "toggleSpikelines"],
+        "displaylogo": False,
+    },
+}
+
+
+def plot_fock(state, modes, cutoff=None, renderer="browser"):
+    """Plot the Fock state probabilities with Plot.ly.
```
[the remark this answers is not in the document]
antalszava: Makes sense, thanks!

.github/CHANGELOG.md
```diff
@@ -25,6 +25,45 @@
   sf.plot_wigner(state, mode, xvec, pvec, renderer="browser")
   ```
 
+* Fock state probabilities can be plotted directly via Strawberry Fields using
+  Plot.ly.
```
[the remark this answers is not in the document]
antalszava: Fair point, updated! Thanks

strawberryfields/plot.py
antalszava: Good catch, updated! Thank you

.github/CHANGELOG.md
```diff
@@ -25,6 +25,45 @@
...
+  state.all_fock_probs()
+
+  modes = [0]
+
+  sf.plot_fock(state, modes, cutoff=5, renderer="browser")
+  ```
+
+* Position and momentum quadrature probabilities can be plotted directly via
+  Strawberry Fields using Plot.ly.
+  [(#510)](https://github.com/XanaduAI/strawberryfields/pull/510)
```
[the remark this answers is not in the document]
antalszava: There was a nice working version to build upon :100:

strawberryfields/plot.py
antalszava: Great point! Thanks for further clarifying. Updated all 3 plotting functions like that.

strawberryfields/plot.py
antalszava: Updated

strawberryfields/plot.py
```diff
@@ -47,26 +52,24 @@ def plot_wigner(state, mode, xvec, pvec, renderer="browser", contours=True):
...
 
-def generate_wigner_chart(data, xvec, pvec, contours=True):
+def generate_wigner_chart(state, mode, xvec, pvec, contours=True):
     """Populates a chart dictionary with reduced Wigner function surface plot data.
 
     Args:
-        data (array): 2D array of size [len(xvec), len(pvec)], containing reduced
-            Wigner function values for specified x and p values.
+        state (:class:`.BaseState`): the state used for plotting
+        mode (int): mode used to calculate the reduced Wigner function
```
antalszava, line 64: Update as per @josh146's [comment](https://github.com/XanaduAI/strawberryfields/pull/510#discussion_r545012852) (1.).

strawberryfields/plot.py
```diff
@@ -120,3 +123,235 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
...
+        "displaylogo": False,
+    },
+}
+
+
+def plot_fock(state, modes, cutoff=None, renderer="browser"):
+    """Plot the (marginal) Fock state probabilities with Plot.ly.
+
+    Args:
+        state (.BaseState): the state used for plotting
```
thisac, line 178: Always a bit confused, but isn't the correct way like this instead:
```suggestion
        state (:class:`.BaseState`): the state used for plotting
```

Oh, just noticed the review comment from @josh146.

If `.BaseState` is the correct way of doing this, then we probably want to update the Wigner plotting function docstrings to this as well. :slightly_smiling_face:

strawberryfields/plot.py
```diff
@@ -120,3 +123,235 @@ def generate_wigner_chart(data, xvec, pvec, contours=True):
     chart["layout"]["scene"]["zaxis"]["title"] = "W(x,p)"
 
     return chart
+
+
+# Plot.ly default barchart JSON
+barchart_default = {
```
thisac, line 129: Is all of this directly copied from SF interactive? I decided to _not_ use the default chart for the Wigner plotting, and simply used a slightly more minimal dict defined directly in `generate_wigner_chart` instead. I think that quite a few of these options either didn't make sense or didn't seem to change anything (so I removed them in favour of a more streamlined set of plotting options). :thinking:

.github/CHANGELOG.md
```diff
@@ -25,6 +25,45 @@
...
+
+  with prog.context as q:
+      Sgate(0.5) | q[0]
+
+  state = eng.run(prog).state
+  state.all_fock_probs()
+
+  modes = [0]
+
+  sf.plot_fock(state, modes, cutoff=5, renderer="browser")
```
thisac, line 44: The latex when hovering over the bars doesn't seem to work properly for me. :thinking:
fock

thisac: Otherwise it looks great. :smile:

strawberryfields/plot.py
antalszava: Actually, just checked, and it seems that both make it clickable!

`.BaseState`:
image

`:class:`.BaseState``:
image

Both are clickable and redirect to the correct page. Accepting the suggestion, @josh146 feel free to share if there's a downside missed and will change it back.

strawberryfields/plot.py
antalszava: Yes, it's the one used online atm (except for a couple of options which became invalid). Would there be any options that are worth leaving out?

.github/CHANGELOG.md
antalszava: Yeah, seems like an [open `plotly` issue](https://github.com/plotly/plotly.js/issues/559).

thisac: Oh, I see. :disappointed:

strawberryfields/plot.py
thisac: Not really. They actually all seem nice to have here. Just played around with them a bit, and I think they make sense to keep.

antalszava:
Thanks for the comments & suggestions @josh146 @thisac! Let me know if there would still be something to adjust.

Added a new entry for plotting to the docs (appears on the sidebar now too):
https://strawberryfields--510.org.readthedocs.build/en/510/code/sf_plot.html

doc/code/sf_plot.rst
```diff
@@ -0,0 +1,13 @@
...
+.. currentmodule:: strawberryfields.plot
+
+.. warning::
+
+    Unless you are a Strawberry Fields developer, you likely do not need
+    to use these classes directly.
+
+.. automodapi:: strawberryfields.plot
+    :no-heading:
+    :skip: copy, deepcopy
```
thisac, line 13: :100:

thisac (approved):
Looks great @antalszava! The plots look really nice. :100:

strawberryfields/plot.py
josh146: Oh the second is clickable?! That never used to be the case!

josh146 (approved):
📈 💯

antalszava:
[ch1776] [ch1775]

PR #511, opened 2020-12-15
Update fidelity_coherent and fidelity_vacuum

elib20:
Redefines these functions to work with the form of the state (linear combination of Gaussians). Uses the overlap integral between two Gaussians, since coherent states are pure.

### Before submitting

- [ ] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      test directory!

- [ ] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [ ] The Strawberry Fields source code conforms to
      [PEP8 standards](https://www.python.org/dev/peps/pep-0008/).
      We check all of our code against [Pylint](https://www.pylint.org/).
      To lint modified files, simply `pip install pylint`, and then
      run `pylint strawberryfields/path/to/file.py`.

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
+++ b/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
@@ -301,8 +301,8 @@ class BosonicModes:
         """ Returns a function that evaluates the Q function of the given state """
         if modes is None:
             modes = list(range(self.nlen))
-        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
-        alpha_mean = np.append(alpha.real, alpha.imag)
+        mode_ind = np.append(2 * np.array(modes), 2 * np.array(modes) + 1)
+        alpha_mean = np.append(alpha.real, alpha.imag) * np.sqrt(2 * self.hbar)
         deltas = self.means[:, mode_ind] - alpha_mean
         cov_sum = (
             self.covs[:, mode_ind, :][:, :, mode_ind] + self.hbar * np.eye((len(mode_ind))) / 2
--- a/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
+++ b/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
@@ -301,23 +301,24 @@ class BosonicModes:
         """ Returns a function that evaluates the Q function of the given state """
         if modes is None:
             modes = list(range(self.nlen))
-
-        Q = self.qmat(modes)
-        Qi = np.linalg.inv(Q)
-        delta = self.mean[modes] - alpha
-
-        delta = np.concatenate((delta, np.conjugate(delta)))
-        return np.sqrt(np.linalg.det(Qi).real) * np.exp(
-            -0.5 * np.dot(delta, np.dot(Qi, np.conjugate(delta))).real
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        alpha_mean = np.append(alpha.real, alpha.imag)
+        deltas = self.means[:, mode_ind] - alpha_mean
+        cov_sum = (
+            self.covs[:, mode_ind, :][:, :, mode_ind] + self.hbar * np.eye((len(mode_ind))) / 2
         )
+        exp_arg = np.einsum("...j,...jk,...k", deltas, np.linalg.inv(cov_sum), deltas)
+        weighted_exp = np.array(self.weights) * np.exp(-exp_arg)
+        fidelity = np.sum(weighted_exp)
+        return fidelity
 
     def fidelity_vacuum(self, modes=None):
         """fidelity of the current state with the vacuum state"""
         if modes is None:
             modes = list(range(self.nlen))
-
         alpha = np.zeros(len(modes))
-        return self.fidelity_coherent(alpha)
+        fidelity = self.fidelity_coherent(alpha, modes=modes)
+        return fidelity
 
     def Amat(self):
         """ Constructs the A matrix from Hamilton's paper"""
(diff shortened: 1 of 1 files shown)


strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -301,23 +301,24 @@ def fidelity_coherent(self, alpha, modes=None):
...
             modes = list(range(self.nlen))
-
-        Q = self.qmat(modes)
-        Qi = np.linalg.inv(Q)
-        delta = self.mean[modes] - alpha
-
-        delta = np.concatenate((delta, np.conjugate(delta)))
-        return np.sqrt(np.linalg.det(Qi).real) * np.exp(
-            -0.5 * np.dot(delta, np.dot(Qi, np.conjugate(delta))).real
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
```
nquesada, line 304: what's the logic behind this line?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -301,23 +301,24 @@ def fidelity_coherent(self, alpha, modes=None):
         """ Returns a function that evaluates the Q function of the given state """
```
nquesada, line 301: I suspect this function should be hbar dependent, i.e., somewhere you need to check hbar in toder to know that q = \sqrt{2 hbar} Re(alpha) and p = \sqrt{2 hbar} Im(alpha)

elib20: I will change alpha_mean

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: You are right. We do not need np.sort()

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -301,23 +301,24 @@ def fidelity_coherent(self, alpha, modes=None):
...
-
-        Q = self.qmat(modes)
-        Qi = np.linalg.inv(Q)
-        delta = self.mean[modes] - alpha
-
-        delta = np.concatenate((delta, np.conjugate(delta)))
-        return np.sqrt(np.linalg.det(Qi).real) * np.exp(
-            -0.5 * np.dot(delta, np.dot(Qi, np.conjugate(delta))).real
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        alpha_mean = np.append(alpha.real, alpha.imag)
```
elib20, line 305:
```suggestion
        alpha_mean = np.append(alpha.real, alpha.imag) * np.sqrt(2 * self.hbar)
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20, line 304:
```suggestion
        mode_ind = np.append(2 * np.array(modes), 2 * np.array(modes) + 1)
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -301,23 +301,24 @@ def fidelity_coherent(self, alpha, modes=None):
...
-        return np.sqrt(np.linalg.det(Qi).real) * np.exp(
-            -0.5 * np.dot(delta, np.dot(Qi, np.conjugate(delta))).real
+        mode_ind = np.append(2 * np.array(modes), 2 * np.array(modes) + 1)
+        alpha_mean = np.append(alpha.real, alpha.imag) * np.sqrt(2 * self.hbar)
+        deltas = self.means[:, mode_ind] - alpha_mean
+        cov_sum = (
+            self.covs[:, mode_ind, :][:, :, mode_ind] + self.hbar * np.eye((len(mode_ind))) / 2
         )
+        exp_arg = np.einsum("...j,...jk,...k", deltas, np.linalg.inv(cov_sum), deltas)
+        weighted_exp = np.array(self.weights) * np.exp(-exp_arg)
```
elib20, line 311:
```suggestion
        weighted_exp = (
            np.array(self.weights)
            * self.hbar ** len(modes)
            * np.exp(-0.5 * exp_arg)
            / np.sqrt(np.linalg.det(cov_sum))
        )
```

elib20 (commented):
Include covariance determinants in weighted sum.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -301,23 +301,29 @@ def fidelity_coherent(self, alpha, modes=None):
...
             modes = list(range(self.nlen))
-
-        Q = self.qmat(modes)
-        Qi = np.linalg.inv(Q)
-        delta = self.mean[modes] - alpha
-
-        delta = np.concatenate((delta, np.conjugate(delta)))
-        return np.sqrt(np.linalg.det(Qi).real) * np.exp(
-            -0.5 * np.dot(delta, np.dot(Qi, np.conjugate(delta))).real
+        mode_ind = np.append(2 * np.array(modes), 2 * np.array(modes) + 1)
```
nquesada, line 304: I am a bit confused about the quadrature or mode ordering here.

nquesada: Did you try to make a two mode test?

nquesada (commented):
Nice implementation! Wondering if one could also add a parity function using very similar machinery?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I did. The way mode_ind is constructed, it does ends up changing the mode ordering to (q1,...,qn,p1,...,pn) when you call an object like self.means[:,mode_ind], but then this is compensated by the fact that the alpha_mean variable has (q1,...,qn,p1,...,pn) ordering.

nquesada: But then, don't you need to also reorder the covariance matrix? Maybe that is what you do below. In any case I'd suggest you add a comment for future readers of your code so that they are not lost.

PR #512, opened 2020-12-15
Adds api_version property, adds 'Accept-Version' to headers

brownj85:
**Context:**
The Xanadu platform API will soon be implementing versioning, allowing clients to 
request a  particular version of the API using the 'Accept-Version' HTTP header. This header will eventually be mandatory.

Adds an `api_version` str property to `Connection` with the value `1.0.0`. On init, adds
'Accept-Version' header to `Connection._headers` with the value of `api_version`.

**Benefits:**
Future releases of sf will be able to use the platform API once versioning becomes
mandatory. Platform will be able to support older versions of sf following breaking
API changes.

**Possible Drawbacks:**
None, beyond the established drawbacks of versioning the platform API - older
releases of sf will eventually be incompatible with the platform once versioning
becomes mandatory. The current (non-versioned) platform API's behavior is 
unchanged by the 'Accept-Version' header. 

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 <h3>New features since last release</h3>
 
-* `Connection` objects now send versioned requests to the platform API.
-  [(#512)](https://github.com/XanaduAI/strawberryfields/pull/512)
-
 * `TDMProgram` objects can now be compiled and submitted via the API.
   [(#476)](https://github.com/XanaduAI/strawberryfields/pull/476)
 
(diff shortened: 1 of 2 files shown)


.github/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 <h3>New features since last release</h3>
 
+* `Connection` objects now send versioned requests to the platform API.
+  [(#512)](https://github.com/XanaduAI/strawberryfields/pull/512)
```
josh146, line 6: I would move this down to the 'improvements' section, since it is less visible to users!

josh146: Also, don't forget to add yourself to the contributors section!

strawberryfields/api/connection.py
```diff
@@ -96,10 +96,19 @@ def __init__(
...
 
         self.log = create_logger(__name__)
 
+    @property
+    def api_version(self) -> str:
+        """The platform API version to request.
+
+        Returns:
+            str
+        """
```
josh146, line 109: Minor: because this is a property, I think you can also write
```suggestion
        """str: the platform API version to request"""
```

josh146 (approved):
Thanks @brownj85!

jswinarton (approved):
:+1:

PR #517, opened 2021-01-06
Gkps state prep

GDauphinais:
Fixed some hbar factors in real cat state preparation.


elib20:
Looks good to me! I was able to produce the same Wigner function using both representations as well as the Fock representation.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -327,8 +327,89 @@ def prepare_cat(self, alpha, phi, desc="complex"):
...
+            cov = np.repeat(cov[None, :], 4 * z_max + 1, axis=0)
+            cov_real = (
+                0.5
+                * self.circuit.hbar
+                * np.array([[[1, 0], [0, 1]], [[1, 0], [0, 1]]], dtype=float)
+            )
+            cov = np.concatenate((cov_real, cov))
+
+            # filter out 0 components
+            filt = ~np.isclose(weights, 0)
```
elib20, line 398: Should the filter depend on the cutoff parameter?

GDauphinais: This line is to remove peaks with 0 weights in the case where cos(phi) or sin(phi) = 0 (or is very close to it, perhaps for the case where the phase is chosen numerically by some algorithm specificied by the user).
The cutoff used for fixing z_max is sufficient otherwise, I'd think.

GDauphinais: Good point, I forgot to set the atol parameter to be consistent with the cutoff

GDauphinais: fixed

GDauphinais:
Little detail that may come in handy for specific applications: for approximate states the numerical error / tolerance isn't unified. Could be a nice feature to add to have some uniformity there.

Also, running frontend test, I get the message:
 VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
 I have python 3.8.5
 I believe the culprit may be this line :         means = np.array([[a for b in tup for a in b] for tup in mean_combs], dtype=complex) (in backend.py)

elib20:
I am also using Python 3.8.5 and I don't get that error. I am using numpy version 1.18.5. Which version are you using?

elib20 (commented):
To reiterate from before for the purpose of an official review: I was able to reproduce Wigner plots using both representations and the Fock backend. Nice work!

elib20 (approved):
Approved!

PR #520, opened 2021-01-11
Implement OAuth refresh token flow

corvust:
**Context:**

The platform backend is moving to a new OAuth based authentication flow using offline refresh tokens and access tokens. 

* Adds an authentication call to retrieve a short-lived access token using the long lived refresh token stored in the configuration (offline token)
* Wraps all API calls to retrieve a new access token if the current one has expired (resulting in a 401 response)

**Benefits:**

More secure and easier user credential management.

**Possible Drawbacks:**

Not backwards compatible, all users will be required to generate new tokens and update strawberry fields when the new authentication mechanism is activated.

**Related GitHub Issues:**

N/A

### DIFF
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -187,7 +187,9 @@ class Connection:
         circuit = bb.serialize()
 
         path = "/jobs"
-        response = self._request("POST", self._url(path), headers=self._headers, json={"circuit": circuit})
+        response = self._request(
+            "POST", self._url(path), headers=self._headers, json={"circuit": circuit}
+        )
         if response.status_code == 201:
             job_id = response.json()["id"]
             if self._verbose:
@@ -257,8 +259,7 @@ class Connection:
         """
         path = "/jobs/{}/result".format(job_id)
         response = self._request(
-            "GET", self._url(path),
-            headers={"Accept": "application/x-numpy", **self._headers}
+            "GET", self._url(path), headers={"Accept": "application/x-numpy", **self._headers}
         )
         if response.status_code == 200:
             # Read the numpy binary data in the payload into memory
@@ -287,8 +288,10 @@ class Connection:
         """
         path = "/jobs/{}".format(job_id)
         response = self._request(
-            "PATCH", self._url(path),
-            headers=self._headers, json={"status": JobStatus.CANCELLED.value}
+            "PATCH",
+            self._url(path),
+            headers=self._headers,
+            json={"status": JobStatus.CANCELLED.value},
         )
         if response.status_code == 204:
             if self._verbose:
@@ -316,22 +319,25 @@ class Connection:
         self._headers.pop("Authorization", None)
         path = "/auth/realms/platform/protocol/openid-connect/token"
         headers = {**self._headers}
-        response = self._request("POST", self._url(path), headers=headers, data={
-            'grant_type': 'refresh_token',
-            'refresh_token': self._token,
-            'client_id': 'public',
-        })
+        response = self._request(
+            "POST",
+            self._url(path),
+            headers=headers,
+            data={
+                "grant_type": "refresh_token",
+                "refresh_token": self._token,
+                "client_id": "public",
+            },
+        )
         if response.status_code == 200:
-            self._headers["Authorization"] = "Bearer {}".format(response.json().get('access_token'))
+            self._headers["Authorization"] = "Bearer {}".format(response.json().get("access_token"))
         else:
-            raise RequestFailedError(
-                "Authorization failed for request"
-            )
+            raise RequestFailedError("Authorization failed for request")
 
-    def _request(self, method: str, path: str, headers: Dict = {}, **kwargs ):
+    def _request(self, method: str, path: str, headers: Dict = {}, **kwargs):
         """Wrap all API requests with an authentication token refresh if a 401 status
         is received from the initial request.
-        
+
         Args:
             method (str): the HTTP request method to use
             path (str): path of the endpoint to use
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -89,6 +89,7 @@ class Connection:
     ):
         default_config = load_config()
 
+        print(token)
         self._token = token or default_config["api"]["authentication_token"]
         self._host = host or default_config["api"]["hostname"]
         self._port = port or default_config["api"]["port"]
@@ -333,7 +334,7 @@ class Connection:
             access_token = response.json().get("access_token")
             self._headers["Authorization"] = f"Bearer {access_token}"
         else:
-            raise RequestFailedError("Authorization failed for request.")
+            raise RequestFailedError("Authorization failed for request, please check your token provided.")
 
     def _request(self, method: str, path: str, headers: Dict = None, **kwargs):
         """Wrap all API requests with an authentication token refresh if a 401 status
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -319,8 +319,7 @@ class Connection:
         self._headers.pop("Authorization", None)
         path = "/auth/realms/platform/protocol/openid-connect/token"
         headers = {**self._headers}
-        response = self._request(
-            "POST",
+        response = requests.post(
             self._url(path),
             headers=headers,
             data={
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -98,7 +98,6 @@ class Connection:
         self._base_url = "http{}://{}:{}".format("s" if self.use_ssl else "", self.host, self.port)
 
         self._headers = {"Accept-Version": self.api_version}
-        self._refresh_access_token()
 
         self.log = create_logger(__name__)
 
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -16,7 +16,7 @@ This module provides an interface to a remote program execution backend.
 """
 import io
 from datetime import datetime
-from typing import List
+from typing import List, Dict
 
 import numpy as np
 import requests
@@ -96,7 +96,9 @@ class Connection:
         self._verbose = verbose
 
         self._base_url = "http{}://{}:{}".format("s" if self.use_ssl else "", self.host, self.port)
-        self._headers = {"Authorization": self.token, "Accept-Version": self.api_version}
+
+        self._headers = {"Accept-Version": self.api_version}
+        self._refresh_access_token()
 
         self.log = create_logger(__name__)
 
@@ -156,7 +158,7 @@ class Connection:
     def _get_device_dict(self, target: str) -> dict:
         """Returns the device specifications as a dictionary"""
         path = f"/devices/{target}/specifications"
-        response = requests.get(self._url(path), headers=self._headers)
+        response = self._request("GET", self._url(path), headers=self._headers)
 
         if response.status_code == 200:
             self.log.info("The device spec %s has been successfully retrieved.", target)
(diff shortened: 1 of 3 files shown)


strawberryfields/api/connection.py
```diff
@@ -301,12 +305,43 @@ def ping(self) -> bool:
...
+                "Authorization failed for request"
+            )
+
+    def _request(self, method: str, path: str, headers: Dict = {}, **kwargs ):
+        """ Wrap all API requests with an auth token refresh if a 401 is recevied
+        from the initial request.
+        """
+        request_headers = {**headers, **self._headers}
+        response = requests.request(method, self._url(path), headers=request_headers, **kwargs)
+        if response.status_code == 401:
```
antalszava, line 338: For my own understanding: could we get a 403 here or is it only 401 in every case?

strawberryfields/api/connection.py
```diff
@@ -301,12 +305,43 @@ def ping(self) -> bool:
...
         path = "/healthz"
-        response = requests.get(self._url(path), headers=self._headers)
+        response = self._request("GET", self._url(path), headers=self._headers)
         return response.status_code == 200
 
     def _url(self, path: str) -> str:
         return self._base_url + path
 
+    def _refresh_access_token(self):
+        """ Use the offline token to request a new access token """
```
antalszava, line 315:
```suggestion
        """Use the offline token to request a new access token."""
```

strawberryfields/api/connection.py
```diff
@@ -301,12 +305,43 @@ def ping(self) -> bool:
...
+        if response.status_code == 200:
+            self._headers["Authorization"] = "Bearer {}".format(response.cookies["access_token"])
+        else:
+            raise RequestFailedError(
+                "Authorization failed for request"
+            )
+
+    def _request(self, method: str, path: str, headers: Dict = {}, **kwargs ):
+        """ Wrap all API requests with an auth token refresh if a 401 is recevied
+        from the initial request.
```
antalszava, line 334:
```suggestion
        """Wrap all API requests with an authentication token refresh if a 401 status
        is received from the initial request.
        
        Args:
            method (str): the HTTP request method to use
            path (str): path of the endpoint to use
            headers (dict): dictionary containing the headers of the request

        Returns:
            requests.Response: the response received for the sent request
```

antalszava (commented):
Looks great so far! 💯 Had a few suggestions/questions

strawberryfields/api/connection.py
```diff
@@ -301,12 +305,43 @@ def ping(self) -> bool:
...
 
     def _url(self, path: str) -> str:
         return self._base_url + path
 
+    def _refresh_access_token(self):
+        """ Use the offline token to request a new access token """
+        self._headers.pop("Authorization", None)
+        # TODO: Make sure this is the right path
+        path = "/auth/token"
+        headers = {**self._headers}
```
josh146, line 319: Perhaps better to do the following instead of the pop above?
```suggestion
        headers = {**self._headers}
        headers.pop("Authorization", None)
```

It feels slightly 'safer' than mutating the instance attribute above, even if it has the same effect

josh146: This way, if the method fails midway, there are no side effects

strawberryfields/api/connection.py
```diff
@@ -301,12 +305,43 @@ def ping(self) -> bool:
...
+        response = self._request("POST", self._url(path), headers=headers, data={
+            'grant_type': 'refresh_token',
+            'refresh_token': self._token,
+            'client_id': 'public',
+        })
+        if response.status_code == 200:
+            self._headers["Authorization"] = "Bearer {}".format(response.cookies["access_token"])
```
josh146, line 326: Is this the only place where the token is stored?

Originally, I was wondering if it made sense to store it locally (temporarily), but I suppose it doesn't matter and is cleaner this way.

josh146 (commented):
Thanks @corvust

strawberryfields/api/connection.py
```diff
@@ -301,12 +305,43 @@ def ping(self) -> bool:
...
+        else:
+            raise RequestFailedError(
+                "Authorization failed for request"
+            )
+
+    def _request(self, method: str, path: str, headers: Dict = {}, **kwargs ):
```
josh146, line 332: Better to have `headers=None` here, and then inside the method have
```python
headers = headers or {}
```

strawberryfields/api/connection.py
corvust: No, we should only expect a 401. A 403 would indicate that our token was valid, but the request was rejected because we don't have permission (based on our user profile) to access the requested resource, which isn't something that refreshing the token will fix.

strawberryfields/api/connection.py
corvust: My thought here was that since we are in the refresh token method, we already know the existing Authorization header is invalid, so removing it from the instance headers is a housekeeping action before we replace it with a new one. Whether the method fails midway or not, any future requests are still going to fail regardless of if we've popped the header or not.

strawberryfields/api/connection.py
corvust: Yes the access token is only stored in the Authorization header. I decided against storing it independently since it isn't used anywhere else, is short lived, and a new one can be fetched if it is somehow lost.

strawberryfields/api/connection.py
```diff
@@ -89,14 +89,17 @@ def __init__(
...
         self._host = host or default_config["api"]["hostname"]
         self._port = port or default_config["api"]["port"]
         self._use_ssl = use_ssl or default_config["api"]["use_ssl"]
         self._verbose = verbose
 
         self._base_url = "http{}://{}:{}".format("s" if self.use_ssl else "", self.host, self.port)
-        self._headers = {"Authorization": self.token, "Accept-Version": self.api_version}
+
+        self._headers = {"Accept-Version": self.api_version}
+        self._refresh_access_token()
```
antalszava, line 101: It's worth noting, that due to this call from now on instantiating a `Connection` object will already try to establish connection with the server.

As users are encouraged to use the higher level `RemoteEngine`, this would mean that they could face the following:
```python
In [12]: os.environ["SF_API_AUTHENTICATION_TOKEN"] = "WrongToken"

In [13]: sf.RemoteEngine("X8")
RequestFailedError                        Traceback (most recent call last)
<ipython-input-13-a1e528990489> in <module>
----> 1 sf.RemoteEngine("X8")

~//strawberryfields/strawberryfields/engine.py in __init__(self, target, connection, backend_options)
    548         self._target = self.DEFAULT_TARGETS.get(target, target)
    549         self._spec = None
--> 550         self._connection = connection or Connection()
    551         self._backend_options = backend_options or {}
    552         self.log = create_logger(__name__)

~//strawberryfields/strawberryfields/api/connection.py in __init__(self, token, host, port, use_ssl, verbose)
     99
    100         self._headers = {"Accept-Version": self.api_version}
--> 101         self._refresh_access_token()
    102
    103         self.log = create_logger(__name__)

~//strawberryfields/strawberryfields/api/connection.py in _refresh_access_token(self)
    334             self._headers["Authorization"] = f"Bearer {access_token}"
    335         else:
--> 336             raise RequestFailedError("Authorization failed for request, please check your token provided.")
    337
    338     def _request(self, method: str, path: str, headers: Dict = None, **kwargs):

RequestFailedError: Authorization failed for request, please check your token provided.
```

@josh146 would you think that this causes any complications?

tests/api/test_connection.py
```diff
@@ -72,9 +74,11 @@ def content(self):
 class TestConnection:
     """Tests for the ``Connection`` class."""
 
-    def test_init(self):
+    def test_init(self, monkeypatch):
```
antalszava, line 77:
```suggestion
    def test_init(self):
```

tests/api/test_connection.py
```diff
@@ -75,6 +77,8 @@ class TestConnection:
     def test_init(self):
         """Tests that a ``Connection`` is initialized correctly."""
         token, host, port, use_ssl = "token", "host", 123, True
+
+        monkeypatch.setattr(requests, "request", mock_return(MockResponse(200, {})))
```
antalszava, line 81:
```suggestion
```

tests/api/test_connection.py
```diff
@@ -95,12 +97,13 @@ def test_get_device_spec(self, prog, connection, monkeypatch):
...
         monkeypatch.setattr(
             requests,
-            "get",
+            "request",
             mock_return(MockResponse(
                 200,
                 {"layout": "", "modes": 42, "compiler": [], "gate_parameters": {"param": [[0, 1]]}}
             )),
         )
+        monkeypatch.setattr(connection, "_refresh_access_token", lambda: None)
```
antalszava, line 106:
```suggestion
```

strawberryfields/api/connection.py
josh146: :+1:

strawberryfields/api/connection.py
josh146: Nice. Thanks for adding these @antalszava!

antalszava:
[ch3931]

strawberryfields/api/connection.py
```diff
@@ -301,12 +307,56 @@ def ping(self) -> bool:
...
+                "refresh_token": self._token,
+                "client_id": "public",
+            },
+        )
+        if response.status_code == 200:
+            access_token = response.json().get("access_token")
+            self._headers["Authorization"] = f"Bearer {access_token}"
+        else:
+            raise RequestFailedError(
+                "Authorization failed for request, please check your token provided."
```
jswinarton, line 335:
```suggestion
                "Could not retrieve access token. Please check that your API key is correct."
```
?

jswinarton: By the way, you could do `response.raise_for_status()` instead on line 330, which will raise a requests exception automatically if the status code isn't 200. That could be helpful because you could catch and re-raise the way you're doing here, but chain the exceptions together to get a more detailed traceback that includes the actual status code it failed with.

jswinarton (approved):

strawberryfields/api/connection.py
antalszava: Oh @jswinarton, it seems that it [would allow other non-error response status code](https://2.python-requests.org/en/master/_modules/requests/models/#Response.raise_for_status) though, correct?

strawberryfields/api/connection.py
```diff
@@ -301,12 +307,56 @@ def ping(self) -> bool:
...
-        response = requests.get(self._url(path), headers=self._headers)
+        response = self._request("GET", self._url(path), headers=self._headers)
         return response.status_code == 200
 
     def _url(self, path: str) -> str:
         return self._base_url + path
 
+    def _refresh_access_token(self):
+        """Use the offline token to request a new access token."""
+        self._headers.pop("Authorization", None)
```
thisac, line 318: (minor / thinking aloud) I'm always partial to using `del` instead of `pop` (since the latter unnecessarily returns the item as well), but I see that `pop` might be a cleaner solution, since it works even if the item doesn't exist. :laughing:

strawberryfields/api/connection.py
```diff
@@ -301,12 +307,56 @@ def ping(self) -> bool:
...
+        is received from the initial request.
+
+        Args:
+            method (str): the HTTP request method to use
+            path (str): path of the endpoint to use
+            headers (dict): dictionary containing the headers of the request
+
+        Returns:
+            requests.Response: the response received for the sent request
+        """
```
thisac, line 349: Oh, this explains my question above! :100:

thisac (approved):
Looking great @corvust and @antalszava! As far as I can tell, this seems to do what it's supposed to. :smile: :100:

PR #531, opened 2021-01-20
Fix gaussian ops

nquesada:
Previously the indices used internally in a number of functions in `ops.py` where generated as sets without necessarily being ordered. This PR turns those sets into lists and sorts them.

### DIFF
--- a/strawberryfields/backends/gaussianbackend/ops.py
+++ b/strawberryfields/backends/gaussianbackend/ops.py
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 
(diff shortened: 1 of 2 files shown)


strawberryfields/backends/gaussianbackend/ops.py
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 
-#     http://www.apache.org/licenses/LICENSE-2.0
+#      http://www.apache.org/licenses/LICENSE-2.0
```
nquesada, line 7:
```suggestion
#     http://www.apache.org/licenses/LICENSE-2.0
```

josh146 (approved):

PR #534, opened 2021-01-27
Gkps tests

elib20:
**Context:** Tidies up and tests bosoniccircuit.py

- Tidies up existing and writes new docstrings for bosoniccircuit.py
- Fixes minor bugs, notably a factor error in post-select generaldyne
- Writes unit tests or small integration tests for functions and methods in bosoniccircuit.py

**Benefits:**
- Achieves 100% code coverage for bosoniccircuit.py

**Possible Drawbacks:**

**Related GitHub Issues:**


strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -191,6 +299,8 @@ def mbsqueeze(self, k, r, phi, r_anc, eta_anc, avg):
             X2, Y2 = self.expandXY([k], X, Y)
             self.apply_channel(X2, Y2)
 
+        # Add new ancilla mode, interfere it and measure it
+        # Delete ancilla mode from active list
         if not avg:
             self.add_mode()
```
nquesada, line 305: is it possible to the map without having to add the extra mode?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -337,15 +467,24 @@ def fidelity_coherent(self, alpha, modes=None):
...
 
     def parity_val(self, modes=None):
-        """Expectation value of the parity operator"""
+        r"""Returns the expectation value of the parity operator, which is the
+        value of the Wigner function at the origin.
+
+        Args:
+            modes (list or None): modes to use for parity calculation
+        """
         if modes is None:
```
nquesada, line 488: wondering if we need an `hbar` somewhere in here. The value of parity is hbar independent, while the covariance matrices themselves are. Probably there has to be an `hbar`-depedent scaling of the value of the Wigner function to get the parity.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: We need to sample a value for the homodyne measurement, so this seems like a straightforward way. The extra mode is Gaussian, so it doesn't increase the number of weights, so I don't think it is too memory intensive.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        std = vals[:, 0].std()
+        assert np.isclose(mean, example.hbar * r * np.cos(phi), rtol=0, atol=10 / np.sqrt(500))
+        assert np.isclose(std, 1, rtol=0, atol=10 / np.sqrt(500))
+
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    def test_heterodyne(self, num_weights, r, phi):
+        r"""Checks heterodyne."""
+        # Create coherent state with multiple weights
```
nquesada, line 296: when you say multiple weights for a coherent state you mean that one of them 1 and the rest are zero?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: You are correct! I had changed it in the calculation of parity in states.py, but had forgotten it here. Will add now.

nquesada (approved):

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: I mean all the weights are 1/num_weights, so that they sum to the Wigner function of a coherent state. Each term has the same mean and covariance.

PR #538, opened 2021-01-29
Adds bosoniccircuit.py and its tests

elib20:
**Context:** Adds core functions for the bosonic backend, and relevant tests

**Description of the Change:** The bosoniccircuit.py file contains the relevant functions for manipulating modes in the bosonic backend as a circuit is performed.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -35,6 +35,8 @@ jobs:
           access_token: ${{ github.token }}
 
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
 
       - name: Set up Python
         uses: actions/setup-python@v2
--- a/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
+++ b/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
@@ -943,7 +943,7 @@ class BosonicModes:
         transformations for the whole system.
 
         Args:
-            modes (list): list of modes on which ``(X,Y)``
+            modes (list): list of modes on which ``(X,Y)`` act
             X (array): matrix for mutltiplicative part of transformation
             Y (array): matrix for additive part of transformation.
         """
--- a/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
+++ b/strawberryfields/backends/bosonicbackend/bosoniccircuit.py
@@ -1,0 +1,829 @@
+# Copyright 2020 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Bosonic circuit operations"""
+# pylint: disable=duplicate-code,attribute-defined-outside-init
+import numpy as np
+from scipy.linalg import block_diag
+from thewalrus.quantum import Xmat
+import thewalrus.symplectic as symp
+
+import itertools as it
+from . import ops
+
+from ..shared_ops import changebasis
+
+
+# Shape of the weights, means, and covs arrays.
+def w_shape(num_modes, num_gauss):
+    r"""Calculate total number of weights. Assumes same number of weights per mode.
+
+    Args:
+        num_modes (int): number of modes
+        num_gauss (int): number of gaussian peaks per mode
+
+    Returns:
+        int: number of weights for all Gaussian peaks in phase space
+    """
+    return num_gauss ** num_modes
+
+
+def m_shape(num_modes, num_gauss):
+    r"""Shape of array of mean vectors. Assumes same number of weights per mode.
+
+    Args:
+        num_modes (int): number of modes
+        num_gauss (int): number of gaussian peaks per mode
+
+    Returns:
+        tuple: (number of weights, number of quadratures)
+    """
+    return (num_gauss ** num_modes, 2 * num_modes)
+
+
+def c_shape(num_modes, num_gauss):
+    r"""Shape of array of covariance matricess. Assumes same number of weights per mode.
+
+    Args:
+        num_modes (int): number of modes
+        num_gauss (int): number of gaussian peaks per mode
+
+    Returns:
+        tuple: (number of weights, number of quadratures, number of quadratures)
+    """
+    return (num_gauss ** num_modes, 2 * num_modes, 2 * num_modes)
+
+
+def to_xp(num_modes):
+    r"""Provides array of indices to order quadratures as all x
+    followed by all p, starting from (x1,p1,...,xn,pn) ordering.
+
+    Args:
+        num_modes (int): number of modes
+
+    Returns:
+        array: quadrature ordering for all x followed by all p
+    """
+    return np.concatenate((np.arange(0, 2 * num_modes, 2), np.arange(0, 2 * num_modes, 2) + 1))
+
+
+def from_xp(num_modes):
+    r"""Provides array of indices to order quadratures as (x1,p1,...,xn,pn)
+    starting from all x followed by all p.
+
+    Args:
+        num_modes (int): number of modes
+
+    Returns:
+        list: quadrature ordering for (x1,p1,...,xn,pn)
+    """
+    perm_inds_list = [(i, i + num_modes) for i in range(num_modes)]
+    perm_inds = [a for tup in perm_inds_list for a in tup]
+    return perm_inds
+
+
+def update_means(means, X, perm_out):
+    r"""Apply a linear transformation ``X`` to the array of means. The
+    quadrature ordering can be specified by ``perm_out`` to match ordering
+    of ``X`` to means.
+
+    Args:
+        means (array): array of mean vectors
+        X (array): matrix for linear transformation
+        perm_out (array): indices for quadrature ordering
+
+    Returns:
+        array: transformed array of mean vectors
+    """
+    X_perm = X[:, perm_out][perm_out, :]
+    return (X_perm @ means.T).T
+
+
+def update_covs(covs, X, perm_out, Y=0):
+    r"""Apply a linear transformation parametrized by ``(X,Y)`` to the
+    array of covariance matrices. The  quadrature ordering can be specified
+    by ``perm_out`` to match ordering of ``(X,Y)`` to the covariance matrices.
+
+    If Y is not specified, it defaults to 0.
+
+    Args:
+        covs (array): array of covariance matrices
+        X (array): matrix for mutltiplicative part of transformation
+        perm_out (array): indices for quadrature ordering
+        Y (array or 0): matrix for additive part of transformation.
+
+    Returns:
+        array: transformed array of covariance matrices
+    """
+    X_perm = X[:, perm_out][perm_out, :]
+    if not isinstance(Y, int):
+        Y = Y[:, perm_out][perm_out, :]
+    return (X_perm @ covs @ X_perm.T) + Y
+
+
+class BosonicModes:
+    """A Bosonic circuit class."""
+
+    # pylint: disable=too-many-public-methods
+
+    def __init__(self, num_subsystems=1, num_weights=1):
+        self.hbar = 2
+
+    def add_mode(self, peak_list=[1]):
+        r"""Add len(peak_list) modes to the circuit. Each mode has a number of
+        weights specified by peak_list, and the means and covariances are set
+        to vacuum.
+
+        Args:
+            peak_list (list): list of weights per mode.
+        """
+
+        num_modes = len(peak_list)
+        num_gauss = np.prod(peak_list)
+        self.nlen += num_modes
+
+        # Updated mode index permutation list
+        self.to_xp = to_xp(self.nlen)
+        self.from_xp = from_xp(self.nlen)
+        self.active.append(self.nlen - 1)
+
+        # Weights are set equal to each other and normalized
+        vac_weights = np.array([1 / num_gauss for i in range(num_gauss)], dtype=complex)
+        # New mode means and covs set to vacuum
+        vac_means = np.zeros((num_gauss, 2 * num_modes)).tolist()
+        vac_covs = [
+            ((self.hbar / 2) * np.identity(2 * num_modes)).tolist() for i in range(num_gauss)
+        ]
+
+        # Find all possible combinations of means and combs of the
+        # Gaussians between the modes.
+        mean_combs = it.product(self.means.tolist(), vac_means)
+        cov_combs = it.product(self.covs.tolist(), vac_covs)
+
+        # Tensor product of the weights.
+        weights = np.kron(self.weights, vac_weights)
+        # De-nest the means iterator.
+        means = np.array([[a for b in tup for a in b] for tup in mean_combs])
+        # Stack covs appropriately.
+        covs = np.array([block_diag(*tup) for tup in cov_combs])
+
+        self.weights = weights
+        self.means = means
+        self.covs = covs
+
+    def del_mode(self, modes):
+        r"""Delete modes from the circuit.
+
+        Args:
+            modes (int or list): modes to be deleted.
+        """
+        if isinstance(modes, int):
+            modes = [modes]
+
+        for mode in modes:
+            if self.active[mode] is None:
+                raise ValueError("Cannot delete mode, mode does not exist")
+
+            self.loss(0.0, mode)
+            self.active[mode] = None
+
+    def reset(self, num_subsystems=None, num_weights=1):
+        """Reset the simulation state.
+
+        Args:
+            num_subsystems (int, optional): Sets the number of modes in the reset
+                circuit. ``None`` means unchanged.
+            num_weights (int): Sets the number of gaussians per mode in the
+                superposition.
+        """
+        if num_subsystems is not None:
+            if not isinstance(num_subsystems, int):
+                raise ValueError("Number of modes must be an integer")
+            self.nlen = num_subsystems
+
+        if not isinstance(num_weights, int):
+            raise ValueError("Number of weights must be an integer")
+
+        self.active = list(np.arange(self.nlen, dtype=int))
+        # Mode index permutation list to and from XP ordering.
+        self.to_xp = to_xp(self.nlen)
+        self.from_xp = from_xp(self.nlen)
+
+        self.weights = np.ones(w_shape(self.nlen, num_weights), dtype=complex)
+        self.weights = self.weights / (num_weights ** self.nlen)
+
+        self.means = np.zeros(m_shape(self.nlen, num_weights), dtype=complex)
+        id_covs = [np.identity(2 * self.nlen, dtype=complex) for i in range(len(self.weights))]
+        self.covs = np.array(id_covs)
+
+    def get_modes(self):
+        r"""Return the modes currently active."""
+        return [x for x in self.active if x is not None]
+
+    def displace(self, r, phi, i):
+        r"""Displace mode ``i`` by the amount ``r*np.exp(1j*phi)``.
+
+        Args:
+            r (float): displacement magnitude
+            phi (float): displacement phase
+            i (int): mode to be displaced
+        """
+        if self.active[i] is None:
+            raise ValueError("Cannot displace mode, mode does not exist")
+        self.means += symp.expand_vector(r * np.exp(1j * phi), i, self.nlen)[self.from_xp]
+
+    def squeeze(self, r, phi, k):
+        r"""Squeeze mode ``k`` by the amount ``r*exp(1j*phi)``.
+
+        Args:
+            r (float): squeezing magnitude
+            phi (float): squeezing phase
+            k (int): mode to be squeezed
+        """
+        if self.active[k] is None:
+            raise ValueError("Cannot squeeze mode, mode does not exist")
+
+        sq = symp.expand(symp.squeezing(r, phi), k, self.nlen)
+        self.means = update_means(self.means, sq, self.from_xp)
+        self.covs = update_covs(self.covs, sq, self.from_xp)
+
+    def mbsqueeze(self, k, r, phi, r_anc, eta_anc, avg):
+        r"""Squeeze mode ``k`` by the amount ``r*exp(1j*phi)`` using measurement-based squeezing.
+
+        Either the average map, described by a Gaussian CPTP transformation, or a single-shot map
+        with ancillary measurement outcomes can be simulated.
+
+        Args:
+            k (int): mode to be squeezed
+            r (float): target squeezing magnitude
+            phi (float): target squeezing phase
+            r_anc (float): squeezing magnitude of the ancillary mode
+            eta_anc(float): detection efficiency of the ancillary mode
+            avg (bool): whether to apply the average map or single-shot
+        Returns:
+            float: if single-shot map selected, returns the measurement outcome of the ancilla
+        """
+
+        if self.active[k] is None:
+            raise ValueError("Cannot squeeze mode, mode does not exist")
+
+        # antisqueezing corresponds to an extra phase shift
+        if r < 0:
+            phi += np.pi
+        r = np.abs(r)
+        # beamsplitter angle
+        theta = np.arccos(np.exp(-r))
+        self.phase_shift(-phi / 2, k)
+
+        # Construct (X,Y) for Gaussian CPTP of average map
+        if avg:
+            X = np.diag([np.cos(theta), 1 / np.cos(theta)])
+            Y = np.diag(
+                [
+                    (np.sin(theta) ** 2) * (np.exp(-2 * r_anc)),
+                    (np.tan(theta) ** 2) * (1 - eta_anc) / eta_anc,
+                ]
+            )
+            Y *= self.hbar / 2
+            X2, Y2 = self.expandXY([k], X, Y)
+            self.apply_channel(X2, Y2)
+
+        # Add new ancilla mode, interfere it and measure it
+        # Delete ancilla mode from active list
+        if not avg:
+            self.add_mode()
+            new_mode = self.nlen - 1
+            self.squeeze(r_anc, 0, new_mode)
+            self.beamsplitter(theta, 0, k, new_mode)
+            self.loss(eta_anc, new_mode)
+            self.phase_shift(np.pi / 2, new_mode)
+            val = self.homodyne(new_mode)
+            self.del_mode(new_mode)
+            self.covs = np.delete(self.covs, [2 * new_mode, 2 * new_mode + 1], axis=1)
+            self.covs = np.delete(self.covs, [2 * new_mode, 2 * new_mode + 1], axis=2)
+            self.means = np.delete(self.means, [2 * new_mode, 2 * new_mode + 1], axis=1)
+            self.nlen = self.nlen - 1
+            self.from_xp = from_xp(self.nlen)
+            self.to_xp = to_xp(self.nlen)
+            self.active = self.active[:new_mode]
+            prefac = -np.tan(theta) / np.sqrt(2 * self.hbar * eta_anc)
+            self.displace(prefac * val[0][0], np.pi / 2, k)
+        self.phase_shift(phi / 2, k)
+
+        if not avg:
+            return val
+
+    def phase_shift(self, phi, k):
+        r"""Implement a phase shift in mode k.
+
+        Args:
+           phi (float): phase
+           k (int): mode to be phase shifted
+        """
+        if self.active[k] is None:
+            raise ValueError("Cannot phase shift mode, mode does not exist")
+
+        rot = symp.expand(symp.rotation(phi), k, self.nlen)
+        self.means = update_means(self.means, rot, self.from_xp)
+        self.covs = update_covs(self.covs, rot, self.from_xp)
+
+    def beamsplitter(self, theta, phi, k, l):
+        r"""Implement a beam splitter operation between modes k and l.
+
+        Args:
+            theta (float): real beamsplitter angle
+            phi (float): complex beamsplitter angle
+            k (int): first mode
+            l (int): second mode
+        """
+        if self.active[k] is None or self.active[l] is None:
+            raise ValueError("Cannot perform beamsplitter, mode(s) do not exist")
+
+        if k == l:
+            raise ValueError("Cannot use the same mode for beamsplitter inputs")
+
+        # Cross-check with Gaussian backend.
+        bs = symp.expand(symp.beam_splitter(theta, phi), [k, l], self.nlen)
+        self.means = update_means(self.means, bs, self.from_xp)
+        self.covs = update_covs(self.covs, bs, self.from_xp)
+
+    def scovmatxp(self):
+        r"""Returns the symmetric ordered array of covariance matrices
+        in the :math:`q_1,...,q_n,p_1,...,p_n` ordering.
+        """
+        return self.covs[:, self.to_xp][..., self.to_xp]
+
+    def smeanxp(self):
+        r"""Returns the symmetric ordered array of means in the
+        :math:`q_1,...,q_n,p_1,...,p_n` ordering.
+        """
+        return self.means.T[self.to_xp].T
+
+    def scovmat(self):
+        r"""Returns the symmetric ordered array of covariance matrices
+        in the :math:`q_1,p_1,...,q_n,p_n` ordering.
+        """
+        return self.covs
+
+    def smean(self):
+        r"""Returns the symmetric ordered array of means
+        in the :math:`q_1,p_1,...,q_n,p_n` ordering.
+        """
+        return self.means
+
+    def sweights(self):
+        """Returns the array of weights."""
+        return self.weights
+
+    def fromsmean(self, r, modes=None):
+        r"""Populates the array of means from a provided array of means.
+
+        The input must already have performed the scaling of the means by self.hbar,
+        and must be sorted in ascending order.
+
+        Args:
+            r (array): vector of means in :math:`(x_1,p_1,x_2,p_2,\dots)` ordering
+            modes (Sequence): sequence of modes corresponding to the vector of means
+        """
+        if modes is None:
+            modes = range(self.nlen)
+
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        self.means[:, mode_ind] = r
+
+    def fromscovmat(self, V, modes=None):
+        r"""Populates the array of covariance matrices from a provided array of covariance matrices.
+
+        The input must already have performed the scaling of the means by self.hbar,
+        and must be sorted in ascending order.
+
+        Args:
+            V (array): covariance matrix in :math:`(x_1,p_1,x_2,p_2,\dots)` ordering
+            modes (sequence): sequence of modes corresponding to the covariance matrix
+        """
+        if modes is None:
+            n = len(V) // 2
+            modes = np.arange(self.nlen)
+
+            if n != self.nlen:
+                raise ValueError(
+                    "Covariance matrix is the incorrect size, does not match means vector."
+                )
+        else:
+            n = len(modes)
+            modes = np.array(modes)
+            if n > self.nlen:
+                raise ValueError("Covariance matrix is larger than the number of subsystems.")
+
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        # Delete current entries of covariance and any correlations to other modes
+        self.covs[:, mode_ind, :] = 0
+        self.covs[:, :, mode_ind] = 0
+        # Set new covariance elements
+        self.covs[np.ix_(np.arange(self.covs.shape[0], dtype=int), mode_ind, mode_ind)] = V
+
+    def fidelity_coherent(self, alpha, modes=None):
+        r"""Returns the fidelity to a coherent state.
+
+        Args:
+            alpha (array): amplitudes for coherent states
+            modes (list or None): modes to use for fidelity calculation
+        """
+        if modes is None:
+            modes = self.get_modes()
+        # Sort by (q1,p1,q2,p2,...)
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        # Construct mean vector for coherent state
+        alpha_mean = np.array([])
+        for i in range(len(modes)):
+            alpha_mean = np.append(alpha_mean, alpha.real[i] * np.sqrt(2 * self.hbar))
+            alpha_mean = np.append(alpha_mean, alpha.imag[i] * np.sqrt(2 * self.hbar))
+        # Construct difference of coherent state mean vector with means of all peaks in the state
+        deltas = self.means[:, mode_ind] - alpha_mean
+        # Construct sum of coherent state covariance matrix and all covariances in the state
+        cov_sum = (
+            self.covs[:, mode_ind, :][:, :, mode_ind] + self.hbar * np.eye((len(mode_ind))) / 2
+        )
+        # Sum all Gaussian peaks
+        exp_arg = np.einsum("...j,...jk,...k", deltas, np.linalg.inv(cov_sum), deltas)
+        weighted_exp = (
+            np.array(self.weights)
+            * self.hbar ** len(modes)
+            * np.exp(-0.5 * exp_arg)
+            / np.sqrt(np.linalg.det(cov_sum))
+        )
+        fidelity = np.sum(weighted_exp)
+        return fidelity
+
+    def fidelity_vacuum(self, modes=None):
+        r"""Returns the fidelity to the vacuum.
+
+        Args:
+            modes (list or None): modes to use for fidelity calculation
+        """
+        if modes is None:
+            modes = self.get_modes()
+        alpha = np.zeros(len(modes))
+        fidelity = self.fidelity_coherent(alpha, modes=modes)
+        return fidelity
+
+    def parity_val(self, modes=None):
+        r"""Returns the expectation value of the parity operator, which is the
+        value of the Wigner function at the origin.
+
+        Args:
+            modes (list or None): modes to use for parity calculation
+        """
+        if modes is None:
+            modes = list(range(self.nlen))
+        # Sort by (q1,p1,q2,p2,...)
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        exp_arg = np.einsum(
+            "...j,...jk,...k",
+            self.means[:, mode_ind],
+            np.linalg.inv(self.covs[:, mode_ind, :][:, :, mode_ind]),
+            self.means[:, mode_ind],
+        )
+        weighted_exp = (
+            np.array(self.weights)
+            * np.exp(-0.5 * exp_arg)
+            / np.sqrt(np.linalg.det(self.covs[:, mode_ind, :][:, :, mode_ind]))
+        )
+        parity = np.sum(weighted_exp) * (self.hbar / 2) ** len(modes)
+        return parity
+
+    def loss(self, T, k):
+        r"""Implements a loss channel in mode k.
+
+        Args:
+            T (float between 0 and 1): loss amount is \sqrt{T}
+            k (int): mode that loses energy
+        """
+
+        if self.active[k] is None:
+            raise ValueError("Cannot apply loss channel, mode does not exist")
+
+        X = np.sqrt(T) * np.identity(2)
+        Y = self.hbar * (1 - T) * np.identity(2) / 2
+        X2, Y2 = self.expandXY([k], X, Y)
+        self.apply_channel(X2, Y2)
+
+    def thermal_loss(self, T, nbar, k):
+        r"""Implements the thermal loss channel in mode k.
+
+        Args:
+            T (float between 0 and 1): loss amount is \sqrt{T}
+            nbar (float): mean photon number of the thermal bath
+            k (int): mode that undegoes thermal loss
+        """
+        if self.active[k] is None:
+            raise ValueError("Cannot apply loss channel, mode does not exist")
+        X = np.sqrt(T) * np.identity(2)
+        Y = self.hbar * (1 - T) * (2 * nbar + 1) * np.identity(2) / 2
+        X2, Y2 = self.expandXY([k], X, Y)
+        self.apply_channel(X2, Y2)
+
+    def init_thermal(self, nbar, mode):
+        r"""Initializes a state of mode in a thermal state with the given population.
+
+        Args:
+            nbar (float): mean photon number of the thermal state
+            mode (int): mode that get initialized
+        """
+        self.thermal_loss(0.0, nbar, mode)
+
+    def is_vacuum(self, tol=0.0):
+        r"""Checks if the state is vacuum by calculating its fidelity with vacuum.
+
+        Args:
+            tol (float): the closeness tolerance to fidelity of 1
+        """
+        fid = self.fidelity_vacuum()
+        return np.abs(fid - 1) <= tol
+
+    def measure_dyne(self, covmat, indices, shots=1):
+        r"""Performs general-dyne measurements on a set of modes.
+
+        For more information see Quantum Continuous Variables: A Primer of Theoretical Methods
+        by Alessio Serafini page 129.
+
+        Args:
+            covmat (array): covariance matrix of the generaldyne measurement
+            indices (list): modes to be measured
+            shots (int): how many measurements are performed
+
+        Returns:
+            array: measurement outcome corresponding to a point in phase space
+        """
+        if covmat.shape != (2 * len(indices), 2 * len(indices)):
+            raise ValueError("Covariance matrix size does not match indices provided")
+
+        if np.linalg.det(covmat) < (self.hbar / 2) ** (2 * len(indices)):
+            raise ValueError("Measurement covariance matrix is unphysical.")
+
+        if self.covs.imag.any():
+            raise NotImplementedError("Covariance matrices must be real")
+
+        for i in indices:
+            if self.active[i] is None:
+                raise ValueError("Cannot apply measurement, mode does not exist")
+
+        expind = np.concatenate((2 * np.array(indices), 2 * np.array(indices) + 1))
+        vals = np.zeros((shots, 2 * len(indices)))
+        imag_means_ind = np.where(self.means[:, expind].imag.any(axis=1))[0]
+        nonneg_weights_ind = np.where(np.angle(self.weights) != np.pi)[0]
+        ub_ind = np.union1d(imag_means_ind, nonneg_weights_ind)
+        ub_weights = np.abs(np.array(self.weights))
+        if len(imag_means_ind):
+            ub_weights[imag_means_ind] *= np.exp(
+                0.5
+                * np.einsum(
+                    "...j,...jk,...k",
+                    (self.means[imag_means_ind, :][:, expind].imag),
+                    np.linalg.inv(
+                        self.covs[imag_means_ind, :, :][:, expind, :][:, :, expind].real + covmat
+                    ),
+                    (self.means[imag_means_ind, :][:, expind].imag),
+                )
+            )
+        ub_weights = ub_weights[ub_ind]
+        ub_weights_prob = ub_weights / np.sum(ub_weights)
+
+        for i in range(shots):
+            drawn = False
+            while not drawn:
+                peak_ind_sample = np.random.choice(ub_ind, size=1, p=ub_weights_prob)[0]
+
+                cov_meas = self.covs[peak_ind_sample, expind, :][:, expind].real + covmat
+                peak_sample = np.random.multivariate_normal(
+                    self.means[peak_ind_sample, expind].real, cov_meas
+                )
+
+                exp_arg = np.einsum(
+                    "...j,...jk,...k",
+                    (peak_sample - self.means[:, expind]),
+                    np.linalg.inv(self.covs[:, expind, :][:, :, expind].real + covmat),
+                    (peak_sample - self.means[:, expind]),
+                )
+                ub_exp_arg = np.copy(exp_arg)
+                if len(imag_means_ind):
+                    ub_exp_arg[imag_means_ind] = np.einsum(
+                        "...j,...jk,...k",
+                        (peak_sample - self.means[imag_means_ind, :][:, expind].real),
+                        np.linalg.inv(
+                            self.covs[imag_means_ind, :, :][:, expind, :][:, :, expind].real
+                            + covmat
+                        ),
+                        (peak_sample - self.means[imag_means_ind, :][:, expind].real),
+                    )
+                prob_dist_val = np.real_if_close(
+                    np.sum(
+                        (
+                            np.array(self.weights)
+                            / np.sqrt(
+                                np.linalg.det(
+                                    2
+                                    * np.pi
+                                    * (self.covs[:, expind, :][:, :, expind].real + covmat)
+                                )
+                            )
+                        )
+                        * np.exp(-0.5 * exp_arg)
+                    )
+                )
+                prob_upbnd = np.real_if_close(
+                    np.sum(
+                        (
+                            ub_weights
+                            / np.sqrt(
+                                np.linalg.det(
+                                    2
+                                    * np.pi
+                                    * (
+                                        self.covs[ub_ind, :, :][:, expind, :][:, :, expind].real
+                                        + covmat
+                                    )
+                                )
+                            )
+                        )
+                        * np.exp(-0.5 * ub_exp_arg[ub_ind])
+                    )
+                )
+                vertical_sample = np.random.random(size=1) * prob_upbnd
+                if vertical_sample < prob_dist_val:
+                    drawn = True
+                    vals[i] = peak_sample
+        # The next line is a hack in that it only updates conditioned on the first samples value
+        # should still work if shots = 1
+        if len(indices) < len(self.active):
+            self.post_select_generaldyne(covmat, indices, vals[0])
+
+        # If all modes are measured, set them to vacuum
+        if len(indices) == len(self.active):
+            for i in indices:
+                self.loss(0, i)
+
+        return vals
+
+    def homodyne(self, n, shots=1, eps=0.0002):
+        r"""Performs an x-homodyne measurement on a mode, simulated by a generaldyne
+        onto a highly squeezed state.
+
+        Args:
+            n (int): mode to be measured
+            shots (int): how many measurements are performed
+            eps (int): squeezing of the measurement state
+
+        Returns:
+            array: homodyne outcome
+        """
+        covmat = self.hbar * np.diag(np.array([eps ** 2, 1.0 / eps ** 2])) / 2
+        return self.measure_dyne(covmat, [n], shots=shots)
+
+    def heterodyne(self, n, shots=1):
+        r"""Performs a heterodyne measurement on a mode, simulated by a generaldyne
+        onto a coherent state.
+
+        Args:
+            n (int): mode to be measured
+            shots (int): how many measurements are performed
+
+        Returns:
+            array: heterodyne outcome
+        """
+        covmat = self.hbar * np.eye(2) / 2
+        return self.measure_dyne(covmat, [n], shots=shots)
+
+    def post_select_generaldyne(self, covmat, indices, vals):
+        r"""Simulates general-dyne measurement on a set of modes with a specified measurement
+        outcome.
+
+        Args:
+            covmat (array): covariance matrix of the generaldyne measurement
+            indices (list): modes to be measured
+            vals (array): measurement outcome to postselect
+        """
+        if covmat.shape != (2 * len(indices), 2 * len(indices)):
+            raise ValueError("Covariance matrix size does not match indices provided")
+
+        for i in indices:
+            if self.active[i] is None:
+                raise ValueError("Cannot apply measurement, mode does not exist")
+
+        expind = np.concatenate((2 * np.array(indices), 2 * np.array(indices) + 1))
+        mp = self.scovmat()
+        (A, B, C) = ops.chop_in_blocks_multi(mp, expind)
+        V = A - B @ np.linalg.inv(C + covmat) @ B.transpose(0, 2, 1)
+        self.covs = ops.reassemble_multi(V, expind)
+
+        r = self.smean()
+        (va, vc) = ops.chop_in_blocks_vector_multi(r, expind)
+        va = va + np.einsum("...ij,...j", B @ np.linalg.inv(C + covmat), (vals - vc))
+        self.means = ops.reassemble_vector_multi(va, expind)
+
+        reweights_exp_arg = np.einsum(
+            "...j,...jk,...k", (vals - vc), np.linalg.inv(C + covmat), (vals - vc)
+        )
+        reweights = np.exp(-0.5 * reweights_exp_arg) / (
+            np.sqrt(np.linalg.det(2 * np.pi * (C + covmat)))
+        )
+        self.weights *= reweights
+        self.weights /= np.sum(self.weights)
+
+        self.means = self.means[abs(self.weights) > 0]
+        self.covs = self.covs[abs(self.weights) > 0]
+        self.weights = self.weights[abs(self.weights) > 0]
+
+    def post_select_homodyne(self, n, val, eps=0.0002, phi=0):
+        r"""Simulates a homodyne measurement on a mode, postselecting on an outcome.
+
+        Args:
+            n (int): mode to be measured
+            val (array): measurement value to post-select
+            eps (int): squeezing of the measurement state
+            phi (float): homodyne angle
+        """
+        if self.active[n] is None:
+            raise ValueError("Cannot apply homodyne measurement, mode does not exist")
+        self.phase_shift(phi, n)
+        covmat = self.hbar * np.diag(np.array([eps ** 2, 1.0 / eps ** 2])) / 2
+        indices = [n]
+        vals = np.array([val, 0])
+        self.post_select_generaldyne(covmat, indices, vals)
+
+    def post_select_heterodyne(self, n, alpha_val):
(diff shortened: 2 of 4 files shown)


strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,831 @@
...
+
+    def is_vacuum(self, tol=0.0):
+        r"""Checks if the state is vacuum by calculating its fidelity with vacuum.
+
+        Args:
+            tol (float): the closeness tolerance to fidelity of 1
+        """
+        fid = self.fidelity_vacuum()
+        return np.abs(fid - 1) <= tol
+
```
nquesada, line 556:
```suggestion
# pylint: disable=too-many-branches
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,831 @@
...
+    Returns:
+        array: transformed array of covariance matrices
+    """
+    X_perm = X[:, perm_out][perm_out, :]
+    if not isinstance(Y, int):
+        Y = Y[:, perm_out][perm_out, :]
+    return (X_perm @ covs @ X_perm.T) + Y
+
+
```
nquesada, line 131:
```suggestion
# pylint: too-many-instance-attributes
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+    def expandS(self, modes, S):
+        """Expands symplectic matrix on subset of modes to symplectic matrix for the whole system.
+
+        Args:
+            modes (list): list of modes on which S acts
+            S (array): symplectic matrix
+        """
+        return symp.expand(S, modes, self.nlen)
+
+    def expandXY(self, modes, X, Y):
```
nquesada, line 814: can't you do this by calling `symp.expand` twice?

nquesada: I think I get the point now, for `Y2` you pad with zeros even along the diagonal.

nquesada: You could do that explicitly though, by calling `symp.expand` and then doing a for loop loop over the diagonal elements of the untouched modes.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+import numpy as np
+from scipy.linalg import block_diag
+import thewalrus.symplectic as symp
+
+from strawberryfields.backends.bosonicbackend import ops
+
+
+# Shape of the weights, means, and covs arrays.
+def w_shape(num_modes, num_gauss):
+    r"""Calculate total number of weights. Assumes same number of weights per mode.
```
antalszava, line 27: (Minor): some of the docstrings might not necessarily need to be raw strings (no special characters)
```suggestion
    """Calculate total number of weights. Assumes same number of weights per mode.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+    r"""Shape of array of covariance matricess. Assumes same number of weights per mode.
+
+    Args:
+        num_modes (int): number of modes
+        num_gauss (int): number of gaussian peaks per mode
+
+    Returns:
+        tuple: (number of weights, number of quadratures, number of quadratures)
+    """
+    return (num_gauss ** num_modes, 2 * num_modes, 2 * num_modes)
```
antalszava, line 62: (Minor)
```suggestion
    num_quad = 2 * num_modes
    return (num_gauss ** num_modes, num_quad, num_quad)
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+
+        Args:
+            peak_list (list): list of weights per mode.
+        """
+        if peak_list is None:
+            peak_list = [1]
+
+        num_modes = len(peak_list)
+        num_gauss = np.prod(peak_list)
+        self.nlen += num_modes
```
antalszava, line 154: How about initializing this attribute in `__init__`? (Perhaps by calling `reset`)

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        Either the average map, described by a Gaussian CPTP transformation, or a single-shot map
+        with ancillary measurement outcomes can be simulated.
+
+        Args:
+            k (int): mode to be squeezed
+            r (float): target squeezing magnitude
+            phi (float): target squeezing phase
+            r_anc (float): squeezing magnitude of the ancillary mode
+            eta_anc(float): detection efficiency of the ancillary mode
+            avg (bool): whether to apply the average map or single-shot
```
antalszava, line 273:
```suggestion
            avg (bool): whether to apply the average map or single-shot

```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            k (int): mode to be squeezed
+        """
+        if self.active[k] is None:
+            raise ValueError("Cannot squeeze mode, mode does not exist")
+
+        sq = symp.expand(symp.squeezing(r, phi), k, self.nlen)
+        self.means = update_means(self.means, sq, self.from_xp)
+        self.covs = update_covs(self.covs, sq, self.from_xp)
+
+    def mbsqueeze(self, k, r, phi, r_anc, eta_anc, avg):
```
antalszava, line 261: How about having `mbsqueeze` -> `mb_squeeze`?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+                    (np.tan(theta) ** 2) * (1 - eta_anc) / eta_anc,
+                ]
+            )
+            Y *= self.hbar / 2
+            X2, Y2 = self.expandXY([k], X, Y)
+            self.apply_channel(X2, Y2)
+
+        # Add new ancilla mode, interfere it and measure it
+        # Delete ancilla mode from active list
+        elif not avg:
```
antalszava, line 304:
```suggestion
        else:
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            self.squeeze(r_anc, 0, new_mode)
+            self.beamsplitter(theta, 0, k, new_mode)
+            self.loss(eta_anc, new_mode)
+            self.phase_shift(np.pi / 2, new_mode)
+            val = self.homodyne(new_mode)
+            self.del_mode(new_mode)
+            self.covs = np.delete(self.covs, [2 * new_mode, 2 * new_mode + 1], axis=1)
+            self.covs = np.delete(self.covs, [2 * new_mode, 2 * new_mode + 1], axis=2)
+            self.means = np.delete(self.means, [2 * new_mode, 2 * new_mode + 1], axis=1)
+            self.nlen = self.nlen - 1
```
antalszava, line 316: Would we not like to make this change in `del_mode`?
```suggestion
            self.nlen -= 1
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        if imag_means_ind.size > 0:
+            ub_weights[imag_means_ind] *= np.exp(
+                0.5
+                * np.einsum(
+                    "...j,...jk,...k",
+                    (self.means[imag_means_ind, :][:, expind].imag),
+                    np.linalg.inv(
+                        self.covs[imag_means_ind, :, :][:, expind, :][:, :, expind].real + covmat
+                    ),
+                    (self.means[imag_means_ind, :][:, expind].imag),
```
antalszava, line 601: Would suggest creating objects with descriptive names so that this operation is more readable (e.g., an object for the result of `einsum`, etc.).

This would also apply to upcoming occurrences of more complex multi-line function calls (`einsum`, `np.sum`)

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            self.active = self.active[:new_mode]
+            prefac = -np.tan(theta) / np.sqrt(2 * self.hbar * eta_anc)
+            self.displace(prefac * val[0][0], np.pi / 2, k)
+
+        self.phase_shift(phi / 2, k)
+
+        if not avg:
+            return val
+
+        return None
```
antalszava, line 328: Maybe this method could be split into two based on the `avg` boolean (separate average map and single-shot cases).

Although there would be some code duplication, it is somewhat unexpected that based on the value of `avg` we either return the measurement result or simply just update the circuit.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+NUM_NEW_WEIGHTS_VALS = [1, 2, 3]
+R_VALS = np.array([-1.0, -1 / 3, 1 / 2, 1.2])
+PHI_VALS = np.array([-1.2, -1 / 5, 0.6, 1.3])
+NBAR_VALS = np.array([0, 1, 2.5])
+NUM_SHOTS_VALS = np.array([1, 2, 10])
+R_ANC_VALS = np.array([0.1, 0.8, 1.2, 5])
+ETA_ANC_VALS = np.array([0.1, 0.4, 0.9, 1])
+
+
+class TestBosonicCircuit:
```
antalszava, line 36: Suggest aligning this with the name of the class (currently `BosonicModes`, but perhaps that will change?)

antalszava (commented):
Hi @elib20 leaving a couple of preliminary comments, will continue further.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            k (int): first mode
+            l (int): second mode
+        """
+        if self.active[k] is None or self.active[l] is None:
+            raise ValueError("Cannot perform beamsplitter, mode(s) do not exist")
+
+        if k == l:
+            raise ValueError("Cannot use the same mode for beamsplitter inputs")
+
+        # Cross-check with Gaussian backend.
```
antalszava, line 359: What does this comment refer to, would it be the remnant of a todo? :thinking:

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        bs = symp.expand(symp.beam_splitter(theta, phi), [k, l], self.nlen)
+        self.means = update_means(self.means, bs, self.from_xp)
+        self.covs = update_covs(self.covs, bs, self.from_xp)
+
+    def scovmatxp(self):
+        r"""Returns the symmetric ordered array of covariance matrices
```
antalszava, line 365: Perhaps worth noting in this and upcoming docstrings what the original ordering was?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+
+    def sweights(self):
+        """Returns the array of weights."""
+        return self.weights
+
+    def fromsmean(self, r, modes=None):
+        r"""Populates the array of means from a provided array of means.
+
+        The input must already have performed the scaling of the means by self.hbar,
+        and must be sorted in ascending order.
```
antalszava, line 396: Maybe worth being explicit here? Maybe mistaken, but ascending could be confused with the :math:`(x_1,\dots,x_n,p_1,\dots,p_n,)` ordering?
```suggestion
        and must be sorted :math:`(x_1,p_1,x_2,p_2,\dots)` order.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        if modes is None:
+            modes = range(self.nlen)
+
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        self.means[:, mode_ind] = r
+
+    def fromscovmat(self, V, modes=None):
+        r"""Populates the array of covariance matrices from a provided array of covariance matrices.
+
+        The input must already have performed the scaling of the means by self.hbar,
```
antalszava, line 411:
```suggestion
        The input must already have performed the scaling of the covariances by self.hbar,
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+    def is_vacuum(self, tol=0.0):
+        r"""Checks if the state is vacuum by calculating its fidelity with vacuum.
+
+        Args:
+            tol (float): the closeness tolerance to fidelity of 1
+        """
+        fid = self.fidelity_vacuum()
+        return np.abs(fid - 1) <= tol
+
+    def measure_dyne(self, covmat, indices, shots=1):
```
antalszava, line 558: (Though saw that such a renaming would break the convention with the Gaussian `Circuit`, maybe worth considering)
```suggestion
    def measure_dyne(self, covmat, modes, shots=1):
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        onto a coherent state.
+
+        Args:
+            n (int): mode to be measured
+            shots (int): how many measurements are performed
+
+        Returns:
+            array: heterodyne outcome
+        """
+        covmat = self.hbar * np.eye(2) / 2
```
antalszava, line 709: If we now that `self.hbar` would not change in the backend, hardcoding in `covmat` might bring some slight performance advantage

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        if self.active[n] is None:
+            raise ValueError("Cannot apply homodyne measurement, mode does not exist")
+        self.phase_shift(phi, n)
+        covmat = self.hbar * np.diag(np.array([eps ** 2, 1.0 / eps ** 2])) / 2
+        indices = [n]
+        vals = np.array([val, 0])
+        self.post_select_generaldyne(covmat, indices, vals)
+
+    def post_select_heterodyne(self, n, alpha_val):
```
antalszava, line 769:
```suggestion
    def post_select_heterodyne(self, mode, alpha_val):
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        Us = symp.interferometer(U)
+        self.means = update_means(self.means, Us, self.from_xp)
+        self.covs = update_covs(self.covs, Us, self.from_xp)
+
+    def apply_channel(self, X, Y):
+        r"""Transforms the state according to a deterministic Gaussian CPTP map.
+
+        Args:
+            X (array): matrix for mutltiplicative part of transformation
+            Y (array): matrix for additive part of transformation.
```
antalszava, line 800: Would we want to programatically pose some restrictions on `X` and `Y` here?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        X2 = symp.expand(X, modes, self.nlen)
+        M = len(Y) // 2
+        Y2 = np.zeros((2 * self.nlen, 2 * self.nlen), dtype=Y.dtype)
+        w = np.array(modes)
+
+        Y2[w.reshape(-1, 1), w.reshape(1, -1)] = Y[:M, :M].copy()
+        Y2[(w + self.nlen).reshape(-1, 1), (w + self.nlen).reshape(1, -1)] = Y[M:, M:].copy()
+        Y2[w.reshape(-1, 1), (w + self.nlen).reshape(1, -1)] = Y[:M, M:].copy()
+        Y2[(w + self.nlen).reshape(-1, 1), w.reshape(1, -1)] = Y[M:, :M].copy()
```
antalszava, line 831: Might be worth storing some of the repeating reshaped weights as separate objects

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
nquesada: I'd still prefer to use `expand` and set the diagonal elements to zero than do all of this numpy indexing black magiv.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+
+    Returns:
+        array: quadrature ordering for all x followed by all p
+    """
+    return np.concatenate((np.arange(0, 2 * num_modes, 2), np.arange(0, 2 * num_modes, 2) + 1))
+
+
+def from_xp(num_modes):
+    r"""Provides array of indices to order quadratures as (x1,p1,...,xn,pn)
+    starting from all x followed by all p.
```
elib20, line 80:
```suggestion
    starting from all x followed by all p i.e., (x1,...,xn,p1,..., pn).
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        array: transformed array of covariance matrices
+    """
+    X_perm = X[:, perm_out][perm_out, :]
+    if not isinstance(Y, int):
+        Y = Y[:, perm_out][perm_out, :]
+    return (X_perm @ covs @ X_perm.T) + Y
+
+
+# pylint: disable=too-many-instance-attributes
+class BosonicModes:
```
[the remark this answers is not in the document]
elib20: I think we just named it symmetrically to how the gaussiancircuit.py file has GaussianModes as the class.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20:
```suggestion
    def mb_squeeze(self, k, r, phi, r_anc, eta_anc, avg):
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I think the spirit of del_mode is to delete modes that the user programs in, but the record that the mode was there is via self.nlen. Whereas here we really don't want the user to be able to access this ancillary mode, except for the measurement outcome from it.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: yes it was

elib20:
```suggestion
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            raise ValueError("Cannot use the same mode for beamsplitter inputs")
+
+        # Cross-check with Gaussian backend.
+        bs = symp.expand(symp.beam_splitter(theta, phi), [k, l], self.nlen)
+        self.means = update_means(self.means, bs, self.from_xp)
+        self.covs = update_covs(self.covs, bs, self.from_xp)
+
+    def scovmatxp(self):
+        r"""Returns the symmetric ordered array of covariance matrices
+        in the :math:`q_1,...,q_n,p_1,...,p_n` ordering.
```
elib20, line 366:
```suggestion
        in the :math:`q_1,...,q_n,p_1,...,p_n` ordering, given that
        they were stored in the :math:`q_1,p_1,...,q_n,p_n` ordering.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        return self.covs[:, self.to_xp][..., self.to_xp]
+
+    def smeanxp(self):
+        r"""Returns the symmetric ordered array of means in the
+        :math:`q_1,...,q_n,p_1,...,p_n` ordering.
```
elib20, line 372:
```suggestion
        :math:`q_1,...,q_n,p_1,...,p_n` ordering, given that they 
        were stored in the :math:`q_1,p_1,...,q_n,p_n` ordering.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        return self.means.T[self.to_xp].T
+
+    def scovmat(self):
+        r"""Returns the symmetric ordered array of covariance matrices
+        in the :math:`q_1,p_1,...,q_n,p_n` ordering.
```
elib20, line 378:
```suggestion
        in the :math:`q_1,p_1,...,q_n,p_n` ordering, just as 
        they were stored.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        return self.covs
+
+    def smean(self):
+        r"""Returns the symmetric ordered array of means
+        in the :math:`q_1,p_1,...,q_n,p_n` ordering.
```
elib20, line 384:
```suggestion
        in the :math:`q_1,p_1,...,q_n,p_n` ordering, just as
        they were stored.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I see my copy-paste has been found :)

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I think we can just go with Nico's suggestion to do it another way then it will be less confusing to follow.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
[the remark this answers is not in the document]
antalszava: Oh, I see! Seems like the reason is the different names of the Gaussian and Fock circuits (`GaussianModes` in the Gaussian case, while both Fock and TF backends having a `Circuit` class).

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            modes = [modes]
+
+        for mode in modes:
+            if self.active[mode] is None:
+                raise ValueError("Cannot delete mode, mode does not exist")
+
+            self.loss(0.0, mode)
+            self.active[mode] = None
+
+    def reset(self, num_subsystems=None, num_weights=1):
```
antalszava, line 201: Could there be cases where we would like to provision the weights ourselves, instead of initializing them to an equal superposition?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            self.post_select_generaldyne(covmat, indices, vals[0])
+
+        # If all modes are measured, set them to vacuum
+        if len(indices) == len(self.active):
+            for i in indices:
+                self.loss(0, i)
+
+        return vals
+
+    def homodyne(self, n, shots=1, eps=0.0002):
```
antalszava, line 683: Any pointers on the choosing the default value of `eps`?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+    def init_thermal(self, nbar, mode):
+        r"""Initializes a state of mode in a thermal state with the given population.
+
+        Args:
+            nbar (float): mean photon number of the thermal state
+            mode (int): mode that get initialized
+        """
+        self.thermal_loss(0.0, nbar, mode)
+
+    def is_vacuum(self, tol=0.0):
```
antalszava, line 549: Could it be worth setting a non-zero, low enough tolerance here as default (e.g., `1e-10`)? Would think that numeric precision would affect the computation anyways.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("num_modes", NUM_MODES_VALS)
+    def test_reset_circuit(self, num_weights, num_modes):
+        r"""Checks that the reset method instantiates correct number of modes
+        as vacuum, each with a given number of weights."""
+        # Create a state over 3 modes, 2 weights per mode
+        example = circuit.BosonicModes()
+        example.reset(3, 2)
+
+        # Confirm number of modes and weights
```
antalszava, line 48: Would suggest splitting this test case into multiple test cases for `reset`. There is a notion of "one `assert` per test case": though that might be a bit pragmatic idea in this case, it can be worth grouping similar assertions into the test cases (e.g., could create 2-3 short tests from this one).

Also, initializing tests that are similar in nature (e.g., all create a 3 modes, 2 weight bosonic circuit) could be done using pytest fixtures.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        # Check mode-wise and quadrature-wise ordering
+        assert np.allclose(example.to_xp, np.array([0, 2, 4, 1, 3, 5]))
+        assert np.allclose(example.from_xp, np.array([0, 3, 1, 4, 2, 5]))
+
+        # Confirm number of weights, means and covs
+        assert example.sweights().shape == (2 ** 3,)
+        assert example.smean().shape == (2 ** 3, 6)
+        assert example.scovmat().shape == circuit.c_shape(3, 2)
+
+        # Reset with number of weights and modes and perform the same check
```
antalszava, line 70: This is a stand-alone test case

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        for i in range(num_modes):
+            assert np.isclose(example.fidelity_vacuum([i]), 1)
+        assert example.active == [i for i in range(num_modes)]
+        assert example.smean().shape == (num_weights ** num_modes, 2 * num_modes)
+        assert example.scovmat().shape == (num_weights ** num_modes, 2 * num_modes, 2 * num_modes)
+
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("num_modes", NUM_MODES_VALS)
+    @pytest.mark.parametrize("num_weights_new", NUM_NEW_WEIGHTS_VALS)
+    def test_add_del_modes(self, num_weights, num_modes, num_weights_new):
```
antalszava, line 84: There are two separate methods tested in here: addition and deletion. Suggest splitting them up.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        r"""Checks that modes can be added and deleted."""
+        # Create state with number of weights and modes
```
antalszava, line 86:
```suggestion
        # Create a state with the pre-defined number of weights and modes
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        example = circuit.BosonicModes()
+        example.reset(num_modes, num_weights)
+        # Delete first mode and check active modes changed
+        example.del_mode(0)
+        assert example.get_modes() == [i for i in range(1, num_modes)]
```
antalszava, line 91:
```suggestion
        assert example.get_modes() == list(range(1, num_modes))
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+
+        # Add new mode with new number of weights
+        example.add_mode([num_weights_new])
+        tot_weights = (num_weights ** num_modes) * num_weights_new
+        # Check numbers of weights, means and covs
+        assert example.get_modes() == [i for i in range(1, num_modes + 1)]
```
antalszava, line 97:
```suggestion
        assert example.get_modes() == list(range(1, num_modes + 1))
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
antalszava: Although the name is used in the Gaussian case, maybe worth considering `update_means`? The current name suggests that the method can only be used for initialization, however, it can serve as an update any time.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        example.squeeze(r, phi, 1)
+        assert np.allclose(example.scovmat()[:, 0:2, 0:2], example.scovmat()[:, 2:4, 2:4])
+        assert np.allclose(example.smean()[:, 0:2], example.smean()[:, 2:4])
+        example.mbsqueeze(0, r, phi, 9, 1, False)
+        example.squeeze(r, phi, 1)
+        assert np.allclose(example.scovmat()[:, 0:2, 0:2], example.scovmat()[:, 2:4, 2:4])
+
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_complex_weight(self, r):
+        r"""Checks a cat state has correct parity and that mean from homodyne sampling is 0."""
```
antalszava, line 367:
```suggestion
        r"""Checks a cat state has correct parity and that the mean from homodyne sampling is 0."""
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+
+        self.weights = np.ones(w_shape(self.nlen, num_weights), dtype=complex)
+        self.weights = self.weights / (num_weights ** self.nlen)
+
+        self.means = np.zeros(m_shape(self.nlen, num_weights), dtype=complex)
+        id_covs = [np.identity(2 * self.nlen, dtype=complex) for i in range(len(self.weights))]
+        self.covs = np.array(id_covs)
+
+    def get_modes(self):
+        r"""Return the modes currently active."""
```
antalszava, line 231: What is an active mode? What is an inactive mode? Maybe worth describing when a mode can be `None` in `self.active`

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,406 @@
...
+        example.reset(1, num_weights)
+        example.displace(r, phi, 0)
+        # Perform heterodyne on mode 0
+        vals = example.heterodyne(0, shots=500)
+        mean = vals[:, 0].mean()
+        assert np.isclose(mean, example.hbar * r * np.cos(phi), rtol=0, atol=10 / np.sqrt(500))
+
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_post_select_homodyne(self, num_weights, r):
```
antalszava, line 307: Could it be useful to also check `post_select_generaldyne` for a non-homodyne/non-heterodyne measurement?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
antalszava: Might also be worth adding a specific test case for this

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+
+    Returns:
+        list: quadrature ordering for (x1,p1,...,xn,pn)
+    """
+    perm_inds_list = [(i, i + num_modes) for i in range(num_modes)]
+    perm_inds = [a for tup in perm_inds_list for a in tup]
+    return perm_inds
+
+
+def update_means(means, X, perm_out):
```
antalszava, line 93: Suggest adding a specific test case for these helper functions

antalszava (commented):
@elib20 :partying_face: This looks really great overall! In terms of code quality, I really liked that the code is structured and overall well documented.

Left some comments and suggestions. Things that I think would be important to consider:
* Using `reset` in the `BosonicModes.__init__` method.
* Splitting up test cases with many `assert` statements: good to keep tests short and concise. An idea is that one test case should test a single functionality. This also applies to some tests where I didn't not leave explicit comments.
* Adding unit tests for often used auxiliary functions. Although they will be covered, it is good to be on the safe side so that these functions indeed do not introduce any unexpected behaviour.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: The reset function is just to reset the modes to vacuum, but they're "nontraditional" vacuum states, in that they are linear combinations of Gaussians all centered at the origin of phase space with equal weight, such that when you add them up you get the vacuum. We do it this way because the other backends' reset functions revert the modes to vacuum, but we want the additional flexibility, through num_weights, to have arrays of means, weights and covs, that we can then manipulate. 

In general, we certainly want to initialize more complicated linear combinations of Gaussians, and we provide routines of interest in the next PR which includes state prep methods.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: A homodyne measurement is a projection onto an eigenstate of the quadrature, which can be approximated by taking a state that's super-squeezed in that quadrature, hence setting eps to be so small.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
nquesada: you mean checking that X,Y satisfy some mathematical constraints or some shape constraints?

thisac, line 800:
```suggestion
            Y (array): matrix for additive part of transformation
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        Us = symp.interferometer(U)
+        self.means = update_means(self.means, Us, self.from_xp)
+        self.covs = update_covs(self.covs, Us, self.from_xp)
+
+    def apply_channel(self, X, Y):
+        r"""Transforms the state according to a deterministic Gaussian CPTP map.
+
+        Args:
+            X (array): matrix for mutltiplicative part of transformation
```
thisac, line 799:
```suggestion
            X (array): matrix for multiplicative part of transformation
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
antalszava: > checking that X,Y satisfy some mathematical constraints

Meant this one. Making sure that they define a `deterministic Gaussian CPTP map`. (Though if this is checked later somewhere in the frontend, then that's also nice)

nquesada: I would probably not check that here.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        weighted_exp = (
+            np.array(self.weights)
+            * np.exp(-0.5 * exp_arg)
+            / np.sqrt(np.linalg.det(self.covs[:, mode_ind, :][:, :, mode_ind]))
+        )
+        parity = np.sum(weighted_exp) * (self.hbar / 2) ** len(modes)
+        return parity
+
+    def loss(self, T, k):
+        r"""Implements a loss channel in mode k.
```
elib20, line 510:
```suggestion
        r"""Implements a loss channel in mode k. T is the loss parameter that must be 
        between 0 and 1.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        if self.active[k] is None:
+            raise ValueError("Cannot apply loss channel, mode does not exist")
+
+        X = np.sqrt(T) * np.identity(2)
+        Y = self.hbar * (1 - T) * np.identity(2) / 2
+        X2, Y2 = self.expandXY([k], X, Y)
+        self.apply_channel(X2, Y2)
+
+    def thermal_loss(self, T, nbar, k):
+        r"""Implements the thermal loss channel in mode k.
```
elib20, line 526:
```suggestion
        r"""Implements the thermal loss channel in mode k. T is the loss parameter that must be 
        between 0 and 1.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: As you will see in a future PR, at the frontend level, I just have an sf.ops called MbSqueeze, and it has the same bool option for avg. If I break up the function here, should I also break the front end ops method?

I know it is bizarre to have the method return different types, but it's motivated by the physics in that this is a gate that requires an ancillary resource state, so you can either write down what the gate does on average (averaging over the measurement outcomes of the ancilla), or write down what the state does in a single-shot instance.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+    array of covariance matrices. The  quadrature ordering can be specified
+    by ``perm_out`` to match ordering of ``(X,Y)`` to the covariance matrices.
+
+    If Y is not specified, it defaults to 0.
+
+    Args:
+        covs (array): array of covariance matrices
+        X (array): matrix for mutltiplicative part of transformation
+        perm_out (array): indices for quadrature ordering
+        Y (array or 0): matrix for additive part of transformation.
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+            self.active[mode] = None
+
+    def reset(self, num_subsystems=None, num_weights=1):
+        """Reset the simulation state.
+
+        Args:
+            num_subsystems (int, optional): Sets the number of modes in the reset
+                circuit. ``None`` means unchanged.
+            num_weights (int): Sets the number of gaussians per mode in the
+                superposition.
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+
+        if k == l:
+            raise ValueError("Cannot use the same mode for beamsplitter inputs")
+
+        # Cross-check with Gaussian backend.
+        bs = symp.expand(symp.beam_splitter(theta, phi), [k, l], self.nlen)
+        self.means = update_means(self.means, bs, self.from_xp)
+        self.covs = update_covs(self.covs, bs, self.from_xp)
+
+    def scovmatxp(self):
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        self.covs[:, :, mode_ind] = 0
+        # Set new covariance elements
+        self.covs[np.ix_(np.arange(self.covs.shape[0], dtype=int), mode_ind, mode_ind)] = V
+
+    def fidelity_coherent(self, alpha, modes=None):
+        r"""Returns the fidelity to a coherent state.
+
+        Args:
+            alpha (array): amplitudes for coherent states
+            modes (list or None): modes to use for fidelity calculation
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,833 @@
...
+        """
+        if modes is None:
+            modes = self.get_modes()
+        # Sort by (q1,p1,q2,p2,...)
+        mode_ind = np.sort(np.append(2 * np.array(modes), 2 * np.array(modes) + 1))
+        # Construct mean vector for coherent state
+        alpha_mean = np.array([])
+        for i in range(len(modes)):
+            alpha_mean = np.append(alpha_mean, alpha.real[i] * np.sqrt(2 * self.hbar))
+            alpha_mean = np.append(alpha_mean, alpha.imag[i] * np.sqrt(2 * self.hbar))
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I have added extensive commenting and clearer variable names now

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: It seems safer to me to explicitly leave hbar everywhere in case someone wants to use the backend functions directly.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: The apply_channel method directly applies these functions and we test that.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
antalszava: Hmm, this might be more connected to what the preferred UI could be for users. If it's more likely, that people would like to perform measurement-based squeezing and this is a small detail only regarding the operation, then maybe it's good to be kept as is in the frontend. Otherwise, it could also be made into two separate operations.

Under the hood, separation would just help with future maintenance of the code.

Usually from a code quality point of view, when there's a boolean argument which greatly influences the logic of the function, it tells that the function is doing at least two or more things (which sort of goes against the [Single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) extended to functions/methods).

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: In the test for loss, we test creating a thermal state via the thermal loss method which applied this method directly.

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: I purposefully tested them together because they both change the list of active modes. I wanted to ensure that no conflict was created between them.

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: I think that would just be another covariance matrix. With homodyne and heterodyne we're already checking two very different cases of covariance matrices so I don't see what else that would add.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I've clarified now.

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Thanks for the info! I will separate it in the backend then, but leave it as one call in the frontend.

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: I have split it now.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: I have fixed this now.

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: Fixed.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+            Y (array): matrix for additive part of transformation.
+        """
+        X2 = symp.expand(X, modes, self.nlen)
+        Y2 = symp.expand(Y, modes, self.nlen)
+        for i in range(self.nlen):
+            if i not in modes:
+                Y2[i, i] = 0
+                Y2[i + self.nlen, i + self.nlen] = 0
+
+        return X2, Y2
```
nquesada, line 957: This is great!

nquesada (approved):

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+            S (array): symplectic matrix
+        """
+        return symp.expand(S, modes, self.nlen)
+
+    def expandXY(self, modes, X, Y):
+        """Expands deterministic Gaussian CPTP matrices ``(X,Y)`` on subset of modes to
+        transformations for the whole system.
+
+        Args:
+            modes (list): list of modes on which ``(X,Y)``
```
elib20, line 946:
```suggestion
            modes (list): list of modes on which ``(X,Y)`` act
```
```suggestion
            modes (list): list of modes on which ``(X,Y)``
```

elib20, line 946:
```suggestion
            modes (list): list of modes on which ``(X,Y)`` act
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+            / np.sqrt(np.linalg.det(cov_sum))
+        )
+        fidelity = np.sum(weighted_exp)
+        return fidelity
+
+    def fidelity_vacuum(self, modes=None):
+        r"""Returns the fidelity to the vacuum.
+
+        Args:
+            modes (list or None): modes to use for fidelity calculation
```
antalszava, line 540:
```suggestion
            modes (list): modes to use for fidelity calculation
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        alpha = np.zeros(len(modes))
+        fidelity = self.fidelity_coherent(alpha, modes=modes)
+        return fidelity
+
+    def parity_val(self, modes=None):
+        r"""Returns the expectation value of the parity operator, which is the
+        value of the Wigner function at the origin.
+
+        Args:
+            modes (list or None): modes to use for parity calculation
```
antalszava, line 556:
```suggestion
            modes (list): modes to use for parity calculation
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+
+        Returns:
+            float: parity of the state in modes
```
antalszava, line 559:
```suggestion
            float: parity of the state for the subsystem defined by modes
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+            raise ValueError("Cannot apply loss channel, mode does not exist")
+
+        X = np.sqrt(T) * np.identity(2)
+        Y = self.hbar * (1 - T) * np.identity(2) / 2
+        X2, Y2 = self.expandXY([k], X, Y)
+        self.apply_channel(X2, Y2)
+
+    def thermal_loss(self, T, nbar, k):
+        r"""Implements the thermal loss channel in mode k. T is the loss parameter that must
+        be between 0 and 1.
```
antalszava, line 601: Should it be checked that `T` is within the correct range?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        Y = self.hbar * (1 - T) * (2 * nbar + 1) * np.identity(2) / 2
+        X2, Y2 = self.expandXY([k], X, Y)
+        self.apply_channel(X2, Y2)
+
+    def init_thermal(self, nbar, mode):
+        r"""Initializes a state of mode in a thermal state with the given population.
+
+        Args:
+            nbar (float): mean photon number of the thermal state
+            mode (int): mode that get initialized
```
antalszava, line 623:
```suggestion
            mode (int): mode that gets initialized
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        """
+        if self.active[k] is None:
+            raise ValueError("Cannot apply loss channel, mode does not exist")
+        X = np.sqrt(T) * np.identity(2)
+        Y = self.hbar * (1 - T) * (2 * nbar + 1) * np.identity(2) / 2
+        X2, Y2 = self.expandXY([k], X, Y)
+        self.apply_channel(X2, Y2)
+
+    def init_thermal(self, nbar, mode):
+        r"""Initializes a state of mode in a thermal state with the given population.
```
antalszava, line 619:
```suggestion
        r"""Initializes a state in a thermal state with the given population.
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        """
+        # pylint: disable=too-many-branches
+        if covmat.shape != (2 * len(modes), 2 * len(modes)):
+            raise ValueError("Covariance matrix size does not match indices provided")
+
+        if np.linalg.det(covmat) < (self.hbar / 2) ** (2 * len(modes)):
+            raise ValueError("Measurement covariance matrix is unphysical.")
+
+        if self.covs.imag.any():
+            raise NotImplementedError("Covariance matrices must be real")
```
antalszava, line 671:
```suggestion
            raise NotImplementedError("Covariance matrices must be real.")
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        Raises:
+            ValueError: if the dimension of covmat does not match number of quadratures
+                        associated with modes
+            ValueError: if covmat does not respect the uncertainty relation
+            NotImplementedError: if any of the covariances of the state are imaginary
+            ValueError: if any of the modes are not in the list of active modes
+        """
+        # pylint: disable=too-many-branches
+        if covmat.shape != (2 * len(modes), 2 * len(modes)):
+            raise ValueError("Covariance matrix size does not match indices provided")
```
antalszava, line 665:
```suggestion
            raise ValueError("Covariance matrix size does not match indices provided.")
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+            shots (int): how many measurements are performed
+
+        Returns:
+            array: heterodyne outcome
+        """
+        covmat = self.hbar * np.eye(2) / 2
+        return self.measure_dyne(covmat, [mode], shots=shots)
+
+    def post_select_generaldyne(self, covmat, modes, vals):
+        r"""Simulates general-dyne measurement on a set of modes with a specified measurement
```
antalszava, line 828:
```suggestion
        r"""Simulates a general-dyne measurement on a set of modes with a specified measurement
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        outcome.
+
+        Args:
+            covmat (array): covariance matrix of the generaldyne measurement
+            modes (list): modes to be measured
+            vals (array): measurement outcome to postselect
+
+        Raises:
+            ValueError: if the dimension of covmat does not match number of quadratures
```
antalszava, line 837:
```suggestion
            ValueError: if the dimension of covmat does not match the number of quadratures
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+                        associated with modes
+            ValueError: if any of the modes are not in the list of active modes
+        """
+        if covmat.shape != (2 * len(modes), 2 * len(modes)):
+            raise ValueError("Covariance matrix size does not match indices provided")
```
antalszava, line 842:
```suggestion
            raise ValueError("The size of the covariance matrix does not match the indices provided.")
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        self.means = self.means[abs(self.weights) > 0]
+        self.covs = self.covs[abs(self.weights) > 0]
+        self.weights = self.weights[abs(self.weights) > 0]
+
+    def post_select_homodyne(self, mode, val, eps=0.0002, phi=0):
+        r"""Simulates a homodyne measurement on a mode, postselecting on an outcome.
+
+        Args:
+            mode (int): mode to be measured
+            val (array): measurement value to post-select
```
antalszava, line 879: Seems to have both ` post-select` and ` postselect` within the file, would suggest going with only one

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+            eps (int): squeezing of the measurement state
+            phi (float): homodyne angle
+
+        Raises:
+            ValueError: if the mode are not in the list of active modes
+        """
+        if self.active[mode] is None:
+            raise ValueError("Cannot apply homodyne measurement, mode does not exist")
```
antalszava, line 887:
```suggestion
            raise ValueError("Cannot apply homodyne measurement, mode does not exist.")
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+
+        Args:
+            mode (int): mode to be measured
+            alpha_val (array): measurement value to post-select
+
+        Raises:
+            ValueError: if the mode are not in the list of active modes
+        """
+        if self.active[mode] is None:
+            raise ValueError("Cannot apply heterodyne measurement, mode does not exist")
```
antalszava, line 905:
```suggestion
            raise ValueError("Cannot apply heterodyne measurement, mode does not exist.")
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
antalszava: The `thermal_loss` method does some other things too (initiates a particular `X` and `Y`, calls this method and calls `apply_chanel`). Writing a unit test for that can have a different effect to writing a unit test just for `expandXY`.

tests/bosonic_files/test_backend_bosoniccircuit.py
antalszava: This way, it is not ensured that deletion and addition work separately with the test. It could arise that some effects that are caused by deletion are pruned by the followed addition, which could cause the test to pass falsely.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        # Check numbers of weights, means and covs
+        assert example.get_modes() == list(range(1, num_modes + 1))
+        assert np.isclose(sum(example.get_weights()), 1)
+        assert example.get_weights().shape == (tot_weights,)
+        assert example.get_mean().shape == (tot_weights, 2 * (num_modes + 1))
+        assert example.get_covmat().shape == (tot_weights, 2 * (num_modes + 1), 2 * (num_modes + 1))
+
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_displace_squeeze_rotate(self, r, phi):
```
antalszava, line 126: This case is testing too much, e.g., squeezing and applying a phase shift could be tested separately.

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        )
+
+        # Initialize third mode as thermal and check it's the same as second mode
+        example.init_thermal(nbar, 2)
+        assert np.allclose(example.get_covmat()[:, 4:6, 4:6], example.get_covmat()[:, 2:4, 2:4])
+
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_gaussian_symplectic(self, r, phi):
+        r"""Checks symplectic transformations, and get_covtmat_xp, get_mean_xp."""
```
antalszava, line 232:
```suggestion
        r"""Checks symplectic transformations, and get_covmat_xp, get_mean_xp."""
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        assert np.allclose(example.get_mean()[:, 0:2], example.get_mean()[:, 2:4])
+        assert np.allclose(
+            example.get_covmat_xp()[:, 0::2, 0::2], example.get_covmat_xp()[:, 1::2, 1::2]
+        )
+        assert np.allclose(example.get_mean_xp()[:, 0::2], example.get_mean_xp()[:, 1::2])
+
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_beamsplitter(self, r, phi):
+        r"""Checks beamsplitter."""
```
antalszava, line 258:
```suggestion
        r"""Checks applying a beamsplitter."""
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        example.beamsplitter(phi, phi, 0, 1)
+        alpha = r * (np.cos(phi) - np.exp(-1j * phi) * np.sin(phi))
+        beta = r * (np.cos(phi) + np.exp(1j * phi) * np.sin(phi))
+        assert np.isclose(example.fidelity_coherent(np.array([alpha, beta])), 1)
+
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    @pytest.mark.parametrize("num_shots", NUM_SHOTS_VALS)
+    def test_generaldyne(self, num_weights, r, num_shots):
+        r"""Checks generaldyne."""
```
antalszava, line 272:
```suggestion
        r"""Checks a generaldyne measurement."""
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        # Check output has right shape
+        assert vals.shape == (num_shots, 2)
+        # Check measured mode was set to vacuum
+        assert np.isclose(example.fidelity_vacuum(modes=[0]), 1)
+
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    def test_homodyne(self, num_weights, r, phi):
+        r"""Checks homodyne."""
```
antalszava, line 293:
```suggestion
        r"""Checks a homodyne measurement."""
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        mean = vals[:, 0].mean()
+        std = vals[:, 0].std()
+        assert np.isclose(mean, example.hbar * r * np.cos(phi), rtol=0, atol=10 / np.sqrt(500))
+        assert np.isclose(std, 1, rtol=0, atol=10 / np.sqrt(500))
+
+    @pytest.mark.parametrize("num_weights", NUM_WEIGHTS_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    def test_heterodyne(self, num_weights, r, phi):
+        r"""Checks heterodyne."""
```
antalszava, line 310:
```suggestion
        r"""Checks a heterodyne measurement."""
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+        assert example.is_vacuum(tol=1e-10)
+
+        # Test high-quality mbsqueezing
+        example.mb_squeeze_single_shot(0, r, phi, 9, 1)
+        example.squeeze(r, phi, 1)
+        assert np.allclose(example.get_covmat()[:, 0:2, 0:2], example.get_covmat()[:, 2:4, 2:4])
+
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_complex_weight(self, r):
+        r"""Checks a cat state has correct parity and that the mean from homodyne sampling is 0."""
```
antalszava, line 390:
```suggestion
        r"""Checks that a cat state has correct parity and that the mean from homodyne sampling is 0."""
```

tests/bosonic_files/test_backend_bosoniccircuit.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # Perform homodyne
+        vals = example.homodyne(0, shots=500)
+        mean = vals[:, 0].mean()
+        assert np.isclose(mean, 0, rtol=0, atol=10 * np.sqrt(abs(r)) / np.sqrt(500))
+
+    @pytest.mark.parametrize("phi", PHI_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_linear_optical_unitary(self, r, phi):
+        r"""Checks linear optical unitary implementation of a phase shift."""
```
antalszava, line 418:
```suggestion
        r"""Checks the linear optical unitary implementation of a phase shift."""
```

antalszava (approved):
Looks good to me! :100: Exciting changes :rocket: Left some comments, some tests (where explicitly commented) could be further split up. Once the suggestions are addressed, this is merge-ready from my side.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -0,0 +1,957 @@
...
+        self.weights = self.weights / (num_weights ** self.nlen)
+
+        self.means = np.zeros(m_shape(self.nlen, num_weights), dtype=complex)
+        id_covs = [np.identity(2 * self.nlen, dtype=complex) for i in range(len(self.weights))]
+        self.covs = np.array(id_covs)
+
+    def get_modes(self):
+        r"""Return the modes that are currently active. Active modes
+        are those created by the user which have not been deleted.
+        If a mode is deleted, its entry in the list is ``None``."""
```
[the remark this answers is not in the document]
elib20: This is copying how the Gaussian backend deletes modes. Not sure about Fock or TF backends.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Will do.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Will do

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: Will split.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
elib20: Done.

tests/bosonic_files/test_backend_bosoniccircuit.py
elib20: Done.

.github/workflows/tests.yaml
```diff
@@ -35,6 +35,8 @@ jobs:
           access_token: ${{ github.token }}
 
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
thisac, line 39: :crossed_fingers: :100:

PR #540, opened 2021-02-10
Rewrite XQC API version to 0.2.0 and add User-Agent

jswinarton:
**Context:**
The current version of StrawberryFields sends an incorrect version header to the Xanadu Quantum Cloud API.

- Update the `Accept-Version` header of requests sent from the `strawberryfields.api.connection` module to `0.2.0` instead of `1.0.0`
- Add a custom User-Agent string which sends a header in the standard format `StrawberryFields/0.17.0` where `0.17.0` is the current version of StrawberryFields

**Benefits:**
- Allows compability with the XQC platform in future iterations where API versioning will be more strict
- The `User-Agent` string gives important contextual information to the XQC API which can be used to provide custom behaviour in the future (for example, rewriting behaviour in the case of problematic requests, or providing deprecation notices in the event of lifecycle changes in the API)

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 <h3>Bug fixes</h3>
 
+* `Connection` objects now send requests to the platform API at version `0.2.0`
+  instead of the incorrect version number `1.0.0`.
+  [(#540)](https://github.com/XanaduAI/strawberryfields/pull/540)
+
 <h3>Documentation</h3>
 
 <h3>Contributors</h3>
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 This release contains contributions from (in alphabetical order):
 
+Jeremy Swinarton.
+
 # Release 0.17.0 (current release)
 
 <h3>New features since last release</h3>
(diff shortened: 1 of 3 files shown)


corvust (approved):

antalszava (approved):
Looks good to me, thank you! :100:

In the future, we'd like to change the API version sent by Strawberry Fields when there's a new API version of XQC that we would like to start using, correct? Would that potentially influence release cycles?

thisac (approved):
Looks good as far as I can tell. :+1: As @antalszava mentioned, I assume this is something that we'll need to update each time we want to start using a new version of the api.

jswinarton:
@antalszava @thisac Yeah, that's the idea. Depending on the speed of work on our end they may be frequent, but how you decide to release these on your end is up to you. For example we'll have version 0.3.0 coming out soon with our new device API, which will give you access to dynamic device information, but maybe you don't have time to integrate that on your side for a while. That's fine because we'll continue to support old versions for a period.

josh146 (approved):

josh146:
@jswinarton before you merge this in, can you update the changelog? Thanks!

PR #541, opened 2021-02-17
Bosonic backend nongaussian

elib20:
**Context:** The bosonic backend simulates states as linear combinations of Gaussians in phase space

- Adds preparation backend and frontend methods for GKP, cat and Fock states to the bosonic backend
- Adds the run_prog and init_circuit methods to bosonic/backend.py which are the engines for running sf Programs in the bosonic backend
- Adds the bosonic compiler
- Adds tests for these new backend methods

**Benefits:**
- Simulation of bosonic qubits

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # pylint: disable=too-many-public-methods
+# pylint: disable=import-outside-toplevel
 
 """Bosonic backend"""
 import itertools as it
@@ -28,8 +29,6 @@ from strawberryfields.backends.states import BaseBosonicState
 
 from strawberryfields.backends.bosonicbackend.bosoniccircuit import BosonicModes
 from strawberryfields.backends.base import NotApplicableError
-
-import strawberryfields.ops as ops
 
 
 def kron_list(l):
@@ -85,6 +84,16 @@ class BosonicBackend(BaseBosonic):
             NotApplicableError: if an op in the program does not apply to the bosonic backend
             NotImplementedError: if an op in the program is not implemented in the bosonic backend
         """
+        from strawberryfields.ops import (
+            Bosonic,
+            Catstate,
+            DensityMatrix,
+            Fock,
+            GKP,
+            Ket,
+            MbSgate,
+            _New_modes,
+        )
 
         # Initialize the circuit. This applies all non-Gaussian state-prep
         self.init_circuit(prog)
@@ -97,15 +106,15 @@ class BosonicBackend(BaseBosonic):
         all_samples = {}
 
         non_gauss_preps = (
-            ops.Bosonic,
-            ops.Catstate,
-            ops.DensityMatrix,
-            ops.Fock,
-            ops.GKP,
-            ops.Ket,
-            ops._New_modes,
+            Bosonic,
+            Catstate,
+            DensityMatrix,
+            Fock,
+            GKP,
+            Ket,
+            _New_modes,
         )
-        ancilla_gates = (ops.MbSgate,)
+        ancilla_gates = (MbSgate,)
         for cmd in prog.circuit:
             # For ancilla-assisted gates, if they return measurement values, store
             # them in ancillae_samples_dict
@@ -164,8 +173,17 @@ class BosonicBackend(BaseBosonic):
         Raises:
             NotImplementedError: if ``Ket`` or ``DensityMatrix`` preparation used
         """
+        from strawberryfields.ops import (
+            Bosonic,
+            Catstate,
+            DensityMatrix,
+            Fock,
+            GKP,
+            Ket,
+            _New_modes,
+        )
 
-        non_gauss_preps = (ops.Bosonic, ops.Catstate, ops.DensityMatrix, ops.Fock, ops.GKP, ops.Ket)
+        non_gauss_preps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket)
         nmodes = prog.init_num_subsystems
         self.begin_circuit(nmodes)
         # Dummy initial weights, means and covs
@@ -187,25 +205,25 @@ class BosonicBackend(BaseBosonic):
                 pars = cmd.op.p
                 for reg in labels:
                     # All the possible preparations should go in this loop
-                    if isinstance(cmd.op, ops.Bosonic):
+                    if isinstance(cmd.op, Bosonic):
                         weights, means, covs = [pars[i] for i in range(3)]
 
-                    elif isinstance(cmd.op, ops.Catstate):
+                    elif isinstance(cmd.op, Catstate):
                         weights, means, covs = self.prepare_cat(*pars)
 
-                    elif isinstance(cmd.op, ops.GKP):
+                    elif isinstance(cmd.op, GKP):
                         weights, means, covs = self.prepare_gkp(*pars)
 
-                    elif isinstance(cmd.op, ops.Fock):
+                    elif isinstance(cmd.op, Fock):
                         weights, means, covs = self.prepare_fock(*pars)
 
-                    elif isinstance(cmd.op, (ops.Ket, ops.DensityMatrix)):
+                    elif isinstance(cmd.op, (Ket, DensityMatrix)):
                         raise NotImplementedError(
                             "Ket and DensityMatrix preparation not implemented in the bosonic backend."
                         )
 
                     # If a new mode is added in the program context, then add it here
-                    elif isinstance(cmd.op, ops._New_modes):
+                    elif isinstance(cmd.op, _New_modes):
                         cmd.op.apply(cmd.reg, self)
                         init_weights.append([0])
                         init_means.append([0])
@@ -225,7 +243,7 @@ class BosonicBackend(BaseBosonic):
                 # Add the mode to the list of already prepared modes, unless the command was
                 # just to create the new mode, in which case it checks again to see if there is
                 # a subsequent non-Gaussian state creation
-                if not isinstance(cmd.op, ops._New_modes):
+                if not isinstance(cmd.op, _New_modes):
                     reg_list += labels
 
             else:
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -27,6 +27,16 @@ from strawberryfields.backends.states import BaseBosonicState
 from strawberryfields.backends.bosonicbackend.bosoniccircuit import BosonicModes
 from strawberryfields.backends.base import NotApplicableError
 
+from strawberryfields.ops import (
+    Bosonic,
+    Catstate,
+    DensityMatrix,
+    Fock,
+    GKP,
+    Ket,
+    MbSgate,
+)
+
 
 def kron_list(l):
     """Take Kronecker products of a list of lists."""
@@ -86,16 +96,6 @@ class BosonicBackend(BaseBosonic):
             NotImplementedError: if an op in the program is not implemented
                                  in the bosonic backend
         """
-
-        from strawberryfields.ops import (
-            Bosonic,
-            Catstate,
-            DensityMatrix,
-            Fock,
-            GKP,
-            Ket,
-            MbSgate,
-        )
 
         # Initialize the circuit. This applies all non-Gaussian state-prep
         self.init_circuit(prog)
@@ -165,15 +165,6 @@ class BosonicBackend(BaseBosonic):
         Raises:
             NotImplementedError: if Ket or DensityMatrix preparation used
         """
-
-        from strawberryfields.ops import (
-            Bosonic,
-            Catstate,
-            DensityMatrix,
-            Fock,
-            GKP,
-            Ket,
-        )
 
         nmodes = prog.num_subsystems
         self.begin_circuit(nmodes)
@@ -474,7 +465,7 @@ class BosonicBackend(BaseBosonic):
         return weights, means, cov
 
     def prepare_gkp(self, state, epsilon, cutoff, desc="real", shape="square"):
-        """Prepares the arrays of weights, means and covs for a finite energy GKP state.
+        r"""Prepares the arrays of weights, means and covs for a finite energy GKP state.
 
         GKP states are qubits, with the qubit state defined by:
 
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -267,9 +267,7 @@ class BosonicBackend(BaseBosonic):
         # Tensor product of the weights.
         tensored_weights = kron_list(init_weights)
         # De-nest the means iterator.
-        tensored_means = np.array(
-            [[a for b in tup for a in b] for tup in mean_combos], dtype=complex
-        )
+        tensored_means = np.array([np.concatenate(tup) for tup in mean_combos], dtype=complex)
         # Stack covs appropriately.
         tensored_covs = np.array([block_diag(*tup) for tup in cov_combos])
 
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -131,7 +131,7 @@ class BosonicBackend(BaseBosonic):
                 applied.append(cmd)
 
             # Rest of operations applied as normal
-            if not isinstance(cmd.op, non_gauss_preps + ancilla_gates):
+            elif not isinstance(cmd.op, non_gauss_preps):
                 try:
                     # try to apply it to the backend and if op is a measurement, store outcome in values
                     val = cmd.op.apply(cmd.reg, self, **kwargs)
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -458,12 +458,11 @@ class BosonicBackend(BaseBosonic):
         p_means = 0.5 * np.array(range(-2 * z_max, 2 * z_max + 1), dtype=float)
 
         # Creating and calculating the weights array for oscillating terms
-        odd_terms = np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int) % 2
+        term_inds = np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int)
+        odd_terms = term_inds % 2
         even_terms = (odd_terms + 1) % 2
-        even_phases = (-1) ** ((np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int) % 4) // 2)
-        odd_phases = (-1) ** (
-            1 + ((np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int) + 2) % 4) // 2
-        )
+        even_phases = (-1) ** ((term_inds % 4) // 2)
+        odd_phases = (-1) ** (1 + ((term_inds + 2) % 4) // 2)
         weights = np.cos(phi) * even_terms * even_phases * np.exp(
             -0.5 * coef_sigma * p_means ** 2
         ) - np.sin(phi) * odd_terms * odd_phases * np.exp(-0.5 * coef_sigma * p_means ** 2)
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -95,6 +95,7 @@ class BosonicBackend(BaseBosonic):
             GKP,
             Ket,
             MbSgate,
+            _New_modes,
         )
 
         # Initialize the circuit. This applies all non-Gaussian state-prep
@@ -107,7 +108,7 @@ class BosonicBackend(BaseBosonic):
         samples_dict = {}
         all_samples = {}
         for cmd in prog.circuit:
-            nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket)
+            nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket, _New_modes)
             ancilla_gates = (MbSgate,)
             # For ancilla-assisted gates, if they return measurement values, store
             # them in ancillae_samples_dict
@@ -172,9 +173,11 @@ class BosonicBackend(BaseBosonic):
             Fock,
             GKP,
             Ket,
+            _New_modes,
         )
 
-        nmodes = prog.num_subsystems
+        nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket)
+        nmodes = prog.init_num_subsystems
         self.begin_circuit(nmodes)
         # Dummy initial weights, means and covs
         init_weights, init_means, init_covs = [[0] * nmodes for i in range(3)]
@@ -187,7 +190,7 @@ class BosonicBackend(BaseBosonic):
 
         # Go through the operations in the circuit
         for cmd in prog.circuit:
-            # Check if an operation has already acted on these modes.
+            # Check if an operation other than New() has already acted on these modes.
             labels = [label.ind for label in cmd.reg]
             isitnew = 1 - np.isin(labels, reg_list)
             if np.any(isitnew):
@@ -212,6 +215,15 @@ class BosonicBackend(BaseBosonic):
                             "Ket and DensityMatrix preparation not implemented in bosonic backend."
                         )
 
+                    # If a new mode is added in the program context, then add it
+                    # but don't add it to the reg_list in case there is a subsequent
+                    # non-Gaussian state prep added
+                    elif isinstance(cmd.op, _New_modes):
+                        cmd.op.apply(cmd.reg, self)
+                        init_weights.append([0])
+                        init_means.append([0])
+                        init_covs.append([0])
+
                     # The rest of the preparations are gaussian.
                     # TODO: initialize with Gaussian |vacuum> state
                     # directly by asking preparation methods below for
@@ -223,7 +235,17 @@ class BosonicBackend(BaseBosonic):
                     init_means[reg] = means
                     init_covs[reg] = covs
 
-                reg_list += labels
+                # Add the mode to the list of already prepared modes, unless the command was
+                # just to create the new mode, in which case it checks again to see if there is
+                # a subsequent non-Gaussian state creation
+                if not isinstance(cmd.op, _New_modes):
+                    reg_list += labels
+
+            else:
+                if type(cmd.op) in nongausspreps:
+                    raise NotImplementedError(
+                        "Non-gaussian state preparations must be the first operation for each register."
+                    )
 
         # Assume unused modes in the circuit are vacua.
         # If there are any Gaussian state preparations, these will be handled
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -337,7 +337,7 @@ class BosonicBackend(BaseBosonic):
              tuple: arrays of the weights, means and covariances for the state
         """
 
-        if desc != "complex" or desc != "real":
+        if desc != "complex" and desc != "real":
             raise ValueError(r'``desc`` accepts only "real" or "complex" as arguments.')
 
         # Case alpha = 0, prepare vacuum
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -330,22 +330,22 @@ class BosonicBackend(BaseBosonic):
 
     def prepare_cat(self, alpha, phi, cutoff, desc, D):
         r"""Prepares the arrays of weights, means and covs for a cat state:
-        
-        .. math::
-       \ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha}).
 
-        Args:
-            alpha (float): alpha value of cat state
-            phi (float): phi value of cat state
-            cutoff (float): if using the 'real' representation, this determines
-                how many terms to keep
-            desc (string): whether to use the 'real' or 'complex' representation
-            D (float): for 'real rep., quality parameter of approximation
+         .. math::
+        \ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha}).
 
-        Returns:
-            tuple: arrays of the weights, means and covariances for the state
+         Args:
+             alpha (float): alpha value of cat state
+             phi (float): phi value of cat state
+             cutoff (float): if using the 'real' representation, this determines
+                 how many terms to keep
+             desc (string): whether to use the 'real' or 'complex' representation
+             D (float): for 'real rep., quality parameter of approximation
+
+         Returns:
+             tuple: arrays of the weights, means and covariances for the state
         """
-        
+
         if desc != "complex" or desc != "real":
             raise ValueError(r'``desc`` accepts only "real" or "complex" as arguments.')
 
@@ -375,33 +375,33 @@ class BosonicBackend(BaseBosonic):
             covs = 0.5 * hbar * np.identity(2, dtype=float)
             covs = np.repeat(covs[None, :], weights.size, axis=0)
             return weights, means, covs
-        
-        #The only remaining option is a real-valued cat state
+
+        # The only remaining option is a real-valued cat state
         return self.prepare_cat_real_rep(alpha, phi, cutoff, D)
-        
+
     def prepare_cat_real_rep(self, alpha, phi, cutoff, D):
         r"""Prepares the arrays of weights, means and covs for a cat state:
-        
-        .. math::
-       \ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha}).
-            
-        For this representation, weights, means and covariances are real-valued.
 
-        Args:
-            alpha (float): alpha value of cat state
-            phi (float): phi value of cat state
-            cutoff (float): if using the 'real' representation, this determines
-                how many terms to keep
-            D (float): for 'real rep., quality parameter of approximation
+         .. math::
+        \ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha}).
 
-        Returns:
-            tuple: arrays of the weights, means and covariances for the state
+         For this representation, weights, means and covariances are real-valued.
+
+         Args:
+             alpha (float): alpha value of cat state
+             phi (float): phi value of cat state
+             cutoff (float): if using the 'real' representation, this determines
+                 how many terms to keep
+             D (float): for 'real rep., quality parameter of approximation
+
+         Returns:
+             tuple: arrays of the weights, means and covariances for the state
         """
         # Normalization factor
         norm = 1 / (2 * (1 + np.exp(-2 * np.absolute(alpha) ** 2) * np.cos(phi)))
         phi = np.pi * phi
         hbar = self.circuit.hbar
-        
+
         # Defining useful constants
         a = np.absolute(alpha)
         phase = np.angle(alpha)
@@ -410,9 +410,7 @@ class BosonicBackend(BaseBosonic):
         num_mean = 8 * a * np.sqrt(hbar) / (np.pi * D * np.sqrt(2))
         denom_mean = 16 * a ** 2 / (np.pi ** 2 * D) + 2
         coef_sigma = np.pi ** 2 * hbar / (8 * a ** 2 * (E + v))
-        prefac = (
-            np.sqrt(np.pi * hbar) * np.exp(0.25 * np.pi ** 2 * D) / (4 * a) / (np.sqrt(E + v))
-        )
+        prefac = np.sqrt(np.pi * hbar) * np.exp(0.25 * np.pi ** 2 * D) / (4 * a) / (np.sqrt(E + v))
         z_max = int(
             np.ceil(
                 2
@@ -474,16 +472,15 @@ class BosonicBackend(BaseBosonic):
             cov = S @ cov @ S.T
 
         return weights, means, cov
-            
 
     def prepare_gkp(self, state, epsilon, cutoff, desc="real", shape="square"):
         """Prepares the arrays of weights, means and covs for a finite energy GKP state.
-        
+
         GKP states are qubits, with the qubit state defined by:
-        
+
         .. math::
         \ket{\psi}_{gkp} = \cos\frac{\theta}{2}\ket{0}_{gkp} + e^(-i\phi)\sin\frac{\theta}{2}\ket{1}_{gkp}
-        
+
         where the computational basis states are :math:`\ket{\mu}_{gkp} = \sum_{n} \ket{(2n+\mu)\sqrt{\pi\hbar}}_{q}`.
 
         Args:
@@ -501,10 +498,10 @@ class BosonicBackend(BaseBosonic):
             NotImplementedError: if the complex representation or a non-square lattice
                                 is attempted
         """
-        
+
         if desc == "complex":
             raise NotImplementedError("The complex description of GKP is not implemented")
-            
+
         if shape != "square":
             raise NotImplementedError("Only square GKP are implemented for now")
 
@@ -572,9 +569,7 @@ class BosonicBackend(BaseBosonic):
 
         # Create set of means before finite energy effects
         means_gen = it.tee(
-            it.starmap(
-                lambda l, m: l + 1j * m, it.product(range(-z_max, z_max + 1), repeat=2)
-            ),
+            it.starmap(lambda l, m: l + 1j * m, it.product(range(-z_max, z_max + 1), repeat=2)),
             2,
         )
         means = np.concatenate(
@@ -607,7 +602,7 @@ class BosonicBackend(BaseBosonic):
         )
         covs = np.repeat(covs[None, :], weights.size, axis=0)
 
-        return weights, means, covs          
+        return weights, means, covs
 
     def prepare_fock(self, n, r=0.05):
         """Prepares the arrays of weights, means and covs of a Fock state.
--- a/strawberryfields/backends/bosonicbackend/backend.py
+++ b/strawberryfields/backends/bosonicbackend/backend.py
@@ -78,9 +78,8 @@ class BosonicBackend(BaseBosonic):
             prog (object): sf.Program instance
 
         Returns:
-            tuple: list of applied commands,
-                    dictionary of measurement samples,
-                    dictionary of ancilla measurement samples
+            tuple: a tuple of the list of applied commands, the dictionary of measurement samples,
+            and the dictionary of ancilla measurement samples
 
         Raises:
             NotApplicableError: if an op in the program does not apply to the bosonic backend
@@ -106,12 +105,13 @@ class BosonicBackend(BaseBosonic):
         applied = []
         samples_dict = {}
         all_samples = {}
+
+        nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket, _New_modes)
+        ancilla_gates = (MbSgate,)
         for cmd in prog.circuit:
-            nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket, _New_modes)
-            ancilla_gates = (MbSgate,)
             # For ancilla-assisted gates, if they return measurement values, store
             # them in ancillae_samples_dict
-            if type(cmd.op) in ancilla_gates:
+            if isinstance(cmd.op, ancilla_gates):
                 # if the op returns a measurement outcome store it in a dictionary
                 val = cmd.op.apply(cmd.reg, self, **kwargs)
                 if val is not None:
@@ -124,9 +124,9 @@ class BosonicBackend(BaseBosonic):
                 applied.append(cmd)
 
             # Rest of operations applied as normal
-            if type(cmd.op) not in nongausspreps + ancilla_gates:
+            if not isinstance(cmd.op, nongausspreps + ancilla_gates):
                 try:
-                    # try to apply it to the backend and, if op is a measurement, store outcome in values
+                    # try to apply it to the backend and if op is a measurement, store outcome in values
                     val = cmd.op.apply(cmd.reg, self, **kwargs)
                     if val is not None:
                         for i, r in enumerate(cmd.reg):
@@ -157,10 +157,10 @@ class BosonicBackend(BaseBosonic):
 
     def init_circuit(self, prog):
         """Instantiate the circuit and initialize weights, means, and covs
-        depending on the Preparation classes.
+        depending on the ``Preparation`` classes.
 
         Args:
-            prog (object): sf.Program instance
+            prog (object): :class:`~.Program` instance
 
         Raises:
             NotImplementedError: if ``Ket`` or ``DensityMatrix`` preparation used
@@ -179,10 +179,10 @@ class BosonicBackend(BaseBosonic):
         nmodes = prog.init_num_subsystems
         self.begin_circuit(nmodes)
         # Dummy initial weights, means and covs
-        init_weights, init_means, init_covs = [[0] * nmodes for i in range(3)]
+        init_weights, init_means, init_covs = [[0] * nmodes for _ in range(3)]
 
-        vac_means = np.zeros((1, 2), dtype=complex)  # .tolist()
-        vac_covs = np.array([0.5 * self.circuit.hbar * np.identity(2)])
+        vac_means = np.zeros((1, 2), dtype=complex)
+        vac_covs = np.expand_dims(0.5 * self.circuit.hbar * np.identity(2), axis=0)
 
         # List of modes that have been traversed through
         reg_list = []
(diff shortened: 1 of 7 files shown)


strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +327,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+            cov = S @ cov @ S.T
+
+        return weights, means, cov
+
+    def prepare_gkp(self, state, epsilon, cutoff, desc="real", shape="square"):
+        r"""Prepares the arrays of weights, means and covs for a finite energy GKP state.
+
+        GKP states are qubits, with the qubit state defined by:
+
+        :math:`\ket{\psi}_{gkp} = \cos\frac{\theta}{2}\ket{0}_{gkp} + e^(-i\phi)\sin\frac{\theta}{2}\ket{1}_{gkp}`
```
elib20, line 478:
```suggestion
        :math:`\ket{\psi}_{gkp} = \cos\frac{\theta}{2}\ket{0}_{gkp} + e^{-i\phi}\sin\frac{\theta}{2}\ket{1}_{gkp}`
```

strawberryfields/ops.py
```diff
@@ -859,6 +878,40 @@ def _apply(self, reg, backend, **kwargs):
...
+
+    In their ideal form, square lattice GKP states are linear combinations of position eigenkets :math:`\ket{\cdot}_q`
+    spaced every :math:`\sqrt{\pi\hbar}`. Finite energy GKPs are attained by applying the Fock damping
+    operator :math:`e^{-\epsilon\hat{n}}` to the ideal states.
+
+    GKP states are qubits, with the qubit state defined by:
+
+    .. math::
+
+    \ket{\psi}_{gkp} = \cos\frac{\theta}{2}\ket{0}_{gkp} + e^(-i\phi)\sin\frac{\theta}{2}\ket{1}_{gkp}
```
elib20, line 892:
```suggestion
                   \ket{\psi}_{gkp} = \cos\frac{\theta}{2}\ket{0}_{gkp} + e^(-i\phi)\sin\frac{\theta}{2}\ket{1}_{gkp}
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+            # Check if an operation other than New() has already acted on these modes.
+            labels = [label.ind for label in cmd.reg]
+            isitnew = 1 - np.isin(labels, reg_list)
+            if np.any(isitnew):
+                # Operation parameters
+                pars = cmd.op.p
+                for reg in labels:
+                    # All the possible preparations should go in this loop
+                    if isinstance(cmd.op, Bosonic):
+                        weights, means, covs = [pars[i].tolist() for i in range(3)]
```
thisac, line 203: Would it be possible to transform `pars[i]` to `ndarrays` (does `[np.array(pars[i]) for i in range(3)]` work instead)? It seems like most `weights`, `means` and `covs` below will be `ndarrays`, so just to be consistent and avoid potential issues down the road.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +350,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+
+        :math:`\ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha})`.
+
+        Args:
+            alpha (float): alpha value of cat state
+            phi (float): phi value of cat state
+            cutoff (float): if using the 'real' representation, this determines
+                 how many terms to keep
+            desc (string): whether to use the 'real' or 'complex' representation
+            D (float): for 'real rep., quality parameter of approximation
```
thisac, line 364:
```suggestion
            D (float): for 'real' representation, quality parameter of approximation
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +350,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+            rcomplex = np.sqrt(2 * hbar) * np.array([1j * alpha.imag, -1j * alpha.real])
+            # Coefficient for complex Gaussians
+            cplx_coef = np.exp(-2 * np.absolute(alpha) ** 2 - 1j * phi)
+            # Arrays of weights, means and covs
+            weights = norm * np.array([1, 1, cplx_coef, np.conjugate(cplx_coef)])
+            weights /= np.sum(weights)
+            means = np.array([rplus, -rplus, rcomplex, np.conjugate(rcomplex)])
+            covs = 0.5 * hbar * np.identity(2, dtype=float)
+            covs = np.repeat(covs[None, :], weights.size, axis=0)
+            return weights, means, covs
```
thisac, line 398: Empty lines for legibility.
```suggestion
        if desc == "complex":
            phi = np.pi * phi

            # Mean of |alpha><alpha| term
            rplus = np.sqrt(2 * hbar) * np.array([alpha.real, alpha.imag])
            # Mean of |alpha><-alpha| term
            rcomplex = np.sqrt(2 * hbar) * np.array([1j * alpha.imag, -1j * alpha.real])
            # Coefficient for complex Gaussians
            cplx_coef = np.exp(-2 * np.absolute(alpha) ** 2 - 1j * phi)

            # Arrays of weights, means and covs
            weights = norm * np.array([1, 1, cplx_coef, np.conjugate(cplx_coef)])
            weights /= np.sum(weights)
            means = np.array([rplus, -rplus, rcomplex, np.conjugate(rcomplex)])
            covs = 0.5 * hbar * np.identity(2, dtype=float)
            covs = np.repeat(covs[None, :], weights.size, axis=0)
            
            return weights, means, covs
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +350,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+        :math:`\ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha})`.
+
+        For this representation, weights, means and covariances are real-valued.
+
+        Args:
+            alpha (float): alpha value of cat state
+            phi (float): phi value of cat state
+            cutoff (float): if using the 'real' representation, this determines
+                 how many terms to keep
+            D (float): for 'real rep., quality parameter of approximation
```
thisac, line 415:
```suggestion
            D (float): for 'real' representation, quality parameter of approximation
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +350,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+
+        where the computational basis states are :math:`\ket{\mu}_{gkp} = \sum_{n} \ket{(2n+\mu)\sqrt{\pi\hbar}}_{q}`.
+
+        Args:
+            state (list): [theta,phi] for qubit definition above
+            epsilon (float): finite energy parameter of the state
+            cutoff (float): if using the 'real' representation, this determines
+                how many terms to keep
+            desc (string): 'real' or 'complex' reprsentation
+            shape (string): 'square' lattice or otherwise
```
thisac, line 511: What do you mean by this? Can it be any shape or must it be a square lattice? Maybe better have something like "shape (string): shape of the lattice; default 'square'".

thisac: Also, (minor) I think we more often use `(str)` instead of `(string)` in the code base, although I'm not sure what's actually preferred, or if it matters much.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +350,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+
+        Returns:
+            tuple: arrays of the weights, means and covariances for the state
+
+        Raises:
+            NotImplementedError: if the complex representation or a non-square lattice
+                                is attempted
```
thisac, line 518:
```suggestion
            NotImplementedError: if the complex representation or a non-square lattice is attempted
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +350,319 @@ def prepare_gaussian_state(self, r, V, modes):
...
+
+        Args:
+            n (int): photon number
+            r (float): quality parameter for the approximation
+
+        Returns:
+            tuple: arrays of the weights, means and covariances for the state
+
+        Raises:
+            ValueError: if 1/r**2 is less than n
```
thisac, line 637:
```suggestion
            ValueError: if :math:`1/r^2` is less than n
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -143,27 +672,33 @@ def displacement(self, r, phi, mode):
     def squeeze(self, r, phi, mode):
         self.circuit.squeeze(r, phi, mode)
 
-    def mb_squeeze(self, mode, r, phi, r_anc, eta_anc, avg):
+    def mb_squeeze_avg(self, mode, r, phi, r_anc, eta_anc):
```
thisac, line 675: 💯

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -143,27 +672,33 @@ def displacement(self, r, phi, mode):
...
         r"""Squeeze mode by the amount ``r*exp(1j*phi)`` using measurement-based squeezing.
+        Here, the average, deterministic Gaussian CPTP map is applied.
+
+        Args:
+            k (int): mode to be squeezed
+            r (float): target squeezing magnitude
+            phi (float): target squeezing phase
+            r_anc (float): squeezing magnitude of the ancillary mode
+            eta_anc(float): detection efficiency of the ancillary mode
```
thisac, line 684:
```suggestion
            eta_anc (float): detection efficiency of the ancillary mode
```

strawberryfields/ops.py
```diff
@@ -859,6 +878,40 @@ def _apply(self, reg, backend, **kwargs):
...
+    lattices, such has hexagonal GKPs, can be obtained by applying symplectic transformations to the
+    square lattice GKPs.
+
+    Args:
+        state (list): [theta,phi] for qubit definition above
+        epsilon (float): finite energy parameter of the state
+        cutoff (float): if using the 'real' representation, this determines
+                how many terms to keep
+        desc (string): 'real' or 'complex' reprsentation
+        shape (string): 'square' lattice or otherwise
```
thisac, line 906: Same comment here as above. Should perhaps write something more akin to "the shape of the lattice".

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+        with prog.context as q:
+            sf.ops.Catstate(alpha,phi) | q[0]
+              
+        backend = bosonic.BosonicBackend()
+        backend.run_prog(prog)
+        state = backend.state()
+        
+       
+        if alpha != 0:
+            #Check shapes
```
thisac, line 58:
```suggestion
            # Check shapes
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+            assert state.num_weights == 4
+            assert state.weights().shape == (4,)
+            assert np.allclose(sum(state.weights()),1)
+            assert state.means().shape == (4,2)
+            assert state.covs().shape == (4,2,2)
+            covs_compare = np.tile(np.eye(2)*sf.hbar/2,(4,1,1))
+            assert np.allclose(state.covs(), covs_compare)
+            
+            #Weights not real if phi != 0 or 1
```
thisac, line 67:
```suggestion
            # Weights not real if phi != 0 or 1
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+                sf.ops.Catstate(alpha,phi) | qf[0]
+            else:
+                sf.ops.Vacuum() | qf[0]
+        eng = sf.Engine("fock", backend_options={"cutoff_dim": 20})
+        results = eng.run(prog_cat_fock)
+        wigner_fock = results.state.wigner(0,x,p)
+        
+        assert np.allclose(wigner_complex,wigner_real,rtol=1e-3,atol=1e-6)
+        assert np.allclose(wigner_complex,wigner_fock,rtol=1e-3,atol=1e-6)
+        assert np.allclose(wigner_fock,wigner_real,rtol=1e-3,atol=1e-6)
```
thisac, line 162:
```suggestion
        assert np.allclose(wigner_complex, wigner_real, rtol=1e-3, atol=1e-6)
        assert np.allclose(wigner_complex, wigner_fock, rtol=1e-3, atol=1e-6)
        assert np.allclose(wigner_fock, wigner_real, rtol=1e-3, atol=1e-6)
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+        r"""Checks fock states in the bosonic representation."""
+        prog = sf.Program(1)
+        with prog.context as q:
+            sf.ops.Fock(n) | q[0]
+              
+        backend = bosonic.BosonicBackend()
+        backend.run_prog(prog)
+        state = backend.state()
+        
+        #Check shapes
```
thisac, line 227:
```suggestion
        # check shapes
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+        assert state.num_weights == n+1
+        assert state.weights().shape == (n+1,)
+        assert np.allclose(sum(state.weights()),1)
+        assert state.means().shape == (n+1,2)
+        assert state.covs().shape == (n+1,2,2)
+        
+        #Check mean photon is close to n
```
thisac, line 234:
```suggestion
        # check mean photon is close to n
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+        mean,var = state.mean_photon(0)
+        assert np.allclose(mean,n,atol=r_fock)
+        assert np.allclose(var,0,atol=r_fock)
```
thisac, line 237:
```suggestion
        assert np.allclose(mean, n, atol=r_fock)
        assert np.allclose(var, 0, atol=r_fock)
```

thisac: There are a lot of missing spaces after `,` and between operations. Might be good to run a linter to make it a bit more legible?

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+        prog2 = sf.Program(1)
+        with prog2.context as q2:
+            sf.ops.Catstate(alpha) | q2[0]
+        backend2.run_prog(prog2)
+        state2 = backend2.state()
+        
+        #Check the two states are equal
+        assert state.__eq__(state2)
+        
+class TestBosonicPrograms():
```
thisac, line 405: No need for parentheses if there is no inheritance.
```suggestion
class TestBosonicPrograms:
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+
+    @pytest.mark.parametrize("alpha", ALPHA_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_init_circuit(self,alpha,r):
+        """Checks init_circuit only prepares non-Gaussian states and vacuum."""
+        prog = sf.Program(3)
+        with prog.context as q:
+            sf.ops.Catstate(alpha) | q[0]
+            #This line should be ignored since it is a Gaussian prep
+            #that would be picked up in run_prog, but not init_circuit
```
thisac, line 416:
```suggestion
            # this line should be ignored since it is a Gaussian prep
            # that would be picked up in run_prog, but not init_circuit
```

tests/backend/test_mbsqueeze_operation.py
```diff
@@ -36,7 +36,7 @@ def test_avg_zero_mbsqueezing(self, setup_backend, r_anc, eta, tol):
         """Tests average map of mbsqueezing operation for no squeezing,
         where the result should be a vacuum state."""
         backend = setup_backend(2)
-        backend.mb_squeeze(0, 0, 0, r_anc, eta, True)
+        backend.mb_squeeze_avg(0, 0, 0, r_anc, eta)
```
thisac, line 39: 👍

tests/backend/test_bosonic_backend.py
thisac: (minor) but there are a few more occurrences below.

thisac (commented):
Mainly checked code quality, and had a few comments. Most fairly minor, or me just noticing small things that could be slightly improved. 🙂

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+        """
+        from strawberryfields.ops import (
+            Bosonic,
+            Catstate,
+            DensityMatrix,
+            Fock,
+            GKP,
+            Ket,
+            MbSgate,
+            _New_modes,
```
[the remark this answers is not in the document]
elib20: I will double-check with Ilan, but I believe that was the issue. I tried importing it at the top, and got a host of errors.

strawberryfields/backends/bosonicbackend/backend.py
elib20, line 511:
```suggestion
            desc (str): 'real' or 'complex' reprsentation
            shape (str): shape of the lattice; default 'square'
```

strawberryfields/ops.py
elib20, line 906:
```suggestion
        desc (str): 'real' or 'complex' reprsentation
        shape (str): shape of the lattice; default 'square'
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,560 @@
...
+        implemented."""   
+        with pytest.raises(NotImplementedError):
+            prog = sf.Program(1)
+            with prog.context as q:
+                sf.ops.GKP(desc="complex") | q[0]
+                  
+            backend = bosonic.BosonicBackend()
+            backend.run_prog(prog)
+
+class TestBosonicUserSpecifiedState():
```
elib20, line 374:
```suggestion
class TestBosonicUserSpecifiedState:
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+                        "The operation {} has not been implemented in the Bosonic Backend for the arguments {}.".format(
+                            cmd.op, kwargs
+                        )
+                    ) from None
+
+        return applied, samples_dict, all_samples
+
+    def init_circuit(self, prog):
+        """Instantiate the circuit and initialize weights, means, and covs
+        depending on the Preparation classes.
```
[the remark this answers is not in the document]
elib20: The various ops that we employ in this section all inherit from the sf.ops.Preparation class.

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+        assert np.allclose(sum(state.weights()), 1)
+        assert state.means().shape == (num_weights_low, 2)
+        assert state.covs().shape == (num_weights_low, 2, 2)
+
+        # Weights, means and covs should be real
+        assert np.allclose(state.weights().real, state.weights())
+        assert np.allclose(state.means().real, state.means())
+        assert np.allclose(state.covs().real, state.covs())
+
+        prog = sf.Program(1)
```
antalszava, line 109: This seems to be a separate test case, worth creating a separate test method.

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+
+        backend_real = bosonic.BosonicBackend()
+        backend_real.run_prog(prog_real_cat)
+        state_real = backend_real.state()
+        parity_real = state_real.parity_expectation([0])
+
+        assert np.allclose(parity_complex, 1)
+        assert np.allclose(parity_real, 1)
+
+        # for phi = 1, should yield parity of -1 unless alpha == 0
```
antalszava, line 192: Seems to be a separate test case, could separate. Also taking `0` out of `ALPHA_VALS` for this test will help eliminate the `if alpha != 0:` condition.

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+        assert np.allclose(state.covs().real, state.covs())
+
+        # Covariance should be diagonal with these entries
+        cov_val = (1 - np.exp(-2 * eps)) / (1 + np.exp(-2 * eps)) * sf.hbar / 2
+        covs_compare = np.tile(cov_val * np.eye(2), (num_weights, 1, 1))
+        assert np.allclose(state.covs(), covs_compare)
+
+        # Means should be integer multiples of sqrt(pi*hbar)/2 times a damping
+        damping = 2 * np.exp(-eps) / (1 + np.exp(-2 * eps))
+        mean_ints = state.means() / (damping * np.sqrt(np.pi * sf.hbar) / 2)
```
antalszava, line 317:
```suggestion
        mean_ints = state.means() / (damping * np.sqrt(np.pi * sf.hbar) / 2)

```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+        # Round to make sure numerical errors are washed out
+        mean_ints = np.round(np.real_if_close(mean_ints), 10)
+        assert np.allclose(mean_ints % 1, np.zeros(mean_ints.shape))
+
+    @pytest.mark.parametrize("eps", EPS_VALS)
+    def test_gkp_logical(self, eps):
+        r"""Checks that logically equivalent GKP states have
+        Wigner functions that agree."""
+        x = np.linspace(-3 * np.sqrt(np.pi), 3 * np.sqrt(np.pi), 40)
+        p = np.linspace(-3 * np.sqrt(np.pi), 3 * np.sqrt(np.pi), 40)
```
antalszava, line 327:
```suggestion
        p = np.copy(x)
```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+        state = backend.state()
+        assert np.allclose(state.parity_expectation([0]), 1, atol=r_fock)
+
+    def test_gkp_complex(self):
+        r"""Checks that trying to call a complex representation
+        raises an error.
+
+        This test can be updated once the complex representation is
+        implemented."""
+        with pytest.raises(NotImplementedError):
```
antalszava, line 370: Let's shrink the scope of this context. Right now any statement within in could raise a `NotImplementedError` error (e.g., even `prog = sf.Program(1)`).

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+    """Tests that small programs run and return the correct output."""
+
+    @pytest.mark.parametrize("alpha", ALPHA_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_init_circuit(self, alpha, r):
+        """Checks init_circuit only prepares non-Gaussian states and vacuum."""
+        prog = sf.Program(3)
+        with prog.context as q:
+            sf.ops.Catstate(alpha) | q[0]
+            # this line should be ignored since it is a Gaussian prep
```
antalszava, line 421: After reading the comment: could it be removed?

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+            # that would be picked up in run_prog, but not init_circuit
+            sf.ops.Squeezed(r) | q[1]
```
antalszava, line 423:
```suggestion
            sf.ops.Squeezed(r) | q[1]

```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+        assert np.allclose(covs, covs_compare)
+
+    @pytest.mark.parametrize("alpha", ALPHA_VALS)
+    @pytest.mark.parametrize("r", R_VALS)
+    def test_different_preparations(self, alpha, r):
+        """Runs a program with non-Gaussian and Gaussian preparations."""
+        prog = sf.Program(3)
+        with prog.context as q:
+            sf.ops.Catstate(alpha) | q[0]
+            sf.ops.Squeezed(r) | q[1]
```
antalszava, line 446: Added new lines, recommend doing it at other places too.
```suggestion
        prog = sf.Program(3)

        with prog.context as q:
            sf.ops.Catstate(alpha) | q[0]
            sf.ops.Squeezed(r) | q[1]

```

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+            assert np.allclose(sum(state.weights()), 1)
+            assert state.means().shape == (1, 4)
+            assert state.covs().shape == (1, 4, 4)
+
+    def test_raised_errors(self):
+        """Runs programs with operations that are not applicable
+        or have not been implemented."""
+        from strawberryfields.backends.base import NotApplicableError
+
+        with pytest.raises(NotApplicableError):
```
antalszava, line 543: Suggest reducing the context scope here too.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+                        weights, means, covs = self.prepare_fock(*pars)
+
+                    elif type(cmd.op) in (Ket, DensityMatrix):
+                        raise NotImplementedError(
+                            "Ket and DensityMatrix preparation not implemented in the bosonic backend."
+                        )
+
+                    # If a new mode is added in the program context, then add it here,
+                    # but don't add its label to the reg_list in case there is a subsequent
+                    # non-Gaussian state prep added
```
antalszava, line 219: Adding the label to the reg_list doesn't seem to be explicitly done until later, how exactly is that avoided here?

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+        for cmd in prog.circuit:
+            # Check if an operation other than New() has already acted on these modes.
+            labels = [label.ind for label in cmd.reg]
+            isitnew = 1 - np.isin(labels, reg_list)
+            if np.any(isitnew):
+                # Operation parameters
+                pars = cmd.op.p
+                for reg in labels:
+                    # All the possible preparations should go in this loop
+                    if isinstance(cmd.op, Bosonic):
```
antalszava, line 200: Would consider moving the branching if statement structure into a separate method if feasible (maybe not due to the new modes branch).

That would also help with:
* removing the too many branches linter check,
* shortening this method,
* including the comments as the docstring of the new method.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+            ancilla_gates = (MbSgate,)
+            # For ancilla-assisted gates, if they return measurement values, store
+            # them in ancillae_samples_dict
+            if type(cmd.op) in ancilla_gates:
+                # if the op returns a measurement outcome store it in a dictionary
+                val = cmd.op.apply(cmd.reg, self, **kwargs)
+                if val is not None:
+                    for i, r in enumerate(cmd.reg):
+                        if r.ind not in self.ancillae_samples_dict.keys():
+                            self.ancillae_samples_dict[r.ind] = [val[:, i]]
```
antalszava, line 120: Suggest checking if using a `defaultdict` could help to decrease the nested nature of this part of the code (that would allow having each member to be initialized as an empty list by default, though suggest checking the quirks of it).

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+    def run_prog(self, prog, **kwargs):
+        """Runs a strawberryfields program using the bosonic backend.
+
+        Args:
+            prog (object): sf.Program instance
+
+        Returns:
+            tuple: list of applied commands,
+                    dictionary of measurement samples,
+                    dictionary of ancilla measurement samples
```
antalszava, line 83: Or going with a bullet point structure, though worth checking the rendering.
```suggestion
            tuple: a tuple of the list of applied commands, the dictionary of measurement samples,
            and the dictionary of ancilla measurement samples
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
         self._supported["mixed_states"] = True
         self._init_modes = None
         self.circuit = None
+        self.ancillae_samples_dict = {}
+
+    def run_prog(self, prog, **kwargs):
```
antalszava, line 74: How come this method is defined? Mostly curious about how it is going to be used with the engine as other backends do not seem to define such a method.

strawberryfields/ops.py
```diff
@@ -1286,6 +1340,40 @@ def _apply(self, reg, backend, **kwargs):
         backend.thermal_loss(p[0], p[1], *reg)
 
 
+class MbSgate(Channel):
```
antalszava, line 1343: Just a thought: how about `MBSgate`? just because `B` also stands for a word.

antalszava: Also, this gate needs to be added to `channels` (doesn't appear in the docs atm)

antalszava (commented):
Hi @elib20, left some comments. Looking awesome! 💯 😍 

Nothing major, mostly suggestions related to code quality. Some cases where there still seems to be ambiguity is where there are multiple `for`/`if` nested blocks. In such cases may be worth trying to simplify them if easily feasible.

strawberryfields/ops.py
nquesada: It almost sounds like `M-BSgate`. Maybe just `MSgate`?

strawberryfields/compilers/bosonic.py
```diff
@@ -0,0 +1,68 @@
...
+        "DensityMatrix",
+        "Bosonic",
+        "GKP",
+        "Catstate",
+        "Comb",
+        # measurements
+        "MeasureHomodyne",
+        "MeasureHeterodyne",
+        # "MeasureFock",
+        # "MeasureThreshold",
```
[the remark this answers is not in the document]
nquesada: Yes, or that is the hope at least.

strawberryfields/ops.py
antalszava: Hmm, could that be confused with the [Mølmer–Sørensen gate](https://en.wikipedia.org/wiki/M%C3%B8lmer%E2%80%93S%C3%B8rensen_gate) (not CV though)? Also, if the original `MbSgate` seems like the best candidate, I'm happy to second to that 😉

strawberryfields/compilers/bosonic.py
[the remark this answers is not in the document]
elib20: Not in the remaining PRs to come in the next week, but it's a top item to implement in the next iteration of this backend.

strawberryfields/backends/bosonicbackend/backend.py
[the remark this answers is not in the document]
nquesada: yup, that is the case. I *think* @josh146 suggested to do it this way.

strawberryfields/compilers/bosonic.py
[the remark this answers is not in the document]
antalszava: 👍 Maybe worth adding a TODO comment there? Just so that it's apparent that they were included intentionally.

strawberryfields/backends/bosonicbackend/backend.py
nquesada: @antalszava : the idea here is that we want to have a custom `run_prog` that checks the type of initialization that happens for the different modes. If you know in advance how the different modes are initialized then you can directly allocate the required arrays. Otherwise, everytime a new mode is parsed one would have to generate a new array with the extra dimensions required for the particular preparation. For example if you have a circuit with two cats, this method will recognize that it needs 4x4 = 16 coefficients, means and covariances. If you instead used the usual `run_prog` you would first allocate 4 coefficients, means and covariances and then later when you read the preparation of the second cat you would have to generate the new arrays having a first index running from 0 to 16-1.

elib20: It was decided early on to build a custom engine for the bosonic backend, which is essentially what this method is, mainly since we had some complications with the initializations for each mode. You will see the changes made to the engine file in the next PR. I'm tagging @ilan-tz since he can give a more thorough answer.

strawberryfields/ops.py
nquesada: Yeah that one is only for qubits, so no chance for confusion.

strawberryfields/backends/bosonicbackend/backend.py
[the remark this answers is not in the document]
josh146: Yep, circular imports :(

strawberryfields/backends/bosonicbackend/backend.py
josh146: Yep. The original backends (Fock, Gaussian, and TF) all are run 'step-by-step'. The `LocalEngine` will initialize the backend with the number of modes and some other program independent properties, and will then separately loop over the program operations and call the corresponding backend methods.

This does not work well for the bosonic backend; information about the program is needed at initialization, but the current `LocalEngine` does not provide! In fact, the `LocalEngine` tries to do too much by itself.

A better approach is how the `RemoteEngine` works - it passes the program _directly_ to the backend, waits for the backend to interpret and execute it, and receives the results. This is the approach we have decided to go with for the Bosonic backend.

josh146: Sidenote: we did look into attempting to modify the local engine to provide this information to the backend, but you end up going down a rabbit hole of needing to patch the local engine, patch the _existing_ backends, and still have the whole thing look like a bit of a hotfix.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+        # TODO: Deal with Preparation classes in the middle of a circuit.
+        applied = []
+        samples_dict = {}
+        all_samples = {}
+        for cmd in prog.circuit:
+            nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket, _New_modes)
+            ancilla_gates = (MbSgate,)
+            # For ancilla-assisted gates, if they return measurement values, store
+            # them in ancillae_samples_dict
+            if type(cmd.op) in ancilla_gates:
```
josh146, line 114: Better to make this a call to `isinstance`, which is a lot safer that `type`:

```suggestion
            if isinstance(cmd.op, ancilla_gates):
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+                    for i, r in enumerate(cmd.reg):
+                        if r.ind not in self.ancillae_samples_dict.keys():
+                            self.ancillae_samples_dict[r.ind] = [val[:, i]]
+                        else:
+                            self.ancillae_samples_dict[r.ind].append(val[:, i])
+
+                applied.append(cmd)
+
+            # Rest of operations applied as normal
+            if type(cmd.op) not in nongausspreps + ancilla_gates:
```
josh146, line 127: (same here re: isinstance). Alternatively, you could also compare against the class name (`cmd.op.__class__.__name__`), which would avoid the need to have the imports above.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,205 @@ def __init__(self):
...
+                    elif isinstance(cmd.op, Catstate):
+                        weights, means, covs = self.prepare_cat(*pars)
+
+                    elif isinstance(cmd.op, GKP):
+                        weights, means, covs = self.prepare_gkp(*pars)
+
+                    elif isinstance(cmd.op, Fock):
+                        weights, means, covs = self.prepare_fock(*pars)
+
+                    elif type(cmd.op) in (Ket, DensityMatrix):
```
josh146, line 212: Recommend changing all checks of this type to use `isinstance` :) Note that `isinstance` supports the second argument being a _sequence_ of allowed classes!

strawberryfields/backends/bosonicbackend/backend.py
elib20:
```suggestion
            if not isinstance(cmd.op, nongausspreps + ancilla_gates):
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+            _New_modes,
+        )
+
+        nongausspreps = (Bosonic, Catstate, DensityMatrix, Fock, GKP, Ket)
+        nmodes = prog.init_num_subsystems
+        self.begin_circuit(nmodes)
+        # Dummy initial weights, means and covs
+        init_weights, init_means, init_covs = [[0] * nmodes for i in range(3)]
+
+        vac_means = np.zeros((1, 2), dtype=complex)  # .tolist()
```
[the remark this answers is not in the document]
elib20: In general the means are complex-valued, so even in this case where they are real-valued we set them to be complex because the could ultimately be multiplied with complex-valued means or covariances in the simulation, and we don't want a type error popping up.

strawberryfields/backends/bosonicbackend/backend.py
elib20:
```suggestion
                    elif isinstance(cmd.op, (Ket, DensityMatrix)):
```

strawberryfields/backends/bosonicbackend/backend.py
elib20: I was previewing what was going to happen. I'll amend the comment.

elib20, line 219:
```suggestion
                    # If a new mode is added in the program context, then add it here
```

strawberryfields/compilers/bosonic.py
elib20, line 47:
```suggestion
        # TODO: "MeasureFock",
        # TODO: "MeasureThreshold",
```

strawberryfields/backends/bosonicbackend/backend.py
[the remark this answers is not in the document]
elib20: I'm just using what it was already called in the ops.py file.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -134,6 +348,321 @@ def prepare_gaussian_state(self, r, V, modes):
...
+
+            Args:
+                peak_loc (array): location of the ideal peak in phase space
+
+            Returns:
+                float: weight of the peak
+            """
+            l, m = peak_loc[:, 0], peak_loc[:, 1]
+            t = np.zeros(peak_loc.shape[0], dtype=complex)
+            t += np.logical_and(l % 2 == 0, m % 2 == 0)
```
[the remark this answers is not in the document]
elib20: This is an implementation of Appendix B of the manuscript. We can cite that once it is out?

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -197,12 +732,12 @@ def measure_homodyne(self, phi, mode, shots=1, select=None, **kwargs):
             val = select * 2 / np.sqrt(2 * self.circuit.hbar)
             self.circuit.post_select_homodyne(mode, val, **kwargs)
 
-        return np.array([val * np.sqrt(2 * self.circuit.hbar) / 2])
+        return np.array([[val * np.sqrt(2 * self.circuit.hbar) / 2]])
```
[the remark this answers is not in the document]
elib20: I realized that the way the values get unpacked later, it expects this shape.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+        from strawberryfields.ops import (
+            Bosonic,
+            Catstate,
+            DensityMatrix,
+            Fock,
+            GKP,
+            Ket,
+            MbSgate,
+            _New_modes,
+        )
```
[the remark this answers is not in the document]
elib20: This unfortunately didn't work :(

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -12,14 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # pylint: disable=too-many-public-methods
+# pylint: disable=import-outside-toplevel
+# pylint: disable=too-many-branches
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -12,14 +12,29 @@
...
-from .bosoniccircuit import BosonicModes
+from strawberryfields.backends.bosonicbackend.bosoniccircuit import BosonicModes
+from strawberryfields.backends.base import NotApplicableError
+
+
+def kron_list(l):
+    """Take Kronecker products of a list of lists."""
+    if len(l) == 1:
+        return l[0]
+    return np.kron(l[0], kron_list(l[1:]))
```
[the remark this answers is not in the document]
elib20: Done.

strawberryfields/backends/bosonicbackend/backend.py
elib20: Done.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+                        "The operation {} cannot be used with the Bosonic Backend.".format(cmd.op)
+                    ) from None
+
+                except NotImplementedError:
+                    # command not directly supported by backend API
+                    raise NotImplementedError(
+                        "The operation {} has not been implemented in the Bosonic Backend for the arguments {}.".format(
+                            cmd.op, kwargs
+                        )
+                    ) from None
```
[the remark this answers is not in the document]
elib20: This structure was copied from the Engine class, so I am not sure of the reason for its implementation.

strawberryfields/ops.py
```diff
@@ -859,6 +878,39 @@ def _apply(self, reg, backend, **kwargs):
...
+    where the computational basis states are :math:`\ket{\mu}_{gkp} = \sum_{n} \ket{(2n+\mu)\sqrt{\pi\hbar}}_{q}`.
+
+    Square lattice GKPs have Wigner functions with peaks arranged on a square lattice, whereas alternative
+    lattices, such has hexagonal GKPs, can be obtained by applying symplectic transformations to the
+    square lattice GKPs.
+
+    Args:
+        state (list): [theta,phi] for qubit definition above
+        epsilon (float): finite energy parameter of the state
+        cutoff (float): if using the 'real' representation, this determines
```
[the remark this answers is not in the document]
elib20: Done

strawberryfields/ops.py
elib20: I'll change to MSgate

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,570 @@
...
+            # Check shapes
+            assert state.num_weights == 4
+            assert state.weights().shape == (4,)
+            assert np.allclose(sum(state.weights()), 1)
+            assert state.means().shape == (4, 2)
+            assert state.covs().shape == (4, 2, 2)
+            covs_compare = np.tile(np.eye(2) * sf.hbar / 2, (4, 1, 1))
+            assert np.allclose(state.covs(), covs_compare)
+
+            # Weights not real if phi != 0 or 1
```
elib20, line 69:
```suggestion
            # Weights should be real if phi is an integer
```

tests/backend/test_bosonic_backend.py
elib20: I would like to test the case where alpha = 0, though.

elib20: It also future-proofs against someone adding 0 to the list of alphas to test.

tests/backend/test_bosonic_backend.py
elib20: I want to test that the line gets ignored.

tests/backend/test_bosonic_backend.py
elib20: Done.

tests/backend/test_bosonic_backend.py
elib20: Done.

tests/backend/test_bosonic_backend.py
elib20: Done.

tests/backend/test_bosonic_backend.py
```diff
@@ -0,0 +1,566 @@
...
+            if phi % 1 == 0:
```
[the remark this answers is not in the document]
elib20: You are right. I've modified the comment to indicate that it will be true for all integers.

strawberryfields/backends/bosonicbackend/backend.py
elib20: Let's keep this on the list of things to investigate in the next round of improvements.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -54,6 +69,207 @@ def __init__(self):
...
+
+        # Find all possible combinations of means and combs of the
+        # Gaussians between the modes.
+        mean_combos = it.product(*init_means)
+        cov_combos = it.product(*init_covs)
+
+        # Tensor product of the weights.
+        weights = kron_list(init_weights)
+        # De-nest the means iterator.
+        means = np.array([[a for b in tup for a in b] for tup in mean_combos], dtype=complex)
```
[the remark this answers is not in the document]
elib20: I've modified to make the array construction analogous to what is done for the covariances.

strawberryfields/compilers/bosonic.py
```diff
@@ -0,0 +1,68 @@
...
+        # TODO: "MeasureThreshold",
+        # channels
+        "LossChannel",
+        "ThermalLossChannel",
+        # single mode gates
+        "Dgate",
+        "Sgate",
+        "Rgate",
+        "BSgate",
+        "MbSgate",
```
elib20, line 56:
```suggestion
        "MSgate",
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -445,13 +457,12 @@ def prepare_cat_real_rep(self, alpha, phi, cutoff, D):
...
+        # Creating and calculating the weights array for oscillating terms
+        term_inds = np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int)
+        odd_terms = term_inds % 2
         even_terms = (odd_terms + 1) % 2
-        even_phases = (-1) ** ((np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int) % 4) // 2)
-        odd_phases = (-1) ** (
-            1 + ((np.array(range(-2 * z_max, 2 * z_max + 1), dtype=int) + 2) % 4) // 2
-        )
+        even_phases = (-1) ** ((term_inds % 4) // 2)
+        odd_phases = (-1) ** (1 + ((term_inds + 2) % 4) // 2)
```
antalszava: 💯

tests/backend/test_bosonic_backend.py
```diff
@@ -421,6 +436,7 @@ def test_init_circuit(self, alpha, r):
             # this line should be ignored since it is a Gaussian prep
```
antalszava, line 436:
```suggestion
            # this line should have no effect since it is a Gaussian prep
```

strawberryfields/ops.py
```diff
@@ -2796,11 +2906,17 @@ def _decompose(self, reg, **kwargs):
     Catstate,
     Thermal,
 )  # have __init__ methods with default arguments
-state_preparations = simple_state_preparations + (Ket, DensityMatrix)
+state_preparations = simple_state_preparations + (Ket, DensityMatrix, Bosonic, GKP)
```
antalszava, line 2909: Just curious for my own understanding: `Gaussian` can be decomposed, could `Bosonic` be also decomposed? (Or that would be more complex.)

antalszava (approved):
@elib20  Looks good to me, amazing! 🥇 😍 So good to see this addition. 🙂

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -198,7 +207,7 @@ def init_circuit(self, prog):
                 for reg in labels:
                     # All the possible preparations should go in this loop
                     if isinstance(cmd.op, Bosonic):
-                        weights, means, covs = [pars[i].tolist() for i in range(3)]
+                        weights, means, covs = [pars[i] for i in range(3)]
```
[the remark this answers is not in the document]
elib20: The other preparations return weights, means and covs as arrays. I think `.tolist()` was a legacy from when the other preparations were returning these as lists. But since the others return as arrays, this should too.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -348,51 +355,57 @@ def prepare_gaussian_state(self, r, V, modes):
         self.circuit.from_covmat(cov, modes)
         self.circuit.from_mean(means, modes)
 
-    def prepare_cat(self, alpha, phi, cutoff, desc, D):
+    def prepare_cat(self, alpha, phi, representation, cutoff, D):
         r"""Prepares the arrays of weights, means and covs for a cat state:
 
-        :math:`\ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi} \ket{-\alpha})`.
+        :math:`\ket{\text{cat}(\alpha)} = \frac{1}{N} (\ket{\alpha} +e^{i\phi\pi} \ket{-\alpha})`.
```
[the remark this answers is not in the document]
elib20: I don't know why it's the case that phi needs to be multiplied by pi (this is the case for the Fock backend too), but the equation should reflect this.

strawberryfields/ops.py
elib20: When you say `Gaussian` can be decomposed, do you mean that you can write a general gaussian preparation in terms of interferometers and squeezers? If that's what you mean, then no, Bosonic cannot be decomposed in the same way. I am not aware of a general set of operations that can that can decompose a non-Gaussian state preparation.

antalszava: Thanks! Yes, was considering how [the `Gaussian` operation is decomposed in SF](https://strawberryfields.readthedocs.io/en/stable/code/api/strawberryfields.ops.Gaussian.html?highlight=gaussian#strawberryfields.ops.Gaussian).

PR #545, opened 2021-02-23
Simulon device spec `mode` validation

thisac:
**Context:**
`program.compile` doesn't support device specifications with dictionary mode entries (currently only Simulon and TD2). The TD2 compilation is handled by `tdmprogram.compile`, and the device spec validation works fine, while the Simulon target compilation is handled by `program.compile` which currently will not perform any validation based on the modes entry since it is a dictionary.

The current device spec mode entry for Simulon looks like this:
```
"modes": {
    "max": {
        "pnr": 20,
        "homodyne": 1000,
        "heterodyne": 1000
    }
},
```
and is currently ignored in SF, since it's a dictionary instead of an integer.

A new `assert_number_of_measurements` method is added that, similarly to what the existing `assert_number_of_modes` does for integer mode entries, asserts that the program conforms to the mode entry when it contains the maximum number of allowed measurements as written in the example above.

This is only run if the target is Simulon, but could easily be updated in the future if other targets will have a similar entry for modes.

**Benefits:**
The number of measurements are checked to be below the allowed number of measurements according to the device specification for Simulon.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -500,17 +500,20 @@ class Program:
             elif "MeasureHeterodyne" in op_name or "MeasureHD" in op_name:
                 num_heterodyne += len(c.reg)
 
-        too_many_measurments = (
-            num_pnr > max_pnr,
-            num_homodyne > max_homodyne,
-            num_heterodyne > max_heterodyne,
-        )
-        if any(too_many_measurments):
+        if num_pnr > max_pnr:
             raise CircuitError(
-                f"This program contains {num_pnr} fock measurements, {num_homodyne} homodyne "
-                f"measurements and {num_heterodyne} heterodyne measurements.\n"
-                f"A maximum of {max_pnr} fock measurements, {max_homodyne} homodyne "
-                f"measurements, {max_heterodyne} heterodyne measurements are supported."
+                f"This program contains {num_pnr} fock measurements. "
+                f"A maximum of {max_pnr} fock measurements are supported."
+            )
+        if num_homodyne > max_homodyne:
+            raise CircuitError(
+                f"This program contains {num_homodyne} homodyne measurements. "
+                f"A maximum of {max_homodyne} homodyne measurements are supported."
+            )
+        if num_heterodyne > max_heterodyne:
+            raise CircuitError(
+                f"This program contains {num_heterodyne} heterodyne measurements. "
+                f"A maximum of {max_heterodyne} heterodyne measurements are supported."
             )
 
     def compile(self, *, device=None, compiler=None, **kwargs):
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -507,8 +507,8 @@ class Program:
             raise CircuitError(
                 f"This program contains {num_pnr} fock measurements, {num_homodyne} homodyne "
                 f"measurements and {num_heterodyne} heterodyne measurements.\n"
-                f"Simulon supports a maximum of {max_pnr} fock measurements, {max_homodyne} "
-                f"homodyne measurements, {max_heterodyne} heterodyne measurements respectively."
+                f"A maximum of {max_pnr} fock measurements, {max_homodyne} homodyne "
+                f"measurements, {max_heterodyne} heterodyne measurements are supported."
             )
 
 
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -597,8 +597,8 @@ class Program:
                     # is provided as an integer
                     self.assert_number_of_modes(device)
                 elif device.target == "simulon":
-                    # check that the number of measurements are withing the allowed
-                    # limits; device.modes will be a dictionary
+                    # check that the number of measurements is within the allowed
+                    # limits for each measurement type; device.modes will be a dictionary
                     self.assert_number_of_measurements(device)
 
         else:
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -600,7 +600,7 @@ class Program:
                     # Check that the number of modes is correct, if device.modes
                     # is provided as an integer
                     self.assert_number_of_modes(device)
-                elif device.target == "simulon":
+                else:
                     # check that the number of measurements is within the allowed
                     # limits for each measurement type; device.modes will be a dictionary
                     self.assert_max_number_of_measurements(device)
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -458,11 +458,39 @@ class Program:
         # of the program registers represents the total number of modes that has ever existed.
         modes_total = len(self.reg_refs)
 
-        if modes_total > device.modes:
+        if isinstance(device.modes, int) and modes_total > device.modes:
             raise CircuitError(
                 f"This program contains {modes_total} modes, but the device '{device.target}' "
                 f"only supports a {device.modes}-mode program."
             )
+
+    def assert_number_of_measurements(self, device):
+        """TODO"""
+        num_pnr, num_homodyne, num_heterodyne = 0, 0, 0
+
+        max_pnr = device.modes["max"]["pnr"]
+        max_homodyne = device.modes["max"]["homodyne"]
+        max_heterodyne = device.modes["max"]["heterodyne"]
+
+        for c in self.circuit:
+            if "MeasureFock" in str(c.op):
+                num_pnr += len(c.reg)
+            if "MeasureHomodyne" in str(c.op):
+                num_homodyne += len(c.reg)
+            if "MeasureHeterodyne" in str(c.op):
+                num_heterodyne += len(c.reg)
+
+        too_many_measurments = (
+            num_pnr > max_pnr, num_homodyne > max_homodyne, num_heterodyne > max_heterodyne
+        )
+        if any(too_many_measurments):
+            raise CircuitError(
+                f"This program contains {num_pnr} fock measurements, {num_homodyne} homodyne "
+                f"measurements and {num_heterodyne} heterodyne measurements.\n"
+                f"Simulon supports a maximum of {max_pnr} fock measurements, {max_homodyne} "
+                f"homodyne measurements, {max_heterodyne} heterodyne measurements respectively."
+            )
+
 
     def compile(self, *, device=None, compiler=None, **kwargs):
         """Compile the program given a Strawberry Fields photonic compiler, or
@@ -543,9 +571,15 @@ class Program:
             else:
                 compiler = _get_compiler(compiler)
 
-            # TODO: add validation for device specs that provide a dictionary for `device.modes`.
-            if device.modes is not None and isinstance(device.modes, int):
-                self.assert_number_of_modes(device)
+            if device.modes is not None:
+                if isinstance(device.modes, int):
+                    # Check that the number of modes is correct, if device.modes
+                    # is provided as an integer
+                    self.assert_number_of_modes(device)
+                elif device.target == "simulon":
+                    # check that the number of measurements are withing the allowed
+                    # limits; device.modes will be a dictionary
+                    self.assert_number_of_measurements(device)
 
         else:
             compiler = _get_compiler(compiler)
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -597,7 +597,7 @@ class Program:
 
             if device.modes is not None:
                 if isinstance(device.modes, int):
-                    # Check that the number of modes is correct, if device.modes
+                    # check that the number of modes is correct, if device.modes
                     # is provided as an integer
                     self.assert_number_of_modes(device)
                 else:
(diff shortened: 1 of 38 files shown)


tests/frontend/test_program.py
```diff
@@ -402,6 +402,52 @@ class DummyCircuit(Compiler):
...
+            ops.MeasureHomodyne(phi=0),  # MeasureX
+            ops.MeasureHomodyne(phi=42),  # MeasureHomodyne
+            ops.MeasureHomodyne(phi=np.pi/2),  # MeasureP
+            ops.MeasureHeterodyne(),  # MeasureHD
+            ops.MeasureHeterodyne(select=0)  # MeasureHeterodyne
+        ],
+    )
+    def test_incorrect_number_of_measurements(self, measure_op):
+        """Test that an exception is raised if the compiler is called with a
+        device spec with an incorrect number of measurements"""
```
thisac, line 417: I added this test here since the other, similar tests were also here (using the same structure), although I think that it might be better to have tests directly for both `assert_number_of_measurements` and `assert_number_of_modes`, instead of implicitly through `prog.compile`.

Perhaps these tests could be updated in a later PR?

strawberryfields/program.py
```diff
@@ -464,6 +469,49 @@ def assert_number_of_modes(self, device):
...
+            raise KeyError(
+                "Device specification must contain an entry for the maximum allowed number "
+                "of measurments. Have you specified the correct target?"
+            ) from e
+
+        for c in self.circuit:
+            op_name = str(c.op)
+            if "MeasureFock" in op_name:
+                num_pnr += len(c.reg)
+            if "MeasureHomodyne" in op_name or "MeasureX" in op_name or "MeasureP" in op_name:
```
antalszava, line 498: `elif` could be useful, the rest is minor suggestion
```suggestion
            elif any(homodyne in op_name for homodyne in ("MeasureHomodyne", "MeasureX", "MeasureP")):
```

strawberryfields/program.py
```diff
@@ -464,6 +469,49 @@ def assert_number_of_modes(self, device):
...
+                num_homodyne += len(c.reg)
+            if "MeasureHeterodyne" in op_name or "MeasureHD"in op_name:
```
antalszava, line 500:
```suggestion
            elif "MeasureHeterodyne" in op_name or "MeasureHD"in op_name:
```

strawberryfields/program.py
```diff
@@ -543,9 +591,15 @@ def _get_compiler(compiler_or_name):
...
-            if device.modes is not None and isinstance(device.modes, int):
-                self.assert_number_of_modes(device)
+            if device.modes is not None:
+                if isinstance(device.modes, int):
+                    # Check that the number of modes is correct, if device.modes
+                    # is provided as an integer
+                    self.assert_number_of_modes(device)
+                elif device.target == "simulon":
+                    # check that the number of measurements are withing the allowed
+                    # limits; device.modes will be a dictionary
```
antalszava, line 601:
```suggestion
                    # check that the number of measurements is within the allowed
                    # limits for each measurement type; device.modes will be a dictionary
```

tests/frontend/test_program.py
```diff
@@ -402,6 +402,52 @@ class DummyCircuit(Compiler):
...
+
+        class DummyCircuit(Compiler):
```
antalszava, line 419:
```suggestion
        class DummyCompiler(Compiler):
```

tests/frontend/test_program.py
```diff
@@ -402,6 +402,52 @@ class DummyCircuit(Compiler):
...
+
+        prog = sf.Program(3)
+        with prog.context as q:
+            for reg in q:
+                measure_op | reg
+
+        with pytest.raises(
+            program.CircuitError, match="Simulon supports a maximum of 2 fock measurements"
+        ):
+            prog.compile(device=spec, compiler=DummyCircuit())
```
antalszava, line 449:
```suggestion
            prog.compile(device=spec, compiler=DummyCompiler())
```

strawberryfields/program.py
```diff
@@ -464,6 +469,49 @@ def assert_number_of_modes(self, device):
...
+
+        too_many_measurments = (
+            num_pnr > max_pnr, num_homodyne > max_homodyne, num_heterodyne > max_heterodyne
+        )
+        if any(too_many_measurments):
+            raise CircuitError(
+                f"This program contains {num_pnr} fock measurements, {num_homodyne} homodyne "
+                f"measurements and {num_heterodyne} heterodyne measurements.\n"
+                f"A maximum of {max_pnr} fock measurements, {max_homodyne} homodyne "
+                f"measurements, {max_heterodyne} heterodyne measurements are supported."
```
antalszava, line 511: Would we want to have a separate error for each measurement type? It's inexpensive for the user to check & update the code because this is still caught locally.

strawberryfields/program.py
```diff
@@ -543,9 +591,15 @@ def _get_compiler(compiler_or_name):
...
 
-            # TODO: add validation for device specs that provide a dictionary for `device.modes`.
-            if device.modes is not None and isinstance(device.modes, int):
-                self.assert_number_of_modes(device)
+            if device.modes is not None:
+                if isinstance(device.modes, int):
+                    # Check that the number of modes is correct, if device.modes
+                    # is provided as an integer
+                    self.assert_number_of_modes(device)
+                elif device.target == "simulon":
```
antalszava, line 599: It's as if we'd prefer not to use explicit target names here, but cannot quite remember why (perhaps was just due to the fact that it could break easily) 🤔 @josh146? Perhaps it's also less relevant now that we'll have API versioning.

tests/frontend/test_program.py
```diff
@@ -402,6 +402,52 @@ class DummyCircuit(Compiler):
...
+                "max": {
+                    "pnr": 2,
+                    "homodyne": 2,
+                    "heterodyne": 2
+                }
+            },
+            "layout": None, "gate_parameters": {}, "compiler": [None]
+        }
+        # set the target to simulon manually to avoid device spec being loaded from api;
+        # needed for the measurements check to run
```
antalszava, line 437: Hmm, isn't this achieved by having `connection=None`? It seems like other than that, `DeviceSpec.__init__` just stores the `target`.

tests/frontend/test_program.py
antalszava: That seems like a good idea! Would it be worth adding a separate unit test just for `assert_number_of_measurements`?

antalszava (commented):
@thisac this is looking great, thank you for adding this! 😍 🥇 No major blockers had some suggestions/questions, will revisit for a quick other round.

strawberryfields/program.py
```diff
@@ -464,6 +469,49 @@ def assert_number_of_modes(self, device):
                 f"only supports a {device.modes}-mode program."
             )
 
+    def assert_number_of_measurements(self, device):
```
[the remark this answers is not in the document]
thisac: Thanks! 😊

strawberryfields/program.py
thisac: Oh, true, I had `if`s instead of `elif`s there since I was thinking that there could be several type of measurements in a circuit, but this only checks one statement at a time, which will only contain a single measurement. 🤦  Thanks for catching this!

strawberryfields/program.py
thisac: True, could fix that. I don't mind it either way tbh.

strawberryfields/program.py
thisac: I think it's because names may change. 

In this case, it's not that crucial though because of, as you mentioned, the API versioning. This part would need to be updated anyway as soon as we decide to use API v0.3.0, so for now I think it's fine to make sure that only Simulon calls `assert_num_of_measurments`, since it's the only target that actually has the correct type of modes entry in API v.0.2.0.

tests/frontend/test_program.py
thisac: Not sure actually. I just noticed that the device spec was being updated when using `target="simulon"`, so the tests wouldn't work, so I added this as a solution.

tests/frontend/test_program.py
thisac: I could do that. In practice, it would do the same thing, just a bit more unit-test-like than this one. I wouldn't mind attempting to refactor some tests at a later stage and moving these tests into an integration tests folder, and making sure that all functions are explicitly tested separately. 

Do you think I should add separate tests for `assert_number_of_measurements` and `assert_number_of_modes` as well? @antalszava

antalszava: From a testing point of view, that would be great because it would constitute a unit test and the test that checks the use of `compile` would be the integration test. Though specifically for these functions, if we deem that testing both "ways" is not too crucial, perhaps we can get away with not having it. Still could be nice to have.

tests/frontend/test_program.py
antalszava: That's odd 🤔 Maybe there's something else going one there? As done here, it should be a matter of setting an attribute 🤔

thisac: It works as you said. I was probably just testing it locally, forgetting the `compiler=None`, when it didn't work (and didn't look into it). Removed it now. 🙂

tests/frontend/test_program.py
thisac: Added two extra tests for the assertion functions directly (they're basically the same tests, just avoiding the compilation part). Also added a TODO to the compiler tests.

strawberryfields/program.py
```diff
@@ -500,17 +500,20 @@ def assert_number_of_measurements(self, device):
             elif "MeasureHeterodyne" in op_name or "MeasureHD" in op_name:
                 num_heterodyne += len(c.reg)
 
-        too_many_measurments = (
-            num_pnr > max_pnr,
-            num_homodyne > max_homodyne,
-            num_heterodyne > max_heterodyne,
-        )
-        if any(too_many_measurments):
+        if num_pnr > max_pnr:
```
antalszava, line 503: 💯

tests/frontend/test_program.py
```diff
@@ -435,16 +491,15 @@ class DummyCompiler(Compiler):
...
+        spec = sf.api.DeviceSpec(target="simulon", connection=None, spec=device_dict)
 
         prog = sf.Program(3)
         with prog.context as q:
             for reg in q:
                 measure_op | reg
 
         with pytest.raises(
-            program.CircuitError, match="Simulon supports a maximum of 2 fock measurements"
+            program.CircuitError, match=f"contains 3 {measure_name} measurements"
```
antalszava, line 502: Nice!

tests/frontend/test_program.py
```diff
@@ -202,6 +202,58 @@ def test_bind_params(self, prog):
...
+                "max": {
+                    "pnr": 2,
+                    "homodyne": 2,
+                    "heterodyne": 2
+                }
+            },
+            "layout": None, "gate_parameters": {}, "compiler": [None]
+        }
+        # set the target to simulon manually to avoid device spec being loaded from api;
+        # needed for the measurements check to run
```
antalszava, line 244:
```suggestion
```

antalszava (approved):
Looks good to me! 💯 Nice one 😊

josh146 (approved):
Nice, thanks for catching this @thisac! I've preliminarily approved, but would be good to hear from @jswinarton re: the simulon target name first.

Once this PR is merged, is there any other work required to support submitting to simulon?

strawberryfields/program.py
josh146: For now, you could just have an `else:` here, and use it to just catch _everything?_ assuming TDM compilation does not go via this pipeline.

strawberryfields/program.py
```diff
@@ -452,8 +452,13 @@ def _linked_copy(self):
         return p
 
     def assert_number_of_modes(self, device):
-        """Check that the number of modes in the program is valid for the given device."""
+        """Check that the number of modes in the program is valid for the given device.
 
+        Args:
+            device (~strawberryfields.api.DeviceSpec): Device specification object to use.
```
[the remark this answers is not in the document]
thisac: Seems to work fine as it is. I'm honestly not sure about the syntax here (I just copied this from the compile method). Could both work? 🤔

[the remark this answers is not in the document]
josh146: Potentially! If sphinx is finding the correct link, feel free to leave it as is

strawberryfields/program.py
thisac: Yes, that'd work as well. @jswinarton mentioned that the name will likely change in the future, so it's probably best to remove the hard-coded "simulon" from here.

PR #547, opened 2021-02-24
Add Simulon remote access to docs

thisac:
**Context:**
Simulon is now accessible to users directly via `RemoteEngine` but this is not described anywhere in the docs, and thus should be added.

* The "Hardware" page is renamed to "Xanadu Quantum Cloud" to encompass both hardware and cloud simulators. Since running jobs on Simulon is almost identical to running on hardware, I thought this made the most sense.
* A new entry for "Cloud simulator" is added, describing how to use Simulon via the cloud.

**Benefits:**
Users will now know that Simulon is accessible as a target for `RemoteEngine`.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/doc/introduction/photonic_hardware.rst
+++ b/doc/introduction/photonic_hardware.rst
@@ -146,6 +146,12 @@ Blackbird script with the ``target`` set to ``simulon`` in the Blackbird header.
 
 See the `Submitting jobs on hardware`_ section above for more details.
 
+.. note::
+
+    Simulon runs on the ``gaussian`` backend (see `simulating_your_program`) and thus only supports
+    Gaussian operations, including homodyne and heterodyne measurements, as well terminal
+    Fock measurements.
+
 Tutorials
 ---------
 
--- a/doc/introduction/photonic_hardware.rst
+++ b/doc/introduction/photonic_hardware.rst
@@ -136,7 +136,7 @@ Cloud simulator
 Instead of submitting a job to be run on quantum hardware, it is also possible to run jobs on
 our cloud simulator Simulon. The process is very similar to running jobs on hardware. You
 will need to configure your account, as described above, and submit a job via the ``RemoteEngine``,
-using `"simulon"` as the target, instead of a specific chip:
+using `"simulon"` as the target instead of a specific chip:
 
 >>> eng = sf.RemoteEngine("simulon")
 >>> result = eng.run(prog)
--- a/doc/introduction/photonic_hardware.rst
+++ b/doc/introduction/photonic_hardware.rst
@@ -1,5 +1,5 @@
-Xanadu Quantum Cloud
-====================
+Hardware and cloud
+==================
 
 Using Strawberry Fields, you can submit quantum programs to be executed on photonic
 hardware or to be run on a cloud simulator via the Xanadu cloud platform.
@@ -69,8 +69,8 @@ or the ``ping`` flag can be used on the command line:
 
    :doc:`command line interface </code/sf_cli>`
 
-Submitting jobs
----------------
+Submitting jobs on hardware
+---------------------------
 
 Once connected to the Xanadu cloud platform, the Strawberry Fields
 :class:`~.RemoteEngine` can be used to submit quantum programs to available
@@ -144,7 +144,7 @@ using `"simulon"` as the target instead of a specific chip:
 Alternatively, it's also possible to run jobs asynchronously with `eng.run_async` or by submitting a
 Blackbird script, setting the ``target`` to ``simulon`` in the Blackbird header.
 
-See the `Submitting jobs`_ section above for more details.
+See the `Submitting jobs on hardware`_ section above for more details.
 
 Tutorials
 ---------
--- a/doc/introduction/photonic_hardware.rst
+++ b/doc/introduction/photonic_hardware.rst
@@ -148,7 +148,7 @@ See the `Submitting jobs on hardware`_ section above for more details.
 
 .. note::
 
-    Simulon runs on the ``gaussian`` backend (see `simulating_your_program`) and thus only supports
+    Simulon runs on the ``gaussian`` backend (see :ref:`simulating_your_program`) and thus only supports
     Gaussian operations, including homodyne and heterodyne measurements, as well terminal
     Fock measurements.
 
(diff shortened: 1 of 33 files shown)


doc/introduction/photonic_hardware.rst
```diff
@@ -1,8 +1,8 @@
-Hardware
-========
+Xanadu Quantum Cloud
+====================
```
josh146, line 2: This change makes sense, but at the same time I am hesitant - for a new user, seeing 'Hardware' in the side bar is much more descriptive and enticing than XQC.

Can we preserve 'hardware' somehow?

doc/introduction/photonic_hardware.rst
```diff
@@ -1,8 +1,8 @@
...
 
-Using Strawberry Fields, you can submit quantum programs
-to be executed on photonic hardware using the Xanadu cloud platform.
+Using Strawberry Fields, you can submit quantum programs to be executed on photonic
+hardware or to be run on a cloud simulator via the Xanadu cloud platform.
```
josh146, line 5:
```suggestion
hardware or to be run on a cloud simulator via the Xanadu Quantum Cloud.
```

doc/introduction/photonic_hardware.rst
```diff
@@ -130,6 +130,22 @@ the Strawberry Fields command line interface:
 After executing the above command, the result will be stored in ``out.txt`` in the current working directory.
 You can also omit the ``--output`` parameter to print the result to the screen.
 
+Cloud simulator
+---------------
+
+Instead of submitting a job to be run on quantum hardware, it is also possible to run jobs on
+our cloud simulator Simulon. The process is very similar to running jobs on hardware. You
```
josh146, line 137:
```suggestion
In addition to submitting jobs to be run on quantum hardware, it is also possible to run jobs on
the Xanadu Quantum Cloud simulator Simulon. The process is very similar to running jobs on hardware. You
```

doc/introduction/photonic_hardware.rst
```diff
@@ -130,6 +130,22 @@ the Strawberry Fields command line interface:
...
+will need to configure your account, as described above, and submit a job via the ``RemoteEngine``,
+using `"simulon"` as the target instead of a specific chip:
```
josh146, line 139:
```suggestion
using ``"simulon"`` as the target instead of a specific chip:
```

doc/introduction/photonic_hardware.rst
```diff
@@ -130,6 +130,22 @@ the Strawberry Fields command line interface:
...
+
+>>> eng = sf.RemoteEngine("simulon")
+>>> result = eng.run(prog)
+
+Alternatively, it's also possible to run jobs asynchronously with `eng.run_async` or by submitting a
+Blackbird script, setting the ``target`` to ``simulon`` in the Blackbird header.
```
josh146, line 145:
```suggestion
Simulon jobs can also be submitted asynchronously using ``eng.run_async``, or by submitting a
Blackbird script with the ``target`` set to ``simulon`` in the Blackbird header.
```

josh146 (approved):
Thanks @thisac! Left a few minor comments. In addition, it would be nice to preserve 'Hardware' in the sidebar if possible. Something like `Hardware and cloud`?

antalszava (commented):
@thisac Looking good! 🙂 💪 Not too major, it seems like `introduction/circuits.rst` is also written with having a remote hardware target in mind. Would we want to extend the text there too? 🤔

doc/introduction/photonic_hardware.rst
thisac: Sure, "Hardware and cloud" might work? I cannot think of anything better really. 🤔

doc/introduction/photonic_hardware.rst
```diff
@@ -130,6 +130,22 @@ the Strawberry Fields command line interface:
...
+
+See the `Submitting jobs`_ section above for more details.
```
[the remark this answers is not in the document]
thisac: I'm thinking that a user might scroll down to read just the "Cloud simulator" section, and not understand that everything is detailed in the section above.

thisac:
Thanks @josh146 and @antalszava. I've updated the text and heading you commented on. Do you think "Hardware and cloud" works better as a main heading?

> it seems like introduction/circuits.rst is also written with having a remote hardware target in mind

I think this part looks fine as it is. It already mentions "remote simulators/hardware" in the "Simulating your program" section, and the other mentions of remote access mainly discusses hardware and hardware compilation, so I think it's fine not mentioning simulators there.

Let me know if you disagree. 🙂

antalszava (approved):
Looks good to me, thanks for addressing the suggestions! 🙂 

> Do you think "Hardware and cloud" works better as a main heading?

No real preference here, any of the mentioned versions would work at least for me. 🙂 

> I think this part looks fine as it is. It already mentions "remote simulators/hardware" in the "Simulating your program" section, and the other mentions of remote access mainly discusses hardware and hardware compilation, so I think it's fine not mentioning simulators there.

That's a good point! One reason why it seemed relevant is that `simulon` also uses a compiler (`gaussian`). However, the description itself is indeed more tailored towards compilation for hardware targets.

josh146 (commented):
Looks good @thisac! I like `Hardware and cloud` as a title.

I just have a request for the `Cloud Simulator` section - can you add details about simulon that will be important for the user to know?

This includes things such as:

- Supported operations (basically just need to say it only supports gaussian operations)
- Supported measurements (need to say which measurements - not sure if it is worth also adding the maximum number of allowed measurements)
- Allowed number of modes.

doc/introduction/photonic_hardware.rst
```diff
@@ -146,6 +146,12 @@ Blackbird script with the ``target`` set to ``simulon`` in the Blackbird header.
 
 See the `Submitting jobs on hardware`_ section above for more details.
 
+.. note::
+
+    Simulon runs on the ``gaussian`` backend (see :ref:`simulating_your_program`) and thus only supports
+    Gaussian operations, including homodyne and heterodyne measurements, as well terminal
+    Fock measurements.
```
antalszava, line 153: Nice! 🙂 Could it be worth adding just a small comment that the number of modes depends on the type & number of measurements used? Without stating anything explicitly.

PR #548, opened 2021-02-25
Replace `pop` with indexing in `reshape_samples`

thisac:
**Context:**
The `reshape_samples` function in `tdmprogram.py` currently works by popping elements from the `all_samples` dictionary that's being sent in as an argument, and then inserting them into a new dictionary which is then returned. This causes the sent in `all_samples` dictionary to be empty after having reshaped the samples (which might be confusing for the user, since this is nowhere stated).

This is not really an optimal way of handling things, and should probably be fixed. 

The popping of samples from the input `all_samples` dictionary is changed to indexing, while keeping track of used indices. Thanks to @lneuhaus for [pointing this out](https://github.com/XanaduAI/strawberryfields/pull/542#discussion_r580376229)!

**Benefits:**
This method doesn't change the input `all_samples` and it should be equally fast as before, if not faster.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


lneuhaus (approved):
Great! Nothing to add here.

Issue #554, opened 2021-03-10
Some cat state functions use int32 arrays


tguillaume:
The functions `strawberryfields.ops.Catstate `and`strawberryfields.utils.states.cat_state `do not work properly with a large Fock dimension cut-off. For example, when I try to run `strawberryfields.utils.states.cat_state(a=4, p=0, fock_dim=50)`, the returned Fock probabilities over n=16 are zero.

Inside these functions, the ket's array is initialized as type int32 due to (I think) an earlier use of np.arange with int32 argument. As a result, some of the maths that is performed further on in the function evaluates to 0 when the number is sufficiently large. 

I was able to fix this by initializing the array describing the ket as a float, e.g.` k = np.arange(fock_dim, dtype='float')` or `l = np.arange(D, dtype='float').`

antalszava:
With #556 merged, closing this, thank you so much @tguillaume @nquesada! 🙂

PR #556, opened 2021-03-10
Fix cat state function array type without doing casting

nquesada:
Solves issue https://github.com/XanaduAI/strawberryfields/issues/554


nquesada:
Hi @tguillaume : I have fixed the bug you found by multiplying `a` bu 1.0 to automatically convert things to floats. I'd prefer this strategy to casting the "photon numbers" into floats.

strawberryfields/ops.py
```diff
@@ -867,8 +867,8 @@ def _apply(self, reg, backend, **kwargs):
         N = temp / pf.sqrt(2 * (1 + pf.cos(phi) * temp ** 4))
 
         # coherent states
-        c1 = (alpha ** l) / np.sqrt(ssp.factorial(l))
-        c2 = ((-alpha) ** l) / np.sqrt(ssp.factorial(l))
+        c1 = ((1.0 * alpha) ** l) / np.sqrt(ssp.factorial(l))
```
antalszava, line 870:
```suggestion
        # Need to cast  alpha to float before exponentiation to avoid overflow
        c1 = ((1.0 * alpha) ** l) / np.sqrt(ssp.factorial(l))
```

strawberryfields/utils/states.py
```diff
@@ -305,8 +305,8 @@ def cat_state(a, p=0, fock_dim=5):
 
     # coherent states
     k = np.arange(fock_dim)
-    c1 = (a ** k) / np.sqrt(fac(k))
-    c2 = ((-a) ** k) / np.sqrt(fac(k))
+    c1 = ((1.0 * a) ** k) / np.sqrt(fac(k))
```
antalszava, line 308:
```suggestion
    # Need to cast a to float before exponentiation to avoid overflow
    c1 = ((1.0 * a) ** k) / np.sqrt(fac(k))
```

antalszava (approved):
Looks good to me! 💯

tguillaume:
Hey @nquesada, I think this does the trick nicely. Cheers!

PR #558, opened 2021-03-18
Clean shared ops

nquesada:
Most of the functionality of `shared_ops.py` is now obtained from the walrus. This PR is checking how much of `shared_ops` can be eliminated.


josh146 (approved):
image

This is great @nquesada 🤩 Just needs to be blacked and then can be merged.

Do you think there are other parts of SF which are no longer required because of The Walrus?

PR #560, opened 2021-03-19
More cleaning shared ops

nquesada:
Removes `sympmat` and `rotation_matrix` from `shared_ops.py`. These functions are now imported from TW.


josh146 (approved):
💯

PR #562, opened 2021-03-20
Tf engine dtype

### LINKED ISSUE #486: Make tensorflow dtype adjustable by user
It would be nice to allow users to set a custom value for the tensorflow dtype if they are really pushing their simulations (and aware of the memory tradeoffs). This would be a very straightforward feature to add, most likely by adjusting `backend_options` to accept a `dtype` argument or something similar

See, e.g., the discussion below:

You could change this line https://github.com/XanaduAI/strawberryfields/blob/9800240e014d9b2844444cf92fd1b17b581dfa06/strawberryfields/backends/tfbackend/ops.py#L67

I tested it and you do get the correct result. Not surprisingly you also get a massive spike in memory usage.

_Originally posted by @nquesada in https://github.com/XanaduAI/strawberryfields/issues/485#issuecomment-722786561_

Aaron-Robertson:
**Context:**
Resolves #486 

Make tensorflow dtype adjustable by user by adding a `dtype` optional argument to `backend_options`. Follows existing patterns for optional arguments.

**Benefits:**
User can adjust `dtype`, i.e.:
```
  prog = sf.Program(2)
  eng  = sf.Engine("tf", backend_options={"cutoff_dim": 50, "dtype": tf.complex128})
  with prog.context as q:
      Sgate(0.8)         | q[0]
      Sgate(0.8)         | q[1]
      BSgate(0.5,0.5)  | (q[0], q[1])
      BSgate(0.5,0.5)  | (q[0], q[1])
  state    = eng.run(prog).state
  N0,N0var = state.mean_photon(0)
  N1,N1var = state.mean_photon(1)
  print(N0)
  print(N1)
  print('analytical:',np.sinh(0.8)**2)
```

**Possible Drawbacks:**
N/A?

**Related GitHub Issues:**
#486

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -80,16 +80,16 @@
   prog = sf.Program(2)
   eng  = sf.Engine("tf", backend_options={"cutoff_dim": 50, "dtype": tf.complex128})
   with prog.context as q:
-      Sgate(0.8)         | q[0]
-      Sgate(0.8)         | q[1]
-      BSgate(0.5,0.5)  | (q[0], q[1])
-      BSgate(0.5,0.5)  | (q[0], q[1])
-  state    = eng.run(prog).state
-  N0,N0var = state.mean_photon(0)
-  N1,N1var = state.mean_photon(1)
+      Sgate(0.8) | q[0]
+      Sgate(0.8) | q[1]
+      BSgate(0.5,0.5) | (q[0], q[1])
+      BSgate(0.5,0.5) | (q[0], q[1])
+  state = eng.run(prog).state
+  N0, N0var = state.mean_photon(0)
+  N1, N1var = state.mean_photon(1)
   print(N0)
   print(N1)
-  print('analytical:',np.sinh(0.8)**2)
+  print("analytical:" ,np.sinh(0.8)**2)
   ```
 
 <h3>Breaking Changes</h3>
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -68,8 +68,29 @@
   plt.contourf(xvec, pvec, wigner)
   plt.show()
   ```
+
 * The `fock` backend now supports the `GKP` preparation [(#553)](https://github.com/XanaduAI/strawberryfields/pull/553)
 
+* The `tf` backend now accepts the Tensor DType as argument.
+  [(#562)](https://github.com/XanaduAI/strawberryfields/pull/562)
+
+  Allows high cutoff dimension to give numerically correct calculations:
+
+  ```python
+  prog = sf.Program(2)
+  eng  = sf.Engine("tf", backend_options={"cutoff_dim": 50, "dtype": tf.complex128})
+  with prog.context as q:
+      Sgate(0.8)         | q[0]
+      Sgate(0.8)         | q[1]
+      BSgate(0.5,0.5)  | (q[0], q[1])
+      BSgate(0.5,0.5)  | (q[0], q[1])
+  state    = eng.run(prog).state
+  N0,N0var = state.mean_photon(0)
+  N1,N1var = state.mean_photon(1)
+  print(N0)
+  print(N1)
+  print('analytical:',np.sinh(0.8)**2)
+  ```
 
 <h3>Breaking Changes</h3>
 
--- a/strawberryfields/backends/tfbackend/backend.py
+++ b/strawberryfields/backends/tfbackend/backend.py
@@ -278,7 +278,13 @@ class TFBackend(BaseFock):
 
             modenames = ["q[{}]".format(i) for i in np.array(self.get_modes())[modes]]
             state_ = FockStateTF(
-                s, len(modes), pure, self.circuit.cutoff_dim, batched=batched, mode_names=modenames, dtype=dtype
+                s,
+                len(modes),
+                pure,
+                self.circuit.cutoff_dim,
+                batched=batched,
+                mode_names=modenames,
+                dtype=dtype,
             )
         return state_
 
--- a/strawberryfields/backends/tfbackend/backend.py
+++ b/strawberryfields/backends/tfbackend/backend.py
@@ -21,7 +21,7 @@ TensorFlow backend interface
 import numpy as np
 import tensorflow as tf
 
-from tensorflow import DType
+# from tensorflow import DType
 from strawberryfields.backends import BaseFock, ModeMap
 from .circuit import Circuit
 from .ops import mixed, partial_trace, reorder_modes
@@ -83,7 +83,8 @@ class TFBackend(BaseFock):
                 For each mode, the simulator can represent the Fock states :math:`\ket{0}, \ket{1}, \ldots, \ket{\text{cutoff_dim}-1}`.
             pure (bool): If True (default), use a pure state representation (otherwise will use a mixed state representation).
             batch_size (None or int): Size of the batch-axis dimension. If None, no batch-axis will be used.
-            dtype (DType): Tensorflow element types in a tensor.
+            dtype (DType): Complex Tensorflow Tensor type representation, either complex64 (default) or complex128.
+                Note, complex128 will increase memory usage substantially.
         """
         cutoff_dim = kwargs.get("cutoff_dim", None)
         pure = kwargs.get("pure", True)
@@ -99,8 +100,8 @@ class TFBackend(BaseFock):
             raise ValueError("Argument 'cutoff_dim' must be a positive integer")
         if not isinstance(pure, bool):
             raise ValueError("Argument 'pure' must be either True or False")
-        if not isinstance(dtype, DType):
-            raise ValueError("Argument 'dtype' must be a Tensorflow DType")
+        if not (dtype == tf.complex64 or dtype == tf.complex128):
+            raise ValueError("Argument 'dtype' must be a complex Tensorflow DType")
         if batch_size == 1:
             raise ValueError(
                 "batch_size of 1 not supported, please use different batch_size or set batch_size=None"
@@ -236,6 +237,7 @@ class TFBackend(BaseFock):
             pure = self.circuit.state_is_pure
             num_modes = self.circuit.num_modes
             batched = self.circuit.batched
+            dtype = self.circuit.dtype
 
             # reduce rho down to specified subsystems
             if modes is None:
@@ -276,7 +278,7 @@ class TFBackend(BaseFock):
 
             modenames = ["q[{}]".format(i) for i in np.array(self.get_modes())[modes]]
             state_ = FockStateTF(
-                s, len(modes), pure, self.circuit.cutoff_dim, batched=batched, mode_names=modenames
+                s, len(modes), pure, self.circuit.cutoff_dim, batched=batched, mode_names=modenames, dtype=dtype
             )
         return state_
 
--- a/strawberryfields/backends/tfbackend/backend.py
+++ b/strawberryfields/backends/tfbackend/backend.py
@@ -83,7 +83,7 @@ class TFBackend(BaseFock):
                 For each mode, the simulator can represent the Fock states :math:`\ket{0}, \ket{1}, \ldots, \ket{\text{cutoff_dim}-1}`.
             pure (bool): If True (default), use a pure state representation (otherwise will use a mixed state representation).
             batch_size (None or int): Size of the batch-axis dimension. If None, no batch-axis will be used.
-            dtype (DType): Tensorflow element types in a tensor. 
+            dtype (DType): Tensorflow element types in a tensor.
         """
         cutoff_dim = kwargs.get("cutoff_dim", None)
         pure = kwargs.get("pure", True)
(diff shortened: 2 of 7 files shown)


Aaron-Robertson:
@josh146 I'm just coming back to this one today and it looks to me like the simple change to `ops.def_type` is not a valid proof of concept anymore (errors coming back from thewalrus). Want me to push up that line change here so you can take a look at test errors?

josh146:
> Want me to push up that line change here so you can take a look at test errors?

Yes, that would be helpful!

co9olguy:
@Aaron-Robertson do you see issues if you replace the built-in type with another precision of complex number? (I doubt things would work with floats, a lot of things under the hood leverage complex numbers)

Aaron-Robertson:
@co9olguy gotcha, that makes sense. I was confused by the `float64` use case from the issue/related comment, but see now it's in reference to the Tensor rather than `ops.def_type`. And yep, `complex128` does complete in tests.

josh146:
> TF's DType only defines complex128 and complex64 though, so maybe just allow user to specify a bool mem_intense or something similar?

Nice idea, but I wonder if there is a need for that level of control? From feedback I've heard in the past, I would say it's sufficient to stick with `complex64` as the default dtype, but allows users to change it to `complex128` if they desire

Aaron-Robertson:
> Nice idea, but I wonder if there is a need for that level of control?

I just meant True => `complex128` and False => `complex64` as default.

> I would say it's sufficient to stick with complex64 as the default dtype, but allows users to change it to complex128 if they desire

Can do, I just thought it would be confusing if there's a dtype arg but the only accepted values are `complex64` and `complex128`. Maybe call it complex_dtype?

josh146:
Oh I see! Thanks for the clarification @Aaron-Robertson. I'm happy with either solution, but tend to lean more towards the `complex_dtype` solution, since it makes it more explicit which dtype is being used by the backend

co9olguy:
Since having `mem_intense` as an argument would already require a docstring describing what exactly that means, I'd be in favour of instead just having `dtype` as the name, saying the acceptable values in the docstring (i.e., the two complex variants), and throwing an direct error message if any other dtype is received

Aaron-Robertson:
Alright, made enough bandwidth for this one in the next few days (and probably want to clarify before I get too much deeper). Do we want to follow existing patterns for backend args? Essentially add to each `tfbackend/ops.py` function as optional and pass down from circuit/states?

I can push what I have, but that's about where I landed after testing a couple other approaches. Happy to switch it up too if there's a better recommendation!

.github/CHANGELOG.md
```diff
@@ -68,8 +68,29 @@
...
 * The `fock` backend now supports the `GKP` preparation [(#553)](https://github.com/XanaduAI/strawberryfields/pull/553)
 
+* The `tf` backend now accepts the Tensor DType as argument.
+  [(#562)](https://github.com/XanaduAI/strawberryfields/pull/562)
+
+  Allows high cutoff dimension to give numerically correct calculations:
+
+  ```python
+  prog = sf.Program(2)
+  eng  = sf.Engine("tf", backend_options={"cutoff_dim": 50, "dtype": tf.complex128})
```
josh146, line 81: This looks great @Aaron-Robertson

.github/CHANGELOG.md
```diff
@@ -68,8 +68,29 @@
...
+      Sgate(0.8)         | q[0]
+      Sgate(0.8)         | q[1]
+      BSgate(0.5,0.5)  | (q[0], q[1])
+      BSgate(0.5,0.5)  | (q[0], q[1])
+  state    = eng.run(prog).state
+  N0,N0var = state.mean_photon(0)
+  N1,N1var = state.mean_photon(1)
+  print(N0)
+  print(N1)
+  print('analytical:',np.sinh(0.8)**2)
```
josh146, line 92:
```suggestion
  state = eng.run(prog).state
  N0, N0var = state.mean_photon(0)
  N1 ,N1var = state.mean_photon(1)
  print(N0)
  print(N1)
  print("analytical:", np.sinh(0.8)**2)
```

Small modification here, as we try to stick to PEP8 formatting for our examples :)

strawberryfields/backends/tfbackend/backend.py
```diff
@@ -82,10 +82,13 @@ def begin_circuit(self, num_subsystems, **kwargs):
                 For each mode, the simulator can represent the Fock states :math:`\ket{0}, \ket{1}, \ldots, \ket{\text{cutoff_dim}-1}`.
             pure (bool): If True (default), use a pure state representation (otherwise will use a mixed state representation).
             batch_size (None or int): Size of the batch-axis dimension. If None, no batch-axis will be used.
+            dtype (DType): Complex Tensorflow Tensor type representation, either complex64 (default) or complex128.
+                Note, complex128 will increase memory usage substantially.
```
josh146, line 86: @Aaron-Robertson is this something you noticed during testing?

strawberryfields/backends/tfbackend/backend.py
```diff
@@ -96,14 +99,16 @@ def begin_circuit(self, num_subsystems, **kwargs):
             raise ValueError("Argument 'cutoff_dim' must be a positive integer")
         if not isinstance(pure, bool):
             raise ValueError("Argument 'pure' must be either True or False")
+        if not dtype in (tf.complex64, tf.complex128):
+            raise ValueError("Argument 'dtype' must be a complex Tensorflow DType")
```
josh146, line 103: Nice!

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -94,7 +98,7 @@ def _valid_modes(self, modes):
         for mode in modes:
             if mode < 0:
                 raise ValueError("Specified mode number(s) cannot be negative.")
-            elif mode >= self._num_modes:
+            if mode >= self._num_modes:
```
josh146, line 101: good catch!

strawberryfields/backends/tfbackend/backend.py
```diff
@@ -82,10 +82,13 @@ def begin_circuit(self, num_subsystems, **kwargs):
                 For each mode, the simulator can represent the Fock states :math:`\ket{0}, \ket{1}, \ldots, \ket{\text{cutoff_dim}-1}`.
             pure (bool): If True (default), use a pure state representation (otherwise will use a mixed state representation).
             batch_size (None or int): Size of the batch-axis dimension. If None, no batch-axis will be used.
+            dtype (DType): Complex Tensorflow Tensor type representation, either complex64 (default) or complex128.
```
josh146, line 85: Small suggestion to make it clear that you need to pass TensorFlow dtypes only:

```suggestion
            dtype (tf.dtype): Complex Tensorflow Tensor type representation, either ``tf.complex64`` (default) or ``tf.complex128``.
```

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -222,6 +225,7 @@ def reset(self, **kwargs):
             pure (bool): if True, the reset circuit will represent its state as a pure state. If False, the representation will be mixed.
             cutoff_dim (int): new Fock space cutoff dimension to use.
             batch_size (None, int): None means no batching. An integer value >= 2 sets the batch size to use.
+            dtype (DType): Complex Tensorflow Tensor type representation, either complex64 (default) or complex128.
```
josh146, line 228:
```suggestion
            dtype (tf.dtype): Complex Tensorflow Tensor type representation, either ``tf.complex64`` (default) or ``tf.complex128``.
```

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -249,6 +253,13 @@ def reset(self, **kwargs):
             self._batch_size = batch_size
             self._batched = batch_size is not None
 
+        if "dtype" in kwargs:
+            dtype = kwargs["dtype"]
+            if dtype is not None:
+                if not dtype in (tf.complex64, tf.complex128):
+                    raise ValueError("Argument 'dtype' must be a complex Tensorflow DType")
```
josh146, line 260: Optional suggestion: you could make this code block slightly more concise by doing something like:

```python
self._dtype = kwargs.get("dtype", tf.complex64)

if self._dtype is not in (tf.complex64, tf.complex128):
    raise ValueError("Argument 'dtype' must be a complex Tensorflow DType")
```

strawberryfields/backends/tfbackend/circuit.py
```diff
@@ -966,6 +1027,11 @@ def batch_size(self):
         """Returns the batch size"""
         return self._batch_size
 
+    @property
+    def dtype(self):
+        """Returns the circuit dtype"""
+        return self._dtype
```
josh146, line 1033: 👍

strawberryfields/backends/tfbackend/ops.py
```diff
@@ -302,19 +304,22 @@ def grad(dy):  # pragma: no cover
...
     if batched:
         return tf.stack(
-            [single_displacement_matrix(r_, phi_, cutoff) for r_, phi_ in tf.transpose([r, phi])]
+            [
+                single_displacement_matrix(r_, phi_, cutoff, dtype=dtype.as_numpy_dtype)
+                for r_, phi_ in tf.transpose([r, phi])
+            ]
         )
-    return single_displacement_matrix(r, phi, cutoff)
+    return single_displacement_matrix(r, phi, cutoff, dtype=dtype.as_numpy_dtype)
```
josh146, line 318: Is this because the `single_displacement_matrix` uses NumPy, and not tensorflow?

strawberryfields/backends/tfbackend/states.py
```diff
@@ -43,19 +43,34 @@ class FockStateTF(BaseFockState):
             num_modes (int): the number of modes in the state
             pure (bool): True if the state is a pure state, false if the state is mixed
             cutoff_dim (int): the Fock basis truncation size
+            batched (bool): (optional) default False means no batching
             mode_names (Sequence): (optional) this argument contains a list providing mode names
-                    for each mode in the state.
+                    for each mode in the state
+            dtype (DType): (optional) complex Tensorflow Tensor type representation, either
+                    complex64 (default) or complex128
```
josh146, line 50:
```suggestion
            dtype (tf.dtype): (optional) complex Tensorflow Tensor type representation, either
                    ``tf.complex64`` (default) or ``tf.complex128``
```

tests/frontend/test_engine.py
```diff
@@ -43,12 +44,14 @@ def prog():
...
     sf.Engine("fock", backend_options={"batch_size": 2, "cutoff_dim": 6}),
     sf.Engine("tf", backend_options={"batch_size": 2, "cutoff_dim": 6}),
+    sf.Engine("tf", backend_options={"batch_size": 2, "cutoff_dim": 6, "dtype": tf.complex64}),
 ]
 
 engines = [
     sf.Engine("gaussian", backend_options={"cutoff_dim": 6}),
     sf.Engine("fock", backend_options={"cutoff_dim": 6}),
     sf.Engine("tf", backend_options={"cutoff_dim": 6}),
+    sf.Engine("tf", backend_options={"cutoff_dim": 6, "dtype": tf.complex128}),
```
josh146, line 54: nice, this is the correct place to add this parametrization as far as I can tell.

josh146 (approved):
Thanks @Aaron-Robertson! This PR is great work. It looks good to me (bar a few minor suggestions and comments), however if possible I might get @ziofil (architect of the TF backend) to have a quick lookover as well

strawberryfields/backends/tfbackend/backend.py
Aaron-Robertson: I noticed a small increase, but mostly included because it's noted in the issue(s).

strawberryfields/backends/tfbackend/ops.py
Aaron-Robertson: They use Tensorflow to calculate the gradients, but TheWalrus to calculate matrix representations. So my interpretation is that TheWalrus uses NumPy.

ziofil: For some operations we use TW also for the gradients. And the implementation in TW itself uses numba.

Aaron-Robertson: Yeah I think that's correct as far as TW. But numba uses numpy dtypes? That's why they get converted, as far as I know.

ziofil: yes, [we use numpy dtypes](https://github.com/XanaduAI/thewalrus/blob/f779cb91d57bd700aeb903fcb1c3132082a3f156/thewalrus/fock_gradients.py#L41). Then numba compiles the code to C upon first call.

strawberryfields/backends/tfbackend/circuit.py
Aaron-Robertson: I like it, good to do similar for the other circuit args?

ziofil (approved):
This looks like a great contribution. I never liked that `def_type` hacky solution.

Aaron-Robertson:
Tried a couple cleanups and oddly ran into pipeline test failures that didn't occur locally. Let's merge this one and I can make a cleanup PR for some of the todos in the tf backend, and to similarly replace `def_type` with `dtype` param for the fock backend 😄

PR #561, opened 2021-03-19
Docs cleanup for contribution

Aaron-Robertson:
**Context:**

Cleanup documentation for ease of contribution based on this [slack conversation](https://xanadu-quantum.slack.com/archives/CA2U56VDE/p1616180499012000).

**Description of the Change:**

- [x] Fix outdated contributing link
- [x] Development guide test section includes pytest-randomly and pytest-mock
- [x] Contribution challenges, PR template, and issue template links are not dead

**Benefits:**

Makes it easier for new contributors.

**Possible Drawbacks:**

N/A

**Related GitHub Issues:**

N/A

PR #563, opened 2021-03-22
Development requirements and formatting

Aaron-Robertson:
**Context:**

Discussed on #561 

Adds a development environment section to development docs, and an associated `dev_requirements.txt` for installation. Includes a `black` format script run by `make format` with associated instructions in development docs and PR template.

**Benefits:**

Makes contribution easier.

**Possible Drawbacks:**

N/A

**Related GitHub Issues:**

N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,8 +13,8 @@ Please complete the following checklist when submitting a PR:
 - [ ] Ensure that the test suite passes, by running `make test`.
 
 - [ ] Ensure that code is properly formatted, by running `make format` or `black -l 100
-      strawberryfields`.
-      You will need to have the Black code format installed: `pip install black`
+      strawberryfields`. You will need to have the Black code format installed: `pip install
+      black`.
 
 - [ ] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
       change, and including a link back to the PR.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,6 +12,8 @@ Please complete the following checklist when submitting a PR:
 
 - [ ] Ensure that the test suite passes, by running `make test`.
 
+- [ ] Ensure that code is properly formatted, by running `make format`.
+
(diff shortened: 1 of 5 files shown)


co9olguy:
Thanks @Aaron-Robertson :smiley:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,6 +12,8 @@ Please complete the following checklist when submitting a PR:
 
 - [ ] Ensure that the test suite passes, by running `make test`.
 
+- [ ] Ensure that code is properly formatted, by running `make format`.
+
```
josh146, line 16: Good catching adding this in! 

I wonder if it is useful to also say:
```suggestion
- [ ] Ensure that code is properly formatted, by running `make format` or `black -l 100 strawberryfields`.
      You will need to have the Black code format installed: pip install black
```

This way, Windows users can run black manually (as they might not have `make` installed), and users that don't have black installed will know what to do.

Makefile
```diff
@@ -54,6 +55,10 @@ clean-docs:
 	make -C doc clean
 	rm -rf doc/code/api
 
+.PHONY : format
+format: 
+	black -l 100 --check --diff strawberryfields
```
josh146, line 60: What does the --check and --diff do here?

josh146 (approved):
Nice, this will be super useful @Aaron-Robertson!

Makefile
thisac: I believe `check` would simply output which files would be updated, without actually writing to the files (`diff` might be similar, just printing the diff as well; not sure if it writes or not though).

Should probably not need either here if we actually wish to have the command update the files.

Aaron-Robertson: Yep, they show files to change and a diff rather than autoformatting. Happy to remove them.

Makefile
```diff
@@ -54,6 +55,10 @@ clean-docs:
 	make -C doc clean
 	rm -rf doc/code/api
 
+.PHONY : format
+format: 
+	black -l 100 strawberryfields
```
thisac, line 60: 👍

thisac (approved):
Thanks @Aaron-Robertson! This is a really nice addition. 🎉

PR #564, opened 2021-03-31
Avoid `VisibleDeprecationWarning` by specifying `dtype` for `ndarray`

antalszava:
Newer numpy version (e.g., `numpy==1.20.2`) raise a warning for constructing ragged arrays:
```
strawberryfields/backends/states.py:1633: VisibleDeprecationWarning:
Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
```
This PR applies the suggestion from the warning message at two places.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -83,6 +83,8 @@
 
 * Cleanup docs to make contribution easier.
   [(#561)](https://github.com/XanaduAI/strawberryfields/pull/561)
+* Add development requirements and format script to make contribution easier.
+  [(#563)](https://github.com/XanaduAI/strawberryfields/pull/563)
 
 <h3>Contributors</h3>
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -92,6 +92,12 @@
   print("analytical:" ,np.sinh(0.8)**2)
   ```
 
+<h3>Improvements</h3>
+
+* No `VisibleDeprecationWarning` is raised when using the state `wigner`
+  method.
+  [(#564)](https://github.com/XanaduAI/strawberryfields/pull/564)
+
 <h3>Breaking Changes</h3>
 
 <h3>Bug fixes</h3>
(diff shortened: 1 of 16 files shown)


josh146 (commented):
Thanks @antalszava for finding and catching this!

I have a question though - is the created array _always_ ragged? I think it is only ragged if the user-provided `xvec` and `pvec` are different lengths.

In which case, we should branch depending on this condition. That way, a standard NumPy array is still created if X and P are the same length.

antalszava:
@josh146 had a closer look at this, a ragged array is created even if the user provided vectors have the same length:
```python
import numpy as np
import strawberryfields as sf

nmodes = 2
epsilon = 0.1
sf.hbar = 1
prog_linear_cluster = sf.Program(nmodes)

with prog_linear_cluster.context as q:
    sf.ops.GKP(state=[0.5*np.pi, 0], epsilon=epsilon) | q[0]
    sf.ops.Squeezed(r = -5) | q[1]

eng = sf.Engine("bosonic")
result = eng.run(prog_linear_cluster)
vec = np.linspace(-5, 5, num=200)
sf.plot.plot_wigner(result.state, 0, vec, vec)
```
Here we use `vec` twice, and the warning still arises. The reason is, that under the hood we use `X, P = np.meshgrid(xvec, pvec, sparse=True)`, where having `sparse=True` will change the shape of `X` and `P`.

```python
vec = np.linspace(-5, 5, num=200)
mesh = np.meshgrid(vec, vec, sparse=True)
print(mesh[0].shape, mesh[1].shape)

mesh = np.meshgrid(vec, vec)
print(mesh[0].shape, mesh[1].shape)
```
```
(1, 200) (200, 1)
(200, 200) (200, 200)
```
Added a condition nonetheless. :slightly_smiling_face:

josh146 (approved):
Ah, thanks for solving this @antalszava! I didn't realize about the `sparse` argument 🙂

josh146:
P.S., don't forget to update the changelog!

PR #566, opened 2021-04-08
Fix TDM "modes" validation to match XQC API

lneuhaus:
**Context:**
Device specifications for TDM programs that are exposed by Xanadu's device API used to include a field `modes` with data like
```
{
    "concurrent": 2,
    "spatial": 1,
    "max": {"temporal": 100},
}
```
This format was recently changed into (ex generis):
```
{
    "concurrent": 2,
    "spatial": 1,
    "temporal_max": 100
}
```

This PR brings Strawberry Fields up to date with the new format.

**Benefits:**
TDM programs can run on Xanadu hardware.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -101,6 +101,7 @@
   [(#540)](https://github.com/XanaduAI/strawberryfields/pull/540)
 * TDM programs now expect a flat (not nested) dictionary of `modes` in device 
   specifications obtained from the XQC platform API.
+  [(#566)](https://github.com/XanaduAI/strawberryfields/pull/566)
 
 <h3>Documentation</h3>
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -99,6 +99,8 @@
 * `Connection` objects now send requests to the platform API at version `0.2.0`
   instead of the incorrect version number `1.0.0`.
   [(#540)](https://github.com/XanaduAI/strawberryfields/pull/540)
+* TDM programs now expect a flat (not nested) dictionary of `modes` in device 
+  specifications obtained from the XQC platform API.
 
 <h3>Documentation</h3>
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -99,6 +99,7 @@
 * `Connection` objects now send requests to the platform API at version `0.2.0`
   instead of the incorrect version number `1.0.0`.
   [(#540)](https://github.com/XanaduAI/strawberryfields/pull/540)
+
 * TDM programs now expect a flat (not nested) dictionary of `modes` in device 
   specifications obtained from the XQC platform API.
   [(#566)](https://github.com/XanaduAI/strawberryfields/pull/566)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -115,9 +115,9 @@
 
 This release contains contributions from (in alphabetical order):
 
-J. Eli Bourassa, Guillaume Dauphinais, Ish Dhand, Theodor Isacsson, Josh Izaac, 
-Nicolás Quesada, Aaron Robertson, Krishna Kumar Sabapathy, Jeremy Swinarton, Antal Száva,
-Ilan Tzitrin.
+J. Eli Bourassa, Guillaume Dauphinais, Ish Dhand, Theodor Isacsson, Josh Izaac,
+Leonhard Neuhaus, Nicolás Quesada, Aaron Robertson, Krishna Kumar Sabapathy, 
+Jeremy Swinarton, Antal Száva, Ilan Tzitrin.
 
 # Release 0.17.0 (current release)
 
--- a/strawberryfields/tdm/tdmprogram.py
+++ b/strawberryfields/tdm/tdmprogram.py
@@ -643,7 +643,6 @@ class TDMProgram(sf.Program):
         self.append(cmd.op.__class__(*params), get_modes(cmd, q))
 
     def assert_number_of_modes(self, device):
-        return
         if self.timebins > device.modes["temporal_max"]:
             raise CircuitError(
                 f"This program contains {self.timebins} temporal modes, but the device '{device.target}' "
--- a/strawberryfields/tdm/tdmprogram.py
+++ b/strawberryfields/tdm/tdmprogram.py
@@ -644,7 +644,7 @@ class TDMProgram(sf.Program):
 
     def assert_number_of_modes(self, device):
         return
-        if self.timebins > device.modes["temporal"]["max"]:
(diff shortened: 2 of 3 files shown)


lneuhaus:
Is the weird behavior of codecov expected? I just changed a few lines, no way to have decreased coverage by 40% ;)

.github/CHANGELOG.md
```diff
@@ -99,6 +99,8 @@
 * `Connection` objects now send requests to the platform API at version `0.2.0`
   instead of the incorrect version number `1.0.0`.
   [(#540)](https://github.com/XanaduAI/strawberryfields/pull/540)
+* TDM programs now expect a flat (not nested) dictionary of `modes` in device 
+  specifications obtained from the XQC platform API.
```
lneuhaus, line 103:
```suggestion
  specifications obtained from the XQC platform API.
  [(#566)](https://github.com/XanaduAI/strawberryfields/pull/566)
```

heltluke:
I guess `test_tdmprogram.py` needs to be updated too, but otherwise looks good Leo!

.github/CHANGELOG.md
```diff
@@ -99,6 +99,9 @@
...
+  [(#566)](https://github.com/XanaduAI/strawberryfields/pull/566)
```
antalszava, line 104: Don't forget to add yourself as a contributor :slightly_smiling_face:

antalszava:
Looks good overall! :slightly_smiling_face: Thank you! Indeed, as Luke mentioned, probably [this line](https://github.com/XanaduAI/strawberryfields/blob/a0f8b5ba03d56b43cf3cc744e54886cba6fc9eca/tests/frontend/test_tdmprogram.py#L528) needs an update.

.github/CHANGELOG.md
```diff
@@ -99,6 +99,9 @@
 * `Connection` objects now send requests to the platform API at version `0.2.0`
   instead of the incorrect version number `1.0.0`.
   [(#540)](https://github.com/XanaduAI/strawberryfields/pull/540)
+* TDM programs now expect a flat (not nested) dictionary of `modes` in device
```
josh146, line 102:
```suggestion

* TDM programs now expect a flat (not nested) dictionary of `modes` in device 
```

josh146 (commented):
Hey @lneuhaus, since SF now uses API versioning, my understanding is this _shouldn't_ be changed unless the pinned API version is also increased as needed.

But maybe someone from platform can clarify this (@jswinarton?)

.github/CHANGELOG.md
lneuhaus: done - so far the least amount of characters that got me into the contributors list :)

lneuhaus:
> since SF now uses API versioning, my understanding is this shouldn't be changed unless the pinned API version is also increased as needed.
> Unless somehow the current API version that SF targets was changed?
> But maybe someone from platform can clarify this (@jswinarton?)

Definitely curious to get @jswinarton, @gtr8, or @Mandrenkov 's input on this. What I know for sure:
- all API versions >= 0.3.0 serve un-nested dicts, so should be in agreement with this patch.
- API version 0.1.0 does not have a device spec endpoint, so is irrelevant.
- API version 0.2.0 specifies a nested dictionary with structure `{"max": {"key": value}}` (not entirely clear to me what the specification allows here apart from "max" in the top-level dict).

fab1an-q (approved):

Mandrenkov:
> * all API versions >= 0.3.0 serve un-nested dicts, so should be in agreement with this patch.

True, but I think this statement is weaker than you think.  Here is how the 0.3.0 endpoint is currently implemented:
* If `modes` is not a dictionary (e.g., an integer), return it without modification.
* If `modes` is a dictionary and the target starts with `simulon`, flatten the modes into a `Dict[str, Any]` and return it.
* If `modes` is a dictionary and the target does *not* start with `simulon`, return an empty dictionary

If you try and request the device specifications for a TD2 device using the 0.3.0 endpoint, you will get an empty dictionary.

> * API version 0.2.0 (the one SF currently requests (see [here](https://github.com/XanaduAI/strawberryfields/blob/a0f8b5ba03d56b43cf3cc744e54886cba6fc9eca/strawberryfields/api/connection.py#L111)) specifies a nested dictionary with structure `{"max": {"key": value}}`. It's not entirely clear to me what keys the specification allows here apart from "max" in the top-level dict, but the only allowed value types here should be dictionaries (i.e. also not integers). From that perspective the current SF implementation is not aligned with the spec for API version 0.2.0 already (since it expects `{"concurrent": 2}` for example), so I believe we are in a gray zone here anyways.

The platform 0.2.0 API specification does indeed state that `modes` is either an integer or a dictionary of the form `{"max": {"key": value}}`; however, the 0.2.0 implementation does *not* perform any validation in this regard.  So, if you request the TD2 device specifications using the 0.2.0 endpoint, you will receive
```json
"modes": {
  "equal": {
    "concurrent": 2,
    "spatial": 1
  },
  "max": {
    "temporal": 100
  }
}
```

> Either we admit that the API 0.2.0 spec is wrong and expand it
> 
> * from `modes: Union[int, Dict[str, Dict[str, int]]]`
> * onto `modes: Union[int, Dict[str, Union[int, Dict[str, int]]]]`
> 
> to be consistent with what's being returned by the API anyways, or we take the API 0.2.0 as the ultimate source of truth, in which case I could modify this PR to take that into account.

I think performing either change is not great for semantic versioning so we should do the one that leaves the API in the most desirable state.  That is, we either
1. Fix the 0.2.0 API implementation to transform the TD2 device spec into a `Dict[str, int]`, or
2. Update the 0.2.0 API specification to allow for a more flexible device spec.

I'm not sure which of these options is more convenient for Strawberry Fields.

lneuhaus:
@Mandrenkov and I have run a test with this branch - it is now in line with how the platform presents the `modes` data, from API version 0.2.0 onwards. So 

> Fix the 0.2.0 API implementation to transform the TD2 device spec into a Dict[str, int]

is what both the platform and this PR went with, so I suggest to simply merge this PR and move on.

Mandrenkov (approved):
Looks good to me.

As @lneuhaus mentioned, the discrepancy can be fixed by updating the TD2 hardware certificate version; however, it will be a week or so before we deploy support for the new certificate version on the platform side.  I think it would sense to hold off merging this PR until then (to minimize the TD2 downtime).

lneuhaus:
> to minimize the TD2 downtime

TD2 is effectively down until this fix is applied since a platform upgrade around January (introduction of device API) - therefore I suggest we merge this now in order to be sure to integrate this behavior into the next Strawberry Fields release.

Mandrenkov:
Okay, sounds good!

PR #576, opened 2021-04-27
Change simulon target to simulon_gaussian

jswinarton:
**Context:**
The Xanadu Quantum Cloud simulator, formerly called `simulon`, now has the target name `simulon_gaussian`. This change was made to support many different simulon types in the future.

- References to `simulon` in documentation and tests have been changed to `simulon_gaussian`.
- A formatting change is made in `strawberryfields/backends/gaussianbackend/gaussiancircuit.py`. It appears that the black version on CI has been upgraded since the last commit, which now checks new elements in docstrings. This change has been made here instead of in a separate PR.

**Benefits:**
People reading the docs will not try to submit jobs to a simulator that doesn't exist :slightly_smiling_face: 

**Possible Drawbacks:**
n/a

**Related GitHub Issues:**
n/a

### DIFF
--- a/doc/introduction/photonic_hardware.rst
+++ b/doc/introduction/photonic_hardware.rst
@@ -136,13 +136,14 @@ Cloud simulator
 In addition to submitting jobs to be run on quantum hardware, it is also possible to run jobs on
 simulators via the Xanadu Quantum Cloud. The process is very similar to running jobs on hardware. You
 will need to configure your account, as described above, and submit a job via the ``RemoteEngine``,
-using a simulator as the target instead of a specific chip:
+using a simulator (a "simulon") as the target instead of a specific chip:
 
 >>> eng = sf.RemoteEngine("simulon_gaussian")
 >>> result = eng.run(prog)
 
-Simulon jobs can also be submitted asynchronously using ``eng.run_async``, or by submitting a
-Blackbird script with the ``target`` set to a simulator target in the Blackbird header.
+Simulator jobs can also be submitted asynchronously using ``eng.run_async``, or
+by submitting a Blackbird script with the ``target`` set to a simulator target
+in the Blackbird header.
 
 See the `Submitting jobs on hardware`_ section above for more details.
 
--- a/doc/introduction/photonic_hardware.rst
+++ b/doc/introduction/photonic_hardware.rst
@@ -134,25 +134,27 @@ Cloud simulator
 ---------------
 
 In addition to submitting jobs to be run on quantum hardware, it is also possible to run jobs on
-the Xanadu Quantum Cloud simulator Simulon. The process is very similar to running jobs on hardware. You
+simulators via the Xanadu Quantum Cloud. The process is very similar to running jobs on hardware. You
 will need to configure your account, as described above, and submit a job via the ``RemoteEngine``,
-using ``"simulon"`` as the target instead of a specific chip:
+using a simulator as the target instead of a specific chip:
 
->>> eng = sf.RemoteEngine("simulon")
+>>> eng = sf.RemoteEngine("simulon_gaussian")
 >>> result = eng.run(prog)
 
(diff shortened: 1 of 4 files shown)


doc/introduction/photonic_hardware.rst
```diff
@@ -134,25 +134,27 @@ Cloud simulator
...
+simulators via the Xanadu Quantum Cloud. The process is very similar to running jobs on hardware. You
 will need to configure your account, as described above, and submit a job via the ``RemoteEngine``,
-using ``"simulon"`` as the target instead of a specific chip:
+using a simulator as the target instead of a specific chip:
 
->>> eng = sf.RemoteEngine("simulon")
+>>> eng = sf.RemoteEngine("simulon_gaussian")
 >>> result = eng.run(prog)
 
 Simulon jobs can also be submitted asynchronously using ``eng.run_async``, or by submitting a
```
antalszava, line 144:
```suggestion
Jobs to the ``simulon_gaussian`` simulator can also be submitted asynchronously using ``eng.run_async``, or by submitting a
```

antalszava (approved):
Looks good to me! :100: Thank you for this :blush:

doc/introduction/photonic_hardware.rst
jswinarton: I actually deliberately changed this because in the near future other simulons will be available, like `simulon_jet`, and I figured it would be easier to imply there are multiple simulons now rather than change it again later when this becomes true.

antalszava: Got it! Maybe it's worth mentioning then that `simulon` is a general term for remote simulators.

jswinarton: Good point, I made this change.

doc/introduction/photonic_hardware.rst
```diff
@@ -134,25 +134,28 @@ Cloud simulator
 ---------------
 
 In addition to submitting jobs to be run on quantum hardware, it is also possible to run jobs on
-the Xanadu Quantum Cloud simulator Simulon. The process is very similar to running jobs on hardware. You
+simulators via the Xanadu Quantum Cloud. The process is very similar to running jobs on hardware. You
 will need to configure your account, as described above, and submit a job via the ``RemoteEngine``,
-using ``"simulon"`` as the target instead of a specific chip:
+using a simulator (a "simulon") as the target instead of a specific chip:
```
thisac, line 139: `a "simulon"` sounds a bit strange, although I don't necessarily dislike it. 

Maybe it could be expanded into a separate sentence saying something along the line of "We refer to our remote simulators as 'simulon'".

thisac (approved):
Thanks @jswinarton!

doc/introduction/photonic_hardware.rst
jswinarton: Ok, I added wording similar to this but kept it bracketed to avoid breaking the flow. Let me know what you think.

thisac: Thanks @jswinarton. Looks better, I think! 🙂

strawberryfields/backends/gaussianbackend/gaussiancircuit.py
```diff
@@ -77,7 +77,7 @@ def add_mode(self, n=1):
         self.nlen = newnlen
 
     def del_mode(self, modes):
-        """ delete mode from the circuit"""
+        """delete mode from the circuit"""
```
josh146, line 80: Black now formats docstrings??? 🙏 
This makes me so happy. It always bothered me that it ignored them

josh146 (approved):
Thanks @jswinarton! Note: we will need to make a `0.18.0-post1` release of SF for this to display in the documentation, since the docs are pinned to 'stable'. This will not require an upload to PyPI (since the source code has not changed), just a GitHub release.

PR #579, opened 2021-05-12
Reduce the use of SymPy lambdify

antalszava:
**Context:**
Some optimization tasks using Strawberry Fields were showing increased memory allocation, although no state was being left (i.e., the engine was reset, state objects recreated, etc.).

Parameters in Strawberry Fields might be numeric or symbolic. To represent both, SymPy is used. When using gate parameters that are internally created, Strawberry Fields evaluates the parameter to get the numeric value using SymPy.

Evaluating the parameters depends on the `sympy.lambdify` function, which [internally caches data using `linecache`](https://github.com/sympy/sympy/blob/aa96b23bb98f00542101c08d8ea02ef83949310b/sympy/utilities/lambdify.py#L876). The memory increase was found to be due to this caching.

Changes the `par_evaluate` function which does the parameter evaluation such that:
* If the SymPy object can be converted to a numeric Python type (`float` or `complex`), then we perform the conversion without using `sympy.lambdify`
* Otherwise, when using `sympy.lambdify` we also clear the cache using `linecache`.

**Benefits:**
* Significantly faster parameter evaluation in cases when the use of `sympy.lambdify` can be avoided
* No memory increase over time

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

[See profiling data here.](https://github.com/XanaduAI/strawberryfields/pull/579#issuecomment-840109642)

### DIFF
--- a/strawberryfields/parameters.py
+++ b/strawberryfields/parameters.py
@@ -193,6 +193,8 @@ def par_evaluate(params, dtype=None):
         atoms = list(p.atoms(MeasuredParameter, FreeParameter))
 
         if not atoms:
+            # If there are not atomic values, we just convert to elementary
+            # Python types
             return float(p) if p.is_real else complex(p)
 
         # evaluate the atoms of the expression
@@ -203,6 +205,12 @@ def par_evaluate(params, dtype=None):
         printer = "tensorflow" if any(is_tf) else "numpy"
 
         func = sympy.lambdify(atoms, p, printer)
+
+        # sympy.lambdify caches data using linecache, if called many times this
+        # can make up for a lot of memory used. We clear the cache here to
+        # avoid that.
+        import linecache
+        linecache.clearcache()
 
         if dtype is not None:
             # cast the input values
@@ -216,9 +224,7 @@ def par_evaluate(params, dtype=None):
 
         return func(*vals)
 
-    import linecache
     ret = list(map(do_evaluate, params))
-    linecache.clearcache()
     if scalar:
         return ret[0]
     return ret
--- a/strawberryfields/parameters.py
+++ b/strawberryfields/parameters.py
@@ -96,6 +96,7 @@ What we cannot do at the moment:
 import collections.abc
 import functools
 import types
+import linecache
 
 import numpy as np
 import sympy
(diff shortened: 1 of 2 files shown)


antalszava:
**Profiling**

The following script was profiled for both memory and time:

```python
import strawberryfields as sf
from strawberryfields.ops import *
from strawberryfields.parameters import par_funcs as pf
import numpy as np

eng = sf.Engine("fock", backend_options={"cutoff_dim": 2})

def evx(param, n):
    prog = sf.Program(2)    
    with prog.context as q:
        ntrot = 10
        for i in range(n): 
            step = param/ntrot
            for itrot in range(ntrot):
                
                s = step
                r = pf.asinh(-s / 2)
                theta = 0.5 * pf.atan2(-1.0 / pf.cosh(r), -pf.tanh(r))

                for _ in range(20):
                    # Decomposition of CXgate
                    BSgate(theta, 0) |  (q[0], q[1]),
                    Sgate(r, 0) | q[0],
                    Sgate(-r, 0) | q[1],
                    BSgate(theta + np.pi / 2, 0) | (q[0], q[1]),
                
        state = eng.run(prog).state
        return state.number_expectation([0])
    
parm = 0.05
for i in range(10):

    if eng.run_progs:
        eng.reset()

    res = evx(parm, 40)
    print(i, flush=True)
```
#### `master` branch:
*memory*:
without_clearing_cache

*time*: 88.01 secs

#### calling `linecache.clearcache()`:
with_clearing_cache

*time*:  87.23 secs

#### `fewer_lambdify` branch (calling `linecache.clearcache()` & no call to `lambdify` if not symbolic):

*time*:  34.86 secs

nquesada (approved):

nquesada:
@antalszava : don't forget to update the changelog. Also you might want to make the CodeFactor bot happy.

strawberryfields/parameters.py
```diff
@@ -191,14 +192,26 @@ def do_evaluate(p):
...
         # (we do it like this to avoid importing tensorflow if it's not needed)
         is_tf = (type(v).__module__.startswith("tensorflow") for v in vals)
         printer = "tensorflow" if any(is_tf) else "numpy"
+
         func = sympy.lambdify(atoms, p, printer)
 
+        # sympy.lambdify caches data using linecache, if called many times this
+        # can make up for a lot of memory used. We clear the cache here to
+        # avoid that.
+        linecache.clearcache()
```
josh146, line 213: Very nice catch @antalszava!

josh146 (approved):
Great 🕵️ work @antalszava!

PR #581, opened 2021-05-14
numpy>=1.19.2 as requirement

### LINKED ISSUE #580: Dependency versions error
#### Issue description
I made a fork of this project and tried to setup a new virtual environment.

```
python -m venv sf-venv
source sf-venv/bin/active.fish
pip install -r requirements.txt
```

However, I got the following error
``` 
ERROR: Cannot install -r requirements.txt (line 4) and numpy>=1.20 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested numpy>=1.20
    tensorflow 2.5.0 depends on numpy~=1.19.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

#### Additional information

If it helps, I am using Python 3.9.4 and pip 21.1.1. 

A quick fix would be to downgrade the version of numpy in requirements.txt and solve the issue, but I am not sure it is the best way to go.

antalszava:
**Context**

The new release of TensorFlow version 2.5 pins `numpy ~= 1.19.2`. As Strawberry Fields pins `numpy>=1.20`, installing SF in a clean environment results in version conflicts between TensorFlow's and SF's `numpy` version requirement.

**Changes**

Pins `numpy>=1.19.2` in `requirements.txt`, `doc/requirements.txt` and `setup.py`.

**Related issues**

Closes #580.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ with open("strawberryfields/_version.py") as f:
 
 
 requirements = [
-    "numpy>=1.17.4",
+    "numpy>=1.19.2",
     "scipy>=1.0.0",
     "sympy>=1.5",
     "networkx>=2.0",
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -4,7 +4,7 @@ ipykernel
 sphinx==2.2.2
 m2r
 networkx>=2.0
-numpy>=1.20
+numpy>=1.19.2
 plotly
 quantum-blackbird
 scipy>=1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-numpy>=1.20
+numpy>=1.19.2
 scipy
 sympy>=1.5
 tensorflow>=2.0


setup.py
```diff
@@ -22,7 +22,7 @@
 
 
 requirements = [
-    "numpy>=1.17.4",
+    "numpy>=1.19.2",
```
josh146, line 25: @antalszava how come this was incremented? Does it need to be? Potentially, SF still runs with NumPy 1.17?

josh146 (approved):
Thanks @antalszava! Just a small question about the `setup.py` file.

setup.py
antalszava: Hmm, just thought of adjusting it with `requirements.txt`, to have it consistent. Would it be better to leave it as is? Wasn't sure if there was a reason for the difference between the two versions.

antalszava:
Hi @josh146, thanks for the comments! I'll merge this but can ease the `numpy` version if we'd like.

Not entirely sure why prior to this PR there's a difference of:
* `setup.py`: `numpy>=1.17.4` and
* `requirements.txt`: `numpy>=1.20`.

josh146:
> Not entirely sure why prior to this PR there's a difference of:

Yes, good point. Ideally, `requirements.txt` should be pinning exactly, and `requirements-ci.txt` should match as closely as possible our `setup.py`.

antalszava:
Ohh okay! So maybe reverting to `setup.py`: `numpy>=1.17.4` and noting to have a `requirements-ci.txt` could work?

Also, I was wondering, how come we don't have `tensorflow` as a requirement in `setup.py`? :thinking:

josh146:
> Also, I was wondering, how come we don't have tensorflow as a requirement in setup.py? 🤔

To avoid it being a hard requirement for users that want to install SF but not TF :)

antalszava:
Oh right! The thing that confused me and which caused the initial issue is that we have it in `requirements.txt`. If we have a `requirements-ci.txt`, would we like to remove TF from `requirements.txt`?

josh146:
> If we have a requirements-ci.txt, would we like to remove TF from requirements.txt?

Good question, I would say probably not. Let's have `requirements.txt` be the full environment required to use all SF features

Issue #593, opened 2021-06-11
Gaussian Compilers Primitive and Decomposition definitions need to be fixed


federico0112:
#### Issue description

Gaussian compilers currently define measurement operations as primitives, which should not be the case. Furthermore, Gaussian state decompositions are defined in the Gaussian compilers decomposition, which is not ideal.

nquesada:
It seemed like good idea to make explicit that a compiler only operates on operation and not measurements and preparations, but this is too overzealous. For the moment we will allow compiler to still have these operations.

PR #594, opened 2021-06-11
Remove measurements and preparations from Gaussian and GaussianMerge

nquesada:
The compilers `Gaussian` and `GaussianMerge` should not accept as primitives measurement and at least some class of preparations. This PR removes this objects from the list things these compilers accept.

Addresses https://github.com/XanaduAI/strawberryfields/issues/593


nquesada:
OK. seems like removing measurement is after all not a good idea. For the moment it is fine to leave them there, but eventually it might be useful to think more carefully about this.

PR #596, opened 2021-06-16
Keep symbolic expressions for Blackbird programs

### LINKED ISSUE #595: Invalid Blackbird circuit is generated from a Program
#### Issue description

Running a `Program` loaded from a Blackbird circuit which was generated by another `Program` does not always work.  For example, consider the "Teleportatation" example from Strawberry Fields Interactive:
```python
import math
import strawberryfields as sf
from strawberryfields.ops import *
​
sf_prog = sf.Program(3)
​
# See the "Teleportation" algorithm in https://strawberryfields.ai/interactive/.
with sf_prog.context as q:
    Coherent(math.sqrt(5/4), math.atan(1/2)) | q[0]
    Sgate(2, 0) | q[2]
    Sgate(-2, 0) | q[1]
    BSgate(0.7854, 0) | (q[1], q[2])
    BSgate(0.7854, 0) | (q[0], q[1])
    MeasureP | q[1]
    MeasureX | q[0]
    Xgate(q[0].par*1.41421356237000) | q[2]
    Zgate(q[1].par*1.41421356237000) | q[2]

# Convert the SF program ​into a Blackbird program.
bb_prog = sf.io.to_blackbird(sf_prog)
bb_prog._name = "Teleportation"
bb_prog._version = 1.0
bb_prog._target["name"] = "gaussian"

# Recover the SF program from the Blackbird program.
bb_circuit = bb_prog.serialize()​
re_prog  = sf.io.loads(bb_circuit)

# Run the recovered SF program with the Gaussian engine.
eng = sf.LocalEngine("gaussian")
results = eng.run(re_prog)
```

* *Expected behavior:* A `Results` object is returned by the final line of the script.

* *Actual behavior:* A `TypeError` exception is raised.

* *Reproduces how often:* Every time.

* *System information:*
```

Python version:            3.9.5
Platform info:             Linux-5.4.72-microsoft-standard-WSL2-x86_64-with-glibc2.31
Installation path:         /home/mikhail/.local/lib/python3.9/site-packages/strawberryfields
Strawberry Fields version: 0.18.0
Numpy version:             1.20.3
Scipy version:             1.6.3
SymPy version:             1.8
NetworkX version:          2.5.1
The Walrus version:        0.15.1
Blackbird version:         0.3.0
TensorFlow version:        None
```

Running the script above yields the following error:

```python
TypeError                                 Traceback (most recent call last)
<ipython-input-16-da60180d5e6d> in <module>
     28 # Run the recovered SF program using the Gaussian backend.
     29 eng = sf.LocalEngine("gaussian")
---> 30 results = eng.run(re_prog)

~/.local/lib/python3.9/site-packages/strawberryfields/engine.py in run(self, program, args, compile_options, **kwargs)
    492                     )
    493 
--> 494         result = super()._run(
    495             program, args=args, compile_options=compile_options, **eng_run_options
    496         )

~/.local/lib/python3.9/site-packages/strawberryfields/engine.py in _run(self, program, args, compile_options, **kwargs)
    272             target = self.backend.compiler
    273             if target is not None:
--> 274                 p = p.compile(compiler=target, **compile_options)
    275             p.lock()
    276 

~/.local/lib/python3.9/site-packages/strawberryfields/program.py in compile(self, device, compiler, **kwargs)
    610             target = compiler.short_name
    611 
--> 612         seq = compiler.decompose(self.circuit)
    613 
    614         if kwargs.get("warn_connected", True):

~/.local/lib/python3.9/site-packages/strawberryfields/compilers/compiler.py in decompose(self, seq)
    222                 try:
    223                     kwargs = self.decompositions[op_name]
--> 224                     temp = cmd.op.decompose(cmd.reg, **kwargs)
    225                     # now compile the decomposition
    226                     temp = self.decompose(temp)

~/.local/lib/python3.9/site-packages/strawberryfields/ops.py in decompose(self, reg, **kwargs)
    450         Like :func:`Operation.decompose`, but applies self.dagger.
    451         """
--> 452         seq = self._decompose(reg, **kwargs)
    453         if self.dagger:
    454             # apply daggers, reverse the Command sequence

~/.local/lib/python3.9/site-packages/strawberryfields/ops.py in _decompose(self, reg, **kwargs)
   1500     def _decompose(self, reg, **kwargs):
   1501         # into a displacement
-> 1502         r = self.p[0] / np.sqrt(2 * sf.hbar)
   1503         return [Command(Dgate(r, 0), reg)]
   1504 

TypeError: ufunc 'true_divide' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

#### Additional information

The issue appears to have something to do with the final two lines of the generated Blackbird circuit:
```python
name Teleportation
version 1.0
target gaussian

Coherent(1.118033988749895, 0.4636476090008061) | 0
Sgate(2, 0) | 2
Sgate(-2, 0) | 1
BSgate(0.7854, 0) | [1, 2]
BSgate(0.7854, 0) | [0, 1]
MeasureHomodyne(phi=1.5707963267948966) | 1
MeasureHomodyne(phi=0) | 0
Xgate("1.41421356237*q0") | 2
Zgate("1.41421356237*q1") | 2
```

thisac:
**Context:**
There are a few issues when attempting to convert an SF circuit with feed-forwarding to Blackbird. If any parameter is found to be symbolic, it's cast as a string rather than keeping it as a `blackbird.RegRefTransform` as it would be if loaded directly from a script to Blackbird. For example, the following SF circuit

```python
with prog.context as q:
    Coherent(math.sqrt(5/4), math.atan(1/2)) | q[0]
    Sgate(2, 0) | q[2]
    Sgate(-2, 0) | q[1]
    BSgate(0.7854, 0) | (q[1], q[2])
    BSgate(0.7854, 0) | (q[0], q[1])
    MeasureP | q[1]
    MeasureX | q[0]
    Xgate(q[0].par*1.41421356237000) | q[2]
    Zgate(q[1].par*1.41421356237000) | q[2]
```
will, when transformed into a Blackbird program, converting `q[0].par*1.41421356237000` and `q[1].par*1.41421356237000` into the strings `"1.41421356237000*q0"` and `"1.41421356237000*q1"`, rather than, as it should, into `blackbird.RegRefTransform` objects.

If attempting to load the following blackbird script

```
name Teleportation
version 1.0
target gaussian

Coherent(1.118033988749895, 0.4636476090008061) | 0
Sgate(2, 0) | 2
Sgate(-2, 0) | 1
BSgate(0.7854, 0) | [1, 2]
BSgate(0.7854, 0) | [0, 1]
MeasureHomodyne(phi=1.5707963267948966) | 1
MeasureHomodyne(phi=0) | 0
Xgate(1.41421356237*q0) | 2
Zgate(1.41421356237*q1) | 2
```

both `1.41421356237*q0` and `1.41421356237*q0` will be stored as `blackbird.RegRefTransform` objects.

The `strawberryfields.to_blackbird` function is changed so that if an SF parameter contains a regref (e.g. `q0` or `q1`, as in the string representation of the symbolic expression in SF) it stores it as a `blackbird.RegRefTransform` instead of as a string.

When converting from a Blackbird program to an SF program, the symbolic expression needs to be extracted from the `RegRefTransform`, so that SF can understand it. This is being added in [this Blackbird PR #46](https://github.com/XanaduAI/blackbird/pull/46).

**Benefits:**
SF circuits with feed-forwarding can now be transformed and serialized into Blackbird and back, without any issues or loss of information.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
Closes #595

### DIFF
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -18,9 +18,9 @@ code.
 """
 # pylint: disable=protected-access,too-many-nested-blocks
 import os
-import regex as re
 from numbers import Number
 
+import re
 import numpy as np
 
 import blackbird
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -20,7 +20,6 @@ code.
 import os
 from numbers import Number
 
-import re
 import numpy as np
 
 import blackbird
@@ -84,8 +83,9 @@ def to_blackbird(prog, version="1.0"):
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
-                    if re.search(r"(?<!\w)q\d*(?!\w)", str(a)):
-                        # check if there are any regref statements (e.g. q0, q1)
+                    isMeasuredParameter = lambda x : isinstance(x, sfpar.MeasuredParameter)
+                    if any(map(isMeasuredParameter, a.free_symbols)):
+                        # check if there are any measured parameters in `a`
                         a = blackbird.RegRefTransform(a)
                     else:
                         a = str(a)
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -18,6 +18,7 @@ code.
 """
 # pylint: disable=protected-access,too-many-nested-blocks
 import os
+import regex as re
 from numbers import Number
 
 import numpy as np
@@ -26,7 +27,6 @@ import blackbird
 import strawberryfields.program as sfp
 import strawberryfields.parameters as sfpar
 from . import ops
-import regex as re
 
 
 # for automodapi, do not include the classes that should appear under the top-level strawberryfields namespace
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -46,6 +46,8 @@ def to_blackbird(prog, version="1.0"):
     bb = blackbird.BlackbirdProgram(name=prog.name, version=version)
     bb._modes = list(prog.reg_refs.keys())
 
+    isMeasuredParameter = lambda x: isinstance(x, sfpar.MeasuredParameter)
+
     # TODO not sure if this makes sense: the program has *already been* compiled using this target
     if prog.target is not None:
         # set the target
@@ -80,7 +82,6 @@ def to_blackbird(prog, version="1.0"):
                     op["kwargs"]["dark_counts"] = cmd.op.dark_counts
 
         else:
-            isMeasuredParameter = lambda x: isinstance(x, sfpar.MeasuredParameter)
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -80,7 +80,7 @@ def to_blackbird(prog, version="1.0"):
                     op["kwargs"]["dark_counts"] = cmd.op.dark_counts
 
         else:
-            isMeasuredParameter = lambda x : isinstance(x, sfpar.MeasuredParameter)
+            isMeasuredParameter = lambda x: isinstance(x, sfpar.MeasuredParameter)
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -26,6 +26,7 @@ import blackbird
 import strawberryfields.program as sfp
 import strawberryfields.parameters as sfpar
 from . import ops
+import regex as re
 
 
 # for automodapi, do not include the classes that should appear under the top-level strawberryfields namespace
@@ -83,7 +84,11 @@ def to_blackbird(prog, version="1.0"):
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
-                    a = str(a)
+                    if re.search(r"(?<!\w)q\d*(?!\w)", str(a)):
+                        # check if there are any regref statements (e.g. q0, q1)
+                        a = blackbird.RegRefTransform(a)
+                    else:
+                        a = str(a)
                 op["args"].append(a)
 
         # If program type is "tdm" then add the looped-over arrays to the
--- a/strawberryfields/io.py
+++ b/strawberryfields/io.py
@@ -80,10 +80,10 @@ def to_blackbird(prog, version="1.0"):
                     op["kwargs"]["dark_counts"] = cmd.op.dark_counts
 
         else:
+            isMeasuredParameter = lambda x : isinstance(x, sfpar.MeasuredParameter)
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
-                    isMeasuredParameter = lambda x : isinstance(x, sfpar.MeasuredParameter)
                     if any(map(isMeasuredParameter, a.free_symbols)):
                         # check if there are any measured parameters in `a`
                         a = blackbird.RegRefTransform(a)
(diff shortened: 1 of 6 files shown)


strawberryfields/io.py
```diff
@@ -83,7 +84,11 @@ def to_blackbird(prog, version="1.0"):
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
-                    a = str(a)
+                    if re.search(r"(?<!\w)q\d*(?!\w)", str(a)):
```
albi3ro, line 87: I was wondering if we could do this without having to run a regex on its string form.  Maybe something like:

```
isMeasuredParameter = lambda x : isinstance(x, sf.parameter.MeasuredParameter)
if any(map(isMeasuredParameter, x.free_symbols)):
```

Maybe I just have personal biases against regular expressions.

strawberryfields/parameters.py
```diff
@@ -257,6 +258,8 @@ def par_convert(args, prog):
     """
 
     def do_convert(a):
+        if isinstance(a, blackbird.RegRefTransform):
```
albi3ro, line 261: Should this get tested as well?

strawberryfields/io.py
thisac: Nice, yes, this is a much nicer solution. Thanks @albi3ro for catching this! 🥇 💯

strawberryfields/parameters.py
thisac: There is actually a test for this, but it was skipped. Just un-skipped it. It should pass when [Blackbird #46](https://github.com/XanaduAI/blackbird/pull/46) is in (at least it does for me locally).

strawberryfields/io.py
```diff
@@ -83,7 +83,12 @@ def to_blackbird(prog, version="1.0"):
             for a in cmd.op.p:
                 if sfpar.par_is_symbolic(a):
                     # SymPy object, convert to string
-                    a = str(a)
+                    isMeasuredParameter = lambda x : isinstance(x, sfpar.MeasuredParameter)
```
albi3ro, line 86: One more comment: Should we define this lambda function outside the for loop, so we don't have to redefine it each time?

albi3ro (approved):
Appreciated the opportunity to start digging into sf and blackbird :)

strawberryfields/parameters.py
```diff
@@ -257,6 +258,8 @@ def par_convert(args, prog):
...
+            a = a.expr
         if isinstance(a, sympy.Basic):
```
antalszava, line 263:
```suggestion
        elif isinstance(a, sympy.Basic):
```

strawberryfields/io.py
```diff
@@ -80,10 +80,15 @@ def to_blackbird(prog, version="1.0"):
                     op["kwargs"]["dark_counts"] = cmd.op.dark_counts
 
         else:
+            isMeasuredParameter = lambda x: isinstance(x, sfpar.MeasuredParameter)
```
antalszava, line 83: (Minor): maybe worth moving this out of the for loop

antalszava (approved):
Looks good to me! :tada:

strawberryfields/parameters.py
thisac: This actually needs to stay as an `if` since if `a` is a `blackbird.RegRefTransform` then it'll be changed into a sympy expression, and thus also enter the next if (since `a` is then a `sympy.Basic`).

strawberryfields/io.py
thisac: Oh, haha, @albi3ro actually commented on the same thing, but I only moved it outside the first for-loop. 😆

strawberryfields/parameters.py
antalszava: Oh completely fair point!

PR #599, opened 2021-06-23
Add n-mode gaussian transformation gate in TFbackend

sylviemonet:
This is the work to add an n-mode gaussian transformation gate (class Ggate) basing on TFbackend, with my supervisor @ziofil and with the help of @nquesada.

The main idea is to create the n-mode gaussian gate from its (S,d) parameters [Symplectic matrix and displacement vector] to its (C, mu, Sigma) parameters and then get the matrix of gate recursively.

[Warning:]
This gate has a special optimizer.

This gate haven't finished the decomposition part.


sylviemonet:
Can not pass the tests because it can not import the function from TW for now.

requirements.txt
```diff
@@ -6,7 +6,7 @@ tensorboard>=2.0
 networkx>=2.0
 git+https://github.com/XanaduAI/blackbird@master#egg=quantum-blackbird
 python-dateutil==2.8.0
-thewalrus>=0.15.0
+http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/thewalrus-0.16.0.dev0+20210730000214-cp38-cp38-manylinux2010_x86_64.whl
```
ziofil, line 9: this will change as soon as the walrus 0.16 is released 👍

strawberryfields/backends/tfbackend/ops.py
```diff
@@ -389,6 +395,106 @@ def two_mode_squeezer_matrix(theta, phi, cutoff, batched=False, dtype=tf.complex
...
+@tf.custom_gradient
+def single_gaussian_gate_matrix(R, y, C, cutoff, dtype=tf.complex64.as_numpy_dtype):
+    """creates a N-mode gaussian gate matrix"""
+    C = C.numpy()
+    y = y.numpy()
+    R = R.numpy()
+    gate = gaussian_gate_tw(R, cutoff, y, C=C, dtype=dtype)
+
+    @tf.function
+    def grad(dy):
```
ziofil, line 472: I know `dy` is kind of standard, but it would probably be helpful for future maintenance to call this variable with the da_db syntax (in this case what is it? dL_dG?)

ziofil: oh, it's the conjugate of dL_dG

strawberryfields/backends/tfbackend/ops.py
```diff
@@ -671,6 +777,124 @@ def two_mode_gate(matrix, mode1, mode2, in_modes, pure=True, batched=False):
     return output
 
 
+def n_mode_gate(matrix, modes, in_modes, pure=True, batched=False):
```
ziofil, line 780: nice!

strawberryfields/ops.py
```diff
@@ -2222,6 +2222,44 @@ def __str__(self):
...
+    def __init__(self, S, d):
+        super().__init__([S, d])
+        self.ns = len(d)
+
+    def _apply(self, reg, backend, **kwargs):
+        S, d = par_evaluate(self.p)
+        backend.gaussian_gate(S, d, *reg)
+
+    def _decompose(self, reg, **kwargs):
+        # TODO: decompose?
```
ziofil, line 2259: @nquesada @josh146 what do you think? We do know how to decompose the Gaussian unitary, but what happens if we tell SF how to do it? Will it try to decompose it all the time and we will lose the advantage of doing it in a single go, or will it decompose it only when we try to compile the circuit?

tests/backend/test_gaussian_gates.py
```diff
@@ -0,0 +1,135 @@
...
+        OW = np.block([[W.real, -W.imag], [W.imag, W.real]])
+        OV = np.block([[V.real, -V.imag], [V.imag, V.real]])
+        S = OW@np.diag(np.concatenate([np.exp(-r), np.exp(r)]))@OV
+        expected_C, expected_mu, expected_Sigma = choi_trick(S, alpha, num_mode)
+        assert np.allclose(_C, expected_C, atol=tol, rtol=0)
+        assert np.allclose(_mu, expected_mu, atol=tol, rtol=0)
+        assert np.allclose(_Sigma, expected_Sigma, atol=tol, rtol=0)
+    
+    @pytest.mark.parametrize("cutoff", [5,7])
+    def test_n_mode_gate(self, cutoff, tol):
```
ziofil, line 48: I would not put more than one `assert` per test. It's good to have many, but maybe put them in different test functions?

ziofil (approved):
looks very good! almost ready for release, good job Yuan 💪

strawberryfields/ops.py
josh146: The decomposition will only occur if not supported by the backend :) This is determined by the `sf.compilers` module

sylviemonet:
Hi Filippo @ziofil, could you give a last look at this PR? I think that things are all good on this one.

nquesada:
@sylviemonet : don't forget to update the CHANGELOG before merging.

nquesada:
Also, @sylviemonet , seems like the CodeFactor bot is not happy about a couple of minor things.

sylviemonet:
> Also, @sylviemonet , seems like the CodeFactor bot is not happy about a couple of minor things.

Hi @nquesada Nico, I'm really confused with CodeFactor's issue. The first one is the part that I haven't touched... The second one is about the docstring... Compared with other functions at the same level as gaussian_gate (e.g. two_mode_squeezing, kerr_interaction...), they do not have the docstring either, so I don't know if I need to add that..

nquesada:
Ahh, ok, no worries. Sometimes the bots are not so smart.

nquesada:
Hi @sylviemonet : I suspect you will need to add a few more test to add coverage to the new functionality you implemented. See for example:
image

You can see the breakdown of the coverage by following the detail links in front of codecov/project

.github/CHANGELOG.md
```diff
@@ -1,6 +1,74 @@
...
+eng = sf.Engine("tf", backend_options={"cutoff_dim": cutoff})
+prog = sf.Program(2)
+
+with prog.context as q:
+    sf.ops.Ggate(S,d) | (q[0],q[1])
+
+state_out = eng.run(prog).state.ket()
+print(state_out)
+```
+ It needs to pay attention that, in order to update the two parameters S and d within a optimization task, you can apply the gradients of d on any optimizer, not with the S. (Because the real symplectic matrix is endowed in a Riemannian mannifold, you need to follow its rule to update.) Therefore, there is a function ``update_symplectic`` is prepared inside the library to update the S.
```
ziofil, line 27: "Note that in order to update the parameter S from its gradient, you cannot use gradient descent directly (because it would not be symplectic after the update). Please use the function ``update_symplectic`` which is specifically for this purpose."

tests/backend/test_gaussian_gates.py
```diff
@@ -0,0 +1,152 @@
...
+        """Test if the output state of the gaussian gate has the right effect on states in the Fock basis"""
+        S = sympmat(num_mode)
+        d = np.random.random(num_mode)
+        X = np.zeros((cutoff, cutoff))
+        X[0,0] = 1
+        Ggate_output = gaussian_gate(S, d, [0,1], in_modes = X, cutoff = cutoff, pure = True)
+        ref_state = np.einsum("abcd,bd->ac",fock_tensor(S, d, cutoff), X)
+        assert np.allclose(Ggate_output, ref_state, atol=tol, rtol=0.0)
+        
+#    def test_gaussian_gate_with_backend(self, setup_backend, cutoff, tol):
```
ziofil, line 140: Remember to remove this from the final PR if it's not needed (we don't want commented code to end up in the main branch).

sylviemonet: Yes, you are right. I was testing the "covarage" functionality of the Codedev...

.github/CHANGELOG.md
```diff
@@ -1,6 +1,74 @@
...
+
+def norm_loss(state):
+    return -tf.abs(tf.linalg.norm(state))**2
+
+def loss(state, objective):
+    return overlap_loss(state, objective) + norm_loss(state)
+
+num_mode = 1
+cutoff = 10
+S = tf.Variable(sympmat(num_mode),dtype=tf.complex64)
```
nquesada, line 48: @sylviemonet : I suspect what you are trying to do here is to generate a random symplectic matrix. The function `sympmat` simply generates the xxpp Symplectic form. What you want to use for making random symplectics is `from thewalrus.random import random_symplectic`

.github/CHANGELOG.md
```diff
@@ -1,6 +1,74 @@
 # Release 0.19.0 (development release)
 
 <h3>New features since last release</h3>
+* The generic multimode Gaussian gate ``Ggate`` is now available in the ``sf.ops`` module with the backend choice of ``tf``. N mode ``Ggate`` can be parametrized by a real symplectic matrix S (size 2N * 2N) and a diplacement vector d (size N). You can also get the gradients of this Gaussian gate by using the ``tape.gradient`` of TensorFlow. [(#599)](https://github.com/XanaduAI/strawberryfields/pull/599)
```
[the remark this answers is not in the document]
nquesada: Should the displacement vector be of size `2N` @sylviemonet ?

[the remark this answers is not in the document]
sylviemonet: Yes, the displacement vector is size 2N (d). And the parameter for the displacement operator \alpha is size N (\frac{1}{\sqrt{2\hbar}}(d_j + 1i d_{j+N}))?

strawberryfields/__init__.py
```diff
@@ -32,6 +32,7 @@
 from .tdm import TDMProgram
 from .plot import plot_wigner, plot_fock, plot_quad
 from . import tdm
+from .backends.tfbackend.ops import update_symplectic
```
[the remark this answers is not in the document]
sduquemesa: It will cause a failure in that case indeed. If `update_symplectic` will only be use within the context of having TensorFlow installed, perhaps this import can be moved to [`strawberryfields/backends/tfbackend/__init__.py`](https://github.com/sylviemonet/strawberryfields/blob/6cabcab5143007643489d2bb49ea8207be7caa4e/strawberryfields/backends/tfbackend/__init__.py#L144-L151) after checking if `tf` is installed.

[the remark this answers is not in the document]
thisac: I agree with @sduquemesa that this would better be moved to the `tfbackend` init file instead, if it needs to live withing the `tfbackend.ops.py` file. Alternatively just keep it where it is (and just remove it from any init files), as `strawberryfields.backends.tfbackend.ops.update_symplectic`. 

Although, I'm wondering why this function is inside `tfbackend` at all. Since it doesn't seem to use TensorFlow, wouldn't it be better just to move it outside of the `tfbackend` files? Would this function perhaps even make sense to put in The Walrus, in the `thewalrus.symplectic` submodule? @ziofil ? 🤔

.github/CHANGELOG.md
```diff
@@ -2,6 +2,73 @@
...
+
+  with prog.context as q:
+      sf.ops.Ggate(S, d) | (q[0], q[1])
+
+  state_out = eng.run(prog).state.ket()
+  ```
+
+  Note that in order to update the parameter S by using its gradient, you cannot use gradient 
+  descent directly (as the unitary would not be symplectic after the update).
+  Please use the function `sf.update_symplectic` which is designed specifically for this purpose.
```
thisac, line 31:
```suggestion
  Note that in order to update the parameter `S` by using its gradient, you cannot use gradient 
  descent directly (as the unitary would not be symplectic after the update). Please use the
  function `sf.update_symplectic` which is designed specifically for this purpose.
```

strawberryfields/ops.py
```diff
@@ -2285,6 +2285,43 @@ def __str__(self):
...
+
+        .. admonition:: Definition
+            :class: defn
+
+        .. math::
+            & cov = S*cov*S^T,\\
+            & mean = mean + d.
+
+    """
+    ns = None
```
thisac, line 2311: Why is this line needed? Seems like `self.ns` is set in the `__init__` method. Is `Ggate.ns` at any point called (would only return `None` either way). 🤔

strawberryfields/__init__.py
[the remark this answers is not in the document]
josh146: > Since it doesn't seem to use TensorFlow, wouldn't it be better just to move it outside of the tfbackend files? 

Agree!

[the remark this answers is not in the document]
josh146: Just checking if anyone is available to work on this? Just to avoid this PR being blocked for too much longer 🙂

strawberryfields/ops.py
ziofil: safe to remove? (I've seen `.ns` exists elsewhere)

ziofil (approved):

requirements.txt
```diff
@@ -1,16 +1,15 @@
-numpy==1.19.2
-scipy==1.7.1
-sympy==1.8
-tensorflow==2.6.0
-tensorboard==2.6.0
-typing-extensions==3.7.4
-networkx==2.6.3
-quantum-blackbird==0.4.0
+numpy>=1.18.5
```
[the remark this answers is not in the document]
ziofil: what should the version be?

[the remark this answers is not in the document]
josh146: @ziofil easiest thing to do is just
```python
git checkout master requirements.txt
```

to just replace this file with the `master` branch version

PR #601, opened 2021-06-25
Space unrolling for TDM programs

thisac:
**Context:**
This PR adds support for using a `TDMProgram` to construct TDM circuits with Fock measurements and multiple loops (e.g. TD3). The major change is needed support for space-unrolling.

* Updates `TDMProgram` to support TDM circuits with Fock measurements.
* Adds a method for space-unrolling TDM circuits (works for all TDM circuits).
* Adds a helper function to calculate the mode indices and number of concurrent modes in a TDM program.
* Updates the engine to support compiling TDM programs with the passive compiler.

**Note**: when space-unrolling a circuit, subsystems need to be added to the `TDMProgram`. These are removed when rolling back the circuit.

**Benefits:**
TDM programs with Fock measurements can be constructed (via `TDMProgram`) and all TDM programs can be space-unrolled.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/backends/fockbackend/backend.py
+++ b/strawberryfields/backends/fockbackend/backend.py
@@ -176,7 +176,6 @@ class FockBackend(BaseFock):
         self.circuit.two_mode_squeeze(r, phi, self._remap_modes(mode1), self._remap_modes(mode2))
 
     def beamsplitter(self, theta, phi, mode1, mode2):
-        print("hello from BS")
         self.circuit.beamsplitter(theta, phi, self._remap_modes(mode1), self._remap_modes(mode2))
 
     def measure_homodyne(self, phi, mode, shots=1, select=None, **kwargs):
--- a/strawberryfields/backends/fockbackend/backend.py
+++ b/strawberryfields/backends/fockbackend/backend.py
@@ -176,6 +176,7 @@ class FockBackend(BaseFock):
         self.circuit.two_mode_squeeze(r, phi, self._remap_modes(mode1), self._remap_modes(mode2))
 
     def beamsplitter(self, theta, phi, mode1, mode2):
+        print("hello from BS")
         self.circuit.beamsplitter(theta, phi, self._remap_modes(mode1), self._remap_modes(mode2))
 
     def measure_homodyne(self, phi, mode, shots=1, select=None, **kwargs):
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -428,13 +428,13 @@ class LocalEngine(BaseEngine):
         # pylint: disable=import-outside-toplevel
         from strawberryfields.tdm.tdmprogram import reshape_samples
 
-        recieved_rolled_circuit = False
+        received_rolled_circuit = False
         if isinstance(program, TDMProgram):
             # priority order for the shots value should be kwargs > run_options > 1
             shots = kwargs.get("shots", program.run_options.get("shots", 1))
             # if a tdm program is input in a rolled state, then unroll it
             if not program.is_unrolled:
-                recieved_rolled_circuit = True
+                received_rolled_circuit = True
                 program.unroll(shots=shots)
 
             # Shots >1 for a TDM program simply corresponds to creating
@@ -503,8 +503,8 @@ class LocalEngine(BaseEngine):
                 )
                 # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
                 result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
-            if recieved_rolled_circuit:
-                # if the tdm circuit is recieved in a rolled state, and unrolled
+            if received_rolled_circuit:
+                # if the tdm circuit is received in a rolled state, and unrolled
                 # for execution, roll it back again
                 program.roll()
         modes = temp_run_options["modes"]
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -435,7 +435,8 @@ class LocalEngine(BaseEngine):
             # priority order for the shots value should be kwargs > run_options > 1
             shots = kwargs.get("shots", program.run_options.get("shots", 1))
 
-            program.unroll(shots=shots)
+            if program.unrolled_circuit is None and program.space_unrolled_circuit is None:
+                program.unroll(shots=shots)
             # Shots >1 for a TDM program simply corresponds to creating
             # multiple copies of the program, and appending them to run sequentially.
             # As a result, we set the backend shots to 1 for the Gaussian backend.
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -428,13 +428,13 @@ class LocalEngine(BaseEngine):
         # pylint: disable=import-outside-toplevel
         from strawberryfields.tdm.tdmprogram import reshape_samples
 
-        rolled_input = False
+        recieved_rolled_circuit = False
         if isinstance(program, TDMProgram):
             # priority order for the shots value should be kwargs > run_options > 1
             shots = kwargs.get("shots", program.run_options.get("shots", 1))
             # if a tdm program is input in a rolled state, then unroll it
             if not program.is_unrolled:
-                rolled_input = True
+                recieved_rolled_circuit = True
                 program.unroll(shots=shots)
 
             # Shots >1 for a TDM program simply corresponds to creating
@@ -503,8 +503,9 @@ class LocalEngine(BaseEngine):
                 )
                 # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
                 result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
-            if rolled_input:
-                # if tdm program is input in a rolled state, roll it back again
+            if recieved_rolled_circuit:
+                # if the tdm circuit is recieved in a rolled state, and unrolled
+                # for execution, roll it back again
                 program.roll()
         modes = temp_run_options["modes"]
 
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -434,9 +434,12 @@ class LocalEngine(BaseEngine):
         if valid_tdm_program:
             # priority order for the shots value should be kwargs > run_options > 1
             shots = kwargs.get("shots", program.run_options.get("shots", 1))
-
+            # if a tdm program is input in a rolled state, then unroll it
             if program.unrolled_circuit is None and program.space_unrolled_circuit is None:
+                rolled_input = True
                 program.unroll(shots=shots)
+            else:
+                rolled_input = False
             # Shots >1 for a TDM program simply corresponds to creating
             # multiple copies of the program, and appending them to run sequentially.
             # As a result, we set the backend shots to 1 for the Gaussian backend.
@@ -497,12 +500,15 @@ class LocalEngine(BaseEngine):
         )
 
         if valid_tdm_program:
-            result._all_samples = reshape_samples(
-                result.all_samples, program.measured_modes, program.N, program.timebins
-            )
-            # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
-            result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
-            program.roll()
+            if isinstance(result.all_samples, dict) and len(result.all_samples) > 0:
+                result._all_samples = reshape_samples(
+                    result.all_samples, program.measured_modes, program.N, program.timebins
+                )
+                # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
+                result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
+            if rolled_input:
+                # if tdm program is input in a rolled state, roll it back again
+                program.roll()
         modes = temp_run_options["modes"]
 
         if modes is None or modes:
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -28,6 +28,7 @@ from strawberryfields.api import Connection, Job, Result
 from strawberryfields.api.job import FailedJobError
 from strawberryfields.logger import create_logger
 from strawberryfields.program import Program
+from strawberryfields.tdm.tdmprogram import TDMProgram
 
 from .backends import load_backend
 from .backends.base import BaseBackend, NotApplicableError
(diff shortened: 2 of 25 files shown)


strawberryfields/backends/fockbackend/backend.py
```diff
@@ -176,6 +176,7 @@ def two_mode_squeeze(self, r, phi, mode1, mode2):
         self.circuit.two_mode_squeeze(r, phi, self._remap_modes(mode1), self._remap_modes(mode2))
 
     def beamsplitter(self, theta, phi, mode1, mode2):
+        print("hello from BS")
```
nquesada, line 179:
```suggestion
```

nquesada (approved):

strawberryfields/tdm/tdmprogram.py
```diff
@@ -572,6 +576,38 @@ def unroll(self, shots):
...
 
+        if self._is_space_unrolled:
+            raise ValueError("Program is space-unrolled and must be rolled before unrolling")
+
+        return self._unroll_program(shots)
+
+    def space_unroll(self, shots=1):
+        """Construct the space-unrolled program
+
+        Constructs the space-unrolled single-shot program, storing it in `self.space_unrolled_circuit`
```
antalszava, line 587: Based on this comment, I'd expect to find a line that is `self.space_unrolled_circuit = <somehow unrolling the circuit>`. Since this is not the case, maybe it's worth rephrasing this part a bit.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -554,9 +551,16 @@ def parameters(self):
...
+            if self.added_subsystems > 0:
+                self._delete_subsystems(self.register[-self.added_subsystems :])
+                self.init_num_subsystems -= self.added_subsystems
+                self.added_subsystems = 0
+
+            self._is_space_unrolled = False
         return self
 
-    def unroll(self, shots):
+    def unroll(self, shots=1):
```
antalszava, line 563: How come `shots=1` is the default now?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -572,6 +576,38 @@ def unroll(self, shots):
             self.circuit = self.unrolled_circuit * shots
             return self
 
+        if self._is_space_unrolled:
+            raise ValueError("Program is space-unrolled and must be rolled before unrolling")
```
antalszava, line 580: I would expect that this error message will seem confusing to a user. Maybe naming the methods to call could help? (E.g., "Program is space-unrolled and cannot be unrolled. Call roll() so that it can be unrolled" or similar).

strawberryfields/tdm/tdmprogram.py
```diff
@@ -605,39 +641,52 @@ def unroll(self, shots):
         # q[sm[2]] as concurrent modes of spatial mode C
         # q[sm[3]] as concurrent modes of spatial mode D.
 
+        # save previous mode index of a command to be able to check when modes
+        # are looped back to the start (not allowed when space-unrolling)
```
antalszava, line 645: Is it not allowed or not applicable for space-unrolling?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -392,12 +381,18 @@ def __init__(self, N, name=None):
...
         self.type = "tdm"
+        self._is_space_unrolled = False
+
         self.timebins = 0
         self.spatial_modes = 0
         self.measured_modes = []
         self.rolled_circuit = None
         # `unrolled_circuit` only contains the unrolled single-shot circuit
         self.unrolled_circuit = None
+        # `space_unrolled_circuit` only contains the space-unrolled single-shot circuit
```
antalszava, line 392: It would be great to get an overall description of the idea of space unrolling and to have a comparison with unrolling. This could maybe be done as an example and the change in the program could be explained.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -429,12 +424,13 @@ def compile(self, *, device=None, compiler=None):
             compiler (str, ~strawberryfields.compilers.Compiler): Compiler name or compile strategy
                 to use. If a device is specified, this overrides the compile strategy specified by
                 the hardware :class:`~.DeviceSpec`. If no compiler is passed, the default "TD2"
+                compiler is used unless the program is a TD3 program, in which case the "passive"
                 compiler is used. Currently, the only other allowed compiler is "gaussian".
```
antalszava, line 428: > Currently, the only other allowed compiler is "gaussian".

This line confuses me :thinking: Do we not allow "TD2" and now "passive" as compilers?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -605,39 +641,52 @@ def unroll(self, shots):
...
             if isinstance(cmd.op, ops.Measurement):
                 self.measured_modes.append(cmd.reg[0].ind)
+
         for i in range(self.timebins):
             for cmd in self.rolled_circuit:
-                self.apply_op(cmd, q, i)
-
-            if self.shift == "default":
+                modes = get_modes(cmd, q)
+                if not (self._is_space_unrolled and any(m.ind < last_idx[cmd] for m in modes)):
```
antalszava, line 656: Would consider:
1) breaking up the boolean expression such that `not` acts on it using De Morgan's laws;
2) naming the right expression such that it will be more readable.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -605,39 +641,52 @@ def unroll(self, shots):
         # q[sm[2]] as concurrent modes of spatial mode C
         # q[sm[3]] as concurrent modes of spatial mode D.
 
+        # save previous mode index of a command to be able to check when modes
+        # are looped back to the start (not allowed when space-unrolling)
+        last_idx = dict()
```
antalszava, line 646: How about naming this `previous_mode_index`?

strawberryfields/engine.py
```diff
@@ -495,13 +496,16 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
...
-            program.roll()
+        if isinstance(program, TDMProgram):
+            if isinstance(result.all_samples, dict) and len(result.all_samples) > 0:
+                result._all_samples = reshape_samples(
+                    result.all_samples, program.measured_modes, program.N, program.timebins
+                )
+                # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
+                result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
+            if rolled_input:
+                # if tdm program is input in a rolled state, roll it back again
```
antalszava, line 507: Why? :thinking: I'd think if it's rolled, we don't need to roll it again :thinking:

strawberryfields/engine.py
```diff
@@ -427,15 +428,15 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
         # pylint: disable=import-outside-toplevel
         from strawberryfields.tdm.tdmprogram import reshape_samples
 
-        # At this time we do not support lists of tdm programs
-        valid_tdm_program = (
-            not isinstance(program, collections.abc.Sequence) and program.type == "tdm"
-        )
-        if valid_tdm_program:
+        rolled_input = False
```
antalszava, line 431: I'm confused: what does rolled input mean? :thinking:

strawberryfields/tdm/tdmprogram.py
```diff
@@ -392,12 +381,18 @@ def __init__(self, N, name=None):
...
         self.timebins = 0
         self.spatial_modes = 0
         self.measured_modes = []
         self.rolled_circuit = None
         # `unrolled_circuit` only contains the unrolled single-shot circuit
         self.unrolled_circuit = None
+        # `space_unrolled_circuit` only contains the space-unrolled single-shot circuit
+        self.space_unrolled_circuit = None
+        # `added_subsystems` corresponds to the number of subsystems added when space-unrolling
+        self.added_subsystems = 0
```
antalszava, line 395: How about adding the `num` prefix? `num_added_subsystems` or `num_added_modes`?

strawberryfields/tdm/tdmprogram.py
```diff
@@ -572,6 +576,38 @@ def unroll(self, shots):
...
+        if self.space_unrolled_circuit is not None:
+            self.circuit = self.space_unrolled_circuit * shots
+            return self
+
+        self.added_subsystems = self.timebins - self.init_num_subsystems
+        if self.added_subsystems > 0:
+            self._add_subsystems(self.added_subsystems)
+
+            self.init_num_subsystems += self.added_subsystems
+        self._is_space_unrolled = True
```
antalszava, line 605: Nice! :)

tests/frontend/test_space_unroll.py
```diff
@@ -0,0 +1,250 @@
...
+from strawberryfields.ops import Sgate, Rgate, BSgate, LossChannel, MeasureFock
+from thewalrus.symplectic import reduced_state
+from thewalrus.quantum import is_pure_cov
+
+pytestmark = pytest.mark.frontend
+
+
+def generate_valid_bs_sequence(delays, modes, random_func=np.random.rand):
+    """
+    Generate sequences of valied beamsplitter angles for a time-domain program where
```
antalszava, line 29:
```suggestion
    Generate sequences of valid beamsplitter angles for a time-domain program where
```

tests/frontend/test_space_unroll.py
```diff
@@ -0,0 +1,250 @@
...
+    prog = sf.TDMProgram([N])
+
+    with prog.context(*angles) as (p, q):
+        for i in range(d):
+            Rgate(p[i + d]) | q[n[i]]
+            BSgate(p[i], np.pi / 2) | (q[n[i + 1]], q[n[i]])
+
+    prog.space_unroll()
+
+    compiled = prog.compile(compiler="passive")
```
antalszava, line 92: How about testing `space_unroll` without compiling the program? Could we make sure that
a) The new modes have been added
b) The correct commands have been added to the space unrolled program?

tests/frontend/test_space_unroll.py
```diff
@@ -0,0 +1,250 @@
...
+    passive_elem = compiled.circuit[0]
+    U = passive_elem.op.p[0]
+    # Check that it is indeed the identity
+    assert np.allclose(U, np.identity(len(U)))
```
antalszava, line 96: :100:

antalszava (commented):
Hi @thisac, this is really great! :1st_place_medal: I have some questions which I'd be keen to get your thoughts on. In general, it may be worth clarifying the difference between unrolling and space unrolling a bit further such that there won't be any confusion.

strawberryfields/tdm/tdmprogram.py
```diff
@@ -113,7 +87,7 @@ def _get_mode_order(num_of_values, modes, N, timebins):
 
     """
     all_modes = []
-    for i in range(len(N)):
+    for i, _ in enumerate(N):
```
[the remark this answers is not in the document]
thisac: Codefactor has started complaining (or maybe it always has) about using `for i in range(len(N)):` instead of `for i, _ in enumerate(N):`. I'm honestly happy either way.

strawberryfields/engine.py
thisac: `rolled_input` refers to whether the program that is passed is a rolled up `TDMProgram` (i.e. that `prog.circuit` is a rolled circuit with placeholders instead of values) or if it's already been unrolled by the user. Either way, it should be kept the same as it was before whenever everything has been run, and not change to rolled/unrolled.

strawberryfields/engine.py
thisac: This means if the input program was rolled it should be kept rolled after the engine has done it's work. It will always have to be unrolled to simulate it.

So, it's either already unrolled (`rolled_input=False`) and then it's just used in it's unrolled state and not rolled again. Or it hasn't yet been unrolled (`rolled_input=True`), and then it must be unrolled and then rolled back again when done.

strawberryfields/tdm/tdmprogram.py
thisac: Bad wording on my side. I've rewritten it now.

strawberryfields/tdm/tdmprogram.py
thisac: It seems to be the most common case (at least when looking at the usage so far), and I believe it makes sense having this as default. There where a lot of `prog.unroll(1)` showing up where the `1` seemed to confused some people, which was partly my fault for writing it like that at some point. 🙂

strawberryfields/tdm/tdmprogram.py
thisac: It's not allowed. It'll work to do that here, but it's not physical, and will cause issues with having more applied gates that there should be.

strawberryfields/engine.py
antalszava: Oh I see! That makes sense! :) Maybe it's worth phrasing the comment such that it states that we've now processed the unrolled circuit and want to restore its rolled state (if the input was rolled)? To me, that piece was missing, though maybe because I was looking at the change which is specific to the step.

strawberryfields/engine.py
antalszava: Thank you, that makes sense! I'm thinking that the name might be made a bit more descriptive to reflect what's going on (though now it makes sense). How about `received_as_rolled` or `received_rolled_circuit`?

strawberryfields/engine.py
```diff
@@ -427,15 +428,15 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
         # pylint: disable=import-outside-toplevel
         from strawberryfields.tdm.tdmprogram import reshape_samples
 
-        # At this time we do not support lists of tdm programs
-        valid_tdm_program = (
-            not isinstance(program, collections.abc.Sequence) and program.type == "tdm"
-        )
-        if valid_tdm_program:
+        recieved_rolled_circuit = False
```
antalszava, line 431:
```suggestion
        received_rolled_circuit = False
```

strawberryfields/engine.py
```diff
@@ -427,15 +428,15 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
...
+        if isinstance(program, TDMProgram):
             # priority order for the shots value should be kwargs > run_options > 1
             shots = kwargs.get("shots", program.run_options.get("shots", 1))
+            # if a tdm program is input in a rolled state, then unroll it
+            if not program.is_unrolled:
+                recieved_rolled_circuit = True
```
antalszava, line 437:
```suggestion
                received_rolled_circuit = True
```

strawberryfields/engine.py
```diff
@@ -495,13 +496,17 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
...
-            result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
-            program.roll()
+        if isinstance(program, TDMProgram):
+            if isinstance(result.all_samples, dict) and len(result.all_samples) > 0:
+                result._all_samples = reshape_samples(
+                    result.all_samples, program.measured_modes, program.N, program.timebins
+                )
+                # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
+                result._samples = np.array(list(result.all_samples.values())).transpose(1, 0, 2)
+            if recieved_rolled_circuit:
```
antalszava, line 506:
```suggestion
            if received_rolled_circuit:
```

strawberryfields/tdm/tdmprogram.py
```diff
@@ -605,39 +653,54 @@ def unroll(self, shots):
...
+
         for i in range(self.timebins):
             for cmd in self.rolled_circuit:
-                self.apply_op(cmd, q, i)
-
-            if self.shift == "default":
+                modes = get_modes(cmd, q)
+                has_looped_back = any(m.ind < previous_mode_index[cmd] for m in modes)
+                valid_application = not self._is_space_unrolled or not has_looped_back
+                if valid_application:
```
antalszava, line 670: Nice!

strawberryfields/engine.py
```diff
@@ -495,13 +496,17 @@ def run(self, program, *, args=None, compile_options=None, **kwargs):
...
+                # if the tdm circuit is recieved in a rolled state, and unrolled
```
antalszava, line 507:
```suggestion
                # if the tdm circuit is received in a rolled state, and unrolled
```

PR #603, opened 2021-06-29
faster gaussian unitary compiling 

jakeffbulmer:
**Context:**
The gaussian_unitary compiler takes a full Gaussian circuit and compiles it into an interferometer -> squeezing -> interferometer -> displacement circuit. Previously, it would compute an MxM matrix-matrix multiplication for each additional gate. Here, we make this faster by replacing the expand and multiply operations with a custom function.

the functions ``_apply_symp_one_mode_gate`` and ``_apply_symp_two_mode_gate``, defined in ``compilers/gaussian_unitary.py`` update symplectic transformations and means vectors in linear time complexity.

**Benefits:**
Running the ``gaussian_unitary`` compiler should be much faster for large numbers of modes.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/strawberryfields/compilers/gaussian_unitary.py
+++ b/strawberryfields/compilers/gaussian_unitary.py
@@ -29,6 +29,31 @@ from thewalrus.symplectic import (
 
 from .compiler import Compiler
 
+def _apply_symp_one_mode_gate(S_G, S, i):
+    """In-place applies a one mode gate G to the symplectic operation, S, in mode i
+    Args:
+        S_G (array): 2x2 matrix for one mode symplectic operation
+        S (array): Symplectic operation
+        i (int): index of one mode gate
+    """
+    M = S.shape[0] // 2
+    (S[i], S[i+M]) = (S_G[0,0] * S[i] + S_G[0,1] * S[i+M], S_G[1,0] * S[i] + S_G[1,1] * S[i+M])
+    return S
+
+def _apply_symp_two_mode_gate(S_G, S, i, j):
+    """In-place applies a two mode gate G to the symplectic operation, S, in modes i and j
+    Args:
+        G (array): 2x2 matrix for two mode gate
+        S (array): Symplectic operation
+        i (int): index of first mode of gate
+        j (int): index of second mode of gate
+    """
+    M = S.shape[0] // 2
+    (S[i], S[j], S[i+M], S[j+M]) = (S_G[0,0] * S[i] + S_G[0,1] * S[j] + S_G[0,2] * S[i+M] + S_G[0,3] * S[j+M],
+                                    S_G[1,0] * S[i] + S_G[1,1] * S[j] + S_G[1,2] * S[i+M] + S_G[1,3] * S[j+M],
+                                    S_G[2,0] * S[i] + S_G[2,1] * S[j] + S_G[2,2] * S[i+M] + S_G[2,3] * S[j+M],
+                                    S_G[3,0] * S[i] + S_G[3,1] * S[j] + S_G[3,2] * S[i+M] + S_G[3,3] * S[j+M])
+    return S
 
 class GaussianUnitary(Compiler):
     """Compiler to arrange a Gaussian quantum circuit into the canonical Symplectic form.
@@ -141,9 +166,9 @@ class GaussianUnitary(Compiler):
             params = par_evaluate(operations.op.p)
             modes = [modes_label.ind for modes_label in operations.reg]
             if name == "Dgate":
-                rnet = rnet + expand_vector(
-                    params[0] * (np.exp(1j * params[1])), dict_indices[modes[0]], nmodes
-                )
+                alpha = params[0] * (np.exp(1j * params[1]))
+                rnet[dict_indices[modes[0]]] += alpha.real 
+                rnet[dict_indices[modes[0]] + nmodes] += alpha.imag
             else:
                 if name == "Rgate":
                     S = expand(rotation(params[0]), dict_indices[modes[0]], nmodes)
--- a/strawberryfields/compilers/gaussian_unitary.py
+++ b/strawberryfields/compilers/gaussian_unitary.py
@@ -29,7 +29,7 @@ from .compiler import Compiler
 
 
 def _apply_symp_one_mode_gate(S_G, S, r, i):
-    """In-place applies a one mode gate to a symplectic operation.
+    r"""In-place applies a one mode gate to a symplectic operation.
 
     Args:
         S_G (array): a :math:`2\times 2` matrix representing a one mode symplectic operation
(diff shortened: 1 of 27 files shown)


strawberryfields/compilers/gaussian_unitary.py
```diff
@@ -18,18 +18,62 @@
...
     interferometer,
     beam_splitter,
 )
-
+from numba import jit
 from .compiler import Compiler
 
 
+@jit(nopython=True)
+def _apply_symp_one_mode_gate(S_G, S, r, i):
```
nquesada, line 33:
```suggestion
def _apply_symp_one_mode_gate(S_G, S, r, i): # pragma: no cover
```

strawberryfields/compilers/gaussian_unitary.py
```diff
@@ -18,18 +18,62 @@
...
+    )
+    (r[i], r[i + M]) = (
+        S_G[0, 0] * r[i] + S_G[0, 1] * r[i + M],
+        S_G[1, 0] * r[i] + S_G[1, 1] * r[i + M],
+    )
+    return S, r
+
+
+@jit(nopython=True)
+def _apply_symp_two_mode_gate(S_G, S, r, i, j):
```
nquesada, line 53:
```suggestion
def _apply_symp_two_mode_gate(S_G, S, r, i, j): # pragma: no cover
```

nquesada: These two `#pragma: no cover` are due to the fact that `pytest` does not play nicely with jitted code when preparing coverage reports, so we have to exclude them from the coverage report.

tests/frontend/compilers/test_gaussianunitary.py
```diff
@@ -45,7 +49,7 @@ def random_params(size, sq_bound, disp_bound):
 
 
 @pytest.mark.parametrize("depth", [1, 3, 6])
-@pytest.mark.parametrize("width", [5, 10, 15])
+@pytest.mark.parametrize("width", [1, 2, 5, 10, 15])
```
nquesada, line 52: Nice reuse of the tests @jakeffbulmer !

strawberryfields/compilers/gaussian_unitary.py
nquesada: The other option is to not jit these functions. Could we use useful to quickly benchmark how much difference it makes adding the jit.

nquesada (commented):
Very nice addition @jakeffbulmer ! Left some minor comments. Don't forget to update the changelog too!

strawberryfields/compilers/gaussian_unitary.py
jakeffbulmer: hopefully this won't be needed once the coverage is updated with the new tests i wrote!

strawberryfields/compilers/gaussian_unitary.py
```diff
@@ -18,18 +18,70 @@
...
     interferometer,
     beam_splitter,
 )
-
 from .compiler import Compiler
 
 
+def _apply_symp_one_mode_gate(S_G, S, r, i):
+    """In-place applies a one mode gate G to the symplectic operation, S, in mode i
+    Args:
```
josh146, line 33:
```suggestion
    """In-place applies a one mode gate to a symplectic operation.

    Args:
```

strawberryfields/compilers/gaussian_unitary.py
```diff
@@ -18,18 +18,70 @@
...
+        S_G (array): 2x2 matrix for one mode symplectic operation
+        S (array): 2Mx2M Symplectic operation
+        r (array): M length means vector
+        i (int): index of one mode gate
```
josh146, line 37:
```suggestion
        S_G (array): a :math:`2\times 2` matrix representing a one mode symplectic operation
        S (array): a :math:`2M\times 2M` symplectic operation
        r (array): means vector of length :math:`M`
        i (int): the mode the symplectic operation ``S_G`` is applied to
```

strawberryfields/compilers/gaussian_unitary.py
```diff
@@ -18,18 +18,70 @@
...
+
+    Returns:
+        S (array): updated symplectic operation
+        r (array): updated means vector
```
josh146, line 41: Python functions can only return a single object --- in this case, it is returning a tuple :)
```suggestion
    Returns:
        tuple[array, array]: returns a tuple containing the updated symplectic operation
        and the updated means vector
```

strawberryfields/compilers/gaussian_unitary.py
```diff
@@ -18,18 +18,70 @@
...
+    Args:
+        S_G (array): 2x2 matrix for one mode symplectic operation
+        S (array): 2Mx2M Symplectic operation
+        r (array): M length means vector
+        i (int): index of one mode gate
+        j (int): index of second mode of gate
+
+    Returns:
+        S (array): updated symplectic operation
+        r (array): updated means vector
```
josh146, line 67: Suggest making the same modifications as my comment above!

josh146 (changes requested):
Nice addition @jakeffbulmer! The new logic is very neat, my only requested changes are minor and apply to the documentation. Out of curiosity, how much faster is the new approach? Is it very noticeable in general use?

jakeffbulmer:
The speedup isn't massive, I just tried a 100 mode circuit with 100 SMSV states followed by ~5000 beam splitters. The compile time went from 2.6s to 660ms.

strawberryfields/compilers/gaussian_unitary.py
jakeffbulmer: I'm happy to make this change, but thought I'd just highlight that I've seen multiple object return docstrings elsewhere (e.g. https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html), even though python does pack them into a tuple.

josh146: Oh, that's good to know! The reason we use by convention one return in the docs here is that I remember there being a rendering issue if we tried to return more than one in the past.

This could be because we use Google-style docstrings, rather than NumPy-style? Or it could have been an issue in an old version of Sphinx which is now rectified.

nquesada: yeah, maybe good to stick with the local convention here.

strawberryfields/compilers/gaussian_unitary.py
jakeffbulmer: done!

nquesada:
I'd say a factor of 4 is still pretty nice!

nquesada (approved):

Issue #571, opened 2021-04-19
Add photon number and threshold detection to the Bosonic backend


josh146:
Currently, the Bosonic backend does not support the `MeasureFock` and `MeasureThreshold` measurement operations. It would be great to add support for these two operations to the Bosonic backend.

This would entail writing functions to calculate the different probabilities up to a given cutoff, using numpy to sample from this distribution and writing update rules for updating the post measurement state.

JeyKelly:
Hi @josh146, I've been looking at this for the past week, getting more familiarized with the software and code. I've managed to get a partially working version of measure_fock so far, by recycling the gaussian version of it, and by using the methods of the bosonic backend. It's partially working under the condition that I have a state which requires only one Gaussian state to be expressed and I let go of extra dimension added to the get_covmat_xp() and get_mean_xp() matrices. Specifically, the output of get_covmat_xp() has an extra dimension ([n, 2, 2]) compared with the scovmatxp() method to take in account the different Gaussians needed to express the state. Mu is also in the format [n, 2] instead of [1] in the Gaussian format. This leads me to think I can simply iterate and take in account the different weights which are linked to the different Gaussian modes of the bosonic state expression. However, when debugging, I see that self.state().weights() gives me outrageous numbers for any Fock state I generate with more than one photon (I'm getting a two dimensional weight of 400, -399 for single photon), so I'm not sure if I'm misunderstanding the model, or there is some kind of normalization that needs to be added.

This leads me to ask - could you help me understand why is the mu array output from BosonicBackend.circuit.mean gives a [n, 2] array, why the mean is in the format [n, 1, 2] (as opposed to a [2, 1] sized mean in the gaussian backend), and why the weight seems so big in my tests? Also, maybe the previous question would help me understand this better, but do you think sampling each Gaussian modes of the bosonic backend and summing their weighted (normalized?) contribution renders an accurate photon number at the end of the calculation?

nquesada:
Hi @JeyKelly  --- The outrageous weights from the Fock states are to be expected. Can I suggest, for ease of testing that you try a cat state?

nquesada:
As for your second question, regarding the shape of the means, have you looked into this tutorial: https://strawberryfields.ai/photonics/demos/run_intro_bosonic.html ? For a cat state, for which one need 4 Gaussians, one will get a vector of 4 means (2-dimensional vectors). Does that help?

nquesada:
[...]

PR #604, opened 2021-07-01
[WIP] Bosonic backend support for threshold measurement

JeyKelly:
**Context:** Currently, the Bosonic backend does not support the `MeasureFock` and `MeasureThreshold` measurement operations. This pull request aims to add support for it. Ref : issue [#571](https://github.com/XanaduAI/strawberryfields/issues/571)

**Description of the Change:** Adding support for measure_threshold and measure_fock

**Benefits:**

**Possible Drawbacks:** N/A

**Related GitHub Issues:**


josh146:
Thanks @JeyKelly for this PR! Feel free to ask any questions here, and you can comment directly on the code itself 🙂

JeyKelly:
@josh146 I've checked the result of the test and it seems like the test was designed to raise a non-implemented error when calling the MeasureThreshold operation (~strawberryfields\tests\backend\test_bosonic_backend.py line 664). Assuming the Fock state case works and post-measurement is updated properly, should I update the pull request to include removing the line of code and adding two extra modes to `test_measurement` verifying the proper sample shapes/states after applying `MeasureFock` and `MeasureThreshold`?

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -803,6 +804,59 @@ def measure_dyne(self, covmat, modes, shots=1):
...
+            measurements[i] = tor_i[0][0] 
+
+        index = np.array([np.random.choice(list(range(len(measurements))),1,p=np.abs(weights))])
+        samples = measurements[index]
+        if measurements[index] == 1:
+            print("wip : updating the state in the case there is a click")
+        else:
+            print("wip : updating the state in the case there is no click")     
+        return samples
+
```
nquesada, line 859: dear @JeyKelly : my first suggestion would be to use the method [`fidelity_vacuum`](https://github.com/XanaduAI/strawberryfields/blob/9f6debbb22476bebb7ef6d4c3041e3590aeba1d4/strawberryfields/backends/bosonicbackend/bosoniccircuit.py#L542). This method will tell you the probability `p0` that the detector does not click. The probability that it does click is given by `1-p0`. You can then flip a coin with this bias (using `np.random.choice`) and then based on that one will have to update the vectors of coefficients, means, and covariance matrices. Maybe I'd suggest you start by writing the logic deciding whether to click or not. I will try to give you more detailed info on how to update the state based on the two different outcomes.

JeyKelly: @nquesada This is definitely a more efficient solution, thank you. It works properly except in the case of vacuum where I used  `is_vacuum` to circumvent the rounding error (`p0>1`). Is there a similarly simple solution for the Fock measurement or is sampling each gaussian mode individually like I put in the pull request a decent approach? It still has the issue of using non-normalized weights, so in the case of a pure Fock state, a measurement can't be done using the current definition.

nquesada: Hi @JeyKelly --- As promised here are some notes explaining the algorithm:

[bosonic_threshold.pdf](https://github.com/XanaduAI/strawberryfields/files/6764815/bosonic_threshold.pdf)

nquesada: Also, this update rule was implemented in Fortran in this repo here https://github.com/XanaduAI/torontonian-sampling/blob/master/src/torontonian_samples.f90

JeyKelly: > 
> 
> Hi @JeyKelly --- As promised here are some notes explaining the algorithm:
> 
> [bosonic_threshold.pdf](https://github.com/XanaduAI/strawberryfields/files/6764815/bosonic_threshold.pdf)

I took a look at this, thanks for explaining it. I was wondering though: wouldn't the rewritten cov matrix \sigma^{(l)} have (2M - 2) for matrices dimensions excluding the last mode in \sigma_{(B)} (and similarly for \vec{r}_{(A)})? I'm asking because I have a hard time understanding why the cov matrices have a (M - 2) x (M - 2) sizes post no-click measurements. 

The way I interpret this is we're cutting out the mode completely post-measurement, and so we end up with modified matrices for the remaining states - however, I don't quite understand why in the case where we have a click, the remaining cov/weights/means matrices have sizes which reflect this extra mode that we have measured? Initially, I thought we were replacing the mode with vacuum and adjusting the remaining ones to reflect this depending on the measurement outcome, but as far as I understand it's not what is done here. Do you have an idea where I'm not understanding properly?

JeyKelly: > Also, this update rule was implemented in Fortran in this repo here https://github.com/XanaduAI/torontonian-sampling/blob/master/src/torontonian_samples.f90

@nquesada I am having trouble interfacing this in Python; it's telling me I'm missing the dependencies (use kinds, vars, structures) in the `torontonian_samples` module. Would this ultimately be translated in python code to be implemented in Strawberryfields? If I can manage to make it run on my end I would gladly rewrite it in a python file that can be used directly in the measure_threshold method.

nquesada: Hi @jeykelly --- I am not suggesting you use that Fortran code verbatim. I suggested it because it can give you and idea of how to implement the update rules. Also, I made typo, in the notes, the updated matrices are of size 2M-2 x 2M-2. Sorry about this!

jakeffbulmer: also to answer your question @JeyKelly about why they are 2M-2 x 2M-2, it is because they only describe the modes which were not measured. Since optical measurements are typically _demolition_ measurements, the measured mode is typically reset to vacuum, as you said. This can be done afterwards, just be adding 2 extra rows/cols to the covariance matrix in the positions corresponding to the measured modes (just with hbar/2 on the diagonal and zero everywhere else). And adding 2 extra zeros to the updated 2M-2 length displacement vector.

JeyKelly:
@nquesada I have revisited the measure_threshold method in the way you suggested. I have ran into a few issues when testing, and this is why I didn't update the test file yet. If I ran the circuit with a `Fock()` state parameter of 7, the circuit returns nan values for the weights (hence can't even run without a measurement), and if I put more, the fidelity of vacuum is bigger than one. This doesn't seem to be related to these modifications, but I thought I would mention it before creating an automatic test for the measurement.

That being said, the `measure_fock` method still remains and I'm not sure on how to approach updating the state (and even if the measurement is done properly). Do you have a similar piece theory that would help achieve this one?

nquesada:
Hi @JeyKelly --- The PR is looking really nice. For tests I'd suggest a few:

* Prepare the vacuum and check that when you use threshold measurements you never get a click.
* Prepare the odd cat states (with `p=1`) and also verify that you never get click.
* This somewhat trivial, but check that the shapes of the covariances and vectors of means are correct when you get a click/no click.

JeyKelly:
> 
> 
> Hi @JeyKelly --- The PR is looking really nice. For tests I'd suggest a few:
> 
>     * Prepare the vacuum and check that when you use threshold measurements you never get a click.
> 
>     * Prepare the odd cat states (with `p=1`) and also verify that you never get click.
> 
>     * This somewhat trivial, but check that the shapes of the covariances and vectors of means are correct when you get a click/no click.

@nquesada Hi, thanks for the response. Just to be sure, do you mean for automatic tests?

nquesada:
Hi @JeyKelly --- Yes, I mean automatic tests. You can start by using the backend independent tests here:

https://github.com/XanaduAI/strawberryfields/blob/master/tests/backend/test_threshold_measurement.py

To test the bosonic backend it should be as easy as adding `"bosonic:"` in this line: https://github.com/XanaduAI/strawberryfields/blob/afde2997ce04d44612232be2a354946f608c0a7f/tests/backend/test_threshold_measurement.py#L24

nquesada:
Then you could add tests specific to the bosonic backend in here https://github.com/XanaduAI/strawberryfields/blob/master/tests/bosonic_files/test_backend_bosoniccircuit.py 

This could be the ones I described before involving cat states and the ones checking the shapes of the updated coefficients, vectors of means and covariance matrices are correct.

JeyKelly:
> To test the bosonic backend it should be as easy as adding `"bosonic:"` in this line:
> 
> https://github.com/XanaduAI/strawberryfields/blob/afde2997ce04d44612232be2a354946f608c0a7f/tests/backend/test_threshold_measurement.py#L24

Hi Again @nquesada, by using the test `test_two_mode_squeezed_measurements` in this file, it made me realize that running the circuit with two identical squeezed state (on two uninteracting modes) and measuring them gives me a different result. It seems that the `self.fidelity_vacuum()` is the culprit. Multiple calls to the function in one program increments its value when the state gets updated post-measurement (namely the self.weights) on uninteracting modes, when there is "no click" (measurements decrease the weights so it doesn't seem to run into problems). I'm not sure if the problem lies in the use of it in the code I submitted here or if the function itself has a flaw, but this prevents running the test because the vacuum fidelity goes above one in this case. Like I've mentioned previously, the fidelity is also problematic (above one) when calculated on a Fock state of more than 7 photons, even on a clean version of strawberryfields.

nquesada:
Hi @JeyKelly --- That is because two-mode squeezing is different from applying single-mode squeezing in two modes. As for your second question, you are correct that Fock states have a number of numerical issues. For the moment I suggest you set them aside (both in terms of preparing them or measuring them) and focus in threshold detection. Let's worry about Fock measurements in a future PR.

JeyKelly:
> 
> 
> Hi @JeyKelly --- That is because two-mode squeezing is different from applying single-mode squeezing in two modes.

Yes, but I ran into fidelity problems by running the single-mode squeezing in two modes while trying to see where the problem was for two-mode squeezing (sorry, it might have been superfluous to mention it). Vacuum fidelity of the remaining mode increases (above 1) when there is a "no click" event in the first single-mode squeezed state.

nquesada:
Can you share a minimal non-working example of the issue you are finding with the two single-mode squeezed states @JeyKelly ?

JeyKelly:
> 
> 
> Can you share a minimal non-working example of the issue you are finding with the two single-mode squeezed states @JeyKelly ?

```
import strawberryfields as sf

hbar = 2
nmodes = 2

prog_cat_fock = sf.Program(nmodes)
with prog_cat_fock.context as q:
    sf.ops.Catstate(alpha=0.1)   | q[0]    
    sf.ops.Catstate(alpha=0.1)   | q[1]         
    sf.ops.MeasureThreshold()    | q[0]   
    sf.ops.MeasureThreshold()    | q[1]   
 
eng = sf.Engine("bosonic", backend_options={"hbar": hbar})
result = eng.run(prog_cat_fock)
```

This will do the trick, or any combination of state close to unity vacuum fidelity (for example, by replacing both states with `sf.ops.Squeezed(r=0.1)`). When the weights are adjusted [here](https://github.com/XanaduAI/strawberryfields/pull/604/commits/8932e563cf9582b3d4a321520ad09d484772a66f#diff-ae001ddb89eb23c5f8c0248d18ab080ffdea5c9f32f3ef4ed4328d6cbf7c241cR856) post "no-click" with the `c' = c/p0` line, `vacuum_fidelity` exceeds 1 when calculated for the `q[1]` mode.

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -803,6 +804,58 @@ def measure_dyne(self, covmat, modes, shots=1):
...
+                haf_i = hafnian_sample_state(reduced_cov_i, shots)
+                measurements[i] = haf_i[0][0] 
+            else:
+                haf_i = hafnian_sample_state(reduced_cov_i, shots, mean=reduced_mean_i)
+                measurements[i] = haf_i[0][0] 
+        index = np.array([np.random.choice(list(range(len(measurements))),1,p=np.abs(weights))])
+        samples = measurements[index] 
+        return samples
+    
+    def measure_threshold(self, modes, shots=1):
```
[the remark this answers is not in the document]
JeyKelly: I just checked and for the example I shared the `modes` array is of size 1, with the correct mode in it.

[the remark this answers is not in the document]
nquesada: sure, but you are still checking the fidelity of the whole state to the multimode vacuum state. You want the fidelity to the single mode vacuum in the mode that is being measured.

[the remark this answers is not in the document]
JeyKelly: @nquesada If I run `vacuum_fidelity = np.abs(self.fidelity_vacuum(modes))` instead, if a "no click" occurs (in this case force `measurement = 0` on a `sf.ops.Catstate(alpha=1.4)` circuit), the following state running through measure_threshold is now returning true to `self.is_vacuum()`, even if it's a `sf.ops.Catstate(alpha=1.4)` as well, with the circuit I sent. This is initially why I didn't pass the parameter because it seems to be the case for every circuit that getting a "no click" causes the next mode to be identified as vacuum. It really seems like adjusting the weight is what is causing issues - if I have two identical catstates and I calculate `vacuum_fidelity`, and multiply `self.weights` by its inverse, and calculate the `vacuum_fidelity` again, it gives 1.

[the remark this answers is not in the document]
JeyKelly: @nquesada I also ran into another problem when trying to run the tests - `chop_in_blocks_vector_multi` in ~`\strawberryfields\strawberryfields\backends\bosonicbackend\ops.py` has issues taking a single-mode circuit. I think it should be adjusted to handle the case where `id_to_keep = []`, similarly to how `chop_in_blocks_multi` returns an empty array for A.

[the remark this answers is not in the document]
nquesada: Thanks for finding this interesting edge case @jeykelly . There are two possible solutions: one would be two "fix" the `chop_in_blocks_vector_multi` functions the other is to us an if statement and check for the special cases where the total number of modes is 1. In this case you just need to decide whether to click or not to click and then regardless of the outcomes set the covs to be a single identity matrix, the means to be a single vector of zeros and the coefficients to be a single coefficient with value 1. I suggest you implement the second option, but first make sure that you do pass the argument `modes` to the vacuum fidelity method. That way you don't need the first `if` statement in your `measure_threshold` method.

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -794,10 +795,24 @@ def thermal_loss(self, T, nbar, mode):
...
     def measure_threshold(self, modes, shots=1, select=None, **kwargs):
-        raise NotImplementedError("Bosonic backend does not yet support threshold measurements.")
+        if select is not None:
+            raise NotImplementedError("Bosonic backend currently does not support " "postselection")
+        if shots != 1:
+            raise NotImplementedError(
+                "Bosonic backend currently does not support " "multiple shots"
+            )
+        samples = self.circuit.measure_threshold(modes, shots)
+        return samples
```
nquesada, line 815:
```suggestion
        raise NotImplementedError("Bosonic backend does not yet support threshold measurements.")
```

strawberryfields/backends/bosonicbackend/bosoniccircuit.py
```diff
@@ -803,6 +803,66 @@ def measure_dyne(self, covmat, modes, shots=1):
 
         return vals
 
+    def measure_fock(self, modes, shots=1):
+        r"""Performs photon number measurement on the given modes"""
+
+        raise NotImplementedError("Bosonic backend does not yet support Fock measurements.")
+
```
nquesada, line 810:
```suggestion
```

strawberryfields/backends/bosonicbackend/backend.py
```diff
@@ -794,7 +795,14 @@ def thermal_loss(self, T, nbar, mode):
...
     def measure_fock(self, modes, shots=1, select=None, **kwargs):
-        raise NotImplementedError("Bosonic backend does not yet support Fock measurements.")
+        if select is not None:
+            raise NotImplementedError("Bosonic backend currently does not support " "postselection")
+        if shots != 1:
+            raise NotImplementedError(
+                "Bosonic backend currently does not support " "multiple shots"
+            )
+        samples = self.circuit.measure_fock(modes, shots)
+        return samples
```
nquesada, line 805:
```suggestion
        raise NotImplementedError("Bosonic backend does not yet support Fock measurements.")
```

nquesada:
@elib20 : I take the point that I should be equivalent to heterodyne post selected on the \alpha = 0 outcome. But this implementation is still correct. I guess it is hard to see how one is equivalent to the other since heterodyne calls generaldyne. Is your point that more tests should be added? Or that I should use reuse the functionality from heterodyne for the no click case here?

PR #606, opened 2021-07-09
[WIP] Add a specific symplectic optimizer for gaussian gate

sylviemonet:

**Context:** Add a specific symplectic optimizer for gaussian gate so that one can update the symplectic matrix S along its Riemannian manifold.

**Description of the Change:** Add an optimizer inherit the optimizer_v2 of TensorFlow inside TF backend. Work in progress with @ziofil.

**Benefits:**

**Possible Drawbacks:** Need to warn the user to use it...

**Related GitHub Issues:**


sylviemonet:
The optimizer is simply a function so that I put it in the last commit #599 directly.

PR #614, opened 2021-07-22
Generaldyne sampling in gaussian

elib20:
**Context:** Generaldyne sampling in the gaussian backend

**Description of the Change:** Fixes the formula used for sampling generaldyne outcomes in the gaussian backend.

The covariance matrix of the sampling distribution needed also to depend on the covariance of the state onto which the target state was begin measured.

Updated the test in test_heterodyne to account for this.

**Benefits:** Correct heterodyne sampling (homodyne unaffected).

**Possible Drawbacks:** None.

**Related GitHub Issues:**


nquesada (approved):
Neat. Thanks @elib20 !

Issue #610, opened 2021-07-16
Native support in the Fock backends for the MZgate


nquesada:
The `MZgate` is the natural gate to decompose an arbitrary linear optical interferometer into single- and two-mode passive gates. Currently, interferometers are decomposed into `MZgate`s and `Rgate`s and then internally the `MZgate` is decomposed into two `BSgate`s and two more rotation gates, leading to an increase by a factor of two in terms of the number of two mode gates needed to decompose and apply a given interferometer, making the process potentially less than optimal.

With the recent addition (cf. https://github.com/XanaduAI/thewalrus/pull/256) of the Fock representation of the `MZgate` and its gradient in the `fock_gradients` module of the development version of The Walrus it becomes possible to apply this gate directly in the Fock backends. 

To implement these changes one needs to: 

* Import the relevant functions from The Walrus and add appropriate methods in the Fock and TensorFlow backends. For development and testing purposes this will require to install the development version of The Walrus.
* Exactly like the `BSgate`, the `MZgate` has selection rules that can be exploited to update the relevant tensor representing a pure or mixed quantum state. This rules have already been implemented for the `BSgate` and are furthermore described in Eqs. 78 and 79 of this reference: https://quantum-journal.org/papers/q-2020-11-30-366/pdf/
* Change the method/functions from `MZgate` to make sure this gate is no longer decomposed but it is instead applied natively to the qumodes.

It would also be nice to have some benchmarking for a few modes studying any possible efficiency benefits when this gate is implemented natively.

PR #617, opened 2021-08-09
610 mzgate add native support

sduquemesa:
**Context:** The MZgate is the natural gate to decompose an arbitrary linear optical interferometer into single- and two-mode passive gates. Currently, interferometers are decomposed into MZgates and Rgates and then internally the MZgate is decomposed into two BSgates and two more rotation gates, leading to an increase by a factor of two in terms of the number of two mode gates needed to decompose and apply a given interferometer, making the process potentially less than optimal.

**Description of the Change:** This change implements the MZgate and its gradients natively on the Fock backends using the The Warus implementation (cf. XanaduAI/thewalrus#256).

**Benefits:** Reduces the number of two-mode gates needed to apply the given interferometer.

**Possible Drawbacks:** N/A

**Related GitHub Issues:** Native support in the Fock backends for the MZgate [#610](https://github.com/XanaduAI/strawberryfields/issues/610)


sduquemesa:
This PR should only be merged after a release of The Walrus is made containing the requirements (MZgate and its gradients).

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -328,6 +329,20 @@ def beamsplitter(theta, phi, trunc):
     return BS_tw.transpose((0, 2, 1, 3))
 
 
+@functools.lru_cache()
```
ziofil, line 332: Does it make sense to use `lru_cache` with `float` inputs? Will the cache ever get successful hits?

tests/frontend/test_program.py
```diff
@@ -697,6 +699,38 @@ class DummyCircuit(Compiler):
...
+        prog = sf.Program(3)
+        U = np.array([[0, 1], [1, 0]])
+        with prog.context as q:
+            ops.MZgate(0.6,0.7) | [q[0], q[1]]
+            ops.Interferometer(U) | [q[0], q[1]]
+
+        new_prog = prog.compile(compiler=DummyCircuit())
+
+        # check compiled program now has 5 gates
+        # the S2gate should decompose into two BS and two Sgates
```
ziofil, line 723: you may want to fix this comment (there are no `S2gate`s here)

ziofil (commented):
Thank you for the new gate : )

Could you add some tests to check that the implementation is correct? (unless I've missed them, there are only tests for the decomposition).

strawberryfields/backends/fockbackend/ops.py
sduquemesa: Unless the parameters are hashed with some sort of rounding this does not make sense indeed. 
However, I see that on the same file some other functions accepting `float` params have the annotation.

sduquemesa:
I've added a test file to check if the native implementation using the gate from The Walrus agrees with the gate decomposition.

strawberryfields/backends/fockbackend/circuit.py
```diff
@@ -335,6 +343,37 @@ def _apply_BS(mat, state, trunc):  # pragma: no cover
...
+
+        .. math::
+            c'_{i, j} = \sum\limits_{k = \text{max}(1+i+j-n, 0)}
+                    ^{\text{min}(i+j, n-1) + 1}
+            B_{ik}^{j, i+j-k} c_{k, i+j-k}
+
+        where :math:`c_{kl}` is the tensor representing the two mode state and
+        :math:`n` is the cutoff in Fock space. Similarly for a mixed state it is
+        possible to do the update of the state using two less summations.
+        """
```
nquesada, line 369: wondering if this function is needed since it does exactly the same as the `applyBS` one. Maybe you can delete this one, rename the previous one `apply_two_mode_passive` and update the functions calls in BS and MZ

strawberryfields/backends/fockbackend/ops.py
nquesada: @ziofil : keep in mind the number of times that you can potentially call a 50/50 beamsplitter, `BSgate(np.pi/4, 0)`

.github/CHANGELOG.md
```diff
@@ -141,6 +141,9 @@
 * Speed improvements to ``gaussian_unitary`` compiler
   [(#603)](https://github.com/XanaduAI/strawberryfields/pull/603)
 
+* Added native support in the Fock backends for the MZgate
```
josh146, line 144:
```suggestion
* Added native support in the Fock backends for the MZgate.
```

strawberryfields/backends/fockbackend/ops.py
```diff
@@ -185,7 +186,7 @@ def project_reset(modes, x, state, pure, n, trunc):
 
     def intersperse(lst):
         # pylint: disable=missing-docstring
-        return tuple([lst[i // 2] for i in range(len(lst) * 2)])
+        return tuple(lst[i // 2] for i in range(len(lst) * 2))
```
josh146, line 189: nice catch!

josh146 (commented):
Thanks for this contribution @sduquemesa!

Regarding the documentation check, in order to get this passing, you will need to also update the Walrus link in `doc/requirements.txt`.

nquesada:
@sduquemesa : It would be a nice if you could a simple benchmark of this new implementation. For example, check how long it takes to propagate the evolution of 4 (or any other number) of photon going into an interferometer with unitary drawn from the Haar measure. You could check how long it takes with your implementation and with the old one.

sduquemesa:
> Regarding the documentation check, in order to get this passing, you will need to also update the Walrus link in `doc/requirements.txt`.

@josh146  I have updated the requirements but doc builds keep failing

josh146:
@sduquemesa it looks like you have updated `requirements.txt`, but `doc/requirements.txt` (this is the file used to install the documentation requirements) will also need to be updated 🙂

sduquemesa:
> @sduquemesa it looks like you have updated `requirements.txt`, but `doc/requirements.txt` (this is the file used to install the documentation requirements) will also need to be updated 🙂

@josh146 🤔 I did, [here](https://github.com/sduquemesa/strawberryfields/blob/ce8042371eabb81fa802533495dd7a4b62288b03/doc/requirements.txt#L17) is the file on this branch and the change is already committed. Works properly on my local env. 
Apparently is a git issue... github did not have a good day yesterday,

doc/requirements.txt
```diff
@@ -14,5 +14,5 @@ sphinx-copybutton
 sphinx-automodapi
 sphinxcontrib-bibtex==0.4.2
 tensorflow==2.5.0
-thewalrus>=0.15.0
-toml
+http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/thewalrus-0.16.0.dev0+20210730000214-cp38-cp38-manylinux2010_x86_64.whl#sha256=5eb17cd4530605ac05c96ba2033b6b413d97420e054d16182344128dfa6734f2
```
nquesada, line 17:
```suggestion
thewalrus>=0.16.0
```

requirements.txt
```diff
@@ -6,7 +6,7 @@ tensorboard>=2.0
 networkx>=2.0
 git+https://github.com/XanaduAI/blackbird@master#egg=quantum-blackbird
 python-dateutil==2.8.0
-thewalrus>=0.15.0
+http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/thewalrus-0.16.0.dev0+20210730000214-cp38-cp38-manylinux2010_x86_64.whl#sha256=5eb17cd4530605ac05c96ba2033b6b413d97420e054d16182344128dfa6734f2
```
nquesada, line 9:
```suggestion
thewalrus>=0.16.0
```

nquesada:
@sduquemesa : seems like this lines of your code are not being tested:

image

Can you check what is going on?

sduquemesa:
> @sduquemesa : seems like these lines of your PR are not being tested:
> 
> image
> 
> Can you add a new test?

I'm stuck with an issue that you maybe already faced with other custom gradients implementation: whenever tensorflow tries to call the custom gradient (from backprop) it fails silently and then returning `None`. 

Like decomposition/native-gates, do I need to register this custom gradient somewhere?

strawberryfields/backends/base.py
```diff
@@ -98,7 +98,7 @@ def delete(self, modes):
         if self.valid(modes):
             new_map = []
             ctr = 0
-            for m in range(len(self._map)):
+            for m, _ in enumerate(self._map):
```
ziofil, line 101: pretty

strawberryfields/backends/fockbackend/ops.py
ziofil: just noticed this. isn't this more readable? `tuple([x for el in lst for x in (el,el)])`

strawberryfields/backends/fockbackend/ops.py
ziofil: ok I checked and floats are hit correctly with lru_cache

ziofil (commented):
just a small readability issue

strawberryfields/backends/fockbackend/ops.py
sduquemesa: isn't that less performant as it has two loops?

ziofil: I don't think the difference is even noticeable

ziofil (approved):

sduquemesa:
image

Benchmark of native vs. decomposed MZ gate for different number of modes. 
See benchmark source code [here](https://gist.github.com/sduquemesa/cf574d250d937049a90d742e2d76e493)

PR #622, opened 2021-08-27
Switch measurements kwargs to args in `io.to_blackbird`

thisac:
**Context:**
Measurement arguments are _always_ compiled into Blackbird keyword arguments when using `io.to_blackbird`, while all other gate arguments are being stored as non-keyword arguments. This causes some issues when attempting to match templates with parameter values ([specifically for QKD](https://github.com/XanaduAI/xanadu-og-config/pull/143#event-5215091659)).

Switch measurement arguments from being compiled into keyword arguments in Blackbird to non-keyword arguments, to conform with the rest of the gate arguments.

**Benefits:**
Matching parameter values with measurement arguments works the same as when matching with gate arguments.

**Possible Drawbacks:**
If someone somewhere uses the fact that measurement arguments always are stored as keyword arguments in the resulting Blackbird program, this will no longer work (since the parameters will be found as a non-keyword argument instead).

**Related GitHub Issues:**
None

### DIFF
--- a/tests/frontend/test_io.py
+++ b/tests/frontend/test_io.py
@@ -66,7 +66,7 @@ def prog():
         # measurement
         ops.MeasureX | q[0]
         ops.MeasureHomodyne(0.43, select=0.32) | q[2]
-        ops.MeasureHomodyne(phi=0.43, select=0.32) | q[2]
+        ops.MeasureHomodyne(0.43, select=0.32) | q[2]
 
     return prog
 
@@ -87,9 +87,9 @@ Sgate(1, 0.0) | 0
 Dgate(0.735934779718964, 0.7469555733762603) | 1
 S2gate(0.543, -0.12) | [0, 3]
 Interferometer(A0) | [0, 1, 2, 3]
-MeasureHomodyne(phi=0) | 0
-MeasureHomodyne(select=0.32, phi=0.43) | 2
-MeasureHomodyne(select=0.32, phi=0.43) | 2
+MeasureHomodyne(0) | 0
+MeasureHomodyne(0.43, select=0.32) | 2
+MeasureHomodyne(0.43, select=0.32) | 2
 """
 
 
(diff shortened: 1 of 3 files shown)


tests/frontend/test_io.py
```diff
@@ -66,7 +66,7 @@ def prog():
         # measurement
         ops.MeasureX | q[0]
         ops.MeasureHomodyne(0.43, select=0.32) | q[2]
-        ops.MeasureHomodyne(phi=0.43, select=0.32) | q[2]
```
josh146, line 69: @thisac does this line no longer work?

heltluke: I'm _pretty_ sure it would be fine to leave as it was. Likely overdid it a bit in my updates to the test suite.

tests/frontend/test_io.py
```diff
@@ -87,9 +87,9 @@ def prog():
 Dgate(0.735934779718964, 0.7469555733762603) | 1
 S2gate(0.543, -0.12) | [0, 3]
 Interferometer(A0) | [0, 1, 2, 3]
-MeasureHomodyne(phi=0) | 0
-MeasureHomodyne(select=0.32, phi=0.43) | 2
-MeasureHomodyne(select=0.32, phi=0.43) | 2
+MeasureHomodyne(0) | 0
+MeasureHomodyne(0.43, select=0.32) | 2
+MeasureHomodyne(0.43, select=0.32) | 2
```
heltluke, line 92: Changing this back, on the other hand (cf. https://github.com/XanaduAI/strawberryfields/pull/622/files#r698481744), would not work.

thisac:
I think this is fine to be merged. I'm not sure why codecov is complaining about less coverage (nothing has really been removed nor added), but perhaps we could just override it (@josh146?). 

Otherwise I just need an approval, and then I can merge.

heltluke (approved):

josh146 (approved):

PR #625, opened 2021-09-16
Symbolic parameter support in TDM compiler

thisac:
**Context:**
The TDM compiler is failing when attempting to compile TDM program using device specs with symbolic template parameters in operations. This is due to the compiler seeing anything that is not a string (e.g., a value _or_ a symbolic parameter) as a hard-coded parameter, which a template parameter is not, and concludes that the values in the spec differ from the values in the program.

* Template parameters are no longer seen as a hard-coded parameter type.
* If a template parameter is used instead of a p-type parameter, it can still be used to validate the parameter values in the device spec if it's named the same there.

**Example:** The following works, with all `{s}` values being validated using the gate parameter range `'s': [[0.0, 99999999.]]`. Also works as before with non-template p-types (e.g. `p0`) using the _same name as in the device spec_.
```python
bbt = """\
name template_td3_fake
version 1.0
target TD3_fake (shots=1)
type tdm (temporal_modes=259, copies=1)

float array p0[1, 259] =
    {s}

Sgate({s}, 0.0) | 43
"""

device_spec = {
    'layout': bbt,
    'modes': 
    {
        'spatial': 1, 
        'concurrent': 44, 
        'temporal_max': 259
    }, 
    'compiler': ['TD3_fake'], 
    'gate_parameters': 
    {
        's': [[0.0, 99999999.]], 
    }
}
```

**Benefits:**
TDM and QKD programs can be be more easily run through the compiler.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/tdm/tdmprogram.py
+++ b/strawberryfields/tdm/tdmprogram.py
@@ -170,7 +170,7 @@ def reshape_samples(all_samples, modes, N, timebins):
     idx_tracker = {i: 0 for i in mode_order}
 
     # iterate backwards through all_samples and add them into the correct mode
-    new_samples = dict()
+    new_samples = {}
     timebin_idx = 0
     for i, mode in enumerate(mode_order):
         mode_idx = modes[i % len(N)]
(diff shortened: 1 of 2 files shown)


lneuhaus (approved):
I tested this and found it working as I'd expect it. Thanks @thisac

nquesada (approved):

Issue #284, opened 2020-01-06
Orbit cardinality for large number of modes


hay-k:
-------------------------------------------------------------------------------------------------------------

#### Issue description

The function `sf.apps.similarity.orbit_cardinality` does not work, when the second argument (i.e. 'modes') is larger than 170. The issue happens because of the function `scipy.special.factorial`, which, in the way it is used, calculates factorials only up to (including) 170.

Example:
```
>>> import strawberryfields as sf
>>> sf.apps.similarity.orbit_cardinality([2, 1], 171)
```

* *Expected behavior:* Output `29070`

* *Actual behavior:* 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/hayk/projects/venv/lib/python3.6/site-packages/strawberryfields/apps/similarity.py", line 278, in orbit_cardinality
    return int(factorial(modes) / np.prod(factorial(counts)))
OverflowError: cannot convert float infinity to integer
```

#### Additional information

The reason behind the issue is, that the function `scipy.special.factorial` with default second argument (i.e. `exact=False`), calculates the factorials in float variables. The factorials of numbers larger than 170 do not fit into a float, hence the function returns `inf`. On the other hand, python can handle large integers, so if we force the factorial funcion to use integers like this
```
return int(factorial(modes, exact=True) / np.prod(factorial(counts, exact=True)))
```
then the issue goes away. Here is the description of the argument 'exact', from scipy documentation: 

> exact: bool, optional
> If True, calculate the answer exactly using long integer arithmetic. If False, result is approximated in floating point rapidly using the gamma function. Default is False.

trbromley:
Thanks @hay-k for the detailed report and the suggested fix. We'll work on this!

PR #628, opened 2021-09-23
factorial for large numbers

sduquemesa:
**Context:** Factorials of numbers larger than 170 do not fit into an `int`, hence return float when needed

**Description of the Change:** use the `qarg exact=True/False` on `scipy.special.factorial` when the result cannot be fitted inside an `int`

**Benefits:** Error won't be raised when `factorial` is called with a large number

**Possible Drawbacks:** None

**Related GitHub Issues:** #284


strawberryfields/apps/similarity.py
```diff
@@ -304,7 +304,13 @@ def orbit_cardinality(orbit: list, modes: int) -> int:
...
     sample = orbit + [0] * (modes - len(orbit))
     counts = list(Counter(sample).values())
-    return int(factorial(modes) / np.prod(factorial(counts)))
+
+    # factorials of numbers larger than 170 do not fit into a int,
+    # hence return float using the qarg `exact=True`
+    if modes > 170:
+        return factorial(modes, exact=True) / np.prod(factorial(counts, exact=True))
+    else:
+        return int(factorial(modes, exact=False) / np.prod(factorial(counts, exact=False)))
```
josh146, line 313:
```suggestion

    return int(factorial(modes, exact=False) / np.prod(factorial(counts, exact=False)))
```
The `else` is not needed as the `return` will 'exit' the if condition :)

josh146 (approved):
Nice catch @sduquemesa! Don't forget to add this to the changelog before merging it in 🙂

strawberryfields/apps/similarity.py
sduquemesa: Agree. But isn't the `if/else` more readable?

josh146: To some extent yes, although I guess it comes down to personal preference! in either case, we are already semi-stuck using pylint's convention here (which I assume follows PEP8)

strawberryfields/apps/similarity.py
```diff
@@ -304,7 +304,13 @@ def orbit_cardinality(orbit: list, modes: int) -> int:
     """
     sample = orbit + [0] * (modes - len(orbit))
     counts = list(Counter(sample).values())
-    return int(factorial(modes) / np.prod(factorial(counts)))
+
+    # factorials of numbers larger than 170 do not fit into a int,
+    # hence return float using the qarg `exact=True`
+    if modes > 170:
+        return factorial(modes, exact=True) / np.prod(factorial(counts, exact=True))
```
thisac, line 311: @sduquemesa Would be good to add a test for this!

How come `exact=True` isn't used everywhere instead of having this if-statement? Is this due to it being much slower, or some other reason? 🤔

sduquemesa: @thisac, this is explained on scipy factorila's [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorial.html)
> If True, calculate the answer exactly using long integer arithmetic. If False, result is approximated in floating point rapidly using the gamma function. Default is False.

Just added a test that will cover this case! 🩺

Issue #582, opened 2021-05-18
Create a separate requirements file for the CI


antalszava:
We would like to have a separate `requirements.txt` (pinned exactly to the versions used) and `requirements-ci.txt` (pinned closely to `setup.py`). This latter would be used by the CI.

See the conversation thread at the end of #581.

sduquemesa:
Hi @antalszava, 

From your comment I understand that `requirements-ci.txt` would be used by the CI to test that strawberryfields works as expected when used as a package, hence it will contain the requirements appearing on `setup.py`. On the other hand, `requirements.txt` will have all dependencies required to used all the features of SF.
Is that correct? Could you elaborate a bit more?

When you mention _pinned closely_ and _pinned exactly_, are you referring to semantic versioning?

sduquemesa:
Pennylane has resolved the same issue and is a good reference for this: [Create a separate requirements file for the CI #1292](https://github.com/PennyLaneAI/pennylane/issues/1292)

antalszava:
Hi @sduquemesa,

> When you mention pinned closely and pinned exactly, are you referring to semantic versioning?

Yes! The idea would be to only pin only those packages to a certain version in the `requirements-ci.txt` file where there is a specific constraint (i.e., SF only works with a certain version of the dependency). The rest of the packages would stay unpinned so that when issuing `pip install -r requirements-ci.txt`, the latest version for each package is installed by the continuous integration check. 

In contrast, in the `requirements.txt` file we would like to provide the user with a working setup to use Strawberry Fields. We provide a list of dependencies with pinned versions, such that by doing `pip install -r requirements.txt`, they will get a working combination of dependencies and be able to use Strawberry Fields.

PR #629, opened 2021-09-23
Create a separate requirements file for the CI

sduquemesa:
**Context:** The requirements.txt file has packages unpinned. This is contrary to the convention of pinning packages in the requirements.txt file.

**Description of the Change:** Create a separate `requirements.txt` (pinned) and `requirements-ci.txt` (unpinned). This latter is used by the CI.

**Benefits:** CI environment will behave closely as when installing the package with `setup.py`

**Possible Drawbacks:** Version conflicts on (unpinned) dependencies potentially will only appear on CI environment.

**Related GitHub Issues:** #582


requirements-ci.txt
```diff
@@ -0,0 +1,15 @@
+numpy
+scipy
+sympy
+tensorflow
+tensorboard
+networkx
+git+https://github.com/XanaduAI/blackbird@master#egg=quantum-blackbird
```
thisac, line 7: We should be able to use the released version of Blackbird now.
```suggestion
quantum-blackbird
```

requirements.txt
```diff
@@ -1,14 +1,14 @@
 numpy>=1.19.2
-scipy
+scipy>=1.7.1
 sympy>=1.5
 tensorflow>=2.0
 tensorboard>=2.0
 networkx>=2.0
 git+https://github.com/XanaduAI/blackbird@master#egg=quantum-blackbird
```
thisac, line 7: Same here!
```suggestion
quantum-blackbird==0.4.0
```

thisac (changes requested):
Thanks @sduquemesa. I'd think we should strictly pin all dependencies in the `requirements.txt` file to latest releases (e.g. `numpy==1.21` instead of `numpy>=1.19.2`), unless already pinned (then keeping them is probably good).

I'm not sure about best practice when it comes to completely unpinning `requirements-ci.txt` (i.e. not even having a minimum version). This seems to be the case in PL, though.

sduquemesa:
@thisac, following the discussion on issues #582, this is the reply from @antalszava:

> Yes! The idea would be to only pin only those packages to a certain version in the requirements-ci.txt file where there is a specific constraint (i.e., SF only works with a certain version of the dependency). The rest of the packages would stay unpinned so that when issuing pip install -r requirements-ci.txt, the latest version for each package is installed by the continuous integration check.

In contrast, in the requirements.txt file we would like to provide the user with a working setup to use Strawberry Fields. We provide a list of dependencies with pinned versions, such that by doing pip install -r requirements.txt, they will get a working combination of dependencies and be able to use Strawberry Fields.

thisac:
Thanks @sduquemesa! We should update the recently merged `.github/workflows/upload.yml` file to use `pip install -r requirements-ci.txt` as well, so that we're testing with the same package versions for release. Looks good otherwise! 💯

requirements.txt
```diff
@@ -1,15 +1,16 @@
...
 quantum-blackbird==0.4.0
 python-dateutil==2.8.0
-thewalrus>=0.16.0
-toml
-appdirs
-numba>=0.48.0
+thewalrus==0.16.2
+toml==0.10.2
+appdirs==1.4.4
+numba
```
thisac, line 13: How come `numba` isn't pinned anymore?

.github/workflows/upload.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - name: Install test dependencies
         run: |
-          pip install -r requirements.txt
+          pip install -r requirements-ci.txt
```
thisac, line 25: 🎉

thisac (approved):
Thanks for pinging me on this @sduquemesa!

requirements.txt
sduquemesa: Sorry, my bad 😖 Fixed now!

PR #631, opened 2021-09-23
Fix workflow for uploading to PyPI

thisac:


josh146 (approved):
Oops 🙈 nice catch @thisac. Hopefully this works for the next release!

PR #633, opened 2021-10-05
Add support for npz results

lneuhaus:
**Context:**
Some devices on the Xanadu Quantum Cloud need to return multiple arrays of data with different shapes or data types. Furthermore, sometimes the size of a job result is a limiting factor due to limited bandwidth. This PR adds support for compressed result data, and supports returning dictionaries of result arrays instead of a single array per job.

With this change, job results can be serialized using `numpy.savez` or `numpy.savez_copressed`. In this format, a dictionary of numpy arrays can be transferred, with compression disabled or enabled.

**Benefits:**
Cloud devices can now return dictionaries of numpy arrays, and comporession can be used to reduce the data payload.

**Possible Drawbacks:**
For single result arrays, it is not clear whether the client supports compression or not. 

~**Related GitHub Issues:**~

### DIFF
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -293,8 +293,7 @@ class Connection:
                 samples = int_to_int64(samples)
 
             # The npz format can be used to compress data. In this case samples is
-            # a dict at this point, but the user might expect a single array. For
-            # this case we use the special key ``single_samples_array``.
+            # a dict with one entry but the user might expect a single array.
             if isinstance(samples, dict) and len(samples) == 1:
                 samples = next(iter(samples.values()))
 
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -288,7 +288,7 @@ class Connection:
                 return samples
 
             if isinstance(samples, dict):
-                samples = {key: int_to_int64(value) for key, value in samples}
+                samples = {key: int_to_int64(value) for key, value in samples.items()}
             else:
                 samples = int_to_int64(samples)
 
--- a/strawberryfields/api/connection.py
+++ b/strawberryfields/api/connection.py
@@ -282,20 +282,25 @@ class Connection:
             # Samples represent photon numbers.
             # Convert to int64, to avoid unexpected behaviour
             # when users postprocess these samples.
-            def int_to_int64(samples): 
+            def int_to_int64(samples):
                 if np.issubdtype(samples.dtype, np.integer):
                     return samples.astype(np.int64)
                 else:
                     return samples
+
             if isinstance(samples, dict):
                 samples = {key: int_to_int64(value) for key, value in samples}
             else:
                 samples = int_to_int64(samples)
 
-            # The npz format can be used to compress data. In this case samples is 
-            # a dict at this point, but the user might expect a single array. For 
+            # The npz format can be used to compress data. In this case samples is
+            # a dict at this point, but the user might expect a single array. For
             # this case we use the special key ``single_samples_array``.
-            if isinstance(samples, dict) and "_single_samples_array" in samples and len(samples) == 1:
+            if (
+                isinstance(samples, dict)
+                and "_single_samples_array" in samples
+                and len(samples) == 1
+            ):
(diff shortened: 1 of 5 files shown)


lneuhaus:
Any action needed for fixing CodeFactor or Codecov failures?

heltluke (commented):
Looks good to me!

strawberryfields/api/connection.py
```diff
@@ -276,12 +276,31 @@ def get_job_result(self, job_id: str) -> Result:
...
+
+            # The npz format can be used to compress data. In this case samples is
+            # a dict at this point, but the user might expect a single array. For
+            # this case we use the special key ``single_samples_array``.
+            if (
+                isinstance(samples, dict)
+                and "_single_samples_array" in samples
+                and len(samples) == 1
+            ):
+                samples = samples["_single_samples_array"]
```
lneuhaus, line 303: this choice of convention is arbitrary, and maybe not the best one. If we went with this kind of pattern to transport single arrays in npz format, where would we best document it?

Mandrenkov: Are you sure we need to establish such a convention? If we know that each SF job produces exactly one output (i.e., an array of samples), would it not be more appropriate to use
```python
samples = next(samples.values())
```
in the event where `samples` is a dictionary?  We can raise an exception if the dictionary has more than one value.

tests/api/test_connection.py
```diff
@@ -254,6 +254,86 @@ def test_get_job_result(self, connection, result_dtype, monkeypatch):
...
+            np.int32,
+            np.int64,
+            np.uint8,
+            np.uint16,
+            np.uint32,
+            np.uint64,
+            np.float32,
+            np.float64,
+        ],
+    )
```
Mandrenkov, line 271: Should we also test complex number types?

Same goes for `test_get_job_result_dict_single_samples_array()` below.

tests/api/test_result.py
```diff
@@ -78,3 +78,18 @@ def test_unkown_shape_print(self, stateful, capfd):
...
         assert f"contains state={stateful}" in out
+
+    @pytest.mark.parametrize("stateful", [True, False])
+    def test_dict_print(self, stateful, capfd):
+        """Test that printing a result object with a dict of samples
+        provides the correct output."""
+        samples = {"ones": np.ones((2, 3, 4, 5)), "twos": 2 * np.ones((2, 3, 4, 5))}
+        result = Result(samples, is_stateful=stateful)
+        print(result)
+        out, err = capfd.readouterr()
```
Mandrenkov, line 89: I'm surprised the convention is not to use
```python
out = repr(result)
```
and assert
```python
assert out == f"<Result: keys={{"ones", "twos"}}, contains state={stateful}>"
```

Mandrenkov (commented):
Looks good although I'm curious about the use case for multiple job results.

strawberryfields/api/connection.py
```diff
@@ -276,12 +276,31 @@ def get_job_result(self, job_id: str) -> Result:
...
                 if np.issubdtype(samples.dtype, np.integer):
-                    # Samples represent photon numbers.
-                    # Convert to int64, to avoid unexpected behaviour
-                    # when users postprocess these samples.
-                    samples = samples.astype(np.int64)
+                    return samples.astype(np.int64)
+                return samples
+
+            if isinstance(samples, dict):
+                samples = {key: int_to_int64(value) for key, value in samples.items()}
```
brownj85, line 291: What would these keys be if returned from a QKD device?

strawberryfields/api/connection.py
brownj85: what's wrong with "samples"? would this conflict with the keys used by QKD results?

strawberryfields/api/connection.py
heltluke: Haven't yet decided exactly what to call them, but we'd like to return two sets of numbers to users---the measured "results" (floats), and the implemented values that will likely differ from the requested values due to scaling and hardware limitations (complex floats).

PR #536, opened 2021-01-27
Add strict markers option for pytest

antalszava:
**Context:**
TBD
**Description of the Change:**
TBD
**Benefits:**
TBD
**Possible Drawbacks:**
TBD
**Related GitHub Issues:**
TBD

Raises the following error if no custom marker was specified for a test case:
```python
Strawberry Fields marker specified for the test function:  <Function test_chop_in_blocks_multi[1]>
collected 8095 items
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/anaconda3/lib/python3.8/site-packages/_pytest/main.py", line 269, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
(...)
INTERNALERROR>     item.add_marker('dummy_strawberry_fields_marker')
INTERNALERROR>   File "/anaconda3/lib/python3.8/site-packages/_pytest/nodes.py", line 274, in add_marker
INTERNALERROR>     marker_ = getattr(MARK_GEN, marker)
INTERNALERROR>   File "/anaconda3/lib/python3.8/site-packages/_pytest/mark/structures.py", line 500, in __getattr__
INTERNALERROR>     fail(
INTERNALERROR>   File "/anaconda3/lib/python3.8/site-packages/_pytest/outcomes.py", line 153, in fail
INTERNALERROR>     raise Failed(msg=msg, pytrace=pytrace)
INTERNALERROR> Failed: 'dummy_strawberry_fields_marker' not found in `markers` configuration option
```

PR #441, opened 2020-08-18
Making Catstate take real arguments

antalszava:
**Context:**

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #639, opened 2021-11-02
Remove TensorFlow dependency from frontend tests

Mandrenkov:
**Context:**
Currently, running `make test-frontend` fails if TensorFlow is not installed.

- Surrounded TensorFlow import in [test_states_probabilities.py](https://github.com/XanaduAI/strawberryfields/blob/master/tests/backend/test_states_probabilities.py) with `try` block.
- Removed `TestBackendImport` from [test_tf_import.py](https://github.com/XanaduAI/strawberryfields/blob/master/tests/backend/test_tf_import.py) since it is largely a duplicate of `TestFrontendImport`.
- Fixed `test_incorrect_tf_version` unit test in [test_tf_import.py](https://github.com/XanaduAI/strawberryfields/blob/master/tests/backend/test_tf_import.py) in the event where TF is not installed.
- Safeguarded TensorFlow import in [test_engine_integration.py](https://github.com/XanaduAI/strawberryfields/blob/master/tests/integration/test_engine_integration.py) and [tfbackend/\_\_init__.py](https://github.com/XanaduAI/strawberryfields/blob/master/strawberryfields/backends/tfbackend/__init__.py).
- Fixed obsolete `mock` import in [test_utils_integration.py](https://github.com/XanaduAI/strawberryfields/blob/master/tests/integration/test_utils_integration.py).

**Benefits:**
- TensorFlow is no longer a hard dependency for running the frontend tests.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.


strawberryfields/backends/tfbackend/__init__.py
```diff
@@ -139,7 +139,6 @@
 
 """
 import sys
-from .backend import TFBackend
```
josh146, line 142: Wait, when did this get moved up here? 🤔 

I'm looking at an older version of the file I have (locally), and it looks like this:

image

Should we revert it to how it used to be?

josh146: This should allow the old `TestBackendImport` you removed below to keep working?

josh146 (approved):
Thanks @Mandrenkov, this is great! I only have one comment regarding `tfbackend/__init__.py`, but it is relatively minor and not needed for approval.

I'm curious when that changed, however --- it might mean users using SF purely for XQC access would have received error messages for not having TF installed 🤔

strawberryfields/backends/tfbackend/__init__.py
Mandrenkov: > Wait, when did this get moved up here? 🤔

It looks like this change was introduced by d8ac13e in #441.

> Should we revert it to how it used to be?

Sure, let's do it! 

> This should allow the old `TestBackendImport` you removed below to keep working?

Actually, I decided to remove `TestBackendImport` because it didn't have a mark (and therefore was never actually run) and only tested a subset of the code path of `TestFrontendImport` anyway.

josh146: > Actually, I decided to remove TestBackendImport because it didn't have a mark (and therefore was never actually run) and only tested a subset of the code path of TestFrontendImport anyway.

🤦 This has come up before that the SF test suite is set up so that tests _have_ to be marked to be run.

.github/CHANGELOG.md
```diff
@@ -6,12 +6,17 @@
 
 <h3>Bug fixes</h3>
 
+* The `frontend` tests can now be run without having TensorFlow installed.
+  [(#639)](https://github.com/XanaduAI/strawberryfields/pull/639)
```
antalszava, line 10: Minor: not sure if this is a bug fix per se that we'd like to include in the changelog to notify users. :thinking:

If it affects the way TF is imported in `strawberryfields/backends/tfbackend/__init__.py`, then it is relevant. :ok_hand:

strawberryfields/backends/tfbackend/__init__.py
antalszava: The purpose of #536 would be to ensure that this doesn't happen (though it still needs work).

antalszava (approved):
Looks good to me! :100: Thank you for fixing this. :slightly_smiling_face:

strawberryfields/backends/tfbackend/__init__.py
josh146: Nice! @sduquemesa @antalszava I suddenly realized that #441 is not listed in the changelog, we should definitely add it under 'breaking changes'!

thisac: Oh, missed that this was moved to the top as well. 👍 for the added comment.

thisac (approved):
Thanks for fixing this @Mandrenkov. It's an appreciated change! 💯

strawberryfields/backends/tfbackend/__init__.py
antalszava: Good catch, thank you :) [opened a PR](https://github.com/XanaduAI/strawberryfields/pull/640)

.github/CHANGELOG.md
Mandrenkov: Hmm, good call! I've removed the note from the breaking changes.

Mandrenkov:
Thank you very much for the reviews and feedback, @josh146, @antalszava, and @thisac! 🚀

PR #642, opened 2021-11-03
Replace SF CLI with XCC CLI

Mandrenkov:
**Context:**
Moving forward, the CLI offered by Strawberry Fields (SF) will be replaced by the [Xanadu Cloud Client (XCC)](https://github.com/XanaduAI/xanadu-cloud-client) CLI.

* Removed the CLI package, tests, entry point, and Sphinx documentation.
* Updated the *Hardware and cloud* tutorial to use the XCC CLI.
* Added a *Migration guides* page to the Sphinx documentation with a CLI section under **Version 0.20.0**.

**Benefits:**
* There is no need to maintain a separate CLI for Strawberry Fields.
* Users only need to set up their Xanadu Cloud credentials once (and use them in several different projects).

**Possible Drawbacks:**
* There may be some initial confusion when users discover the `sf` entry point is missing.

**Related GitHub Issues:**
None.


doc/development/migration_guides.rst
```diff
@@ -0,0 +1,27 @@
...
+   :header-rows: 1
+
+   * - **Strawberry Fields v0.19.0**
+     - **Strawberry Fields v0.20.0**
+   * - ``sf configure --token "foo"``
+     - ``xcc config set REFRESH_TOKEN "foo"``
+   * - ``sf --ping``
+     - ``xcc ping``
+   * - ``sf run "foo.xbb"``
+     - ``xcc job submit --name "bar" --target "X8_01" --circuit "$(cat foo.xbb)"``
```
josh146, line 27: This might be worth duplicating in the changelog 🙂 

In terms of communicating breaking changes, I'm very much a fan of 'overcommunication'. Especially since the changelog doesn't need to be maintained!

josh146: (table renders really nicely btw)

josh146: A sudden minor thought, but `--circuit "$(cat foo.xbb)"` would be very unix specific -- what should windows users using `xcc` do here?

josh146 (approved):
Excited for this change!

doc/development/migration_guides.rst
Mandrenkov: Thanks for pointing all of this out, @josh146 and @thisac!

> I wonder if this might be worth duplicating in the changelog 🙂

Done. Unfortunately, the MD → RST converter does not like tables so I've resorted to a list format instead.

> A sudden minor thought, but `--circuit "$(cat foo.xbb)"` would be very unix specific -- what should windows users using `xcc` do here?

Ouch, I hadn't considered that! It turns out that you can run the following in place of `cat foo.xbb` in PowerShell:
```powershell
Get-Content foo.xbb -Raw
```

doc/introduction/photonic_hardware.rst
```diff
@@ -1,73 +1,34 @@
 Hardware and cloud
 ==================
 
-Using Strawberry Fields, you can submit quantum programs to be executed on photonic
-hardware or to be run on a cloud simulator via the Xanadu Quantum Cloud.
+Using Strawberry Fields, you can submit quantum programs to be executed on
+photonic hardware or to be run on a cloud simulator via the Xanadu Cloud.
```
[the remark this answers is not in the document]
Mandrenkov: Indeed, Xanadu Cloud (XC) is the preferred branding now! The latter shouldn't be publicly used anymore.

Mandrenkov:
Let me know what you think of the new release notes, @josh146 or @thisac. The rest may follow a similar format.

doc/development/migration_guides.rst
josh146: > Ouch, I hadn't considered that! It turns out that you can run the following in place of cat foo.xbb in PowerShell:

I'm not sure if submitting files from the terminal is a common workflow, but if it is, it might make sense to include the file read as part of the `xcc` library, since it can then be written in Python/be platform independent?

If it is not a common workflow, we could probably just remove it from the docs altogether.

.github/CHANGELOG.md
```diff
@@ -75,6 +75,83 @@
...
+          "finished_at": null,
+          "running_time": null
+      }
+      $ xcc job get 0b0f5a46-46d8-4157-8005-45a4764361ba --result
+      {
+          "output": [
+              "[[0 0 0 0]\n[0 0 0 0]\n[0 0 0 0]\n[0 0 0 0]]"
+          ]
+      }
+      ```
```
josh146, line 152: 👨‍🍳

doc/introduction/photonic_hardware.rst
```diff
@@ -1,73 +1,34 @@
...
-
-      The above configuration file need only be created once. Once created,
-      Strawberry Fields will automatically load the stored authentication
-      credentials.
-
-* **Environment variables**. Set the ``SF_API_AUTHENTICATION_TOKEN`` environment variable
-  with your API token.
+Before using the Xanadu Cloud platform, you will need to configure your
+credentials using the `Xanadu Cloud Client
+<https://github.com/XanaduAI/xanadu-cloud-client#setup>`__. The simplest way
```
josh146, line 18: good idea to just link to the XCC docs!

josh146 (approved):
New changelog looks good from my end @Mandrenkov!

doc/introduction/photonic_hardware.rst
Mandrenkov: Thanks!

doc/development/migration_guides.rst
Mandrenkov: > I'm not sure if submitting files from the terminal is a common workflow, but if it is, it might make sense to include the file read as part of the `xcc` library, since it can then be written in Python/be platform independent?

Yeah, in hindsight, I think you're right. We should implement this feature in the following XCC release.

.github/CHANGELOG.md
```diff
@@ -76,9 +76,82 @@
...
+          "created_at": "2021-11-02 03:04:05,10",
+          "finished_at": null,
+          "running_time": null
+      }
+      $ xcc job get 0b0f5a46-46d8-4157-8005-45a4764361ba --result
+      {
+          "output": [
+              "[[0 0 0 0]\n[0 0 0 0]\n[0 0 0 0]\n[0 0 0 0]]"
+          ]
+      }
```
thisac, line 151: Neat! 💯

doc/development/migration_guides.rst
```diff
@@ -24,4 +24,8 @@ each ``sf`` (v0.19.0) command:
    * - ``sf --ping``
      - ``xcc ping``
    * - ``sf run "foo.xbb"``
-     - ``xcc job submit --name "bar" --target "X8_01" --circuit "$(cat foo.xbb)"``
\ No newline at end of file
+     - ``xcc job submit --name "bar" --target "X8_01" --circuit "$(cat foo.xbb)"``
+
+.. note::
+
+  To submit a job using Windows PowerShell, simply replace ``cat foo.xbb`` with  ``Get-Content foo.xbb -Raw``.
```
thisac, line 31: Does PowerShell also support the `$(...)` expansion syntax? I also wonder how many people tend to use PowerShell instead of cmd, or some other cmd-based alternative. 🤔

thisac (approved):
The changelog entry looks great! 💯

.github/CHANGELOG.md
Mandrenkov: Phew! 😌

doc/development/migration_guides.rst
Mandrenkov: > Does PowerShell also support the `$(...)` expansion syntax?

Yes! I'm not sure if the semantics are equivalent to Bash but it's close enough to make it work here.

> I also wonder how many people tend to use PowerShell instead of cmd, or some other cmd-based alternative. 🤔

My understanding is that Microsoft has preferred PowerShell over cmd for some time now so I think it's a safe bet to just show a PowerShell command as the Windows example.

thisac: > My understanding is that Microsoft has preferred PowerShell over cmd for some time now

Good point, I'm just not so sure most Windows users actually use Powershell. 😆

PR #644, opened 2021-11-09
Update DeviceSpec to work with XCC

thisac:
**Context:**
As part of the migration to the XCC API, the `DeviceSpec` class needs to be updated.

* The connection object is removed from the `DeviceSpec` class, so that it only contains the specification with which is was initialized. If the spec needs to be refreshed a new `DeviceSpec` object needs to be created.
* The `target` is now retrieved from the device specification rather than passed separately to the `__init__()` method. This avoids any ambiguities.

**Benefits:**
* The `DeviceSpec` class works with XCC.
* The `DeviceSpec` class only encapsulates the device specification (providing some helper methods), and does not handle any connections.

**Possible Drawbacks:**
* The target is now taken from the device specification rather than explicitly passed. Shouldn't cause any issues, but would change the primary target if they had differed. 

**Related GitHub Issues:**
None


strawberryfields/api/devicespec.py
```diff
@@ -15,62 +15,66 @@
...
     """The specifications for a specific hardware device.
 
     Args:
-        target (str): name of the target hardware device
         spec (dict): dictionary representing the raw device specification.
-            This dictionary should contain the following key-value pairs:
+            This dictionary must contain the following key-value pairs:
 
-            - layout (str): string containing the Blackbird circuit layout
+            - layout (str): string containing the Blackbird or XIR circuit layout
```
Mandrenkov, line 34: I think it would be more accurate to state
```suggestion
            - layout (str): string containing the Blackbird circuit layout or XIR manifest
```

strawberryfields/api/devicespec.py
```diff
@@ -15,62 +15,66 @@
...
+            missing_keys = {
+                "target",
+                "layout",
+                "modes",
+                "compiler",
+                "gate_parameters",
+            } - spec.keys()
+            raise ValueError(
+                f"Device specification is missing the following keys: {sorted(missing_keys)}"
+            )
```
Mandrenkov, line 51: I love the idea of the `if` condition here but it turns out to be excessive if we want to avoid duplication:
```suggestion
        missing_keys = {"target", "layout", "modes", "compiler", "gate_parameters"} - spec.keys()
        if missing_keys:
            raise ValueError(
                f"Device specification is missing the following keys: {sorted(missing_keys)}"
            )
```

strawberryfields/api/devicespec.py
```diff
@@ -15,62 +15,66 @@
...
+
         self._spec = spec
 
     @property
-    def target(self):
+    def target(self) -> str:
         """str: The name of the target hardware device."""
-        return self._target
+        return self._spec["target"]
```
Mandrenkov, line 58: Finally! 🥳

strawberryfields/api/devicespec.py
```diff
@@ -15,62 +15,66 @@
...
 
     @property
-    def modes(self):
+    def modes(self) -> int:
         """int: Number of modes supported by the device."""
         return self._spec["modes"]
 
     @property
-    def compiler(self):
+    def compiler(self) -> Sequence[str]:
```
Mandrenkov, line 71: I noticed in some of the RemoteEngine tests that `DeviceSpec.compiler` can also just be a string. Is that true?

strawberryfields/api/devicespec.py
```diff
@@ -100,25 +104,7 @@ def gate_parameters(self):
...
-        if self.layout_is_formatted():
-            return
-
-        if isinstance(program, TDMProgram):
-            self._spec["layout"] = self._spec["layout"].format(
-                target=self.target, tm=program.timebins
-            )
-        else:
-            # TODO: update when `self._spec["layout"]` is returned as an unformatted string
-            raise NotImplementedError("Formatting not required or supported for non-TDM programs.")
```
Mandrenkov, line 119: Just to confirm, these are no longer needed, right?

strawberryfields/api/devicespec.py
```diff
@@ -100,25 +104,7 @@ def gate_parameters(self):
...
-
-    def validate_parameters(self, **parameters):
+    def validate_parameters(self, **parameters: complex) -> None:
```
Mandrenkov, line 107: Can we add a `Raises` section to this docstring? Otherwise, it's not clear how errors are communicated.

strawberryfields/api/devicespec.py
```diff
@@ -136,9 +122,9 @@ def validate_parameters(self, **parameters):
             if p not in self.gate_parameters:
                 raise ValueError(f"Parameter {p} not a valid parameter for this device")
 
-    def create_program(self, **parameters):
-        """Create a Strawberry Fields program matching the low-level layout of the
-        device.
+    def create_program(self, **parameters: complex):
```
Mandrenkov, line 125: Is this right?
```suggestion
    def create_program(self, **parameters: complex) -> sf.Program:
```

tests/api/test_devicespec.py
```diff
@@ -14,21 +14,20 @@
...
-import textwrap
 
 from strawberryfields.api import DeviceSpec
 from strawberryfields.compilers import Ranges
-import strawberryfields as sf
-
-import blackbird
 
-# pylint: disable=bad-continuation,no-self-use,pointless-statement
+# pylints: disable=bad-continuation,no-self-use,pointless-statement
```
Mandrenkov, line 25: What's the difference between `# pylint` and `# pylints`?

tests/api/test_devicespec.py
```diff
@@ -128,60 +130,24 @@ def test_create_program(self, monkeypatch):
...
+        """Test that error is raised when a specification with missing entries is supplied"""
+        invalid_spec = {
+            "target": "abc",
+            "modes": 2,
+            "compiler": ["Xcov"],
+        }
+        with pytest.raises(
+            ValueError, match=r"missing the following keys: \['gate_parameters', 'layout'\]"
+        ):
+            DeviceSpec(spec=invalid_spec)
```
Mandrenkov, line 153: Nice! 💯

tests/api/test_remote_engine.py
```diff
@@ -234,7 +234,14 @@ def test_compile_device_invalid_device_error(self, prog, monkeypatch, caplog):
...
 
         # Setting compile_info with a dummy devicespec and compiler name
-        X8_spec = DeviceSpec(target="DummyDevice", connection=None, spec=None)
+        dummy_spec = {
+            "target": "DummyDevice",
+            "modes": 2,
+            "layout": None,
+            "gate_parameters": None,
+            "compiler": [None],
+        }
```
Mandrenkov, line 243: As a heads up, we might not need this after you merge `integrate-xanadu-cloud-client`.

tests/frontend/test_program.py
```diff
@@ -256,8 +263,14 @@ def test_assert_max_number_of_measurements(self, measure_op, measure_name):
...
         """Check that the correct error is raised when calling `prog.assert_number_of_measurements`
         with the incorrect type of device spec mode entry."""
-        device_dict = {"modes": 2, "layout": None, "gate_parameters": None, "compiler": [None]}
-        spec = sf.api.DeviceSpec(target="simulon_gaussian", connection=None, spec=device_dict)
+        device_dict = {
+            "target": "simulon_gaussian",
+            "modes": 2,
+            "layout": None,
+            "gate_parameters": None,
+            "compiler": [None],
```
Mandrenkov, line 271: Are you sure this is legal?
```suggestion
            "compiler": [],
```

Mandrenkov (commented):
Looks good to me! 👍

I'll approve once this PR incorporates the changes from #645.

strawberryfields/api/devicespec.py
thisac: Completely right! I added the `missing_keys` later, so I missed this.

strawberryfields/api/devicespec.py
thisac: `_spec["compiler"]` should always be a list of compiler as far as I'm aware. `_spec["default_compiler"]` is a single compiler (the first in the list). Could you point to anywhere where `_spec["compiler"]` is a string?

strawberryfields/api/devicespec.py
thisac: Shouldn't be. They were only used when the TDM spec was returning an unformatted layout, which is no longer the case. These functions aren't called anywhere in neither SF nor OG.

strawberryfields/api/devicespec.py
thisac: Yes! I must confess that I was almost certain that I added this specific return type hint, but apparently I didn't. 😕

tests/api/test_devicespec.py
thisac: Oh, whoops, I just added the `s` to check whether these actually needed to be disabled.

tests/frontend/test_program.py
thisac: Hmm, you're right. Not really, but it made the tests pass (since prior the compiler was set to `None` in the tests).

thisac: Fixed so that `"layout"` is always a string, `"gate_parameters"` is always a dictionary, and `"compiler"` is always a list of strings or empty list.

strawberryfields/api/devicespec.py
Mandrenkov: Sure! The tests I was referring to are:
* [`TestRemoteEngineIntegration.test_recompilation_run_async`](https://github.com/XanaduAI/strawberryfields/blob/integrate-xanadu-cloud-client/tests/api/test_remote_engine.py#L261)
* [`TestRemoteEngineIntegration.test_recompilation_run`](https://github.com/XanaduAI/strawberryfields/blob/integrate-xanadu-cloud-client/tests/api/test_remote_engine.py#L280)
* [`TestRemoteEngineIntegration.test_validation`](https://github.com/XanaduAI/strawberryfields/blob/integrate-xanadu-cloud-client/tests/api/test_remote_engine.py#L303)

strawberryfields/api/devicespec.py
Mandrenkov: Will using a string resolve the circular import issue? 🤔
```suggestion
    def create_program(self, **parameters: complex) -> "sf.Program":
```

Mandrenkov (approved):
Awesome! 🚀

strawberryfields/api/devicespec.py
thisac: Yes, that could work. But `import strawberryfields as sf` still needs to be added, but never used (which I find a bit ugly just for the sake of a type hint). What do you think?

strawberryfields/api/devicespec.py
thisac: Just updated them to be lists instead. The tests pass either way, and they should be lists as far as I'm concerned. 🙂

PR #645, opened 2021-11-09
Replace Connection class with xcc.Connection

Mandrenkov:
**Context:**
It was decided that the strawberryfields.api.Connection class should be replaced by the [xcc.Connection](https://xanadu-cloud-client.readthedocs.io/en/stable/api/xcc.Connection.html) class.

- **[Main]** Deleted the strawberryfields.api.Connection class and its corresponding unit tests.
- **[Main]** Updated RemoteEngine and the remaining API code to use the xcc.Connection class.
- **[Main]** Fixed, refactored, and enabled the tests in tests/api/test_remote_engine.py.
- **[Main]** Added a "Connection" section to the v0.20.0 migration guide.
- **[Secondary]** Sorted the installation requirements (and added XCC as a dependency).
- **[Secondary]** Added a help entry for the `docs` Makefile target.

**Benefits:**
- Strawberry Fields can leverage the Xanadu Cloud 0.4.0 API.
- There is less API code that needs to be maintained.

**Possible Drawbacks:**
- Job results stored in the .npy format cannot be retrieved.
- Existing code which directly instantiates a strawberryfields.api.Connection object will need to be updated.

**Related GitHub Issues:**
None.


strawberryfields/engine.py
```diff
@@ -575,19 +582,36 @@ def target(self) -> str:
...
+        settings = xcc.Settings()
+
+        return xcc.Connection(
+            refresh_token=settings.REFRESH_TOKEN,
+            access_token=settings.ACCESS_TOKEN,
+            host=settings.HOST,
+            port=settings.PORT,
+            tls=settings.TLS,
+            headers={"User-Agent": f"StrawberryFields/{__version__}"},
+        )
```
thisac, line 605: Nice, I like this lazy loading approach with an updated connection each call, unless passed at init. 👍

strawberryfields/engine.py
```diff
@@ -575,19 +582,36 @@ def target(self) -> str:
...
         """The device specifications for target device"""
         if self._spec is None:
-            self._spec = self._connection.get_device_spec(self.target)
+            target = self.target
+            connection = self.connection
+            device = xcc.Device(target=target, connection=connection)
+            self._spec = DeviceSpec(target=target, connection=connection, spec=device.specification)
         return self._spec
 
     def run(
```
thisac, line 617: What do you think about adding a `Raises:` section to the docstring?

strawberryfields/engine.py
```diff
@@ -635,7 +659,7 @@ def run(
 
                 time.sleep(self.POLLING_INTERVAL_SECONDS)
         except KeyboardInterrupt as e:
-            self._connection.cancel_job(job.id)
+            xcc.Job(id_=job.id, connection=self.connection).cancel()
             raise KeyboardInterrupt("The job has been cancelled.") from e
 
     def run_async(
```
thisac, line 665: Same here: a `Raises:` docstring section.

strawberryfields/engine.py
```diff
@@ -721,11 +745,27 @@ def run_async(
         if "shots" not in run_options:
             raise ValueError("Number of shots must be specified.")
 
-        return self._connection.create_job(self.target, program, run_options)
+        # Serialize a Blackbird circuit for network transmission
+        bb = to_blackbird(program)
+        bb._target["name"] = program.target
+        bb._target["options"] = run_options
+        circuit = bb.serialize()
```
thisac, line 752:
```suggestion
        bb = to_blackbird(program)
        bb._target["options"] = run_options
        circuit = bb.serialize()
```
The target name and options should be set by the `to_blackbird` function, although any potential kwargs passed to the `eng.run()` method wouldn't be in `program.run_options`, so those would still need to be updated afterwards.

thisac: An option (to avoid setting hidden attributes) would be to create a deep copy of the program and update its `run_options` (starting from line 740):
```python
# update the run options if provided
if kwargs:
    program = copy.deepcopy(program)
    program.run_options.update(kwargs)

if "shots" not in program.run_options:
    raise ValueError("Number of shots must be specified.")

# Serialize a Blackbird circuit for network transmission
circuit = to_blackbird(program).serialize()
```
Maybe not as performant though. 🤔

thisac (approved):
It looks good! Just a few thoughts and comments, mainly regarding tangential clean-up. 💯 

The `Job` changes and `DeviceSpec` changes will be up in another PR, so those parts will change from what's there now.

Makefile
```diff
@@ -8,6 +8,7 @@ help:
 	@echo "  install            to install Strawberry Fields"
 	@echo "  wheel              to build the Strawberry Fields wheel"
 	@echo "  dist               to package the source distribution"
+	@echo "  docs               to generate the Sphinx documentation"
```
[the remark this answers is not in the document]
Mandrenkov: Yes! I think it's also missing in PennyLane and a few other places as well.

strawberryfields/engine.py
Mandrenkov: Thanks! This comes in very handy when (short-lived) access tokens are used instead of (long-lived) refresh tokens.

strawberryfields/engine.py
Mandrenkov: Done.

strawberryfields/engine.py
Mandrenkov: Done.

strawberryfields/engine.py
Mandrenkov: Thanks for the tips! Although I generally prefer not to access protected members, I would argue that
```python
bb._target["options"] = run_options
```
is still less sketchy than
```python
program = copy.deepcopy(program)
```
As you pointed out, the latter also has some performance implications (although these are probably negligible given the context in which the code is run).

PR #646, opened 2021-11-09
Use refactored gaussian gate from The Walrus

sduquemesa:
**Context:** [PR #295](https://github.com/XanaduAI/thewalrus/pull/295) on The Walrus deprecates the C++ implementation of  `gaussian_gate` and `grad_gaussian_gate` in favor of the numba implementation. This in turn modifies the signatures of the imported functions.

**Description of the Change:** This PR will use the new implementation of `gaussian_gate` and `grad_gaussian_gate` on The Walrus

**Benefits:** The Walrus has less C++ dependencies

**Possible Drawbacks:** None

**Related GitHub Issues:** None

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: josh146/sphinx-action@master
+    - uses: XanaduAI/sphinx-action@master
       with:
         docs-folder: "doc/"
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -15,5 +15,5 @@ sphinx-copybutton
 sphinx-automodapi
 sphinxcontrib-bibtex==0.4.2
 tensorflow==2.6.1
-http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/thewalrus-0.17.0.dev0+20211117203611-cp37-cp37m-manylinux2014_x86_64.whl#sha256=5e5ebc26d5a4839a7cbcce63bf6b5a4245928f7808a10ceb3e695908db7bc463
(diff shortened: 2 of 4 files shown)


sduquemesa:
**Note:** checks will fail until a new version of the walrus including the mentioned changes ([PR #295](https://github.com/XanaduAI/thewalrus/pull/295) on The Walrus) is released.

josh146:
> Note: checks will fail until a new version of the walrus including the mentioned changes (PR #295 on The Walrus) is released.

@sduquemesa you can temporarily update `requirements.txt` in this branch to point to the correct wheel here: http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/

In this case you probably want

```
http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/thewalrus-0.17.0.dev0+20211101181237-cp38-cp38-manylinux2010_x86_64.whl
```

since the CI uses Python 3.8

.github/workflows/docs.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: josh146/sphinx-action@master
+    - uses: XanaduAI/sphinx-action@master
```
thisac, line 10: Seems better to use the `XanaduAI` organisation here. What's the difference, and why was the one from @josh146 used before? Is it only an older version? 🤔

sduquemesa: `XanaduAI/sphinx-action@master` is a newly created fork from @josh146's action. The former uses `python3.8` whereas the latter `python3.7`.

josh146: 🤔

sduquemesa: @josh146, for context: `josh146/sphinx-action` uses `python3.7` but `readthedocs` requires `python3..8`. Since they both use `docs/requirements.txt`, the build of one of them will fail if we point to _the walrus_ wheel for python 3.7 or 3.8.

Using `python3.8` on `sphinx-action` ensures that the same python version is used across all the CI.

josh146: ah got it! no worries @sduquemesa. Better to use the XanaduAI anyway :)

thisac (approved):
Great job on this @sduquemesa! 💯 We can merge this one now and then I'll change back the TW requirements as soon as the new TW version is released.

PR #650, opened 2021-11-10
Replace `strawberryfields.api.Job` with `xcc.Job`

thisac:
**Context:**
As part of the XCC integration, there's a new `xcc.Job` class that will replace the old `strawberryfields.api.Job` class.

* `strawberryfields/api/job.py` is removed and all uses of `strawberryfields.api.Job` is replaced with `xcc.Job`.
* An `xcc.Job` object is returned when running `RemoteEngine.run_async()`.
* Documentation is updated to explain how to use the returned `xcc.Job` object when running `RemoteEngine.run_async()`.
* `tests/api/test_remote_engine` tests are updated and `tests/api/test_job.py` is removed.

**Benefits:**
The `strawberryfields.api.Job` class is removed in favour of `xcc.Job` and we're one step closer to XCC integration.

**Possible Drawbacks:**
* When running `RemoteEngine.run_async()` the returned `xcc.Job` object doesn't return results as a `Result` object (as before, and as `RemoteEngine.run()` still does). The user will need to pass the `job.result` results manually into a `Result` object.

**Related GitHub Issues:**
None


doc/introduction/photonic_hardware.rst
```diff
@@ -46,39 +46,23 @@ will execute.
 
 Jobs can also be submitted using the **non-blocking** :meth:`eng.run_async() <.RemoteEngine.run_async>`
 method. Unlike :meth:`eng.run() <.RemoteEngine.run>`, which returns a :class:`~.Results` object once the computation is
-complete, this method instead returns a :class:`~.Job` object directly that can be queried
+complete, this method instead returns a `xcc.Job` object directly that can be queried
```
josh146, line 49: It might be nice to set up intersphinx in the `conf.py` file 🙂 That way, you can still do
```
:class:`xcc.job`
```

and it should automatically resolve to the correct page in the XCC docs.

Similar to how we have set up intersphinx between the pennylane and qml sphinx.

This could be quite nice, since I imagine users might want to quickly see the available methods and attributes on an XCC job

doc/introduction/photonic_hardware.rst
```diff
@@ -46,39 +46,23 @@ will execute.
...
 
 >>> job = engine.run_async(prog, shots=3)
 >>> job.status
 "queued"
->>> job.refresh()
+>>> job.clear()
 >>> job.status
 "complete"
->>> result = job.result
+>>> result = sf.api.Result(job.result)
```
josh146, line 58: minor question: should `Result` be part of the XCC rather than SF?

Mandrenkov: Nope. As far as the XCC is concerned, the result of a job is a dictionary (which is parsed from an .npz file); it has no concept of "samples" or the like.

doc/introduction/photonic_hardware.rst
Mandrenkov: Nice! As long as this isn't too tedious to set up or maintain, I say we should go for it.

doc/introduction/photonic_hardware.rst
```diff
@@ -46,39 +46,23 @@ will execute.
...
 method. Unlike :meth:`eng.run() <.RemoteEngine.run>`, which returns a :class:`~.Results` object once the computation is
-complete, this method instead returns a :class:`~.Job` object directly that can be queried
+complete, this method instead returns a `xcc.Job` object directly that can be queried
 to get the job's status.
 
 >>> job = engine.run_async(prog, shots=3)
 >>> job.status
 "queued"
->>> job.refresh()
+>>> job.clear()
```
Mandrenkov, line 55: To avoid race conditions:
```suggestion
>>> job.wait()
```

strawberryfields/api/__init__.py
```diff
@@ -13,13 +13,16 @@
 # limitations under the License.
 """
 This package contains the modules for the low-level Strawberry Fields program
-execution API. Specifically, the :class:`~strawberryfields.api.Job` and
-:class:`~strawberryfields.api.Result` classes provide interfaces for managing
-program execution jobs and job results respectively.
+execution API. Specifically, the :class:`~strawberryfields.api.Result` and
+:class:`~strawberryfields.api.DeviceSpec` classes are provided to represent
+hardware results and device specifications.
```
Mandrenkov, line 18: (Optional)
```suggestion
hardware results and device specifications, respectively.
```

strawberryfields/engine.py
```diff
@@ -543,13 +542,12 @@ class RemoteEngine:
     >>> job = engine.run_async(program, shots=1)
     >>> job.status
     "queued"
-    >>> job.result
-    InvalidJobOperationError
-    >>> job.refresh()
+    >>> job.clear()
```
Mandrenkov, line 545: Similar to the Sphinx documentation, I think [`xcc.Job.wait()`](https://xanadu-cloud-client.readthedocs.io/en/stable/api/xcc.Job.html#xcc.Job.wait) is more appropriate:
```suggestion
    >>> job.wait()
```

strawberryfields/engine.py
```diff
@@ -647,23 +645,19 @@ def run(
...
             program, compile_options=compile_options, recompile=recompile, **kwargs
         )
         try:
-            while True:
-                job.refresh()
-                if job.status == "complete":
-                    self.log.info("The remote job %s has been completed.", job.id)
-                    return job.result
-
+            while not job.finished:
```
Mandrenkov, line 656: Okay, I admit that this part is slightly tricky. In an ideal world, we could just use
```python
job.wait()
```
but this would be problematic for long-running jobs that use a Connection without a refresh token (since the access token could expire before the job finishes). So, just to be safe, we can change the loop to
```python
while True:
    # Refresh the Connection
    job._connection = self.connection
    # Clear the Job cache
    job.clear()

    if job.finished:
        break
    time.sleep(self.POLLING_INTERVAL_SECONDS)
```
or
```python
while True:
    # Refresh the Connection and clear the Job cache
    job = xcc.Job(id_ = job.id, connection=self.connection)

    if job.finished:
        break
    time.sleep(self.POLLING_INTERVAL_SECONDS)
```
or even
```python
while not xcc.Job(job.id, self.connection).finished:
    time.sleep(self.POLLING_INTERVAL_SECONDS)
```
Note that the return value of virtually all `xcc.Job` properties are cached so we need to execute
```python
job.clear()
```
to fetch the updated status of a job.

strawberryfields/engine.py
```diff
@@ -674,12 +668,21 @@ def run(
             xcc.Job(id_=job.id, connection=self.connection).cancel()
             raise KeyboardInterrupt("The job has been cancelled.") from e
 
+        if job.status == "complete":
+            self.log.info(f"The remote job {job.id} has been completed.")
+            return Result(samples=list(job.result.values())[0], is_stateful=False)
+        else:
+            message = f"The remote job {job.id} has failed with status {job.status}."
```
Mandrenkov, line 675: Don't forget to include the job metadata in the message since that field contains the details of the error.
```suggestion
            message = f"The remote job {job.id} has failed with status {job.status}: {job.metadata}."
```
(XCC v0.1.1 will introduce the `xcc.Job.metadata` attribute.)

tests/api/test_remote_engine.py
```diff
@@ -41,21 +41,25 @@ def job(connection, monkeypatch):
...
 
-    result = Result(np.array([[1, 2], [3, 4]]), is_stateful=False)
+    result = {"samples": np.array([[1, 2], [3, 4]]), "foo": [np.array([5, 6])]}
 
     monkeypatch.setattr(xcc.Job, "submit", mock_return(job))
-    monkeypatch.setattr(sf.engine.Job, "refresh", refresh)
-    monkeypatch.setattr(sf.engine.Job, "result", result)
-    monkeypatch.setattr(sf.engine.Job, "meta", {"foo": "bar"})
+    monkeypatch.setattr(xcc.Job, "result", result)
+    monkeypatch.setattr(xcc.Job, "_details", _details)
```
Mandrenkov, line 61: We may as well just use this instead:
```suggestion
    job._details = {"status": "open"}
```

tests/api/test_remote_engine.py
```diff
@@ -120,18 +124,21 @@ def test_run_async(self, prog):
...
 
-        assert job.status == JobStatus.OPEN.value
+        # job.status calls job.finished, incrementing the request counter
+        assert job.status == "open"
 
-        for _ in range(REQUESTS_BEFORE_COMPLETED):
-            job.refresh()
+        for _ in range(REQUESTS_BEFORE_COMPLETED - 1):
+            assert job.finished == False
+        assert job.finished == True
```
Mandrenkov, line 132: It is more Pythonic to compare Booleans using `is`:
```suggestion
            assert job.finished is False
        assert job.finished is True
```

tests/api/test_remote_engine.py
```diff
@@ -141,7 +148,7 @@ def test_run_async_options_from_kwargs(self, prog, blackbird):
         engine.run_async(prog, shots=1234)
         assert blackbird._target["options"] == {"cutoff_dim": 12, "shots": 1234}
 
-    def test_run_asnyc_options_from_program(self, prog, blackbird):
+    def test_run_async_options_from_program(self, prog, blackbird):
```
Mandrenkov, line 151: Whoops! 🤦‍♂️

doc/introduction/photonic_hardware.rst
thisac: I was thinking a bit about this, and thought that `job.clear()` was most similar to `job.refresh()` (keeping the job non-blocked). But I agree this seems good also.

strawberryfields/engine.py
thisac: Nice, didn't know about this. 💯

strawberryfields/engine.py
thisac: Thanks @Mandrenkov. I went with the first suggestion here, and we can revisit this once support for dynamically refreshing config info exist in the `xcc.Connection` class.

doc/introduction/photonic_hardware.rst
thisac: Done! Thanks @josh146. 🥳

doc/introduction/photonic_hardware.rst
josh146: :+1: I suppose it's just a matter of migrating `sf.api.Result` out of the API folder?

Mandrenkov: Hmm, I don't quite follow. 🤔 Can you clarify what is achieved by moving sf.api.Result out of the sf.api?

strawberryfields/engine.py
Mandrenkov: Let's put this PR on hold until the XCC v0.1.1 is released so that we can access the metadata; I would argue that this is very important debugging information.

tests/api/test_remote_engine.py
```diff
@@ -58,7 +58,8 @@ def finished(self):
 
     monkeypatch.setattr(xcc.Job, "submit", mock_return(job))
     monkeypatch.setattr(xcc.Job, "result", result)
-    monkeypatch.setattr(xcc.Job, "_details", _details)
+    monkeypatch.setattr(xcc.Job, "clear", lambda _: None)
+    job._details = {"status": "open"}
```
Mandrenkov, line 62: Very minor point, but I think it would look better if we assigned `job._details` immediately after `job` is instantiated:
```python
    job = xcc.Job(id_="123", connection=connection)
    job._details = {"status": "open"}l

    result = {"samples": np.array([[1, 2], [3, 4]]), "foo": [np.array([5, 6])]}

    monkeypatch.setattr(xcc.Job, "submit", mock_return(job))
    monkeypatch.setattr(xcc.Job, "result", result)
    monkeypatch.setattr(xcc.Job, "clear", mock_return(None))
    monkeypatch.setattr(xcc.Job, "finished", finished)
    return job
```

strawberryfields/engine.py
```diff
@@ -653,26 +653,36 @@ def run(
...
         except KeyboardInterrupt as e:
             xcc.Job(id_=job.id, connection=self.connection).cancel()
             raise KeyboardInterrupt("The job has been cancelled.") from e
 
-        if job.status == "complete":
+        if job.status == "failed":
+            message = (
+                f"The remote job {job.id} failed due to an internal "
+                "server error. Please try again."
+            )
```
Mandrenkov, line 674: Don't forget to output the metadata here too!
```suggestion
            message = (
                f"The remote job {job.id} failed due to an internal "
                f"server error: {job.metadata}. Please try again."
            )
```

doc/conf.py
```diff
@@ -50,7 +50,10 @@
 ]
 
 
-intersphinx_mapping = {"photonics": ("https://strawberryfields.ai/photonics", None)}
+intersphinx_mapping = {
+    "photonics": ("https://strawberryfields.ai/photonics", None),
+    "xcc": ("https://xanadu-cloud-client.readthedocs.io/en/stable/", None)
```
Mandrenkov, line 55: Does this work too? It would be nice to leave the `/ → /en/stable/` redirect up to Read the Docs.
```suggestion
    "xcc": ("https://xanadu-cloud-client.readthedocs.io/", None)
```

.github/CHANGELOG.md
```diff
@@ -186,6 +186,43 @@
...
+  class.
+  [(#645)](https://github.com/XanaduAI/strawberryfields/pull/645)
+
+  A `Job` object is returned when running jobs asynchronously. In previous versions of Strawberry
+  Fields (v0.19.0 and lower), the `Job` object can be used as follows:
+
+  ```pycon
+  >>> job = engine.run_async(program, shots=1)
+  >>> job.status
+  "queued"
```
Mandrenkov, line 200: Minor point, but I _think_ the Python interpreter outputs strings in single quotes.
```suggestion
  'queued'
```
Same goes for the rest below.

.github/CHANGELOG.md
```diff
@@ -186,6 +186,43 @@
...
+  >>> job.wait()
+  >>> job.status
+  "complete"
+  >>> job.result
+  {"output": [array([[0 1 0 2 1 0 0 0]])]}
+  ```
+
+  The `job.wait()` method is a blocking method that will wait for the job to finish. Alternatively,
+  `job.clear()` can be called to clear the cache allowing `job.status` to re-fetch the job status.
```
Mandrenkov, line 224: Grammar:
```suggestion
  `job.clear()` can be called to clear the cache, allowing `job.status` to re-fetch the job status.
```

Mandrenkov (approved):
Looks good to me! 🎉

As mentioned in https://github.com/XanaduAI/strawberryfields/pull/650#discussion_r748322200, however, it may be worth putting this PR on hold until the XCC is updated.

doc/introduction/photonic_hardware.rst
josh146: Just that the `sf.api` namespace seems somewhat redundant with the move to `xcc` as the API client backend 🙂

Mandrenkov: Oh, I see!

Perhaps we can convert `sf.api` from a package to a module with the code for both DeviceSpec and Result.

thisac: > Perhaps we can convert sf.api from a package to a module

Sounds like a good idea! Either that or just move `result.py` and `device_spec.py` into the root SF folder. If keeping it as one module, what would be the suggested name? `api.py`, `device_spec.py`?

Mandrenkov: Hmm, I think I would go with either `api.py` or `remote.py`.

tests/api/test_remote_engine.py
thisac: True, I agree!

doc/conf.py
thisac: I tried that first and it didn't work. Seems like it could find any Sphinx docs at the URL without the `/en/stable` part.

.github/CHANGELOG.md
```diff
@@ -186,6 +186,43 @@
...
+
+  ```pycon
+  >>> job = engine.run_async(program, shots=1)
+  >>> job.status
+  "queued"
+  >>> job.result
+  InvalidJobOperationError
+  >>> job.refresh()
+  >>> job.status
+  "complete"
```
thisac, line 205:
```suggestion
  'complete'
```

.github/CHANGELOG.md
```diff
@@ -186,6 +186,43 @@
...
+  >>> job.result
+  [[0 1 0 2 1 0 0 0]]
+  ```
+
+  In Strawberry Fields v0.20.0, the `Job` object works slightly differently:
+
+  ```pycon
+  >>> job = engine.run_async(program, shots=1)
+  >>> job.status
+  "queued"
```
thisac, line 215:
```suggestion
  'queued'
```

.github/CHANGELOG.md
```diff
@@ -186,6 +186,43 @@
...
+  >>> job.wait()
+  >>> job.status
+  "complete"
+  >>> job.result
+  {"output": [array([[0 1 0 2 1 0 0 0]])]}
```
thisac, line 220:
```suggestion
  "complete"
  >>> job.result
  {"output": [array([[0 1 0 2 1 0 0 0]])]}
```
```suggestion
  'complete'
  >>> job.result
  {'output': [array([[0 1 0 2 1 0 0 0]])]}
```

doc/introduction/photonic_hardware.rst
thisac: I'm not really sure. Would it perhaps even make more sense to have both a `device_spec.py` _and_ a `result.py` module? The `DeviceSpec` is purely remote, but `Result` is actually used for both local and remote runs.

Mandrenkov: That's a good point! I think you've convinced me to extract them as separate modules.

doc/conf.py
Mandrenkov: Aw, okay. 😟

strawberryfields/api/devicespec.py
```diff
@@ -26,7 +26,7 @@
 
 
 class DeviceSpec:
-    """The specifications for a specific hardware device.
+    """The specifications for a specific remote device.
```
Mandrenkov, line 29: Good catch! 💯

doc/development/migration_guides.rst
```diff
@@ -61,3 +61,44 @@ The (semantically) equivalent code in Strawberry Fields v0.20.0 is
...
+    >>> job.status
+    'queued'
+    >>> job.wait()
+    >>> job.status
+    'complete'
+    >>> job.result
+    {'output': [array([[0 1 0 2 1 0 0 0]])]}
+
+  The `job.wait()` method is a blocking method that will wait for the job to finish. Alternatively,
+  `job.clear()` can be called to clear the cache, allowing `job.status` to re-fetch the job status.
```
Mandrenkov, line 104: Typo:
```suggestion
  The ``job.wait()`` method is a blocking method that will wait for the job to finish. Alternatively,
  ``job.clear()`` can be called to clear the cache, allowing ``job.status`` to re-fetch the job status.
```

We should also consider linking to the XCC documentation.

Mandrenkov (approved):
Nice!

doc/conf.py
josh146: Yep, this is required because readthedocs hosts _multiple_ docs versions (e.g., `stable`, `latest`, `v0.19.0`, etc.). So you need to be explicit about the version you want intersphinx to resolve links to

Mandrenkov: Right, but I figured the redirect would just automatically choose one of these versions for us. The advantage of _not_ specifying a version is that changes to the default XCC Read the Docs version would be reflected in SF as well.

PR #652, opened 2021-11-12
Extract API subpackage

thisac:
**Context:**
Due to the XCC integration, most of the Xanadu API subpackage is now gone.

* `strawberryfields.api.results.py` and `strawberryfields.api.devicespec.py` are moved to the root `strawberryfields` folder, and the `strawberryfields.api` folder is removed.
* The `FailedJobError` is moved from `strawberryfields.api.__init__.py` to `strawberryfields.engine.py` (where it's used).

**Benefits:**
Things look cleaner.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


.github/CHANGELOG.md
```diff
@@ -247,6 +247,13 @@
   connection usages, and `DeviceSpec.target` is retrieved from the device specification rather than
   passed at initialization.
 
+* The `api` subpackage has been removed and the contained `DeviceSpec` and `Result` classes have
+  been moved to the root `strawberryfields` folder. They can now be imported as follows:
+
+  ```python
+  from strawberryfields import DeviceSpec, Result
```
Mandrenkov, line 254: Is this the preferred access pattern? Or would we rather users write
```python
import strawberryfields as sf
# sf.DeviceSpec
# sf.Result
```

strawberryfields/engine.py
```diff
@@ -39,6 +40,10 @@
 __all__ = ["BaseEngine", "LocalEngine", "BosonicEngine"]
 
 
+class FailedJobError(Exception):
+    """Raised when a job had a failure on the server side."""
```
Mandrenkov, line 44: I was wondering where this went. Good call! 👍

strawberryfields/__init__.py
```diff
@@ -31,6 +31,8 @@
 from .tdm import TDMProgram
 from .plot import plot_wigner, plot_fock, plot_quad
 from . import tdm
+from .devicespec import DeviceSpec
+from .result import Result
```
Mandrenkov, line 35: Should we add these to `__all__` as well?

tests/api/conftest.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-Test fixtures and shared functions for strawberryfields.api tests
+Test fixtures and shared functions for strawberryfields API tests
```
Mandrenkov, line 15: Right?
```suggestion
Test fixtures and shared functions for Strawberry Fields API tests
```

tests/frontend/test_post_processing.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-Unit tests for strawberryfields.api.post_processing
+Unit tests for strawberryfields.utils.post_processing
```
Mandrenkov, line 15: Whoops! 🤭

Mandrenkov (approved):
Great work, Theo! 🧹

strawberryfields/__init__.py
thisac: I was thinking about that as well. Actually, now when you mention it, I think we probably should!

.github/CHANGELOG.md
thisac: True, I think your suggestion is better.

PR #654, opened 2021-11-18
Merge master into branch

thisac:
**Context:**
Merging master into `integrate-xanadu-cloud-client` branch.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: josh146/sphinx-action@master
+    - uses: XanaduAI/sphinx-action@master
       with:
         docs-folder: "doc/"
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,6 +10,5 @@ jobs:
     - uses: XanaduAI/sphinx-action@master
       with:
         docs-folder: "doc/"
-        repo-token: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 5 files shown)


doc/requirements.txt
```diff
@@ -14,6 +14,6 @@ sphinx-autodoc-typehints==1.10.3
 sphinx-copybutton
 sphinx-automodapi
 sphinxcontrib-bibtex==0.4.2
-tensorflow==2.5.1
-thewalrus>=0.16.0
-toml
\ No newline at end of file
+tensorflow==2.6.1
+http://xanadu-wheels.s3-website-us-east-1.amazonaws.com/thewalrus-0.17.0.dev0+20211117203611-cp38-cp38-manylinux2014_x86_64.whl#sha256=f8cef1f05a903eec2b563758b7542490681dfc899e57cce1542385546e222306
```
thisac, line 18: Will be switched to `thewalrus>=0.17.0` when making the version bump PR, after TW has been released.

.github/workflows/docs.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: josh146/sphinx-action@master
+    - uses: XanaduAI/sphinx-action@master
```
Mandrenkov, line 10: We should also this opportunity to remove the `repo-token` field!

Mandrenkov (approved):
Looks good provided that we adjust the TW dependency version in the v0.20.0 version bump PR.

.github/workflows/docs.yml
thisac: What was the `repo-token` used for and why is it no longer needed? Can we just remove that line (leaving everything else as it is)?

PR #655, opened 2021-11-18
Remove `fill_template` calls in `RemoteEngine.run_async`

thisac:
**Context:**
Some traces of removed methods were left from the `DeviceSpec` XCC integration update.

Removes non-existing `DeviceSpec.fill_template` and `DeviceSpec.is_formatted` calls in `RemoteEngine.run_async`

**Benefits:**
It will now work to run TDM programs remotely using the updated `DeviceSpec` class.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -719,9 +719,6 @@ class RemoteEngine:
         kwargs.update(self._backend_options)
 
         device = self.device_spec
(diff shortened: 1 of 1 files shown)


Mandrenkov (approved):
As discussed offline, TDM layouts are now guaranteed to be valid Blackbird scripts so template filling is not required.

PR #658, opened 2021-11-19
Increment version to 0.21.0-dev

thisac:
Incrementing version number to 0.21.0-dev post release.

### DIFF
--- a/strawberryfields/_version.py
+++ b/strawberryfields/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.20.0"
+__version__ = "0.20.0-dev"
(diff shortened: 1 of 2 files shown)


strawberryfields/_version.py
```diff
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.20.0"
+__version__ = "0.20.0-dev"
```
Mandrenkov, line 19: Is this right?
```suggestion
__version__ = "0.21.0-dev"
```

Mandrenkov (changes requested):
I think the new version number is slightly off.

strawberryfields/_version.py
thisac: Thanks, I think I'm getting a bit tired. 🤦

Mandrenkov (approved):

PR #659, opened 2021-11-23
Compile info and target to `TDMProgram.compile`

thisac:
**Context:**
The `Program.compile` method sets the program target (which is `None` at start) to the device target (e.g., `X8_01`) and updates `Program.compile_info`. This never happens in `TDMProgram.compile`, so `prog.target` is always `None`, which in turn causes the Blackbird serialize method to ignore the passed shots option, since it has no target name.

For example, the BlackbirdProgram.target for a regular `sf.Program` could be `{'name': 'X8_01', 'options': {'shots': 1}}` which would serialize into target `X8_01 (shots=1)`. The corresponding for an `sf.TDMProgram` would be `{'name': None, 'options': {'shots': 1}}`, which wouldn’t serialize since there’s no target name.

`program.compile_info` and `program.target` are set if compilation (i.e., validation) is successful for the `TDMProgram`.

**Benefits:**
The correct target is now passed along correctly to the program when compiling, and is correctly serialized into a Blackbird script before being sent to hardware.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,11 +6,17 @@
 
 <h3>Bug fixes</h3>
 
+* ``program.compile_info`` and ``program.target`` are correctly set if compilation is successfully
+  completed when compiling a ``TDMProgram``.
+  [(#659)](https://github.com/XanaduAI/strawberryfields/pull/659)
+
 <h3>Documentation</h3>
 
 <h3>Contributors</h3>
 
 This release contains contributions from (in alphabetical order):
+
+Theodor Isacsson
 
 # Release 0.20.0 (current release)
 
--- a/strawberryfields/tdm/tdmprogram.py
+++ b/strawberryfields/tdm/tdmprogram.py
@@ -437,9 +437,11 @@ class TDMProgram(Program):
 
     # pylint: disable=too-many-branches
     def compile(self, *, device=None, compiler=None):
-        """Compile the time-domain program given a Strawberry Fields photonic hardware device specification.
+        """Compile the time-domain program given a Strawberry Fields photonic hardware device
+        specification.
 
-        Currently, the compilation is simply a check that the program matches the device.
+        The compilation simply checks that the program matches the device and sets the compile
+        information and the program target to the correct device target.
 
         Args:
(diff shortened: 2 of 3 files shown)


thisac:
[sc-11667]

strawberryfields/tdm/tdmprogram.py
```diff
@@ -557,6 +557,8 @@ def compile(self, *, device=None, compiler=None):
                                     device.target, program_param, param_range
                                 )
                             )
+            self._compile_info = (device, "TDM")
+            self._target = device.target
```
Mandrenkov, line 561: Should we update the docstring to state that compilation also sets the target and compilation information fields of the TDM program? Otherwise, it may seem that calling `TDMProgram.compile()` is always an optional step.

.github/CHANGELOG.md
```diff
@@ -6,12 +6,18 @@
 
 <h3>Bug fixes</h3>
 
+* ``program.compile_info`` and ``program.target`` are correctly set if compilation is successfully
+  completed when compiling a ``TDMProgram``.
```
Mandrenkov, line 10: Given that the changelog is written in Markdown, we can replace \`\` with \`. I would also specify that this change only impacts TDM programs compiled using the "TDM" compiler.
```suggestion
* The `TDMProgram.compile_info` and `TDMProgram.target` fields are now set when
  a `TDMProgram` is compiled using the "TDM" compiler.
```

Mandrenkov (commented):
No major complaints from me! 🎉

heltluke (approved):

strawberryfields/tdm/tdmprogram.py
thisac: It is an optional (manual) step since the `Engine` will call the compile method internally when run (this is the same way that the X8 compilation happens). This happens both on the `LocalEngine` and `RemoteEngine`, compiling either for the simulator backend or the target device.

thisac: Updated the docstring slightly still, since it makes sense to have it there. 🙂

Mandrenkov (approved):
Looks good to me! 👍

PR #661, opened 2021-11-29
Updates to device specfication and program validation

thisac:
**Context:**
Some changes in [this PR in platform-services-shared](https://github.com/XanaduAI/platform-services-shared/pull/60) allows the device specification to have optional entries for `layout` and `gate_parameters`. This PR adds some changes to support this.

* The return type hint for `DeviceSpec.layout` and `DeviceSpec.gate_parameters` are set to optional.
* A reasonable error message is raised if a user attempts to use a device specification that contains gate parameters but no layout (in which case it cannot validate the applied gate parameters against the allowed gate parameters described in the device specification).

**Benefits:**
The `DeviceSpec` conforms to the updates on the platform.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


.github/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 <h3>New features since last release</h3>
 
+* `DeviceSpec.layout` and `DeviceSpec.gate_parameters` can now return `None`, which can happen when
+  e.g., a remote simulator device is used.
+  [(#660)](https://github.com/XanaduAI/strawberryfields/pull/660)
```
Mandrenkov, line 7:
```suggestion
  [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
```

.github/CHANGELOG.md
```diff
@@ -10,6 +14,11 @@
   a `TDMProgram` is compiled using the "TDM" compiler.
   [(#659)](https://github.com/XanaduAI/strawberryfields/pull/659)
 
+* `program.compile` now raises an error if the device specification contains gate parameters but no
+  circuit layout. Without a layout, the gate parameters cannot be validated against the device
+  specification.
+  [(#660)](https://github.com/XanaduAI/strawberryfields/pull/660)
```
Mandrenkov, line 20:
```suggestion
  [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
```

.github/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 <h3>New features since last release</h3>
 
+* `DeviceSpec.layout` and `DeviceSpec.gate_parameters` can now return `None`, which can happen when
+  e.g., a remote simulator device is used.
```
Mandrenkov, line 6: Up to you, but I usually find that using more sentences helps with clarity.
```suggestion
* `DeviceSpec.layout` and `DeviceSpec.gate_parameters` may now return `None`. This can happen
  when a remote simulator device is used.
```

strawberryfields/devicespec.py
```diff
@@ -90,13 +90,13 @@ def gate_parameters(self) -> Mapping[str, Ranges]:
...
+                range_list = [[i] if not isinstance(i, Sequence) else i for i in param_ranges]
+                gate_parameters[gate_name] = Ranges(*range_list)
 
-        for gate_name, param_ranges in self._spec["gate_parameters"].items():
-            # convert gate parameter allowed ranges to Range objects
-            range_list = [[i] if not isinstance(i, Sequence) else i for i in param_ranges]
-            gate_parameters[gate_name] = Ranges(*range_list)
-
-        return gate_parameters
+        return gate_parameters or None
```
Mandrenkov, line 99: What should be the expected behaviour when the gate parameters dictionary is empty (e.g., a device has a layout with no parameters)? Shouldn't we just return an empty dictionary in that case?

Mandrenkov (commented):
Looks good! 👍

One thing to keep in mind is that the layout and gate parameters are only allowed to be `None` in the 0.5.0 Xanadu Cloud API. As a result, the code in this PR should only see use when the XCC dependency is upgraded to use the 0.5.0 API.

strawberryfields/devicespec.py
thisac: That probably makes sense, yes. I'm not sure it matters that much though, or maybe it does. 🤔

thisac:
> One thing to keep in mind is that the layout and gate parameters are only allowed to be `None` in the 0.5.0 Xanadu Cloud API. As a result, the code in this PR should only see use when the XCC dependency is upgraded to use the 0.5.0 API.

Will we be updating to use the 0.5.0 API anytime soon? How regularly do we want to switch to new API versions?

tests/frontend/test_program.py
```diff
@@ -650,6 +650,37 @@ class DummyCircuit(Compiler):
...
             )
 
+    def test_missing_layout(self):
+        """Test that an error is raised if the device spec is missing a layout when compiled"""
+        device_dict = {
+            "target": None,
+            "layout": None,
+            "modes": 2,
+            "compiler": ["DummyCompiler"],
+            "gate_parameters": {
```
heltluke, line 660: Maybe there should now also be a test where the `device_dict` does not contain any `gate_parameters`? I'm somewhat curious if it makes sense, even for a simulator, to have a layout but no gate parameters...

thisac: I can add this. Currently, the layout isn't used at all if the spec doesn't contain any `gate_parameters` (both if it's an empty dict or `None`), unless using a `TDMProgram`, in which case an error would be raised.

Mandrenkov:
> Will we be updating to use the 0.5.0 API anytime soon? How regularly do we want to switch to new API versions?

This is still TBD but we're toying with the idea of releasing a new platform API version on a quarterly cadence (provided there are enough features to justify a release). In the case of the 0.5.0 API, this would be Q1 of 2022 (i.e., SF v0.23.0).

strawberryfields/devicespec.py
Mandrenkov: I think it's worth making the distinction because if the gate parameters are `{}` we know the device is not parameterized; however, if the gate parameters are `None` then we're not sure if the device is not parameterized OR it simply forgot to supply the gate parameters in its specification.

strawberryfields/devicespec.py
```diff
@@ -89,8 +89,10 @@ def gate_parameters(self) -> Mapping[str, Ranges]:
         >>> spec.gate_parameters
         {'squeezing_amplitude_0': x=0, x=1, 'phase_0': x=0, 0≤x≤6.283185307179586}
         """
-        gate_parameters = {}
+        if not self._spec["gate_parameters"]:
+            return None
```
Mandrenkov, line 93: Is this the expected behaviour when the gate parameters is an empty dictionary? Or is it more correct to say
```suggestion
        if self._spec["gate_parameters"] is None:
            return None
```

strawberryfields/devicespec.py
```diff
@@ -107,6 +109,9 @@ def validate_parameters(self, **parameters: complex) -> None:
         Raises:
             ValueError: if an invalid parameter is passed
         """
+        if not self.gate_parameters:
+            return
```
Mandrenkov, line 113: Does it makes sense to emit a warning message in this case?

.github/CHANGELOG.md
```diff
@@ -2,16 +2,27 @@
...
+  when a remote simulator device is used.
+  [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
+
 <h3>Breaking Changes</h3>
 
 <h3>Bug fixes</h3>
 
+* `program.compile` now raises an error if the device specification contains gate parameters but no
+  circuit layout. Without a layout, the gate parameters cannot be validated against the device
+  specification.
```
Mandrenkov, line 15: Nicely put!

tests/frontend/test_program.py
```diff
@@ -672,6 +672,81 @@ class DummyCircuit(Compiler):
...
+                "squeezing_amplitude_0": [0, 1],
+            },
+        }
+
+        class DummyCircuit(Compiler):
```
Mandrenkov, line 687: Minor point, but is this a typo?
```suggestion
        class DummyCompiler(Compiler):
```

tests/frontend/test_program.py
```diff
@@ -672,6 +672,81 @@ class DummyCircuit(Compiler):
...
+
+        device_dict = {
+            "target": None,
+            "layout": mock_layout,
+            "modes": 2,
+            "compiler": ["DummyCompiler"],
+            "gate_parameters": None,  # no gate_parameters, so any value is valid
+        }
+
+        class DummyCircuit(Compiler):
```
Mandrenkov, line 725: See comment above.
```suggestion
        class DummyCompiler(Compiler):
```

tests/frontend/test_program.py
```diff
@@ -672,6 +672,81 @@ class DummyCircuit(Compiler):
...
+            S2gate({squeezing_amplitude_0}, 0.0) | [0, 1]
+            """
+        )
+
+        device_dict = {
+            "target": None,
+            "layout": mock_layout,
+            "modes": 2,
+            "compiler": ["DummyCompiler"],
+            "gate_parameters": None,  # no gate_parameters, so any value is valid
```
Mandrenkov, line 722: Should we add a test for the case where `gate_parameters` is an empty dictionary? This could happen if a new hardware device is registered which is not configurable.

Mandrenkov (approved):
Looks good to me!

As mentioned in a previous review, these changes only make sense in the context of the 0.5.0 API; however, they could still serve to make future SF migrations easier for users (i.e., when the 0.4.0 API is deprecated or removed).

tests/frontend/test_program.py
thisac: Likely, I believe it's been there for some time, but haven't been updated (due to it being quite minor).

strawberryfields/devicespec.py
thisac: I'm not sure if any of the options would be considered more expected than the other, although it's probably safer to return empty dicts instead of `None` in this case.

strawberryfields/devicespec.py
thisac: Perhaps, I'm not sure it's necessary though. Do you think so?

Mandrenkov: My intuition is that if a caller invokes `validate_parameters()`, they expect some sort of validation to take place; if there is nothing to validate, they shouldn't call the method in the first place. The goal is to avoid a situation where a user _thinks_ a set of parameters have been validated when, in fact, no validation has taken place.

Put another way, it comes down to whether we think calling `validate_parameters()` is a mistake when there are no parameters. If the answer is "yes", we should raise a warning. If the answer is "no", the current behaviour is correct.

thisac: I don't see calling `validate_parameters()` as a mistake even when there are no parameter restrictions. It simply means that all parameters values are valid. For example, a user might not know if there are any restrictions, and simply calls this function to make sure that the used parameters are valid.

This should perhaps rather be update to 
```suggestion
        if self.gate_parameters is None:
            return
```
so that an empty `gate_parameters` dict would mean that no parameters are valid (i.e., raising the `Parameter {p} not a valid parameter for this device` error below), although I think I'd prefer not raising a warning.

Mandrenkov: Okay, sounds good!

PR #662, opened 2021-12-01
Update max measurements check

thisac:
**Context:**
The platform returns a flat dictionary entry for the maximum number of measurements allowed when running jobs on Simulon. Strawberry Fields currently expects a nested dictionary entry (which was added in #545).

`Program.assert_max_number_of_measurements` is updated to use the flat modes entry when checking for maximum allowed measurements. 

**Benefits:**
`simulon_gaussian` is one step closer to working again.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -12,7 +12,8 @@
 
 * Updates `Program.assert_max_number_of_measurements` to expect the maximum number
   of measurements from the device specification as a flat dictionary entry instead
-  of a nested one:
+  of a nested one.
+  [(#662)](https://github.com/XanaduAI/strawberryfields/pull/662)
 
   ```python
   "modes": {
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,6 +10,30 @@
   a `TDMProgram` is compiled using the "TDM" compiler.
   [(#659)](https://github.com/XanaduAI/strawberryfields/pull/659)
 
+* Updates `Program.assert_max_number_of_measurements` to expect the maximum number
+  of measurements from the device specification as a flat dictionary entry instead
+  of a nested one:
+
+  ```python
+  "modes": {
+      "pnr_max": 20,
+      "homodyne_max": 1000,
+      "heterodyne_max": 1000,
+  }
+  ```
+
+  instead of
+
+  ```python
+  "modes": {
+      "max": {
+          "pnr": 20,
+          "homodyne": 1000,
+          "heterodyne": 1000,
+      }
+  }
+  ```
+
 <h3>Documentation</h3>
 
 <h3>Contributors</h3>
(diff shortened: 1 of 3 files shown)


.github/CHANGELOG.md
```diff
@@ -10,6 +10,30 @@
   a `TDMProgram` is compiled using the "TDM" compiler.
   [(#659)](https://github.com/XanaduAI/strawberryfields/pull/659)
 
+* Updates `Program.assert_max_number_of_measurements` to expect the maximum number
+  of measurements from the device specification as a flat dictionary entry instead
+  of a nested one:
```
Mandrenkov, line 15: Looks like we're missing a PR link here.

strawberryfields/program.py
```diff
@@ -605,6 +605,8 @@ def compile(self, *, device=None, compiler=None, **kwargs):
                 The default is False.
             warn_connected (bool): If True, the user is warned if the quantum circuit is not weakly
                 connected. The default is True.
+            shots (int): Number of times the program measurement evaluation is repeated. Passed
+                along to the compiled program's `run_options`.
```
Mandrenkov, line 609: I think this is rendered in RST:
```suggestion
                along to the compiled program's ``run_options``.
```

strawberryfields/program.py
```diff
@@ -526,13 +526,13 @@ def assert_max_number_of_measurements(self, device):
...
-            max_homodyne = device.modes["max"]["homodyne"]
-            max_heterodyne = device.modes["max"]["heterodyne"]
+            max_pnr = device.modes["pnr_max"]
+            max_homodyne = device.modes["homodyne_max"]
+            max_heterodyne = device.modes["heterodyne_max"]
         except (KeyError, TypeError) as e:
             raise KeyError(
                 "Device specification must contain an entry for the maximum allowed number "
-                "of measurments. Have you specified the correct target?"
+                "of measurements. Have you specified the correct target?"
```
Mandrenkov, line 535: Are we sure that, "Have you specified the correct target?" is an accurate suggestion? It seems to me that users would only see this message when there is a bug in the SF-XC interface!

Mandrenkov (approved):
Thanks for catching this, Theo! 🔍

I can confirm this PR allows the following script to finish successfully using the `dev` Xanadu Cloud environment:
```python
import strawberryfields as sf
from strawberryfields.ops import MeasureFock

prog = sf.Program(4)

with prog.context as q:
    MeasureFock() | q

eng = sf.RemoteEngine("simulon_gaussian")
results = eng.run(prog, shots=3)
print(results.samples)
```

strawberryfields/program.py
thisac: Hmm... you're probably right. This message can be improved.

strawberryfields/program.py
thisac: Good catch!

PR #663, opened 2021-12-01
Raw results and metadata

thisac:
**Context:**
There's no way to access non-samples results (i.e., metadata) in the `Result` class.

* A `Result.raw` method is added that returns the raw result dictionary.
* A `Result.metadata` method is added that returns the metadata of the results (i.e., everything except the `"output"` entry).
* Tests are added for the `Result` class.
* A setter method is added for `Result.state` allowing a state to be set iff a state has not previously been set and the `Result` object contains `samples_dict` samples, indicating that it's a local simulation.

**Benefits:**
* Results metadata can be accessed without using the private `Result._result` method.
* The `Result` class is properly tested.
* The state can be properly set in the result object.

**Possible Drawbacks:**
* The state setter relies on the fact that all local simulations have a `samples_dict` (which is currently the case).

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -197,7 +197,7 @@ class BaseEngine(abc.ABC):
             print_fn (function): optional custom function to use for string printing.
         """
         for k, r in enumerate(self.run_progs):
-            print_fn("Run {}:".format(k))
+            print_fn(f"Run {k}:")
             r.print(print_fn)
 
     @abc.abstractmethod
@@ -265,7 +265,7 @@ class BaseEngine(abc.ABC):
                 # there was a previous program segment
                 if not p.can_follow(prev):
                     raise RuntimeError(
-                        "Register mismatch: program {}, '{}'.".format(len(self.run_progs), p.name)
+                        f"Register mismatch: program {len(self.run_progs)}, '{p.name}'."
                     )
 
                 # Copy the latest measured values in the RegRefs of p.
@@ -336,7 +336,7 @@ class LocalEngine(BaseEngine):
         return super().__new__(cls)
 
     def __str__(self):
-        return self.__class__.__name__ + "({})".format(self.backend_name)
+        return self.__class__.__name__ + f"({self.backend_name})"
 
     def reset(self, backend_options=None):
         backend_options = backend_options or {}
@@ -374,15 +374,14 @@ class LocalEngine(BaseEngine):
             except NotApplicableError:
                 # command is not applicable to the current backend type
                 raise NotApplicableError(
-                    "The operation {} cannot be used with {}.".format(cmd.op, self.backend)
+                    f"The operation {cmd.op} cannot be used with {self.backend}."
                 ) from None
 
             except NotImplementedError:
                 # command not directly supported by backend API
                 raise NotImplementedError(
-                    "The operation {} has not been implemented in {} for the arguments {}.".format(
-                        cmd.op, self.backend, kwargs
-                    )
+                    f"The operation {cmd.op} has not been implemented in {self.backend} for the "
+                    f"arguments {kwargs}."
                 ) from None
 
         # combine the samples sorted by mode, and cast correctly to array or tensor
@@ -788,9 +787,7 @@ class RemoteEngine:
         return job
 
     def __repr__(self):
-        return "<{}: target={}, connection={}>".format(
-            self.__class__.__name__, self.target, self._connection
-        )
+        return f"<{self.__class__.__name__}: target={self.target}, connection={self._connection}>"
 
     def __str__(self):
         return self.__repr__()
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -523,7 +523,7 @@ class LocalEngine(BaseEngine):
             # state object requested
             # session and feed_dict are needed by TF backend both during simulation (if program
             # contains measurements) and state object construction.
-            result._state = self.backend.state(**temp_run_options)
+            result.state = self.backend.state(**temp_run_options)
             if self.backend_name == "bosonic":
                 result._ancilla_samples = self.backend.ancillae_samples_dict.copy()
         return result
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -274,7 +274,7 @@ class BaseEngine(abc.ABC):
                 # Copy the latest measured values in the RegRefs of p.
                 # We cannot copy from prev directly because it could be used in more than one
                 # engine.
-                for k, v in enumerate(self.samples or []):
+                for k, v in enumerate(self.samples):
                     p.reg_refs[k].val = v
 
             # bind free parameters to their values
--- a/strawberryfields/engine.py
+++ b/strawberryfields/engine.py
@@ -30,9 +30,9 @@ from strawberryfields.result import Result
 from strawberryfields.io import to_blackbird
 from strawberryfields.logger import create_logger
 from strawberryfields.program import Program
-from strawberryfields.tdm.tdmprogram import TDMProgram
+from strawberryfields.tdm.tdmprogram import TDMProgram, reshape_samples
 
-from .backends import load_backend
+from .backends import load_backend, BosonicBackend
 from .backends.base import BaseBackend, NotApplicableError
 from ._version import __version__
 
@@ -221,6 +221,9 @@ class BaseEngine(abc.ABC):
             commands that were applied to the backend, the samples returned from the backend, and
             the samples as a dictionary with the measured modes as keys
         """
+        # `_run_program` is called in `BaseEngine._run` but should never be called
+        # from a `BaseEngine` object, in which case an error should be raised.
+        raise NotImplementedError("'run_program' method must be implemented on child class")
 
     def _run(self, program, *, args, compile_options, **kwargs):
         """Execute the given programs by sending them to the backend.
@@ -271,7 +274,7 @@ class BaseEngine(abc.ABC):
                 # Copy the latest measured values in the RegRefs of p.
                 # We cannot copy from prev directly because it could be used in more than one
                 # engine.
-                for k, v in enumerate(self.samples):
+                for k, v in enumerate(self.samples or []):
                     p.reg_refs[k].val = v
 
             # bind free parameters to their values
@@ -288,8 +291,12 @@ class BaseEngine(abc.ABC):
 
             prev = p
 
-        samples = {"output": [self.samples], **self.samples_dict}
-        return Result(samples)
+        ancillae_samples = None
+        if isinstance(self.backend, BosonicBackend):
+            ancillae_samples = self.backend.ancillae_samples_dict.copy()
+
+        samples = {"output": [self.samples]}
+        return Result(samples, samples_dict=self.samples_dict, ancillae_samples=ancillae_samples)
 
 
 class LocalEngine(BaseEngine):
@@ -386,6 +393,12 @@ class LocalEngine(BaseEngine):
 
         # combine the samples sorted by mode, and cast correctly to array or tensor
         samples, samples_dict = self._combine_and_sort_samples(samples_dict)
+
+        if isinstance(prog, TDMProgram) and samples_dict:
+            samples_dict = reshape_samples(samples_dict, prog.measured_modes, prog.N, prog.timebins)
+            # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
+            samples = np.array(list(samples_dict.values())).transpose(1, 0, 2)
+
         return applied, samples, samples_dict
 
     def _combine_and_sort_samples(self, samples_dict):
@@ -431,9 +444,6 @@ class LocalEngine(BaseEngine):
         Returns:
             Result: results of the computation
         """
-        # pylint: disable=import-outside-toplevel
-        from strawberryfields.tdm.tdmprogram import reshape_samples
-
         received_rolled_circuit = False
         if isinstance(program, TDMProgram):
             # priority order for the shots value should be kwargs > run_options > 1
@@ -484,7 +494,7 @@ class LocalEngine(BaseEngine):
 
         # check that post-selection and feed-forwarding is not used together with shots > 1
         for p in program_lst:
-            for c in p.circuit:
+            for c in p.circuit or []:
                 try:
                     if c.op.select and eng_run_options["shots"] > 1:
                         raise NotImplementedError(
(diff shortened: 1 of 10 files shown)


strawberryfields/result.py
```diff
@@ -73,6 +80,15 @@ def __init__(self, result: Mapping, ancilla_samples: Mapping = None) -> None:
...
         self._ancilla_samples = ancilla_samples
 
+    @property
+    def raw(self) -> Mapping:
+        """The raw results dictionary containing any samples and metadata.
+
+        Returns:
+            dict: samples and metadata
+        """
+        return self._result
```
thisac, line 90: I was thinking about whether it would be better not to return the results directly like this, but rather copy it or simply return a view of the dictionary, to avoid users accidentally changing anything in the `Result` object. Not sure if it's worth it or, in that case, how to best do that.

tests/api/test_result.py
```diff
@@ -16,30 +16,99 @@
...
 pytestmark = pytest.mark.api
 
 
+raw_results = {
+    "output": [np.ones((2, 3, 4, 5))],
+    0: [1, 2, 3],
+    1: [4, 5],
+    "other_data": [1, 0, 1, 2],
+    "string": "metadata",
+}
```
Mandrenkov, line 32: I know this is just a test module, but there are a few issues with this result:
1. The keys `0` and `1` are illegal since all keys must be strings.
 2. The list values are illegal since all values (with the exception of the `"output"` value) must be NumPy arrays.
 3. The string value `"metadata"` is illegal (see previous point).

tests/api/test_result.py
```diff
@@ -16,30 +16,99 @@
...
+
+    def test_raw(self):
+        """Test that raw results are correctly returned."""
+        result = Result(raw_results)
+        assert result.raw == raw_results
+
+    def test_samples_dict(self):
+        """Test that ``samples_dict`` is correctly returned."""
+        result = Result(raw_results)
+        assert result.samples_dict == {0: [1, 2, 3], 1: [4, 5]}
```
Mandrenkov, line 52: I just realized that SF v0.20.0 may not be handling the samples dictionary correctly - the `Result` class should be checking for _strings_ in the raw result which resemble integers.

tests/api/test_result.py
```diff
@@ -16,30 +16,99 @@
...
+
+    def test_ancilla_samples(self):
+        """Test that ancilla samples are correctly returned."""
+        ancilla_samples = {0: [0, 1], 2: [1, 3]}
+        result = Result(raw_results, ancilla_samples=ancilla_samples)
+        assert result.ancilla_samples == ancilla_samples
+
+    def test_state(self):
+        """Test that the state can be correctly set and returned."""
```
Mandrenkov, line 61: This also tests that a state cannot be set twice (unless, of course, the state is set to `None`).

Mandrenkov (commented):
Looks great! 🌟

My only concern is regarding the types of the raw result keys and values.

strawberryfields/result.py
```diff
@@ -146,9 +175,25 @@ def state(self) -> Optional[BaseState]:
...
+
+        Raises:
+            TypeError: if state is already set
+            ValueError: if result have ``samples_dict`` (i.e, do not come from a local simulation)
+        """
+        if self._state:
+            raise TypeError("State already set and cannot be changed.")
+        # Samples from local simulation jobs will always have a samples_dict containing the same
+        # samples organized by measured mode. `self.samples` is either `None` or an array.
+        if self.samples is not None and len(self.samples) != 0 and not self.samples_dict:
```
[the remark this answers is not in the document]
thisac: This one is a bit tricky. If `self.samples` is an ndarray, casting it to a bool raises an error,
```pycon
>>> bool(np.array([1, 2, 3]))
  File "<stdin>", line 1, in <module>
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
so using only `if self.samples ...` doesn't work, and using `len(self.samples)` doesn't work if it's `None`. I agree that this looks a bit ugly, but I couldn't figure out a better way to do this. Do you have any other ideas?

A workaround could be to use `if np.any(self.samples) ...` which technically works, but isn't supported by type hinting if `self.samples` is `None` (which causes annoying squiggles to pop up in my editor), or `if np.any(self.samples or []) ...`, but I'm not too fond of either of them.

tests/api/test_result.py
thisac: I'm not sure we ever decided that keys must be strings and values must be arrays, or did we? At least not on the SF side. The idea behind the `samples_dict` was to check for integer keys and have them correspond to modes, and I though metadata could potentially be strings (if needed).

tests/api/test_result.py
thisac: The usage of integers as keys was deliberate here. See my other comment above.

strawberryfields/result.py
thisac: Sure, I agree. 👍

strawberryfields/result.py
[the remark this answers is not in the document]
Mandrenkov: Ah, okay! I see what you mean. In that case, I agree that the current implementation is the way to go.

tests/api/test_result.py
Mandrenkov: Actually, this restriction is enforced by the .npz format which stores data in the form `dict[str, np.ndarray]`.

tests/api/test_result.py
Mandrenkov: Let's continue this discussion in the above comment!

tests/api/test_result.py
heltluke: Yeah, real data will never look like this. For example, on QKD_01, raw results presently look like
```
{'alphas': array([-0.00023017-1.17238805e-03j,  0.00163755+1.69466410e-03j,
         0.00170546+1.58876506e-03j, ..., -0.00131527+1.67085418e-03j,
         0.0005287 -7.48599295e-05j,  0.00027381+2.05610324e-04j]),
 'output': [array([ 9.29057598e-04,  1.29526854e-03,  2.39390135e-03, ...,
         -1.32498145e-03,  1.77919865e-05, -1.01980567e-03])]}
```
where "alphas" is the metadata coming from the `.npz` (that we save with the key "alphas").

tests/api/test_result.py
```diff
@@ -49,19 +118,19 @@ def test_tdm_print(self, capfd):
...
 
-    def test_unkown_shape_print(self, capfd):
+    def test_unknown_shape_print(self, capfd):
         """Test that printing a result object with samples with an unknown shape
         provides the correct output."""
         samples = np.ones((2, 3, 4, 5))
         result = Result({"output": [samples]})
         print(result)
-        out, err = capfd.readouterr()
+        out, _ = capfd.readouterr()
```
heltluke, line 133: If you wanted to save yourself the throwaway `_`, I note that the `pytest` examples seem to recommend writing this as `captured = capfd.readouterr()` and making assertions against `captured.out`.

strawberryfields/result.py
heltluke: Is there a clear use case for accessing the results like this?

heltluke:
Gotta love `f-strings` :heart:

strawberryfields/result.py
```diff
@@ -146,9 +175,25 @@ def state(self) -> Optional[BaseState]:
         """
         return self._state
 
+    @state.setter
+    def state(self, state) -> None:
```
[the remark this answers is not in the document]
thisac: Added a check that the set state is either a `BasisState` or `None`, simplifying the process of attempting to add a non-existing state which was being made in SF.

tests/api/test_result.py
thisac: Good idea! Updated!

strawberryfields/result.py
thisac: Maybe not. It'd just be a way of easily getting direct access to the results obtained from XCC. 🤔

Mandrenkov: Perhaps it could be useful for debugging? But then again, users can just access `_result` directly.

strawberryfields/engine.py
```diff
@@ -221,6 +221,9 @@ def _run_program(self, prog, **kwargs):
             commands that were applied to the backend, the samples returned from the backend, and
             the samples as a dictionary with the measured modes as keys
         """
+        # `_run_program` is called in `BaseEngine._run` but should never be called
+        # from a `BaseEngine` object, in which case an error should be raised.
+        raise NotImplementedError("'run_program' method must be implemented on child class")
```
Mandrenkov, line 226: Are we sure `abc.abstractmethod` won't raise something similar already?

strawberryfields/result.py
heltluke: This :point_up: was exactly my thinking

strawberryfields/result.py
```diff
@@ -75,17 +75,24 @@ class Result:
...
         """The raw results dictionary containing any samples and metadata.
 
+        .. warning::
+
+            The raw values are returned by reference, meaning that any changes to
+            them will be reflected in the ``Result`` object.
+
         Returns:
-            dict: samples and metadata
+            Mapping[str, Union[np.ndarray, List[np.ndarray]]: samples and metadata
```
Mandrenkov, line 95: Just curious: is the convention in PL and SF to use `list` or `List` in the Sphinx docstrings?
```suggestion
            Mapping[str, Union[np.ndarray, list[np.ndarray]]: samples and metadata
```

strawberryfields/engine.py
```diff
@@ -374,19 +381,24 @@ def _run_program(self, prog, **kwargs):
...
+                    f"The operation {cmd.op} has not been implemented in {self.backend} for the "
+                    f"arguments {kwargs}."
                 ) from None
 
         # combine the samples sorted by mode, and cast correctly to array or tensor
         samples, samples_dict = self._combine_and_sort_samples(samples_dict)
+
+        if isinstance(prog, TDMProgram) and samples_dict:
+            samples_dict = reshape_samples(samples_dict, prog.measured_modes, prog.N, prog.timebins)
+            # transpose the samples so that they have shape `(shots, spatial modes, timebins)`
```
heltluke, line 399: Why not simply have the simulator return results in the desired form? There was certainly a bit of effort on the hardware side to get them in `(shots, sptatial modes, timebins)` format.

Mandrenkov (approved):
Some of the new additions are a little beyond my domain of expertise but the rest looks good!

One thing we should decide before merging is whether `Result.raw` is still needed.

strawberryfields/result.py
```diff
@@ -80,7 +103,7 @@ def samples(self) -> Optional[np.ndarray]:
         Returned measurement samples will have shape ``(shots, modes)``.
 
         Returns:
-            array[array[float, int]]: measurement samples returned from
+            ndarray[ndarray[float, int]]: measurement samples returned from
```
heltluke, line 106: Should this read `np.ndarray` or even `numpy.ndarray`?

strawberryfields/result.py
```diff
@@ -95,22 +118,41 @@ def samples(self) -> Optional[np.ndarray]:
...
+            the prior measured sample on the same mode is replaced. The samples
+            dictionary, on the other hand, keeps _all_ measurements.
+
+        Returns:
+            Mapping[int, List]: mode index associated with the list of measurement outcomes
+        """
+        return self._samples_dict
+
+    @property
+    def metadata(self) -> Mapping[str, np.ndarray]:
```
heltluke, line 142: Nice!

tests/api/test_result.py
```diff
@@ -16,53 +16,122 @@
...
+
+
 class TestResult:
     """Tests for the ``Result`` class."""
 
+    def test_samples(self):
+        """Test that ``samples`` is correctly returned."""
+        result = Result(raw_results)
+        assert result.samples is not None
+        assert np.equal(result.samples, np.ones((2, 3, 4, 5))).all()
```
heltluke, line 40: Why not
```suggestion
        assert np.array_equal(result.samples, np.ones((2, 3, 4, 5)))
```
?

tests/api/test_result.py
```diff
@@ -16,53 +16,122 @@
...
 
 from strawberryfields.result import Result
 
 pytestmark = pytest.mark.api
 
 
+raw_results = {
+    "output": [np.ones((2, 3, 4, 5))],
+    "meta_array": np.array([1, 0, 1, 2]),
+    "meta_matrix": np.array([[1, 2], [3, 4]]),
```
heltluke, line 29: Might make sense to assign the values (`[np.ones((2, 3, 4, 5))]`, `np.array([1, 0, 1, 2])`, `np.array([[1, 2], [3, 4]])`) to variables above, to reduce code repetition when making assertions below.

heltluke (approved):

strawberryfields/result.py
heltluke: Only able to be `list` after Python 3.9, right? So dependent on the Python versions supported?

strawberryfields/engine.py
thisac: Oh, maybe. I basically added this to get rid of the type hint issue with calling `run_engine` from `BaseEngine._run`. Without it, my type-hinting engine assumes that `None` is being returned here, and thus complains. Perhaps I should just ignore it though. 🤔

strawberryfields/result.py
thisac: > Only able to be list after Python 3.9, right? So dependent on the Python versions supported?

I didn't know that Python 3.9 supported nested types in `list`. That's basically why I used `List` here instead, which is probably still a good idea if we'd like to have subtypes (although this technically only affects type hints, and not types in docstrings, I think).

> is the convention in PL and SF to use list or List in the Sphinx docstrings?

I don't think there is a convention here. `list` is definitely more common, but there are quite a few `List` occurrences as well across SF and PL.

strawberryfields/engine.py
thisac: This only affects simulators, which return the results in the wrong order and thus need to be reshaped and transposed. So, if I recall correctly, this is needed for the results to have the same shape as the hardware.

tests/api/test_result.py
thisac: Good point, didn't think of it. 👍

strawberryfields/engine.py
heltluke: I understand. I just find it a little odd that it's been left to SF to get results in the right shape, rather than demanding simulators send results to SF in the desired shape.

strawberryfields/result.py
thisac: Perhaps we should just remove it then? I see no real reason to have it, to be honest, since there will probably not be anything in `Result.raw` that isn't already provided via `Result.samples` or `Result.metadata`. 🤔

strawberryfields/engine.py
thisac: Oh, I see. This is mostly due to the way that TDM programs are simulated. They need to be unrolled before being run, so what is _actually_ being simulated isn't specifically a TDM circuit. The backend itself doesn't know if it's working with a TDM program or not, and thus returns the measurements in the wrong order/on the wrong modes.

I think the `_run_program` method is the deepest we can put this transformation.

strawberryfields/result.py
josh146: > I don't think there is a convention here. list is definitely more common, but there are quite a few List occurrences as well across SF and PL.

Actually, this should be `list`, any instances of `List` were accidental or missed during code review! 

This is primarily because our Sphinx conventions predate Python type hinting

strawberryfields/result.py
josh146: Here our convention is `ndarray`, or in some cases, `array_like`, to match the numpy documentation

josh146: e.g., https://numpy.org/doc/stable/reference/generated/numpy.dot.html#numpy.dot

strawberryfields/engine.py
heltluke: Understood.

Very cool that the backend/simulator is agnostic to the program type.

strawberryfields/engine.py
Mandrenkov: Oh, I see! In that case, I'm okay with either approach (although I guess it's better safe than sorry?).

PR #664, opened 2021-12-08
Update readme to markdown

thisac:
**Context:**
The current readme doesn't work well with PyPI due to the inclusion of raw directives.

* Changes the format of the readme to markdown instead of rst.

**Benefits:**
PyPI uploads should work.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


.github/CHANGELOG.md
```diff
@@ -44,6 +44,9 @@
 
 <h3>Documentation</h3>
 
+* Port README to markdown.
```
josh146, line 47:
```suggestion
* README has been ported to Markdown.
```

setup.py
```diff
@@ -49,7 +49,7 @@
                                           "apps/data/sample_data/*"]},
     "include_package_data": True,
     "description": "Open source library for continuous-variable quantum computation",
-    "long_description": open("README.rst", encoding="utf-8").read(),
+    "long_description": open("README.md", encoding="utf-8").read(),
     "long_description_content_type": "text/x-rst",
```
josh146, line 53: I think this line should also be updated, otherwise PyPI will continue to try parsing it as a ReST file

josh146: See https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/ (I think it should be `text/markdown` if I am reading it correctly?)

josh146 (approved):
Nice, thanks @thisac! Just one line to update in `setup.py`, but apart from that looks good to go

setup.py
thisac: Nice, thanks @josh146. Read through the link and updated it!

PR #669, opened 2021-12-21
Update plot.py - Show plots when not running in a notebook

Hosseinberg:
The code works fine when using it in a notebook but the graph does not show when using other IDEs such as Spyder. The lines are added to make sure that when it is not running on a notebook, the graph will be displayed in a browser.


strawberryfields/apps/plot.py
```diff
@@ -130,7 +130,10 @@ def graph(g: nx.Graph, s: Optional[list] = None, plot_size: Tuple = (500, 500)):
         import plotly.graph_objects as go
     except ImportError:
         raise ImportError(plotly_error)
-
+    
+    if (get_ipython().__class__.__name__ != 'ZMQInteractiveShell'):
+        plotly.io.renderers.default = "browser"
```
thisac, line 135: I believe `get_ipython` is only available in IPython terminals and notebooks, so this would raise an `NameError` if used in regular Python. It would need to be wrapped in a try-except block.

E.g., something like the following could work:
```python
try:
    in_notebook = get_ipython().__class__.__name__ == 'ZMQInteractiveShell'
except NameError:
    in_notebook = False

if not in_notebook:
    pio.renderers.default = "browser"
```
You also need to `import plotly.io as pio` in the above try-except block to be able to access the `pio.renderers.default` property. 🙂

thisac (changes requested):
Thanks for the contribution @Hosseinberg. This is a good catch! 💯 

The format check is currently failing, so it'd be great if you could run the Black autoformatter, with max-line length set to 100, on the file that you've changed. Don't forget to add your change to the changelog as well, and include your name in the list of contributors for the development release. It can be found [here](https://github.com/XanaduAI/strawberryfields/blob/master/.github/CHANGELOG.md)!

thisac:
@Hosseinberg Any updates on this PR? Let us know if you're still working on it or if you'd like some more assistance getting it completed and merged.

Hosseinberg:
The codefactor found 5 issues and most of them are "Consider explicitly re-raising using the 'from' keyword" which I think can be solved using "from None". The other issues are related to trailing whitespace that I could not find. The problem is that the lines that codefactor is referring to for the issues are different from the updated [patch-1](https://github.com/Hosseinberg/strawberryfields/tree/patch-1). I think I need a little help here.

strawberryfields/apps/plot.py
```diff
@@ -130,7 +130,18 @@ def graph(g: nx.Graph, s: Optional[list] = None, plot_size: Tuple = (500, 500)):
...
+    try:
+        in_notebook = get_ipython().__class__.__name__ == "ZMQInteractiveShell"
+    except NameError:
+        in_notebook = False
+
+    if not in_notebook:
+        try:
+            import plotly.io as pio
+        except ImportError:
+            raise ImportError(plotly_error) from None
```
thisac, line 142: The import could just be put inside the try-except block above (right next to the `import plotly.graph_objects as go` line) instead of creating a new one.

thisac (commented):
Thanks for working on this @Hosseinberg! Don't forget to add an entry to the changelog and your name to the contributors list.

> The problem is that the lines that codefactor is referring to for the issues are different from the updated patch-1. I think I need a little help here.

I can handle this, and fix or override the issues, since they weren't caused by any of your code additions.

.github/CHANGELOG.md
```diff
@@ -5,13 +5,17 @@
 <h3>Breaking Changes</h3>
 
 <h3>Bug fixes</h3>
+* Makes it possible to show the graphs when not running as a notebook.
+  [(#669)](https://github.com/XanaduAI/strawberryfields/pull/669)
```
thisac, line 9:
```suggestion
* It's now possible to show graphs using the plot apps layer when not run in notebooks.
  [(#669)](https://github.com/XanaduAI/strawberryfields/pull/669)
```

thisac:
I believe this is almost ready to be merged @Hosseinberg! You only need to run `make format` in the root folder in this branch (make sure that you use the latest version of [Black](https://pypi.org/project/black/), which I believe is 21.12b0). I can take care of the CodeFactor issues. Let me know if you're having any issues!

strawberryfields/apps/plot.py
```diff
@@ -102,34 +97,36 @@ def _edge_coords(g: nx.Graph, l: dict) -> dict:
 
 def graph(g: nx.Graph, s: Optional[list] = None, plot_size: Tuple = (500, 500)):  # pragma: no cover
```
thisac, line 98: These empty lines shouldn't be removed. Some of them (e.g., the ones the `.. image::` directive below) are causing the docs build to fail. Not sure if these were somehow removed when you attempted to format the file, but they need to be added back before merging.

thisac (changes requested):

thisac (approved):
Thanks for all the work on this one! 🏅

PR #674, opened 2022-01-17
fix teleportation.py beamsplitter phase (to 0)

jonschlipf:
**Context:**
The teleportation example fails to teleport the state. Setting the BS phase to 0 apparently fixes this for various numeric examples. The phase was 0 in the previous version of this example as well as the theory part of the html example in the docs.

The beam splitter phase in the teleportation example was set to 0.

**Benefits:**
The teleported state on q[2] can now be confirmed to have its probability distribution around the original state (via Wigner plots etc.).

**Possible Drawbacks:**
None

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -15,7 +15,7 @@
   specification.
   [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
   
-* The teleportation tutorial `examples/teleportation.py` now uses the correct value (now ϕ=0 instead of ϕ=π/2) for the phase shift of the beam     splitters.
+* The teleportation tutorial `examples/teleportation.py` now uses the correct value (now `phi = 0` instead of `phi = np.pi / 2`) for the phase shift of the beamsplitters.
   [(#674)](https://github.com/XanaduAI/strawberryfields/pull/674)
 
 <h3>Documentation</h3>
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -14,6 +14,9 @@
   circuit layout. Without a layout, the gate parameters cannot be validated against the device
   specification.
   [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
+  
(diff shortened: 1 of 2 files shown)


examples/teleportation.py
```diff
@@ -29,7 +29,7 @@
     Squeezed(2) | q[2]
 
     # apply gates
-    BS = BSgate(pi / 4, pi)
+    BS = BSgate(pi / 4,0)
```
thisac, line 32: Just to conform to PEP 8.

```suggestion
    BS = BSgate(pi / 4, 0)
```

thisac (commented):
Thanks for finding this @jonschlipf! It is indeed correct and should be changed. Another option would be to negate the final `Zgate`; both work equally fine. 🙂 

Don't forget to add an entry to the [changelog](https://github.com/XanaduAI/strawberryfields/blob/master/.github/CHANGELOG.md) as well as you name to the list of contributors.

jonschlipf:
Thank you @thisac ! I modified the changelog with a bullet item and my name in the contributor list.

.github/CHANGELOG.md
```diff
@@ -14,14 +14,17 @@
   circuit layout. Without a layout, the gate parameters cannot be validated against the device
   specification.
   [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
+  
+* The teleportation tutorial `examples/teleportation.py` now uses the correct value (now ϕ=0 instead of ϕ=π/2) for the phase shift of the beam     splitters.
+  [(#674)](https://github.com/XanaduAI/strawberryfields/pull/674)
```
thisac, line 19: Although I really like the look of the ϕ and the π, I'd say it's better to go with the names used in the function, programmatically, here instead (`phi` and `np.pi`).
```suggestion
* The teleportation tutorial `examples/teleportation.py` now uses the correct value (now `phi = 0` instead of `phi = np.pi / 2`) for the phase shift of the beamsplitters.
  [(#674)](https://github.com/XanaduAI/strawberryfields/pull/674)
```

thisac (commented):
Fantastic! Thank you very much. Just one last comment, and then it's ready to be merged.

thisac (approved):
Thanks for the contribution @jonschlipf!

PR #675, opened 2022-01-18
fix failing logger tests

sduquemesa:
**Context:**
Tests in `test_logger.py` fail locally when running `pytest` directly.

This PR solves the failing tests by
- fixing the failing stream assertion on `test_default_sf_logger`,
- by setting the level of the root logger instead of the module logger on `test_custom_logger_before_sf_logger_with_higher_level`.

Additional to the above, some unused arguments on test functions were removed.

**Benefits:**
Logger test no longer fail when using `pytest` directly.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -14,6 +14,9 @@
   circuit layout. Without a layout, the gate parameters cannot be validated against the device
   specification.
   [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
+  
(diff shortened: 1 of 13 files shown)


thisac (approved):
Great addition! 🎉 

It'd also be great to add `pytestmark = pytest.mark.frontend
` to the top of `test_logger.py` here, so that the tests are run with `make test` (and also on CI). This seems to have been the reason we didn't catch this earlier.

PR #679, opened 2022-02-07
Certificate in device

thisac:
**Context:**
The certificate is needed for the TD3 pipeline. It makes sense adding it to the `DeviceSpec` class (which contains the device specification). This also warrants a name change of the class, calling it `Device` rather than `DeviceSpec`(this also makes sense since we've had a tendency to call `DeviceSpec` objects simply `device`).

*Note: The plan is to later to remove the `sf.Device` class altogether and only keep the `xcc.Device` class. Some further changes need to be made for this to work properly, although this would probably be the better solution in the long run, avoiding using multiple abstractions for the same thing.*

* The `DeviceSpec` class is renamed `Device`.
* A certificate can optionally be passed to the `Device` initialization along with the specification.
* A `certificate` method is added to the `Device` class, which returns the certificate attribute.

**Benefits:**
* The device certificate can be accessed via `Device` class objects.
* There's less confusion regarding what a device specification is (is it a dictionary or a `DeviceSpec` object?).

**Possible Drawbacks:**
* There's now very little difference between `sf.Device` and `xcc.Device`.

**Related GitHub Issues:**
None


strawberryfields/device.py
```diff
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-This module contains a class that represents the specifications of
-a device available via the API.
+This module contains a class that represents a device available via the API.
```
Mandrenkov, line 15: Can we say "Xanadu Cloud" here to be more specific?
```suggestion
This module contains a class that represents a device on the Xanadu Cloud.
```

strawberryfields/device.py
```diff
@@ -53,12 +57,12 @@ def target(self) -> str:
 
     @property
     def layout(self) -> Optional[str]:
-        """str: Returns a string containing the Blackbird circuit layout."""
+        """str: A string containing the Blackbird circuit layout."""
```
Mandrenkov, line 60: As per the [PennyLane development guide](https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html), should we preserve the "Returns a ..." wording?

strawberryfields/device.py
```diff
@@ -35,16 +34,21 @@ class DeviceSpec:
             - modes (int): number of modes supported by the target
             - compiler (list): list of supported compilers
             - gate_parameters (dict): parameters for the circuit gates
+
+        certificate (dict): the certificate of the device
```
Mandrenkov, line 38: To match the description of `spec` above:
```suggestion
        certificate (dict, optional): dictionary representing the device certificate
```

strawberryfields/device.py
```diff
@@ -35,16 +34,21 @@ class DeviceSpec:
...
     """
 
-    def __init__(self, spec: Mapping[str, Any]) -> None:
+    def __init__(
+        self, spec: Mapping[str, Any], certificate: Optional[Mapping[str, Any]] = None
```
Mandrenkov, line 42: Just curious: if we use `spec`, why not use `cert`?
```suggestion
        self, spec: Mapping[str, Any], cert: Optional[Mapping[str, Any]] = None
```

.github/CHANGELOG.md
```diff
@@ -2,12 +2,18 @@
 
 <h3>New features since last release</h3>
 
-* `DeviceSpec.layout` and `DeviceSpec.gate_parameters` may now return `None`. This can happen
+* `Device.layout` and `Device.gate_parameters` may now return `None`. This can happen
   when a remote simulator device is used.
   [(#661)](https://github.com/XanaduAI/strawberryfields/pull/661)
 
+* A `Device.certificate` method is added which returns the hardware device certificate.
+  [(#XXX)](https://github.com/XanaduAI/strawberryfields/pull/XXX)
```
Mandrenkov, line 10:
```suggestion
  [(#679)](https://github.com/XanaduAI/strawberryfields/pull/679)
```

.github/CHANGELOG.md
```diff
@@ -2,12 +2,18 @@
...
+
 <h3>Breaking Changes</h3>
 
+* `DeviceSpec` is renamed to `Device`.
+  [(#XXX)](https://github.com/XanaduAI/strawberryfields/pull/XXX)
```
Mandrenkov, line 15:
```suggestion
  [(#679)](https://github.com/XanaduAI/strawberryfields/pull/679)
```

tests/api/test_devicespec.py
```diff
@@ -83,17 +92,18 @@
...
 
-class TestDeviceSpec:
-    """Tests for the ``DeviceSpec`` class."""
+class TestDevice:
+    """Tests for the ``Device`` class."""
 
     def test_initialization(self):
         """Test that the device spec class initializes correctly."""
-        spec = DeviceSpec(spec=device_dict)
+        device = Device(spec=device_dict, certificate=device_certificate)
```
Mandrenkov, line 100: Should we rename this too while we're at it?
```suggestion
        device = Device(spec=device_spec, certificate=device_certificate)
```

Mandrenkov (commented):
Looks good to me! 👍

I will approve once the changelog PR references have been updated.

strawberryfields/device.py
thisac: Good point. No reason really. Cert sounds good! 👍

strawberryfields/device.py
thisac: It was changed to follow the rest of the docstrings in the file. Does the guide say that and action "Returns a ..." is preferred to a descriptive first line? I could only find that is should be a "Single sentence that summarizes the function.". 🤔

Mandrenkov: Oh, sorry, I was specifically referring to:
> Document all methods and properties with docstrings below the method declaration, as you would for functions above.

Mandrenkov (approved):

strawberryfields/device.py
thisac: Oh, I see. Yes, definitely. That makes sense. I guess I was just a bit lazy here. 😅

josh146: @Mandrenkov which part of the documentation guide are you referring to?

strawberryfields/engine.py
```diff
@@ -609,22 +609,20 @@ def connection(self) -> xcc.Connection:
         )
 
     @property
-    def device_spec(self) -> DeviceSpec:
+    def device(self) -> Device:
         """The specification of the target device.
 
         Returns:
-            DeviceSpec: the device specification
+            ~strawberryfields.Device: the device specification
```
josh146, line 616:
```suggestion
            .strawberryfields.Device: the device specification
```

strawberryfields/engine.py
```diff
@@ -609,22 +609,20 @@ def connection(self) -> xcc.Connection:
         )
 
     @property
-    def device_spec(self) -> DeviceSpec:
+    def device(self) -> Device:
         """The specification of the target device.
```
josh146, line 613: I must admit, the fact that the docstring summary doesn't change makes me wonder if the renaming is necessary?

strawberryfields/engine.py
```diff
@@ -609,22 +609,20 @@ def connection(self) -> xcc.Connection:
...
         """
-        if self._spec is None:
-            target = self.target
-            connection = self.connection
-            device = xcc.Device(target=target, connection=connection)
-            self._spec = DeviceSpec(spec=device.specification)
-        return self._spec
+        if self._device is None:
+            device = xcc.Device(target=self.target, connection=self.connection)
+            self._device = Device(spec=device.specification, cert=device.certificate)
```
josh146, line 624: Yeah we should definitely aim to merge them into a single abstraction, this is confusing!

At the moment, what is the main difference between the two? Is it that:

- `xcc.Device` is an object that represents the API contract with the Device, and is used for submitting and querying jobs?
- `sf.Device` stores the device specification and certificate?

josh146: If that is the case, should the names reflect this difference?

tests/api/test_devicespec.py
```diff
@@ -46,6 +46,15 @@
...
 }
 
+device_certificate = {
+    "target": "X8_01",
+    "finished_at": "2022-02-07T15:29:48.080220+00:00",
+    "identity_unitary": {},
+    "shots_per_benchmark": 999999,
+    "anti_identity_unitary": {},
+    "laser_wavelength_meters": 42,
+}
```
josh146, line 56: Will all device certificates have these fields? Or is this just an example of possible fields?

josh146 (approved):
Looks good from my end @thisac! Good idea to split this into a separate PR.

strawberryfields/device.py
Mandrenkov: @josh146, I'm referring to the _Documentation_ page of the PennyLane development guide under the [**Classes** section](https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#classes) of the **Docstrings** header.

strawberryfields/engine.py
Mandrenkov: Should this docstring be updated to say "device" instead of "device specification"?

strawberryfields/engine.py
Mandrenkov: > `xcc.Device` is an object that represents the API contract with the Device

Correct! It is the canonical way to access any information about a device on the Xanadu Cloud, including device specifications, certificates, expected uptime, etc.

> and is used for submitting and querying jobs?

Nope. That's what `xcc.Job` is for. You only use `xcc.Device` for querying information about _devices_.

> `sf.Device` stores the device specification and certificate?

Yes, although it also has a few helper functions for interacting with `sf.Program` objects.

thisac: There's actually not much of a difference between what the two classes contain. Apart from the specification and the certificate, `xcc.Device` contains a few the expected uptime and status of the device. So I actually think that them having the same name here makes sense (but we should probably merge them).

> xcc.Device [...] is used for submitting and querying jobs?

Not really, that's the `xcc.Job` class. The `xcc.Device` only fetches the specification, certificate and details dictionaries, and caches them.

strawberryfields/device.py
Mandrenkov: Okay, it turns out I misunderstood the intent of that line in the PennyLane development guide. Carry on! 🙂

PR #682, opened 2022-02-15
`shots=None` will not execute any measurement

sduquemesa:
**Context:**
Simulating a circuit requires to execute the measurements to return results. This execution can take long time, however there are cases where no execution is required, e. g., when calculating the covariance matrix of the circuit.

If `shots=None` the backend bypasses the execution (i.e. the `apply` method) of any op of class `Measurement` that belongs to the circuit.

**Benefits:**
Users can define circuits and bypass measurements to avoid lengthy executions of simulations if not needed.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -9,6 +9,8 @@
 * A `Device.certificate` method is added which returns the hardware device certificate.
   [(#679)](https://github.com/XanaduAI/strawberryfields/pull/679)
 
+* Setting `shots=None` on the engine or program run options will not execute any measurements applied on the circuit. [(#682)](https://github.com/XanaduAI/strawberryfields/pull/682)
+
 <h3>Breaking Changes</h3>
 
 * `DeviceSpec` is renamed to `Device`.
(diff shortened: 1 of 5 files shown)


.github/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 * A `Device.certificate` method is added which returns the hardware device certificate.
   [(#679)](https://github.com/XanaduAI/strawberryfields/pull/679)
 
+* Setting `shots=None` on the engine or program run options will not execute any measurements applied on the circuit. [(#682)](https://github.com/XanaduAI/strawberryfields/pull/682)
```
thisac, line 12:
```suggestion
* Setting `shots=None` in the engine or program run options will not execute any measurements applied on the circuit. [(#682)](https://github.com/XanaduAI/strawberryfields/pull/682)
```

thisac (changes requested):
Realized that tests are missing for the TDM program case when `shots=None`. [These tests](https://github.com/XanaduAI/strawberryfields/blob/master/tests/frontend/test_tdmprogram.py#L705-L731) could be adapted similar to the tests in `test_engine.py`. Other than that, looks 💯 !

sduquemesa:
@thisac, just added the test cases for the TDM program 🚀

thisac (approved):
🥳

PR #685, opened 2022-02-28
Run CI tests without Tensorflow

### LINKED ISSUE #641: CI tests without TensorFlow installed
### 

- [X] I have searched exisisting GitHub issues to make sure the feature request does not already exist.

### Feature details

Add a CI action that runs the relevant tests in an environment without TensorFlow installed.

Currently, no tests are being run without TensorFlow installed (with the exception of some specific tests where TensorFlow is in some way mocked). This makes it easy to mistakenly add code that is dependent on TensorFlow to core Strawberry Fields, which doesn't have TensorFlow as a requirement.

**Example:**
Importing a function from the TensorFlow backend into other parts of Strawberry Fields would bypass the check for whether TensorFlow is installed or not. For example, adding the following in `strawberryfields/__init__.py` would implicitely try to import TensorFlow each time Strawberry Fields is imported (e.g., `import strawberryfields as sf`):
```python
# strawberryfields/__init__.py
from tensorflow.backends.ops import update_symplectic
```

### Implementation

Add a GitHub action that creates an environment without TensorFlow installed, and runs the parts of the test suite that do not depend on TensorFlow. This should be run on each push to PR and merge.

### How important would you say this feature is?

1: Not important. Would be nice to have.

### Additional information

Front-end tests have recently been fixed to be able to run without TensorFlow (#639).

sduquemesa:
**Context:**
Currently, no tests are being run without TensorFlow installed. This makes it easy to mistakenly add code that is dependent on TensorFlow to core Strawberry Fields, which doesn't have TensorFlow as a requirement.

Adds a CI action that runs the relevant tests in an environment without TensorFlow installed. 

This is done by 
- adding a new strategy to the `core-test` job where pytest will run all tests except those marked with `"tf"`. That is, setting `OPTIONS="not tf"`,
- creating a new environment variable `TF_DEPENDENCIES="tensorflow|tensorboard"` listing the dependencies to be removed from the `requirements-ci.txt` file,
- using `awk` to remove dependencies listed on `TF_DEPENDENCIES` from the `requirements-ci.txt` file when `OPTIONS="not tf"`.

**Benefits:**
Avoid to mistakenly add code that is dependent on TensorFlow to core Strawberry Fields.

**Possible Drawbacks:**
Longer CI run times

**Related GitHub Issues:**
Closes #641

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,6 +19,7 @@ jobs:
       SYMPY_USE_CACHE: "no"
       OPTIONS: ${{ matrix.options }}
       BATCHED: ${{ matrix.batched }}
+      SKIP_DEPENDENCIES: "tensorflow|tensorboard"
 
     strategy:
       matrix:
@@ -29,6 +30,7 @@ jobs:
             - {BATCHED: 1, OPTIONS: "tf and pure"}
             - {BATCHED: 1, OPTIONS: "tf and mixed"}
             - {BATCHED: 0, OPTIONS: "bosonic"}
+            - {BATCHED: 0, OPTIONS: "not tf"}
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.4.1
@@ -46,6 +48,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          [ "$OPTIONS" = "not tf" ] && gawk -i inplace '!/('$SKIP_DEPENDENCIES')/' requirements-ci.txt
           python3 -m pip install --upgrade pip
           pip install -r requirements-ci.txt
           pip install wheel codecov pytest pytest-cov pytest-randomly pytest-mock pytest-logger --upgrade
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,7 +19,7 @@ jobs:
       SYMPY_USE_CACHE: "no"
       OPTIONS: ${{ matrix.options }}
       BATCHED: ${{ matrix.batched }}
-      SKIP_DEPENDENCIES: "tensorflow|tensorboard"
+      TF_DEPENDENCIES: "tensorflow|tensorboard"
 
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


josh146:
😍

.github/workflows/tests.yaml
```diff
@@ -19,6 +19,7 @@ jobs:
       SYMPY_USE_CACHE: "no"
       OPTIONS: ${{ matrix.options }}
       BATCHED: ${{ matrix.batched }}
+      SKIP_DEPENDENCIES: "tensorflow|tensorboard"
```
thisac, line 22: Perhaps this is better called something like `TF_DEPENDENCIES`, since it only lists the dependencies that should be skipped if running `"not tf"`. I think that might make more sense here not to confuse it with skipping other dependencies.

It could be slightly confusing is someone looks at the tests and sees the following (which are running _with_ TensorFlow installed but still lists the environment variables).

Screen Shot 2022-03-01 at 5 17 49 PM

.github/workflows/tests.yaml
```diff
@@ -46,6 +48,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          [ "$OPTIONS" = "not tf" ] && gawk -i inplace '!/('$SKIP_DEPENDENCIES')/' requirements-ci.txt
```
thisac, line 51: Could this persist between test runs? E.g., if this is executed before the regular tests (which require TensorFlow), could the `requirements-ci.txt` have been changed due to this?

thisac (changes requested):
Great to have this @sduquemesa! I just have a comment and a question, but overall it's a very neat solution which seems to work great! 💯

.github/workflows/tests.yaml
sduquemesa: When using the [matrix](https://github.com/XanaduAI/strawberryfields/blob/f6038e060756e6948b3173875e566b22581872c3/.github/workflows/tests.yaml#L25-L33) strategy, every entry defines a new job and each job provisions a different virtual environment. This modification will only apply to the env with `OPTIONS="not tf"` and will not persist to other envs.

thisac (approved):
Fantastic! 💯

PR #686, opened 2022-03-04
`sf.Program` equality operator

### LINKED ISSUE #135: Add equality comparison for Programs

thisac:
**Context:**
There's currently no easy way to check for program equality. There's a function in the test suite (existing in three different places) that checks for program equivalence by checking the isomorphism equality of two different programs, but it's not possible (until now) to simply check it using `prog_1 == prog_2`.

Tracked by [[sc-13847]](https://app.shortcut.com/xanaduai/story/13847/program-equality-method?cf_workflow=500000106&ct_workflow=all&vc_group_by=day).

* Moves the `program_equivalence` function to `program.utils.py`.
* Adds an `__eq__` method to `sf.Program` which checks (strictly) that all the operations and parameters are identical.
* Add a `Program.equivalence` convenience method which calls `program_equivalence`.

**Benefits:**
It's now possible to check the equality of two programs much easier; either by using `prog_1 == prog_2`, or `prog_1.equivalence(prog_2)` or `sf.program_utils.program_equivalence`.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
Closes #135


josh146 (commented):
Nice to have this finally integrated into SF 🎉 No major comments from me, happy for this to be merged with @sduquemesa's approval.

One tangential question: the way that this checks for program equivalency will scale really badly, due to the use of graph isomorphism. While this is a good utility function to have, should `==` default to simply checking temporal operations/parameters?

thisac:
> One tangential question: the way that this checks for program equivalency will scale really badly, due to the use of graph isomorphism. While this is a good utility function to have, should == default to simply checking temporal operations/parameters?

That is definitely a good point. I'm not sure, actually. On one hand, it's really neat to be able to check _this_ equivalence by simply using `==`, but on the other hand, as you said, it doesn't scale well. Are there any other options here; e.g., to have only check DAG's for smaller circuits, or to provide some warning for larger circuits (not too fond of either of those, though, to be honest)? 🤔

josh146:
If I had to pick, I would have `==` by default not create a DAG at all, and just compare circuits one by one.

I suppose because this reminds me of other programming environments, where `==` does a _direct_ equivalency check, and doesn't attempt to do any simplification/analysis that might require computational overhead.

thisac:
@josh146 I've updated the `Program.__eq__` method to simply compare the gates and params in the circuits, instead of creating a DAG and checking isomorphism. I also added a convenience method `Program.equivalence`, using `program_equivalence`, which allows users to directly compare programs without needing to import the helper function.

josh146:
Thanks @thisac! That sounds like a really nice compromise 🎉

.github/CHANGELOG.md
```diff
@@ -9,7 +9,21 @@
...
+
+* There's a `program_equivalence` function in `strawberryfields/program_utils.py` which checks
+  Strawberry Fields programs for equivalence.
+  [(#686)](https://github.com/XanaduAI/strawberryfields/pull/686)
+
+* An equality operator is implemented for `strawberryfields.Program`, checking that the exact same
+  gates and respective parameters, are applied in order.
+  [(#686)](https://github.com/XanaduAI/strawberryfields/pull/686)
+
+* A `Progra.equivalence` convenience method is added which calls the `program_equivalence`
```
josh146, line 24:
```suggestion
* A `Program.equivalence` convenience method is added which calls the `program_equivalence`
```

strawberryfields/program.py
```diff
@@ -203,6 +203,66 @@ def __len__(self):
...
 
+    def __eq__(self, prog):
+        """Equality operator for programs.
+
+        .. note::
+
+            This operation does not account for logically equivalent programs where the order
+            of operations or modes do not matter. It simply checks that the operations and
+            parameters are identically applied. For a more thorough check, use the
+            ``Program.equivalence()`` method instead.
```
josh146, line 214: Does this show in the documentation? 🤔 It might have to be added elsewhere, like in the main `Program` docstring.

strawberryfields/program.py
```diff
@@ -203,6 +203,66 @@ def __len__(self):
...
+        """
+        # is the targets differ, the programs are not equal
+        if self.target != prog.target:
+            return False
+
+        # if the registers differ, the programs are not equal
+        if self.register != prog.register:
+            return False
```
josh146, line 222: Good idea!

strawberryfields/program.py
```diff
@@ -203,6 +203,66 @@ def __len__(self):
...
+    def equivalence(self, prog, **kwargs):
+        """Checks if two programs are equivalent.
+
+        This function converts the program lists into directed acyclic graphs,
+        and runs the NetworkX `is_isomorphic` graph function in order
+        to determine if the two programs are equivalent.
+
+        .. note::
+
+            This method is a convenience method, wrapping the `program_equivalence`
```
josh146, line 247:
```suggestion
            This method is a convenience method, wrapping the :func:`.program_equivalence`
```

strawberryfields/program.py
```diff
@@ -203,6 +203,66 @@ def __len__(self):
...
+            if not all((names_eq, param_eq, modes_eq)):
+                return False
+
+        return True
+
+    def equivalence(self, prog, **kwargs):
+        """Checks if two programs are equivalent.
+
+        This function converts the program lists into directed acyclic graphs,
+        and runs the NetworkX `is_isomorphic` graph function in order
```
josh146, line 242:
```suggestion
        and runs the NetworkX ``is_isomorphic`` graph function in order
```

josh146 (approved):
Love it @thisac! I think it is a really clean/nice implementation, and avoids potential bottlenecks occurring from users doing `prog1 == prog2`.

strawberryfields/program.py
thisac: Probably not, no. Moved to `Program` class docstring.

PR #687, opened 2022-03-08
Assert spec and layout target equality

thisac:
**Context:**
Currently, there's no check against using different targets in the specification (`Device.target`, or `Device._spec["target"]`) and in the Blackbird layout (target field in `Device.layout`, or `BlackbirdProgram.target`). This shouldn't be an issue for the user unless the specification is incorrect, although it's safer to have a check for this at device creation directly.

* A `Device.assert_spec` static method is added which checks that the target in the layout is the same as the target field in the specification. This check is also performed at `Device` initialization.
* If no target is specified in the layout, the device target is used instead. This is useful e.g., when compiling a program with a device (currently, if this would be attempted without a target in the layout, an obscure `CircuitError` would be raised.

**Benefits:**
* The target in the layout can either be omitted, or _must_ be the same as the target in the specification.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/device.py
+++ b/strawberryfields/device.py
@@ -59,7 +59,7 @@ class Device:
         layout = spec["layout"]
         target = spec["target"]
 
-        groups = re.findall(r"target (\w+)", layout or "")
+        groups = re.findall(r"\b(?=\w)target (\w+)", layout or "")
         if len(groups) <= 1:
             if len(groups) == 1 and target != groups[0]:
                 raise ValueError(
--- a/strawberryfields/device.py
+++ b/strawberryfields/device.py
@@ -46,17 +46,19 @@ class Device:
                 f"Device specification is missing the following keys: {sorted(missing_keys)}"
             )
 
-        self._spec = self.assert_spec(spec)
+        self._spec = self.validate_target(spec)
         self._certificate = cert
 
     @staticmethod
-    def assert_spec(spec: Mapping[str, Any]):
+    def validate_target(spec: Mapping[str, Any]):
         """Check that the target in the specification is equal to the layout target.
 
         Returns:
-            Dict: dictionary representing the raw device specification."""
+            Dict: dictionary representing the raw device specification.
+        """
         layout = spec["layout"]
         target = spec["target"]
+
         groups = re.findall(r"target (\w+)", layout or "")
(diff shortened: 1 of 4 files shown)


strawberryfields/device.py
```diff
@@ -45,9 +46,28 @@ def __init__(self, spec: Mapping[str, Any], cert: Optional[Mapping[str, Any]] =
                 f"Device specification is missing the following keys: {sorted(missing_keys)}"
             )
 
-        self._spec = spec
+        self._spec = self.assert_spec(spec)
         self._certificate = cert
 
+    @staticmethod
+    def assert_spec(spec: Mapping[str, Any]):
```
josh146, line 53: I wonder if it is clearer to call this `validate_target`?

strawberryfields/device.py
```diff
@@ -45,9 +46,28 @@ def __init__(self, spec: Mapping[str, Any], cert: Optional[Mapping[str, Any]] =
...
+        """Check that the target in the specification is equal to the layout target.
+
+        Returns:
+            Dict: dictionary representing the raw device specification."""
```
josh146, line 57:
```suggestion
            dict: dictionary representing the raw device specification."""
```

tests/api/test_devicespec.py
```diff
@@ -149,6 +149,36 @@ def test_create_program_no_layout(self):
...
+            version 1.0
+            target banana
+
+            S2gate({squeezing_amplitude_0}, 0.0) | [0, 1]
+            MZgate({phase_0}, {phase_1}) | [0, 1]
+            MeasureFock() | [0, 1]
+            """
+        )
+        device_dict_different_target = {
+            "target": "pawpaw",
```
josh146, line 166: 😆

josh146 (approved):
Looks good to me @thisac! I suppose I have one question, though - is there anywhere in the logic where we are making _use_ of the fact that different spec and different layout targets are allowed?

That is, do we have any existing cases where they might already differ?

strawberryfields/device.py
thisac: Yes, that's probably better. I was thinking this as a more general spec validation, but you're correct. `validate_target` makes more sense the way it's written now.

Issue #683, opened 2022-02-24
Register references cached by SymPy


thisac:
### Before posting a bug report

- [X] I have searched exisisting GitHub issues to make sure the issue does not already exist.

### Expected behavior

It works to use feed-forwarding to use measured values as parameters in gate applications. Specifically, there seems to be an issue with older `MeasuredParameter` objects being cached and reused when using together with mathematical operation (such as in the example below).

```python
ops.Rgate(1 + r[0].par) | r[0]
```

### Actual behavior

1. A register reference object calls the `par` method, returning a `MeasuredParameter` object, inheriting from `sympy.Symbol` (e.g., `r[0].par`).
2. A mathematical operation is applied to the `MeasuredParameter` object, and a `sympy.Expr` object is returned (e.g., `1 + r[0].par`).
3. This symbolic expression seems to get cached by SymPy.
4. The cached expression is used instead of a newly created expression when they're identical, implicitly using the _old_ register reference stored in the `MeasuredParameter` object which no longer exists in the programs registers (`Program.registers`).
5. An error is raised in the `Program._test_regrefs` method since the measured parameters register reference doesn't match the ones in the program.

### Reproduces how often

Consistently, but only under very specific circumstances where SymPy seems to be caching symbolic parameters.

### System information

```shell
Strawberry Fields: a Python library for continuous-variable quantum circuits.
Copyright 2018-2020 Xanadu Quantum Technologies Inc.

Python version:            3.9.6
Platform info:             macOS-12.1-arm64-i386-64bit
Installation path:         /Users/theodor/Repos/strawberryfields-td3/strawberryfields
Strawberry Fields version: 0.22.0-dev
Numpy version:             1.21.3
Scipy version:             1.7.0
SymPy version:             1.8
NetworkX version:          2.6.1
The Walrus version:        0.18.0
Blackbird version:         0.5.0-dev
XCC version:               0.2.0-dev
TensorFlow version:        2.4.1
```

### Source code

```shell
"""Needs fixtures in `tests/conftest.py`. Create a file with the following content and put it in the `tests` folder. Run it using pytest: `pytest tests/name_of_test_file.py`"""

import inspect

import blackbird as bb

from strawberryfields import ops
from strawberryfields.program import Program


def test_regref_inconsistency(setup_eng):
    layout = inspect.cleandoc(
        """
        name test
        version 1.0

        float array p0[1, 259] =
            {s}
        float array p1[1, 259] =
            {r0}
[...]

PR #688, opened 2022-03-09
Deep copy from `Program.compile()`

thisac:
**Context:**
The compiler create a linked copy of the program by calling `Program._linked_copy()`, which retains the same regrefs. Since the copy is shallow, it also keeps the same TDM parameters in a TDM program. These parameters are updated by the compiler and thus changed in both the compiled program _and_ the initial program, while the circuit is only changed in the compiled program.

* `Program._linked_copy()` is updated to create a deep-copy of all attributes except the ones containing register references, including e.g., the TDM parameters.
* The initial program's regrefs are linked to the compiled programs regrefs instead of implicitly referenced by using a shallow copy.

**Benefits:**
The compiler actually returns a copy of the program without changing anything in the original program.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -546,7 +546,9 @@ class Program:
             Program: a copy of the Program
         """
         self.lock()
-        p = copy.copy(self)  # shares RegRefs with the source
+        p = copy.deepcopy(self)
+        p.reg_refs = self.reg_refs  # shares RegRefs with the source
+
         # link to the original source Program
         if self.source is None:
             p.source = self
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -553,7 +553,7 @@ class Program:
             # should share the same register references. Program.circuit potentially
             # contains FreeParameters/MeasuredParameters, which contain RegRefs.
             if name not in ("circuit", "reg_refs", "init_reg_refs"):
-                setattr(name, val, copy.deepcopy(val))
+                setattr(p, name, copy.deepcopy(val))
 
         # link to the original source Program
         if self.source is None:
--- a/strawberryfields/program.py
+++ b/strawberryfields/program.py
@@ -546,8 +546,14 @@ class Program:
             Program: a copy of the Program
         """
         self.lock()
-        p = copy.deepcopy(self)
-        p.reg_refs = self.reg_refs  # shares RegRefs with the source
+        p = copy.copy(self)
+
+        for name, val in self.__dict__.items():
+            # Deep-copy all attributes except 'circuit' and 'reg_refs', since the programs
+            # should share the same register references. Program.circuit potentially
+            # contains FreeParameters/MeasuredParameters, which contain RegRefs.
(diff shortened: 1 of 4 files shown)


strawberryfields/program.py
```diff
@@ -546,7 +546,15 @@ def _linked_copy(self):
...
         self.lock()
-        p = copy.copy(self)  # shares RegRefs with the source
+        p = copy.copy(self)
+
+        for name, val in self.__dict__.items():
+            # Deep-copy all attributes except 'circuit' and 'reg_refs', since the programs
+            # should share the same register references. Program.circuit potentially
+            # contains FreeParameters/MeasuredParameters, which contain RegRefs.
+            if name not in ("circuit", "reg_refs", "init_reg_refs"):
+                setattr(p, name, copy.deepcopy(val))
```
thisac, line 556: Optimally, we should deep-copy the circuit as well, but since the may contain regrefs, this isn't as straight-forward to do. Just making a deepcopy here would cause two errors:
1. Due to `MeasuredParameter` not supporting `deepcopy` (because of the different signature of the `__new__()` method). A `__deepcopy__()` method would potentially need to be added.
2. Any copied regrefs would raise the unnerving `"RegRef state has become inconsistent."` issue we had [earlier](https://github.com/XanaduAI/strawberryfields/issues/683).

We could e.g., override the `__deepcopy__` method for the `Command` class, but that would require it to only deep-copy everything _except_ symbolic parameters. Alternatively, add a `copy_everything_except_regref` method to `Command`.

sduquemesa: Because this is a potentially actionable item, maybe is worth to have it as a (TODO) comment in the code. It might get lost here.

sduquemesa: Because this is a potentially actionable item, maybe is worth to have it as a (TODO) comment in the code. It might get lost here.

tests/frontend/test_program.py
```diff
@@ -227,7 +227,7 @@ def test_eq_symmetric_bsgate(self, compare_params):
         assert prog_2.equivalence(prog_1, compare_params=compare_params)
 
     @pytest.mark.parametrize("compare_params", [True, False])
-    def test_neq_operator_equivalent(self, compare_params):
+    def test_equivalence_different_circuits(self, compare_params):
         """Programs with differnet, but equivalent, circuits."""
```
sduquemesa, line 231:
```suggestion
        """Programs with different, but equivalent, circuits."""
```

sduquemesa (approved):
Great @thisac! 
One question only: why programs should share the same references for the registers? Why copy everything else except `reg_refs`?

strawberryfields/program.py
thisac: I've created an issue for this instead (#691). I think it's better to keep track of it there rather than as a TODO which can easily be lost. 🙂

thisac: I've created an issue for this instead (#691). I think it's better to keep track of it there rather than as a TODO which can easily be lost. 🙂

PR #692, opened 2022-03-14
Update random tests and seeds to use `pytest-randomly`

sduquemesa:
### Before submitting

**Context:**
Currently different modules in the test suite use numpy seeds to make the tests in each particular module deterministic. However, using seeds in this manner breaks the possibility of using random ordering (to check for no inter-test dependency) and seeding of the full test suite through `pytest-randomly`; hence strawberry fields has been running its test suite with a fixed seed, i. e., `pytest --randomly-seed=42`.

This PR removes set numpy seeds from test modules when possible (some specific test do requiere to set a seed), leaving `pytest-randomly` as the responsible of setting and configuring the state of the random number generator.

**Benefits:**
Pytest no longer requires a fixed seed in order to execute the test suite allowing `pytest-randomly` to shuffle tests ordering and controlling the global state of the random number generator. In this sense, `pytest`/`pytest-randomly` will become the only "source of truth" regarding random number generation for the test.

**Possible Drawbacks:**
[The answer to the ultimate question of life, the universe, and everything](https://en.wikipedia.org/wiki/42_(number)#The_Hitchhiker's_Guide_to_the_Galaxy) is no longer hardcoded in strawberryfields.

### DIFF
--- a/tests/apps/train/test_cost.py
+++ b/tests/apps/train/test_cost.py
@@ -251,7 +251,6 @@ class TestStochasticIntegrationPNR:
         sampling from single mode squeezed states whose photon number distribution is specified
         by the negative binomial."""
         ps = 1 / (1 + np.array(n_mean_by_mode))
-        np.random.seed(0)
         return np.array([2 * np.random.negative_binomial(0.5, p, n_samples) for p in ps]).T
 
     def h_setup(self, objectives):
--- a/tests/backend/test_multimode_state_preparations.py
+++ b/tests/backend/test_multimode_state_preparations.py
@@ -21,9 +21,6 @@ import numpy as np
 from strawberryfields import ops
 from strawberryfields.utils import random_covariance, displaced_squeezed_state
 
-# make test deterministic
-np.random.seed(42)
-
 
 MAG_ALPHAS = np.linspace(0, 0.8, 3)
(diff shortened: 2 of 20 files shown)


.github/CHANGELOG.md
```diff
@@ -33,6 +33,10 @@
   the target field in the specification. This check is also performed at `Device` initialization.
   [(#687)](https://github.com/XanaduAI/strawberryfields/pull/687)
 
+* Tests are run in random order and the seed for numpy's and python's random number generators is
+  set by `pytest-randomly`.
+  [(#692)](https://github.com/XanaduAI/strawberryfields/pull/692)
```
thisac, line 38:
```suggestion
* Tests are run in random order and the seed for NumPy's and Python's random number generators are
  set by `pytest-randomly`.
  [(#692)](https://github.com/XanaduAI/strawberryfields/pull/692)
```

thisac (approved):
Looks great @sduquemesa. We'll just have to keep in mind to run the tests (in random order) a few times, fixing any potential issues, before setting this in the CI as well.

sduquemesa:
_Note:_ this PR is not removing the set seed from the CI test execution (i. e., `--randomly-seed=42` pytest's option), only the unneeded `np.random.seed` spread across the test modules.

PR #696, opened 2022-03-22
Incrementing the version number to 0.22.0

thisac:
Preparing for the release of Strawberry Fields 0.22.0.
* Changelog has been updated with examples.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Release 0.22.0 (development release)
+# Release 0.22.0 (current release)
 
 <h3>New features since last release</h3>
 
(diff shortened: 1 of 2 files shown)


.github/CHANGELOG.md
```diff
@@ -10,13 +10,52 @@
...
+
+  with prog.context as q:
+      ops.Sgate(0.5) | q[0]
+      ops.MeasureFock() | q
+
+  results = eng.run(prog, shots=None)
+
+  print(results.samples)
+  # will output an empty list []
+  ```
```
sduquemesa, line 57: Here it would be cool to explicitly show that even if no samples are retuned the state is available
```suggestion
  print(results.samples)
  # will output an empty list []
  
  # the resulting Gaussian state is accesible 
  # via its vector of means and covariance matrix
  print(results.state.means())
  print(results.state.cov())
  ```
```

sduquemesa (approved):
@thisac, great to have examples describing the new features! It should become a practice to add them every time when updating the changelog.

thisac:
> It should become a practice to add them every time when updating the changelog.

Yes, this should absolutely be the case, but we've just been a bit bad at including them. 😄

PR #702, opened 2022-04-25
Tidy up (un)rolling

thisac:
**Context:**
Rolling/unrolling/space-unrolling wasn't always working as one might intuitively expect. This PR tweaks the (un)rolling a bit and also fixes a few issues.

* A locked program can now be (un)rolled, and automatically restores the lock if there.
* Rolling and unrolling now _only_ happens in place. It has previously also been returning the (un)rolled circuit, but this was never used.
* Fixed bug when trying to unroll an already unrolled program with a different number of shots.
* Fixed bug with vacuum modes missing.
* Tidied up some docstrings.

**Benefits:**
Things look and work better.

**Possible Drawbacks:**
(un)rolling no longer returns the (un)rolled circuit. It only happens in place.

**Related GitHub Issues:**
None

### DIFF
--- a/strawberryfields/tdm/tdmprogram.py
+++ b/strawberryfields/tdm/tdmprogram.py
@@ -237,7 +237,7 @@ def move_vac_modes(samples, N, crop=False):
 
 
 def get_mode_indices(delays):
-    """Calculates the mode indices for use in a ``TDMProgram``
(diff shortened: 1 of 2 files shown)


sduquemesa (approved):
Making the almighty CodeFactor happy once again! 😶‍🌫️🐲

PR #711, opened 2022-05-30
Validate parameters fix

thisac:
**Context:**
TDM programs sometimes uses arrays instead of single parameter values causing issues when calling the `Device.validate_parameter` method.

* `Device.validate_parameter` is updated to work with nested parameter arrays by flattening any nested array and then checking each parameter separately.

**Benefits:**
`Device.validate_parameter` works with arbitrarily nested parameter arrays.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


.github/CHANGELOG.md
```diff
@@ -28,6 +28,9 @@
 * Fixed bug with vacuum modes missing.
   [(#702)](https://github.com/XanaduAI/strawberryfields/pull/702)
 
+* Validating parameters now work with nested parameter arrays.
```
sduquemesa, line 31:
```suggestion
* Validating parameters now works with nested parameter arrays.
```

strawberryfields/device.py
```diff
@@ -142,18 +142,35 @@ def validate_parameters(self, **parameters: complex) -> None:
         if self.gate_parameters is None:
             return
 
+        def _flatten(iterable: Iterable) -> Generator:
+            for it in iterable:
+                if isinstance(it, Iterable) and not isinstance(it, str):
+                    yield from _flatten(it)
+                else:
+                    yield it
```
sduquemesa, line 150: Why choosing this approach over using an utility function from `itertools`?

sduquemesa: Oh, I think I answered myself: `itertools` won't flatten deeply nested arrays whereas here you use recursion to do so. Great approach!

tests/api/test_devicespec.py
```diff
@@ -213,6 +213,8 @@ def test_invalid_spec(self):
             {"phase_0": 1.23},
             {"phase_0": 0},
             {"phase_0": 6.29999},
+            {"phase_0": [6.29999, 0, 0.14]},
+            {"phase_0": [[0.1, 0, 2], [[0.5]]]},
```
sduquemesa, line 217: yep, recursion for the deeply nested ones ♻️🤓

In which cases the TDM program uses arrays instead of single parameter values?

strawberryfields/device.py
```diff
@@ -142,18 +142,35 @@ def validate_parameters(self, **parameters: complex) -> None:
...
+                else:
+                    if v not in self.gate_parameters[p]:
+                        # parameter is present in the device specifications
+                        # but the user has provided an invalid value
+                        raise ValueError(
+                            f"'{p}' has invalid value {v}. Only {self.gate_parameters[p]} allowed."
+                        )
+
+            else:
+                raise ValueError(f"Parameter '{p}' not a valid parameter for this device")
```
sduquemesa, line 170: A small cosmetic suggestion: I think the code is more readable if the first `if-else` clause is replaced by a guard clause (halting the execution as well when the condition of `p in self.gate_parameters` is not met) and hence removing one level of nesting. 💅

```suggestion
            if p not in self.gate_parameters:
                raise ValueError(f"Parameter '{p}' not a valid parameter for this device")
            
            if isinstance(v, Iterable):
                for i in _flatten(v):
                    if i not in self.gate_parameters[p]:
                        raise ValueError(
                            f"'{p}' has invalid value {i}. Only {self.gate_parameters[p]} allowed."
                        )
            elif v not in self.gate_parameters[p]:
                # parameter is present in the device specifications
                # but the user has provided an invalid value
                raise ValueError(
                    f"'{p}' has invalid value {v}. Only {self.gate_parameters[p]} allowed."
                )
```

sduquemesa (approved):
Great @thisac! I have left a small suggestion. Other than that, nothing blocking this from merging.

strawberryfields/device.py
thisac: Thanks, although it's probably a bit overkill to allow for deeply nested arrays, to be honest.

strawberryfields/device.py
thisac: Thanks, very good suggestion! 💯

PR #722, opened 2022-06-16
Fix release action

thisac:
**Context:**
Automatic PyPI release action is failing due to failing tests when simply running tests with `pytest tests` (instead of `make tests`).

* Adds a random seed to the tests run on the upload action.
* Sets the env var `SYMPY_USE_CACHE` to `"no"` to avoid issue with sympy caching (see #684 for details).
* Fixes tolerances in a few tests that were failing too often (specifically when running without setting a seed).
* Replaces `_einsum_v1` with `tf.einsum` in all places but one (still an issue with `sf.backends.tfbackend.ops.conditional_state()` which I wasn't able to solve) and also removed the einsum test marked XFAIL (was XPASSing).

**Benefits:**
* It should now be possible to run all tests directly using `pytest tests`.
* The automatic PyPI release on GitHub published should now work.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -6,6 +6,10 @@ on:
 jobs:
   upload:
     runs-on: ubuntu-latest
+
+    env:
+      SYMPY_USE_CACHE: "no"
+
     steps:
       - uses: actions/checkout@v2
 
@@ -27,7 +31,7 @@ jobs:
 
       - name: Run tests
         run: |
-          python -m pytest tests --tb=native
+          python -m pytest tests -p no:warnings --randomly-seed=42 --tb=native
 
       - name: Publish
         uses: pypa/gh-action-pypi-publish@master
--- a/strawberryfields/backends/tfbackend/ops.py
+++ b/strawberryfields/backends/tfbackend/ops.py
@@ -33,6 +33,10 @@ from string import ascii_lowercase as indices
 from string import ascii_letters as indices_full
 
 import tensorflow as tf
+
+# only used for `conditional_state`; remove when working with `tf.einsum`
+from tensorflow.python.ops.special_math_ops import _einsum_v1
+
 import numpy as np
 from scipy.special import factorial
 from scipy.linalg import expm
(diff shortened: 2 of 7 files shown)


strawberryfields/backends/tfbackend/ops.py
```diff
@@ -1464,7 +1455,9 @@ def conditional_state(system, projector, mode, state_is_pure, batched=False):
     einsum_args = [system, tf.math.conj(projector)]
     if not state_is_pure:
         einsum_args.append(projector)
-    cond_state = tf.einsum(eqn, *einsum_args)
+
+    # does not work with `tf.einsum`; are the `einsum_args` shapes wrong?
+    cond_state = _einsum_v1(eqn, *einsum_args)
```
thisac, line 1460: This _should_ work using `tf.einsum`, but there were issues with the shape of the `einsum_args` arrays. It looked like they were of the wrong shape, but I couldn't figure it out. It works (for now) using `_einsum_v1` as before, but it'd be nicer to simply use `tf.einsum` instead.

tests/apps/train/test_cost.py
```diff
@@ -324,7 +324,7 @@ def test_gradient(self, dim, n_mean, simple_embedding):
         This can be differentiated to give the derivative:
         d/dx E((s - x) ** 2) = 6 * n_mean + 2 * (1 - x).
         """
-        n_samples = 10000  # We need a lot of shots due to the high variance in the distribution
+        n_samples = 20000  # We need a lot of shots due to the high variance in the distribution
```
thisac, line 327: This takes ~3 seconds to run on my laptop (probably longer on CI). Test failed often with only 10000 samples, and even more than 20000 would be better, and would allow for a lower tolerance below, but would take much longer to run.

strawberryfields/backends/tfbackend/ops.py
sduquemesa: Unless the shape of the input tensors in changed, I there's no way around the issue. You're correct in the sense that the shape is the problem, from the new `einsum` (`_einsum_v2`) [code comments](https://github.com/tensorflow/tensorflow/blob/d8ce9f9c301d021a69953134185ab728c1c248d3/tensorflow/python/ops/special_math_ops.py#L1187-L1191)

```
    # Validate and sanitize the equation and resolve static input shapes, as
    # opt_einsum requires that all shapes be a tuple of positive integers.
    # Also remove ellipsis from the equation as opt_einsum will replace them
    # with named labels. Then broadcasting between different shapes or ranks
    # wouldn't work. (E.g. [1, 1, 2] wouldn't broadcast with [3, 1]).
```

Some of the inputs are, for example, of sizes [1, 6, 6, 6, 6, 6, 6] and [1, 6], and then won't broadcast and work with `einsum`.

.github/workflows/upload.yml
```diff
@@ -27,7 +31,7 @@ jobs:
 
       - name: Run tests
         run: |
-          python -m pytest tests --tb=native
+          python -m pytest tests -p no:warnings --randomly-seed=42 --tb=native
```
sduquemesa, line 34: For future reference I'm attaching this comment from PR #480:

> Optimally, we would remove the `randomly-seed=42` from the tests, causing the seed to be set randomly each time, helping us discover potential bugs due to the stochastic nature of some tests. By resetting the seed before each test, we may miss failures that might be present. This might cause some headaches, though.

sduquemesa (approved):
Thanks @thisac! Hopefully this will make the release pipeline more streamlined 🚀 I left some comments just as potential future references.

PR #751, opened 2025-01-28
Update decomposition function

jama7168:
### Before submitting

- [ ] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      test directory!

- [ ] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [ ] Ensure that code and tests are properly formatted, by running `make format` or `black -l 100
      <filename>` on any relevant files. You will need to have the Black code format installed: 
      `pip install black`.

- [ ] The Strawberry Fields source code conforms to
      [PEP8 standards](https://www.python.org/dev/peps/pep-0008/).
      We check all of our code against [Pylint](https://www.pylint.org/).
      To lint modified files, simply `pip install pylint`, and then
      run `pylint strawberryfields/path/to/file.py`.

**Context:**
During testing the interferometer Class I came across the problem that  Interferometer(unitary_matrix, mesh = 'triangular', tol = 1e-10)  and  Interferometer(unitary_matrix, mesh = 'rectangular_phase_end', tol = 1e-10) give different results when applied within the boson sampling. As the matrix we are starting with is the same that should not be (see also here: https://discuss.pennylane.ai/t/problems-in-understanding-sf-decompositions-triangular/7875/6 for more details and the discussion).

I changed the following things: 
the order of the return in the triangular decomposition from reversed(tlist), np,diag(localV), None to None, np,diag(localV), t_list so that in the Interferometer_decomposition function the order of application is first the diagonal (phase shifters) and then the T_i[0] to T_i[N]. 
Then I changed the interferometer_decompose method such that it checks if the BS1 from the Decomposition is None.
I also provided the minimal example as a test with which i tested the interferometer class. Now the results for both are the same as it should be. 

**Benefits:**

Fixed a problem with the triangular decomposition in the Interferometer.

**Possible Drawbacks:**
Since I changed the return of the triangular decomposition it might affect other functions. Maybe we should check that first!!

**Related GitHub Issues:**
Not a gitub Issue but here the form discussion I had:
https://discuss.pennylane.ai/t/problems-in-understanding-sf-decompositions-triangular/7875/6

### DIFF
--- a/tests/frontend/test_interferometer_triangular.py
+++ b/tests/frontend/test_interferometer_triangular.py
@@ -1,0 +1,54 @@
+import strawberryfields as sf
+from strawberryfields.ops import *
+import numpy as np
+
+M = 4
+# Create a random complex matrix
+random_matrix = np.random.rand(M, M) + 1j * np.random.rand(M, M)
+
+# Perform QR decomposition to get a unitary matrix
+q, r = np.linalg.qr(random_matrix)
+
+# Normalize to ensure unitary property (q should already be unitary, but we'll double-check)
+unitary_matrix = q / np.linalg.norm(q, axis=0)
+
+
+boson_sampling_triangular = sf.Program(4)
+#initiate a starting state |1101>
+with boson_sampling_triangular.context as q:
+    # prepare the input fock states
+    Fock(1) | q[0]
+    Fock(1) | q[1]
+    Vac     | q[2]
+    Fock(1) | q[3]
+
+    # apply the triangular interferometer
+    Interferometer(unitary_matrix, mesh = 'triangular', tol = 1e-10) | q
+
+#run the simulation
+boson_sampling_triangular.compile(compiler="fock")
+eng = sf.Engine(backend="fock", backend_options={"cutoff_dim": 5})
+results = eng.run(boson_sampling_triangular)
+# extract the joint Fock probabilities
+probs_triangular = results.state.all_fock_probs()
+
+#now try with mesh = 'rectangular'
+boson_sampling_rectangular = sf.Program(4)
+
+with boson_sampling_rectangular.context as q:
+    # prepare the input fock states
+    Fock(1) | q[0]
+    Fock(1) | q[1]
+    Vac     | q[2]
+    Fock(1) | q[3]
+
+    # apply the rectangular interferometer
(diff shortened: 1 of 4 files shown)


CatalinaAlbornoz:
Hi @jama7168, 

Thanks again for opening this PR. A member of our team is currently reviewing it. We should get back to you soon with next steps.

Have a good weekend!

tests/frontend/test_interferometer_triangular.py
```diff
@@ -0,0 +1,54 @@
...
+    # apply the rectangular interferometer
+    Interferometer(unitary_matrix, mesh = 'rectangular', tol = 1e-10) | q
+    
+boson_sampling_rectangular.compile(compiler="fock")
+eng = sf.Engine(backend="fock", backend_options={"cutoff_dim": 5})
+results = eng.run(boson_sampling_rectangular)
+# extract the joint Fock probabilities
+probs_rect = results.state.all_fock_probs()
+#all probabilities should be the same here! Sp the output should be True
+print(f"{np.allclose(probs_triangular, probs_rect)}")
```
ziofil, line 54: shouldn't this be an assert statement?

ziofil (commented):
Thanks Jacob for opening the PR! Could you follow up on the errors that are raised in the tests? I also left a comment. Thanks again!

tests/frontend/test_interferometer_triangular.py
jama7168: I think you are right there. I copied the code from my local test where i just printed out the results to check if it works. I replaced it with an assert statement and made the test itself a bit more stable and added that one can change the size of the test matrix. I hope this fixes the problem with the test. As I am very new to this, I tried to run make docs and make test on my local machine but I ran into some troubles setting it up correctly. I still hope that this resolves the problem.

elib20:
Hi Jacob. Can you look into the issues that have been raised by CodeFactor and readthedocs and make sure the rest of the tests pass? We'll rereview once all the checks have passed.

CatalinaAlbornoz:
Hi @jama7168, I wanted to check if you wanted to keep working on this PR. Do you need help with any of the issues that were raised by CodeFactor and readthedocs?

